PR #7, opened 2021-08-23
Fix: USA and China Microwave Tank negative power bug

alanblack166:
USA and China Power Plants, if disabled by a Microwave Tank, will suppress their power production until the Microwave Tank stops.

If they are disabled while being built, however, they will lose power that they never produced. Once construction is complete, the power plant will produce net zero power, upgrades and overcharge notwithstanding, and if that-or any-power plant is sold, the player will have permanent negative power.

This commit fixes that: It will no longer be possible to disable power plants while they are being built.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -9677,11 +9677,24 @@ Object AirF_AmericaPowerPlant
   EnergyBonus      = 5
   VisionRange      = 200.0           ; Shroud clearing distance
   ShroudClearingRange = 200
+
+  ; @bugfix - hanfield
+  ; Added new default armorset that cannot be targeted or suffer SUBDUAL damage.
+  ; This resolves the Microwave Tank exploit but does not prevent being properly disabled later.
+  ; 23/08/2021
+
   ArmorSet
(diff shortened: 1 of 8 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -9752,6 +9765,23 @@ Object AirF_AmericaPowerPlant
...
+  ; @bugfix - hanfield
+  ; Added two extra modules for Microwave Tank bug fix.
+  ;
+  ; MicrowaveTankFix01 grants an object-class upgrade once the building is complete.
+  ; MicrowaveTankFix02 upgrades the armor with said upgrade and allows subdual once again.
+  ;
+  ; 23/08/2021
+
+  Behavior = GrantUpgradeCreate ModuleTag_MicrowaveTankFix01
+    UpgradeToGrant = Upgrade_ChinaOverlordGattlingCannon
```
xezon: Why does it say `ChinaOverlordGattlingCannon` here?

alanblack166: We use an upgrade with Type = OBJECT to enable the ArmorUpgrade module locally. This will be replaced with a dedicated dummy upgrade.

PR #54, opened 2021-08-27
Created PlayerTemplate.ini

alanblack166:
Unedited PlayerTemplate.ini for future modification.


commy2:
We should just add all the relevant files at once.

I also propose that we remove trailing whitespace characters as soon as possible. This will make editing these files later a lot easier and cause less changed lines in the diffs and also makes greping through the files easier.

PR #63, opened 2021-08-27
Added global dummy upgrade and grant object

alanblack166:
Added global dummy upgrade and an object that will grant it to all challenge/skirmish/MP players upon starting the game

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -2841,17 +2841,17 @@ Object GrantGlobalDummyUpgradeObject
   KindOf = NO_COLLIDE IMMOBILE UNATTACKABLE
   
   ; *** ENGINEERING Parameters ***
-  Body = ImmortalBody ModuleTag_040
+  Body = ImmortalBody ModuleTag_Body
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -2829,7 +2829,34 @@ Object EMPPatriotEffectSpheroid
...
+Object GrantGlobalDummyUpgradeObject
+
+  ; ***DESIGN parameters ***
+  VisionRange = 0.0
+  EditorSorting = SYSTEM
+  
+  KindOf = NO_COLLIDE IMMOBILE UNATTACKABLE
+  
+  ; *** ENGINEERING Parameters ***
+  Body = ImmortalBody ModuleTag_040
```
xezon: Is the any rationale behind choosing these `ModuleTag` names with the number behind?

commy2: The name has to be unique. It can also be used in a map to remove or overwrite the module.

Generally, they are just ordered in ascending order. Module_Tag01,  Module_Tag02,  Module_Tag03 etc.

Edit: *unique inside the Object type

PR #67, opened 2021-08-27
Fix: Scud Storm Always Creates Green Particles When Firing

commy2:
1.04

- The particle effects of a firing a Scud Storm are always green, no matter what faction or upgrade.

With this patch:

- Anthrax Beta creates blue particle effects. Anthrax Gamma creates pink particle effects.
- Demo General creates no particle effects.


xezon (changes requested):
Please add comments to added code and changed lines (in INI files)

Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
```diff
@@ -5368,6 +5368,708 @@ Object Boss_ScudStorm
       ParticleSysBone      = SteamS01 SteamVent
     End
 
+    ; Patch104p @bugfix
```
xezon: Please put a description in the same line as Patch104p keyword, so that when searching for lines containing it, there will always be a meaningful result.

xezon: Description looks understandable and sufficient.

xezon (approved):

PR #69, opened 2021-08-27
Fixes burning and poisoned infantry drawing enemy fire

alanblack166:
Burning and poisoned infantry should no longer draw enemy fire.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -1259,7 +1259,7 @@ Object FlamingInfantry
 
   ; *** ENGINEERING Parameters ***
   RadarPriority = NOT_ON_RADAR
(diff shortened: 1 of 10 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -1259,7 +1259,7 @@ Object FlamingInfantry
 
   ; *** ENGINEERING Parameters ***
   RadarPriority = NOT_ON_RADAR
-  KindOf = CAN_CAST_REFLECTIONS INFANTRY
+  KindOf = CAN_CAST_REFLECTIONS INFANTRY UNATTACKABLE
```
xezon, line 1262: You can also add a comment in these modified lines.

xezon (changes requested):
You somehow have multiple changes in this one Request. Could you please check?

PR #72, opened 2021-08-27
Fix: Scud-bug

commy2:
- fixes the famous Scud-bug


xezon (changes requested):
Please add at least 1 comment in each file describing the problem and the fix.

xezon (approved):

xezon (approved):

PR #76, opened 2021-08-27
Change: Adds heroic tracers to GLA Jarmen and USA Pathfinder

alanblack166:
Jarmen Kell and Pathfinders should now fire red tracers when heroic.


xezon (approved):

ExiLe-ZH:
Is this visible for the enemy? If so, then this would be an unintentional side effect and something that should implemented through a dedicated pathfinder nerf pull request instead, if we intent to implement this nerf.

PR #77, opened 2021-08-27
Fix: Destroyed Toxin Tractor Always Creates Green Particles 

commy2:
1.04:

- When a Toxin Tractor is destroyed, it will create a green poison cloud, ~~no matter what upgrade.~~ with the Anthrax Beta upgrade.
- When a Toxin General Tractor is destroyed, a green poison cloud is used before Anthrax Gamma, even though the Tractor uses a blue stream for the weapon.

After this patch:

- Each Toxin Tractor will spawn the correct poison puddle when destroyed that matches the Anthrax upgrade used by the weapon.


Patch104pZH/GameFilesEdited/Data/INI/Object/StealthGeneral.ini
```diff
@@ -17905,9 +17905,10 @@ Object Slth_GLAVehicleToxinTruck
     FX                        = FINAL FX_ToxinTruckExplosion
   End
 
+  ; Patch104p @bugfix Upgrade explosion effect with Anthrax upgrade. commy2 27/08/2021
   Behavior                    = SlowDeathBehavior  ModuleTag_05Anthrax
     DeathTypes                = ALL -CRUSHED -SPLATTED
-    RequiredStatus            = STATUS_RIDER1
+    RequiredStatus            = STATUS_RIDER1   ; Patch104p @bugfix commy2 27/08/2021 Disable effect once upgraded with Anthrax. commy2 27/08/2021
```
alanblack166, line 18781: I've come to discover that having SlowDeathBehavior modules with RequiredStatus fields can make the object remain after a player surrenders or is defeated. I would advise to instead alternate between ExemptStatus-es for the two SlowDeath modules: 
Anthrax Alpha death is exempted by STATUS_RIDER2;
Anthrax Beta death is exempted by STATUS_RIDER1. 
Use StatusBitsUpgrade to set one status whilst clearing the other.

commy2: I tested this and it does not seem to affect this unit, apparently because it does not use the Rider logic.

xezon (approved):

PR #85, opened 2021-08-28
Fix: Toxin General Demo Trap Creates Explosion And Poison Cloud Before It Triggers

commy2:
ZH 1.04:

- The Toxin General's Demo Trap explodes and creates the poison cloud immediately and before even the trigger sound plays.

After this patch:

- The explosion happens properly after the warning delay like every other Demo Trap.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -5444,23 +5444,26 @@ Object Chem_GLADemoTrap
     DetonateWhenKilled        = Yes
   End
 
+  ; Patch104p @bugfix commy2 28/08/2021 Fix Toxin Demo Trap exploding before the warning delay unlike other Demo Traps.
+
   Behavior = SlowDeathBehavior ModuleTag_05
-    ExemptStatus = UNDER_CONSTRUCTION
+    ExemptStatus = UNDER_CONSTRUCTION STATUS_RIDER1
     DestructionDelay = 1000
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
```diff
@@ -5444,23 +5444,26 @@ Object Chem_GLADemoTrap
...
+    DestructionDelay = 1000
+    FX = INITIAL FX_GLADemoTrapWarning
+    Weapon = FINAL Chem_DemoTrapDetonationWeaponGamma
   End
 
-  Behavior = FireWeaponWhenDeadBehavior ModuleTag_Chem02
-    DeathWeapon   = Chem_DemoTrapDetonationWeaponGamma
-    StartsActive  = No                      ; turned on by upgrade
-    TriggeredBy   = Chem_Upgrade_GLAAnthraxGamma
+  Behavior = StatusBitsUpgrade ModuleTag_05Upgrade
```
xezon, line 5464: Maybe we should keep the original `ModuleTag_Chem02` name?

commy2: It's a different module type (StatusBitUpgrade instead of FireWeaponWhenDeadBehavior).

I chose to name `ModuleTag_05Anthrax`, because it's a copy paste of `ModuleTag_05` with the poison field adjusted; and `ModuleTag_05Upgrade`, because it turns off `ModuleTag_05` and turns on `ModuleTag_05Anthrax`.

xezon (approved):

PR #89, opened 2021-08-28
Fix: GLA Buildings Sometimes Don't Create Scrap When They Destroy Enemy Vehicles

commy2:
ZH 1.04:

- GLA buildings do not create scrap consistently. A Tunnel Network will not create scrap while a Stinger Site does... sometimes. Except a Demo Stinger Site when kills by Suicide. And except ... bla bla

With this patch:

- GLA buildings create scrap when they kill enemy vehicles no questions asked.

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -188,7 +188,7 @@ https://github.com/commy2/zerohour/issues/28  [MAYBE]                 Demo Gener
 https://github.com/commy2/zerohour/issues/27  [IMPROVEMENT][NPROJECT] Scud Storm Does Not Damage Itself
 https://github.com/commy2/zerohour/issues/26  [IMPROVEMENT]           Saboteur Uses Wrong Art For Sabotage Building-Ability
 https://github.com/commy2/zerohour/issues/25  [NOTRELEVANT]           Boss Tomahawk Launcher Never Drops Scrap
-https://github.com/commy2/zerohour/issues/24  [IMPROVEMENT]           GLA Buildings Don't Create Scrap
+https://github.com/commy2/zerohour/issues/24  [DONE]                  GLA Buildings Don't Create Scrap
 https://github.com/commy2/zerohour/issues/23  [IMPROVEMENT]           Demo General Jarmen Kell Repeats Voice Line When Planting Timed Demo Charge
 https://github.com/commy2/zerohour/issues/22  [IMPROVEMENT]           Build Scorpion Tank-Button Misspelling
 https://github.com/commy2/zerohour/issues/21  [IMPROVEMENT]           Demo And Tox Rebel Ambush Tooltips Are Wrong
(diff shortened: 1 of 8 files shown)


xezon:
The title says "sometimes". Is it really just sometimes? I am asking because this has gameplay/balance implications if we grant GLA more scrap from kills. If it is not random, then this is a balancing concern and not a bug fix and should not be submitted without permission of the balancing team.

xezon (commented):
See above.

commy2:
> Is it really just sometimes?

Example is that Tunnels do not create scrap, so if the killshot comes from the Tunnel, you don't get scrap even though you popped 10 Quads.

Stingers however do create scrap, because the Stinger Soldiers are set up like any other infantry and those are the objects that actually make the kill.

But the Stinger Site building itself is not set up to create scrap, which means that if you suicide a Demo General Stinger Site, it does not create scrap.

I find this all confusing and would rather have it that GLA creates scrap when they kill a vehicle with their units (and buildings). This takes away that randomness and makes the game more predictable.

xezon:
Yes. These area reasonable arguments. Consistent behavior certainly is desirable.

Jundiyy:
It is already consistent if you look at it from a different angle. Anything which is 'human' based will kill and give scrap.
Normal units kill and give scrap
Stinger Sites have units inside which kill and give scrap
Tunnel Networks have no units, hence no scrap
Stinger Site blowing up is the building being blown up not the units doing suicide, hence no scrap.

Seems logical.

ReLaX82:
If that what you write is true and not a random behaviour, I would simply close this point and just state somewhere to anyone that it has been observed and decided as no bug.

KaSSaDo:
I have always thought GLA buildings don't get scrap from destroyed vehicles whereas GLA units do. Stinger Sites should always create scrap, is it really only "sometimes"? This should be consistent. RPGs shooting from inside Palace also generate scrap.

If tunnels get scrap from vehicles that would be a huge boost for GLA which should be avoided in my opinion.

Jundiyy:
> I have always thought GLA buildings don't get scrap from destroyed vehicles whereas GLA units do. Stinger Sites should always create scrap, is it really only "sometimes"? This should be consistent. RPGs shooting from inside Palace also generate scrap.
> 
> If tunnels get scrap from vehicles that would be a huge boost for GLA which should be avoided in my opinion.

It's not 'sometimes', Stinger Soldiers always give scrap. The one time 'they' don't is when you suicide the Demo Stinger Site (after upgrade) and the reason you don't is because it is the building doing the damage and not the units. Hence in such a state it's following the Tunnel Network.

xezon:
Given the explanation from Jundiyy that buildings cannot create scrap in original game, I think this is perfectly ok and we don't need to change anything here. It's not a bug, but a design decision.

Stubbjax:
Should demo traps create scrap?

Jundiyy:
For wjaywhat rr

Jundiyy:
For what reason? It's not a unit but a building.

PR #92, opened 2021-08-28
Fix: Some American And Chinese Units Use GLA Death Scream When Gamma Poisoned

commy2:
ZH 1.04:

- When poisoned with Anthrax Gamma only (pink puddles), some units use the GLA death scream instead of the intended USA or China death scream.

After this patch:

- All units use the correct scream when dying by Anthrax Gamma poison.


xezon (approved):
This is a really nice fix!

PR #95, opened 2021-08-28
Fix: GLA Stealth Saboteur Death Scream

alanblack166:
Stealth General Saboteur now uses the Saboteur death effect instead of the Rebel.


Patch104pZH/GameFilesEdited/Data/INI/Object/StealthGeneral.ini
```diff
@@ -16617,13 +16617,15 @@ Object Slth_GLAInfantrySaboteur
     SabotageDuration = 30000
   End
 
+  ; Patch104p @bugfix hanfield 28/08/2021 Replaced FX_RebelDie with FX_SaboteurDie.
```
commy2, line 16620: Perhaps inline this? That's what I am doing for one liners.

xezon (approved):

PR #96, opened 2021-08-28
Fix: China Supply Truck mass

alanblack166:
Chinese Supply Trucks show now have a much more reasonable mass.


commy2 (approved):
My first thought was 40 (same as Vee, Technical is 30), but this works as well.

xezon (approved):

PR #98, opened 2021-08-28
Speaker Towers No Longer Ensure Player Survival

alanblack166:
Chinese Speaker Towers no longer contribute towards a player's survival. If a player is reduced to a Speaker Tower - he is defeated.


xezon:
I know I marked this as IMPROVEMENT originally, but on second thought I am not convinced we need this changed, because it does not feel odd to have a player survive on Speaker Tower. Player can also survive on Power Plant. Thoughts?

commy2:
Depends on if you count the Speaker Tower as base defense or not. You don't currently (1.04) survive with e.g. a Gattling Cannon or a Bunker only either.

To me Bunker/Gat/Prop are the Chinese base defenses. Those are the ones you fit on an Overlord and Helix as upgrades.

Jundiyy:
I also feel it should count as a normal building. Yes it's small but it doesn't act as defence. Just a small normal building. So should keep the player in the game.

xezon:
It's reasonable to make the Speaker Tower building behave either way. It becomes a design question. Considering there is no need/benefit to change original behavior, we can just leave it as is.

PR #103, opened 2021-08-28
Fix: Demo General Worker On Bike Missing Demo General Bike Destruction Effect

commy2:
Zero Hour 1.04:

- Blowing up a Demo General's Bike with a Worker does not produce the same red effect as any other rider does.

After this patch:

- Bike uses the same effect no matter what rider.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -17705,7 +17705,7 @@ Object Demo_GLAVehicleCombatBike
     RequiredStatus = STATUS_RIDER1
     DeathTypes = ALL -CRUSHED -SPLATTED -TOPPLED
     DestructionDelay = 0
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -17705,7 +17705,7 @@ Object Demo_GLAVehicleCombatBike
     RequiredStatus = STATUS_RIDER1
     DeathTypes = ALL -CRUSHED -SPLATTED -TOPPLED
     DestructionDelay = 0
-    FX = INITIAL FX_BuggyNewDeathExplosion
+    FX = INITIAL FX_DEMOBuggyNewDeathExplosion  ; Patch104p @bugfix hanfield 27/08/2021 Fix Worker was making Bike use a different death effect.
```
commy2:
```suggestion
    FX = INITIAL FX_DEMOBuggyNewDeathExplosion  ; Patch104p @bugfix commy2 29/08/2021 Fix Worker was making Bike use a different death effect.
```

xezon (approved):

PR #104, opened 2021-08-28
Fix: Demo General Combat Bike With Demolitions Upgrade Deals No Damage When Destroyed By Gamma Poison

commy2:
ZH 1.04

- The Demo General's Combat Bike with a Terrorist and the Demolitions upgrade deals no damage to enemies when destroyed by an Anthrax Gamma weapon only.

With patch:

- The type of poison has no influence on the death weapon being fired.


xezon:
This will make Tox with Gamma upgrade a bit weaker against Demo Bikes. Tunnels would essentially not able to stop demo bikes without taking damage anymore. We need this reviewed by the Balance boys.

ReLaX82:
I don't get it, maybe someone would be so kind to explain?

commy2:
Imagine a Demo Bike upgraded with Demolitions upgrade.

If you kill it with green or blue poison, it explodes as if you had suicided it.
But if you kill it with pink poison, it is just killed without exploding.

Extra weird is that this distinction does not happen before the Demolitions upgrade.

It's a bug / typo in the ini files.

MTKing4:
Never heard of this one before 

If anthrax alpha and beta don't cause this, then so shouldn't gamma

+ It will balance the tox demo MU a bit more so yeah i vote fix

xezon:
I am also in favor of this change for consistency sake. It would avoid random moments, where demo bike explodes infront of army but deals no damage, just because the last shot was dealt by the Gamma Toxin.

Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -17851,6 +17851,9 @@ Object Demo_GLAVehicleCombatBike
 ;    FlingPitchVariance  = 10
 ;  End
 
+  ; Patch104p @bugfix commy2 29/08/2021 Fix poison death not firing expected death weapons.
+  ; This made Bike deal not damage when killed by Anthrax Gamma poison.
```
commy2:
```suggestion
  ; This made Bike deal no damage when killed by Anthrax Gamma poison.
```

xezon (approved):

PR #111, opened 2021-08-29
NProject: Fixed the issues with firing elevation limit. Now both tank…

alanblack166:
…s and artillery can attack very high/low targets alike

All weapons which previously had TargetPitch parameters no longer have this. This allows them to acquire targets regardless of minor elevation differences.


xezon:
Could someone please give an example (or test map?) how and where this can be tested pre and post change? I would like to test its effect.

ThePredatorBG:
Wouldn't this change make tanks on low ground engage tanks on hills, for example on Defcon? It won't look right and it changes the gameplay on some maps. A larger pitch range is probably better than removing TargetPitch constraints entirely.
Original TargetPitch was annoying on Barren Badlands and Dark Mountain maps. Maybe try 45 degree values instead?

xezon:
We tested on **[RANK] Canyon of the Dead v2** and **[RANK] Sleeping Dragon**

There is a legit issue reproducible on Canyon. On the oil ramp, if unit stands there and you click on it, then tank will stop moving and the **attack-cursor** will change into **no-attack-cursor**. This is bad behavior. You can't see cursor in image.

shot_20210830_212648_2

On Sleeping Dragon however we can now also shoot down the cliffs, which has big impact on overall gameplay on all maps with cliffs. So removing pitch settings is not suitable. We should look into tweaking them slightly instead.

shot_20210830_215752_1

PR #132, opened 2021-09-01
Fix: Helix and Overlord addons are now colored red when affected by Frenzy

alanblack166:
With these changes, the Helix and Overlord addons will now be properly tinted red when their parent unit is affected by Frenzy.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1196,6 +1196,10 @@ Weapon GattlingBuildingGun
   WeaponBonus           = CONTINUOUS_FIRE_MEAN RATE_OF_FIRE 200% ; When the object achieves this state, this weapon gets double the rate of fire
   WeaponBonus           = CONTINUOUS_FIRE_FAST RATE_OF_FIRE 300% ; This is not cumulative, so with Delay of 40, and values of 2 and 4 for these  bonuses, you shoot at (40, 20, 10)
   WeaponBonus           = PLAYER_UPGRADE DAMAGE 125%     ; ChainGun upgrade
+  ; Patch104p @bugfix hanfield 01/09/2021 Added WeaponBonus parameters for Frenzy that nullify the bonus but allow it to color the Gattling Cannon.
+  WeaponBonus           = FRENZY_ONE DAMAGE 90%
+  WeaponBonus           = FRENZY_TWO DAMAGE 80%
+  WeaponBonus           = FRENZY_THREE DAMAGE 70%
   AntiAirborneVehicle   = No
   AntiAirborneInfantry  = No
   AntiSmallMissile      = No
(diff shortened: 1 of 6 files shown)


commy2:
Does this mean mobile Gattlings now fire faster due to Avenger blue laser?

alanblack166:
Should not be the case.

xezon:
I tested all factions. All units look good.

shot_20210901_204721_1

shot_20210901_204857_2

shot_20210901_205024_3

shot_20210901_205641_4

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -1196,6 +1196,10 @@ Weapon GattlingBuildingGun
   WeaponBonus           = CONTINUOUS_FIRE_MEAN RATE_OF_FIRE 200% ; When the object achieves this state, this weapon gets double the rate of fire
   WeaponBonus           = CONTINUOUS_FIRE_FAST RATE_OF_FIRE 300% ; This is not cumulative, so with Delay of 40, and values of 2 and 4 for these  bonuses, you shoot at (40, 20, 10)
   WeaponBonus           = PLAYER_UPGRADE DAMAGE 125%     ; ChainGun upgrade
+  ; Patch104p @bugfix hanfield 01/09/2021 Added WeaponBonus parameters for Frenzy that nullify the bonus but allow it to color the Gattling Cannon.
+  WeaponBonus           = FRENZY_ONE DAMAGE 90%
+  WeaponBonus           = FRENZY_TWO DAMAGE 80%
+  WeaponBonus           = FRENZY_THREE DAMAGE 70%
```
xezon, line 1202: Hmm. Are we sure this negates it correctly?
```
  WeaponBonus = FRENZY_ONE    DAMAGE 110% ; Granted by a China Generals power
  WeaponBonus = FRENZY_TWO    DAMAGE 120%
  WeaponBonus = FRENZY_THREE  DAMAGE 130%
```
VS
```
  WeaponBonus           = FRENZY_ONE DAMAGE 90%
  WeaponBonus           = FRENZY_TWO DAMAGE 80%
  WeaponBonus           = FRENZY_THREE DAMAGE 70%
```

It may be calculated like so:
```
1,1 * 0,9 = 0,99 = 99%
1,2 * 0,8 = 0,96 = 96%
1,3 * 0,7 = 0,91 = 91%
```

Right now we count on it being calculated like so:
```
(1,1 + 0,9) / 2 = 1,00 = 100%
(1,2 + 0,8) / 2 = 1,00 = 100%
(1,3 + 0,7) / 2 = 1,00 = 100%
```

But that is not clear (to me).

alanblack166:
The latter calculation is accurate. We use such means in ROTR for weapons that are not meant to gain any bonuses.

commy2:
The actual calculation works like this:

total = (bonus_1 - 1) + (bonus_2 - 1) + 1

So

bonus_1 = 110%
bonus_2 = 90%

=> total = 100%

bonus_1 = 110%
bonus_2 = 120%

=> total = 130%

which is, admittedly, counter intuitive.

PR #133, opened 2021-09-01
Fix: Snow models of Speaker Towers are now being rendered with proper HOUSECOLORs

xamorish:
Tested with all Chinese factions on multiple official snow maps. This appears to be an issue with how the developers compiled W3D models using an invalid HOUSECOLOR material containing a bitmap texture that is completely missing from the game.

Before patch:
b1
b2

After patch:
a1
a2


xezon:
Looks very good. Could you resolve the merge conflicts? The .bat file had its eol characters changed from LF to CR LF.

xamorish:
> 
> 
> Looks very good. Could you resolve the merge conflicts? The .bat file had its eol characters changed from LF to CR LF.

Thanks, done!

commy2:
This PR broke all Propaganda Tower SNOW map shadows. In 1.04, only Night map undamaged Tower is broken.

xezon:
Perhaps a result of using these broken w3d importers.

commy2:
That's what I think. There're more issues with the Speaker Towers. I think they are a good test case for that Blender script. Unfortunately, I will be very busy for the next two weekends.

commy2:
The models use Housecolor2_S.dds instead of Housecolor2.dds.

xezon:
@commy2 
Did you create a follow up change for this or is this yet pending?

PR #134, opened 2021-09-01
Fix: Missing RadarPriority on Bunkers and Propaganda Centers fixed

alanblack166:
NProject fix:  Bunkers and Propaganda Centers now have RadarPriority.


xezon:
What does this fix user facing?

alanblack166:
Defining these parameters makes sure that the objects properly display on the map and are filtered correctly for EVA alerts and events. A requested NProject fix that removes ambiguity.

commy2:
Nice find.

I'd also add this to Boss_ChinaBunker.

commy2 (approved):

xezon (approved):

PR #135, opened 2021-09-01
Fix: Standardised Toxin Tractor primary weapon minimum range

Stubbjax:
# Description

The minimum attack range for all toxin tractor primary weapons is now consistent. It did not make sense that the beta / gamma (blue / purple) variants had a minimum range of 10 while the standard weapon (green) has a minimum range of 0.

Old:

Weapon | Range | Min Range
-- | -- | --
ToxinTruckGun | 100 | 0
ToxinTruckGunPlusOne | 100 | 0
ToxinTruckGunPlusTwo | 100 | 0
ToxinTruckGunUpgraded | 100 | 10
ToxinTruckGunUpgradedPlusOne | 100 | 10
ToxinTruckGunUpgradedPlusTwo | 100 | 10
Chem_ToxinTruckGunGamma | 100 | 10
Chem_ToxinTruckGunGammaPlusOne | 100 | 10
Chem_ToxinTruckGunGammaPlusTwo | 100 | 10

New:

Weapon | Range | Min Range
-- | -- | --
ToxinTruckGun | 100 | 10
ToxinTruckGunPlusOne | 100 | 10
ToxinTruckGunPlusTwo | 100 | 10
ToxinTruckGunUpgraded | 100 | 10
ToxinTruckGunUpgradedPlusOne | 100 | 10
ToxinTruckGunUpgradedPlusTwo | 100 | 10
Chem_ToxinTruckGunGamma | 100 | 10
Chem_ToxinTruckGunGammaPlusOne | 100 | 10
Chem_ToxinTruckGunGammaPlusTwo | 100 | 10

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -3027,7 +3027,7 @@ Weapon ToxinTruckGun
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
-  MinimumAttackRange          = 0 ;20.0
+  MinimumAttackRange          = 10.0 ;20.0
   DamageType                  = POISON
   DeathType                   = POISONED
   WeaponSpeed                 = 600                     ;  dist/sec
@@ -3127,7 +3127,7 @@ Weapon ToxinTruckGunPlusOne
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
-  MinimumAttackRange          = 0 ;20.0
+  MinimumAttackRange          = 10.0 ;20.0
   DamageType                  = POISON
   DeathType                   = POISONED
   WeaponSpeed                 = 600                     ;  dist/sec
@@ -3227,7 +3227,7 @@ Weapon ToxinTruckGunPlusTwo
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
-  MinimumAttackRange          = 0 ;20.0
+  MinimumAttackRange          = 10.0 ;20.0
   DamageType                  = POISON
   DeathType                   = POISONED
   WeaponSpeed                 = 600                     ;  dist/sec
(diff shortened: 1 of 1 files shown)


xezon (commented):
Please add a comment behind each line with **Patch104p** keyword and a short description what was changed and why.

Stubbjax:
> 
> 
> Please add a comment behind each line with **Patch104p** keyword and a short description what was changed and why.

Ah, I was not aware of this process. I will append the comments. Is there a reason for doing it this way and not e.g. using Git Blame to inspect changes?

commy2:
Why not make all 0?

Stubbjax:
> 
> 
> Why not make all 0?

Why? Skilled players can use the minimum attack range to their advantage by molesting the tractor. Setting it to 0 would be removing this skill differentor.

Not only that, but it looks funky that the nozzle can spray in on itself.

xezon:
Documenting our changes will make it easy for everyone to search for them with any text editor.

xezon (approved):

Jundiyy:
I'm not sure but all set to 0 might be more normal for players. Suddenly not being able to attack close targets would feel weird imo.

And perhaps the reason for the difference was that the weaker Toxin was given a bit of an advantage by allowing it to target at 0. Whereas the stronger ones were given more damage but had to pay by having a minimum distance.

xezon:
It may be worthwhile to check how 0 compares to 10. And then see if we should tweak it smaller than 10.

Stubbjax:
> 
> 
> I'm not sure but all set to 0 might be more normal for players. Suddenly not being able to attack close targets would feel weird imo.

Does it feel weird when your RPG Troopers or Scorpions have to reposition when a target gets too close? 

Imagine a pro player's humvee is about to die to a toxin tractor. They micro their humvee to continuously drive close to the tractor to save their humvee from death and kill the tractor. There is an admirable level of skill and knowledge required to pull this off - a crowd-pleaser moment, if you will. By setting the minimum range to 0, you are effectively removing this skill differentiator. How is that beneficial for the game?

> 
> And perhaps the reason for the difference was that the weaker Toxin was given a bit of an advantage by allowing it to target at 0. Whereas the stronger ones were given more damage but had to pay by having a minimum distance.

If anything feels weird, it's that the minimum range is inconsistent and that the base toxin weapon can essentially spray inwards. It's not like different tractors have different ranges (which would be more acceptable) - it actually changes when upgraded (or becoming heroic, which gives the beta weapon). Not only that, but it's a one-off change between default and beta, with no further change between beta and gamma.

It also makes no sense to over-buff a stat and then try to compensate by debuffing another stat. Adding inconsistencies has to be done with careful consideration and make sense, as it is an additional interaction for players to learn. Increasing the minimum range once when upgrading toxin damage is not a logical inconsistency in my book.

Jundiyy:
I don't think it would be bad to make them all 10, or more than 0 so it doesn't spray on itself. You have a point of spraying inwards, of course the micro part is also nice for all Tractors that you have mentioned. Think this should be a good move.

The point I meant about the reasoning from EA was, was the original weapons all have 0 then anything that's upgraded gets 10. Which does make sense for why they would do that but I think what you have said is better.

commy2:
The min range changing is obviously not intentional. It's typical fat-fingering.

PR #136, opened 2021-09-01
Dummy weapons no longer deal damage, triggering *UnderAttack EVA events

xamorish:
Dummy weapons no longer trigger EVA *UnderAttack events while keeping their initial purpose.


commy2:
Conflicts with some of my future unit AI tweaks. I suggest we wait on this and then compare it with my solution.

xamorish:
> 
> 
> Conflicts with some of my future unit AI tweaks. I suggest we wait on this and then compare it with my solution.

Sure! Optionally, if your AI tweaks aren't being merged anytime soon, this PR could be somehow labeled for later review to prevent forgetting about it.

commy2:
"ongoing discussion" serves that purpose.

xezon:
I added new tag :)

xezon:
@xamorish 
commy2 closed it because his change attempts to fix the same issue. Your changes look a bit different. Your change touches more files. Could you please check if commy2's change covers everything necessary?

xezon:
commy2:
> My fix is superior as it

> 1) affects more units, e.g. Outpost and Bus (EVA message would play before rocket guys hit)
> 2) also removes the impact effect, which this PR does not

> Must say that only affecting NEUTRALS is clever though.

PR #157, opened 2021-09-03
Fix: Non-Vanilla USA Avengers Attempt To Retaliate With Their Blue Laser

commy2:
ZH 1.04

- The vUSA Avenger will never retaliate. The other Avengers will retaliate if the option is enabled.

After this patch:

- All Avengers behave the same. None of the Avengers will retaliate.

For the uninitiated:

"Retaliatiation" only happens if the respective option is enabled in the ingame OPTIONS menu. Retaliation is enabled by default. A unit that can retaliate will - if the option is enabled - automatically engage enemies !when attacked! while idle (meaning, standing and not in Guard mode).

This is bad for the Avenger, because it tries to mark the enemy that hit it with the blue laser, causing it to move closer and run into its death. The main purpose however is the passive PDL, which works best if the unit standing behind your other units.

For this reason EA disabled Retaliation for the Avenger, similar to how they disabled automatically engaging enemies in general for the ECM tank, however due to the rushed devolopment, that change never made it to the Laser, Airforce, Superweapon and Boss General variants of the Avenger.


xezon (approved):
Very nice fix.

PR #158, opened 2021-09-03
Fix: Detached Floating Object Orbits Avenger Model When Moving

commy2:
ZH 1.04:

- Part of the driver cabin mounted gun is linked to the wheels, thus leading to a spinning detatched pixel when moving.

With patch:

- Only the wheels turn when moving.


xezon (approved):
Approved. I think we should add the 2 original Avenger models to GameFilesOriginal folder for reference.

commy2:
Added the original models to GameFilesOriginal.

Note that there are already multiple Avenger models with the same filename and virtual file location in the base game installation. The one uploaded here is the one actually seen ingame and is located in the PatchZH.big file.

xezon (approved):

PR #160, opened 2021-09-03
Fix: Pilots no longer have Chemical Suits and Advanced Training cameo and no longer slow down

alanblack166:
Pilots never had Chemical Suits active, and activating them makes the game crash, so they had a cameo for something that was never there. Also, upon reaching heroic veterancy, they became slower. This fixes both issues.


commy2:
While you're at it, the Advanced Training upgrade icon makes no sense on this unit. While it can rank up by walking over veterancy crates, it will never earn EXP that can be multiplied by 2, because it has neither a weapon, nor can crush anything.

xezon (approved):

commy2 (approved):

PR #163, opened 2021-09-03
Fix: Inconsistent Suicide Ability Button Placement

commy2:
ZH 1.04

- Different unit types may or may not use the same slot for the Suicide ability, which means the button to Suicide may be missing when multiple units are selected.

After patch:

- The button is placed on the same slot for all units, so the player can always* suicide as long as only Demo General units are selected.

* The Terrorist and Bomb Truck do not get a Suicide button with this PR, thus meaning that these two still shadow the button.


commy2:
- [x] reminder to myself to fix the commit name when squasing

xezon (approved):
Excellent change!

PR #165, opened 2021-09-03
Fix: Inconsistent Evacuate Button Placement

commy2:
ZH 1.04

- The Evacute buttons on the Humvee and the Ambulance are not at the same position, which means the player can't mass evacuate units if a group of Humvees and Ambulances are selected at once.
- This is not an issue in CCG and was only introduced presumably when the Hellfire Drone was added and the column number was increased from 6 to 7 in ZH 1.0.
- The same happens for the Technicals and Battle Buses.
- The same happens for the Troop Crawlers and Outposts

After patch:

- Same slot is used for all units. No more shadowing.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -18,6 +18,9 @@
 ;+--+--+--+--+--+--+--+
 ;
 ;------------------------------------------------------------------------------
+
+; Patch104p @bugfix commy2 03/09/2021 Align evacuate buttons on same slot.
+
 ; Many things just want these three buttons.
 CommandSet GenericCommandSet
   11 = Command_AttackMove
@@ -365,8 +368,8 @@ CommandSet GLAVehicleBattleBusCommandSet
   6 = Command_TransportExit
   7 = Command_TransportExit
   8 = Command_TransportExit
+  9 = Command_Evacuate
   11 = Command_AttackMove
-  10 = Command_Evacuate
   13 = Command_Guard
   14 = Command_Stop
 End
@@ -384,8 +387,8 @@ CommandSet GLAVehicleTechnicalCommandSet
   3 = Command_TransportExit
   4 = Command_TransportExit
   5 = Command_TransportExit
+  9 = Command_Evacuate
   11 = Command_AttackMove
-  10 = Command_Evacuate
   13 = Command_Guard
   14 = Command_Stop
 End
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -4020,6 +4024,7 @@ CommandSet Infa_ChinaTroopCrawlerCommandSet
   8 = Command_TransportExit
   9 = Command_EmptyCrawler
   11 = Command_AttackMove
+ ;13 = Command_Guard
```
xezon, line 4041: Why is this command removed here?

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -2231,9 +2235,9 @@ CommandSet Demo_GLAVehicleTechnicalCommandSetUpgrade
   3 = Command_TransportExit
   4 = Command_TransportExit
   5 = Command_TransportExit
-  9  = Demo_Command_TertiarySuicide
+  9 = Command_Evacuate
   11 = Command_AttackMove
-  10 = Command_Evacuate
+  12 = Demo_Command_TertiarySuicide
```
xezon, line 2240: Is `Demo_Command_TertiarySuicide` the demo upgrade suicide ability? I see it has overall very wild number assignments. Is this a problem?

commy2: Yes. The original commit comes from my own project and was cherry picked (rather, manually patch ported and `apply`'d).

It comes after https://github.com/xezon/GeneralsGamePatch/pull/163 in my commit history. That commit aligns the suicide button. If that commit is merged first, I will likely have to solve some merge conflict and this change should disappear.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
commy2: Not removed. Just added a commented line. The Troop Crawler is missing the Guard button, which is a similar issue, but out of scope for this PR. I was lazy in the other project and did both things in one commit. Don't want to argue for the Guard button now, but I left the changes in, at least as comments, so they can be added later if desired. I can remove them completely if needed.

The same change is in the `ChinaTroopCrawlerCommandSet`. As it stands now, it's a comment and not actual code that is recognized by the game.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
xezon: Ah ok. Ups I already merged some of your changes now. I think it is better I let you merge because of the order.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
xezon: Ah right this is an added comment. I somehow interpreted this as commented out line. Ok I understand.

xezon (approved):

xezon:
Adding Controversial label for now, since it changes many button placements (for the better).

PR #168, opened 2021-09-03
Fix: Vanilla GLA Hijacker And Saboteur Can Ride Toxin Generals Combat Bike

commy2:
ZH 1.04:

- The Hijacker and Saboteur from the normal GLA faction can ride the bike of the Toxin General, but no other non-Toxin unit can.
- Same applies to the GLA02 campaign Bike

After patch:

- Only the units of the same sub-faction can ride their bikes.

Sadly one cannot make all units cross compatible with all bikes. That's because every bike can only carry 8 different unit types. This is the most sane solution.


xezon:
Hmm. Wouldn't be the expectation that any GLA faction infantry can board any GLA faction bike? Or would there be any weird problems resulting of that?

commy2:
You can only allow 8 different riders on any single Combat Bike, sadly. The most sane to me is to only allow the units from the same faction.

xezon (approved):

PR #171, opened 2021-09-03
Fix: Saboteur Body Disappears From Destroyed Combat Bike

commy2:
ZH 1.04

- When a Saboteur on a Combat Bike is destroyed, the Saboteur is deleted instead of thrown off the bike like every other unit. Note, this purely a visual effect as the thrown off body does not affect the environment / units.

After patch:

- The Saboteur is now ejected like any other unit.

Notes:

- `OCL_SaboteurFlyingOffBikeDeathStart` is referenced by most bikes already, but it was undefined, and thus ignored.
- The Demo General is the only one that is missing the effect entirely. It was added for completeness, because the rider slot 7 is otherwise functional.
- There is no animation specifically for the Saboteur for this. However, the Saboteur uses the same skeleton as the Worker (it is a Worker reskin), so the flying-off animation can be re-used from the Worker and fits perfectly.


xezon:
Very resourceful fix. Well done.

xezon (approved):

PR #173, opened 2021-09-03
Fix: Spy Drone Selected By Q-Shortcut

### LINKED ISSUE #40: USA Spy Drone is selectable by Q (Select All Units)
USA Spy Drone is selectable by Q (Select All Units)

Add `IGNORES_SELECT_ALL` to `KindOf`

commy2:
ZH 1.04

- The Spy Drone is selected by the Select All (Q) shortcut, even though it completely ignores any orders.

After patch:

- The Spy Drone is no longer selected by Q, but can still be selected by click or box.

- Closes: #40


xezon (approved):
Nice change!

PR #174, opened 2021-09-03
Fix: Black Lotus Seleced By Q-Shortcut

commy2:
Same idea as https://github.com/xezon/GeneralsGamePatch/pull/173

Not as obvious as the drone one. This idea was requested to me by multiple (2) people and at least one test person loved it after I explained it. It's not a fighting unit. And Hackers aren't selected either by Q.


commy2:
Note also that the player can specifically select their Hero with the H key. (Or was it Ctrl+H? I forgot...)

xezon:
I think the change is logical and sound, but I am a bit on the fence here. What if player got used to selecting his forces with Q button and expects his Lotus unit to follow the command? Lotus is a unit that can attack (disable) cars so having it run around the troops and tanks can make sense.

Madrageeeeee:
dont really have an opinion about this. Guess it makes sense in spammy games where you have a lotus in someones base and you want all your fighting units to attack / go to a certain point while keeping your lotus in the enemy base / behind enemy lines. 

On the other hand, not too big of an issue and preference, so would keep as is to change as little as possible.

xezon:
I am closing this now. There appears very little interest in this functionality. Also I have never read anyone requesting this in the forums. It is best to not mess with user keys here unnecessarily, because player might get confused if suddenly Lotus moves with all other units when using this shortcut.

xezon:
Jarmen Kell and Burton are also selectable with Q. So from that perspective looks also ok to me to leave as is.

PR #181, opened 2021-09-04
Fix: Spectre Gunship Selected By W-Shortcut

commy2:
ZH 1.04

- The Spectre Gunship is selected by using the Select Aircraft (W) shortcut. It probably shouldn't.

After patch:

- Pressing W will not select the Spectre.
- close https://github.com/xezon/GeneralsGamePatch/issues/148

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -2,6 +2,7 @@ Note: These are issue reports. The titles describe the problem. So the expectati
 
 GAME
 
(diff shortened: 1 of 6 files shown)


commy2:
Note: this is my only PR so far I haven't tested. Couldn't be arsed to get to rank 3 against AI. Blame [hanfield](https://github.com/xezon/GeneralsGamePatch/issues/148#issuecomment-912025992) if this doesn't work ;)

xezon:
We have a test map here that can be used to get 5 Star instantly:
https://github.com/xezon/GeneralsGamePatch/tree/main/Maps/TestRank

xezon (approved):

PR #186, opened 2021-09-04
Fix: Single Player Combat Bike Animation, Model, Sound, Etc Inconsistencies

commy2:
ZH 1.04

- Some single player only Combat Bikes do not use the usual rider animations, voice lines and sound effects.

After patch:

- All Combat Bikes look and behave the same, except for their rider on top.
- The non normal GLA Bikes where missing `DoorOpenTime`. It defaults to 0 anyway, but I added it for completeness.

Examples of fixed issues:

- When the normal GLA Bike with RPG on top (SP only unit, see GLA01 campaign mission) is having an Hijacker on top, the bomb of the Terrorist is shown.
- When the GLA02 Stealthed Combat Bike is selected, it plays voice lines of the Rocket Buggy.
- The same bike has no team colors, but is always the same color.


Patch104pZH/GameFilesEdited/Data/INI/Object/GLAVehicle.ini
```diff
@@ -2797,14 +3006,12 @@ Object GLAVehicleCombatBikeTerrorist
...
   VisionRange     = 180
   ShroudClearingRange = 300
-;  Remove - JMC 8/6/03 - This is not usable by the player but I can use it for any AI if the prereqs are gone.
-;  Prerequisites
-;    Object = GLAArmsDealer
-;    Object = GLAPalace
-;  End
+  Prerequisites
+    Object = GLAArmsDealer
+  End
```
commy2, line 3030: Not really necessary, just put there so the line delta between bikes in minimized. Since the Bike is build from the AD, the prerequesite is always there.

commy2:
- [x] rename when squasing

Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
```diff
@@ -21781,7 +21882,7 @@ Object Boss_VehicleCombatBikeTerrorist
     DamageFX        = TankDamageFX
   End
   BuildCost       = 500
-  BuildTime       = 8.0          ;in seconds
+  BuildTime       = 4.0          ;in seconds
```
xezon: This change looks as if it does not belong here. Or?

xezon (approved):

PR #190, opened 2021-09-04
Fix: Non-Vanilla USA Fire Bases Use Patriot Sound Effect When Selected

commy2:
ZH 1.04

- The Airforce, Super Weapon and Laser General's Fire Bases use the sound effect of the Patriot Battery instead of the unique sound effect of the normal USA Fire Base.

After this patch:

- All Fire Bases use their intended and unqiue sound effect when selected.


commy2:
- [x] rename commit when squashing

xezon (approved):

PR #193, opened 2021-09-04
Fix: Some Chinooks Collect Slower Than Others

commy2:
ZH 1.04

- The Super Weapon and Laser General's Chinook, as well as the Combat Chinook (but not the normal AFG Chinook) take about 0.4 seconds longer to collect and drop a supply load.

After this patch:

- All Chinooks collect with the same speed.


xezon:
I agree it should be consistent, but why not adjust the USA Chinook to all others? This way only one faction would be touched instead of multiple.

commy2:
I do not know which way is preferable. I went with the vUSA one since that seems to be the intended values, since that ini file contains the old values, but commented out and overwritten with the new ones.

It honestly makes not much difference in practice.

Kesanov:
Is this really bug? To me it reads like AFG having super chinooks that collect faster. 

However, it's weird that it isn't mentioned in tooltip and that all chinooks use the same model & name.

xezon:
It is a difference of 100 ms. This is one-tenth of a second. Not a big impact.

commy2:
> Is this really bug? To me it reads like AFG having super chinooks that collect faster.

Yes. AFG Super Chinooks does not explain why normal USA gets the same values.

> It is a difference of 100 ms. This is one-tenth of a second. Not a big impact.

This. To anyone that fears or hopes this changes balance in either direction: it won't. The change is too miniscule. There is more variance due to building placement. **I suggest before anyone argues otherwise, to actually place two of these Supply Stashes ingame and compare collection time. There is a reason essentially no one knew about this.**

xezon (approved):

PR #205, opened 2021-09-04
Fix: Tomahawk Launchers Have Battle And Scout Drone Upgrade Icon Reversed

commy2:
ZH 1.04

- All Tomahawk Launhers have the upgrade icons for Scout Drone and Battle Drone reversed compared to any other USA vehicle.
- The Boss General's Tomahawk Launcher has the upgrade icons for Hellfire Drone and Advanced Training reversed.

After:

- All USA sub-faction vehicles have the drone upgrade icons in the same order.

Not really a bug, but it pisses me off. Other cameo fixes will come soonTM


xezon (approved):
Nice attention to details. Never noticed this.

PR #207, opened 2021-09-04
Fix: Scud Launchers Missing Upgrade Icon For AP Rockets

commy2:
ZH 1.04

- The SCUD Launchers are missing the upgrade icon for AP Rockets, despite them getting the damage bonus for high explosive warheads (though not for Anthrax warheads).

After patch:

- The SCUD Launchers have all their upgrade icons, in a order that makes sense compared to other GLA vehicles, and consistent across sub-factions.

Note that this was marked as MAYBE. This seems to me to need fixing though, because the bonus is real, and people should know it is.

As for public relations, we should make clear that this is not a change to the SCUD Launcher - making it benefit from AP Rockets. It already does in 1.04. All we changed was the missing icon.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -17749,9 +17749,11 @@ Object Chem_GLAVehicleScudLauncher
   SelectPortrait         = SUScudLauncher_L
   ButtonImage            = SUScudLauncher
 
-  UpgradeCameo1 = Upgrade_GLAJunkRepair
+  ; Patch104p @bugfix commy2 04/09/2021 Added missing upgrade icon(s).
+
+  ;UpgradeCameo1 = Upgrade_GLAAPRockets
   UpgradeCameo2 = Chem_Upgrade_GLAAnthraxGamma
-  ;UpgradeCameo3 = NONE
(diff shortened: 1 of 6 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
```diff
@@ -17749,9 +17749,11 @@ Object Chem_GLAVehicleScudLauncher
   SelectPortrait         = SUScudLauncher_L
   ButtonImage            = SUScudLauncher
 
-  UpgradeCameo1 = Upgrade_GLAJunkRepair
+  ; Patch104p @bugfix commy2 04/09/2021 Added missing upgrade icon(s).
+
+  ;UpgradeCameo1 = Upgrade_GLAAPRockets
```
xezon, line 17754: This line is commented out.

commy2: The Tox General Scud Launcher can only use Anthrax Warheads. These do not get the bonus for AP Rockets. I added the commented line to signify that this slot belongs to AP Rockets on other SCUDS, and that the upgrade is not shown intentionally. 

The slot 1 is intentionally left blank so that Junk Repairs and Anthrax occupy the same position as on the other sub-faction SCUD Launchers.

xezon (approved):

PR #208, opened 2021-09-04
Fix GLA Terrorist passenger crash

alanblack166:
Terrorists crashed the game in two scenarios. This commit fixes the crash in both instances.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -13019,8 +13019,10 @@ Object Chem_GLAInfantryTerrorist
 
 
   ; I have just pulled my ripcord, and this ain't no parachute!
-  Behavior = SlowDeathBehavior ModuleTag_Death03
-    DeathTypes          = NONE +SUICIDED
+  ; Patch104p @bugfix hanfield 05/09/2021 Added MASKED status as a requirement for a death without fireworks. This one is used inside of Technicals and Battle Buses.
+  Behavior = SlowDeathBehavior ModuleTag_ContainedDeath_Patch104p
+    DeathTypes          = NONE +SUICIDED +BURNED +EXPLODED +CRUSHED +SPLATTED +LASERED
+    RequiredStatus      = MASKED
     SinkDelay           = 3000
     SinkRate            = 0.5     ; in Dist/Sec
     DestructionDelay    = 8000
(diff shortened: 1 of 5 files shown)


commy2:
This is WIP.

xezon:
What is missing?

alanblack166:
Need to finish the other generals.

commy2:
Also currently the added deaths are handled by two SlowDeath modules. See ModuleTag_Death01. This will make the game chose one at random.

xezon:
Crash is still reproducible.

xezon:
Potentially obsolete by change
* #699

However, does this change still have use? Test if suicide in this branch behaves any better.

commy2:
If the bug is fixed, this PR is superseded.

xezon:
I tested.

I am unable to observe any difference in damage output of Original vs Hanfield change. Both Terror in Technical and Battle Bus incur big damage on blow up. Behaviour on China mines also looks identical. The only difference I can observe are the visuals during blow up, where in this change the terrorists and their particles are visible during the death of the vehicle, which looks visually much more impressive. If we care for this visual, then this change seems cool.

## Original Bang bus blow up

shot_20220720_162908_1

## Original Bang bus aftermath

shot_20220720_162909_2

## New Bang bus blow up

shot_20220720_162742_1

## New Bang bus aftermath

shot_20220720_162744_2

xezon:
I will wrap up this change some time soon. Change is nice.

xezon:
I now tweaked the change as much as I can and tested Terrorist killing, crushing and driving over mines. All looked good. Crushing Terrorist in Vehicle will just kill it without them exploding in both Original and Patch. Sometimes the Terrorist just evacuate and survive when driving over mines, but that also happens in both Original and Patch. I did not apply the changes to non Skirmish Factions, because with GC_Chem the setup would be a nightmare.

commy2:
So what exactly changed here? From my understanding, the CTD was already fixed by removing the bugged suicide weapon from the Tox Terrorist.

xezon:
The only thing I see changed here is what you can see in the images above. The terrorists and their effects are visible when their host vehicle dies. Other than that it appears to behaves the same.

commy2:
I see. So this is an entirely visual change. Pull request title should reflect that.

xezon:
Done.

commy2:
DestroyRidersWhoAreNotFreeToExit  is added to vGLA Bus, but not the others.

alanblack166:
Maybe because the commit is not final?

xezon:
> DestroyRidersWhoAreNotFreeToExit is added to vGLA Bus, but not the others.

Oh right, I forgot. Will add them :)

xezon:
After a bit more conversation on Discord about it, we figured proper fix for the original crash issue. This fix now no longer crashes the game with GLA Terrorists in specific scenarios.

xezon:
I had to revert the changes to GC_Chem because the crashed the game on load.

xezon:
Maybe close. Not sure.

PR #209, opened 2021-09-04
Nuclear Missile exploit fix

alanblack166:
This commit should fix the Nuclear Missile exploit.


commy2 (approved):

xezon (approved):

xezon:
alan says "should" be fixed. Meaning we should still verify the fix, no? I can test later.

alanblack166:
Yes, best to test it. It works on my end, but it's always good to double check.

commy2:
Asuming this works, same shoul be added to:
SupW_NeutronMissile and CruiseMissile

xezon:
Tested. Fix works.

commy2 (approved):
👍

PR #222, opened 2021-09-05
Revert: 1.04+ Composite Armor Tweak

commy2:
- ref https://github.com/xezon/GeneralsGamePatch/pull/213

This PR reverts the Composite Armor tweak from 1.04+ which this project inherited. The tweak seems to be either undesirable or controversial, so ___the original values from 1.04 are now restored.___

This tweak also sort of fixes the bug where Composite Armor is applied twice on the vanilla USA Crusader as well as the Laser Tank (but not the SP only SWG Crusader). Instead double the bonus is applied once, which is identical ingame behaviour, but more explicit code.


Jundiyy:
Yes, this seems better.

xezon (approved):

PR #234, opened 2021-09-06
Scud Storm launch can no longer be cancelled by accident

alanblack166:
Scud Storm, if given a target or ordered to force fire while launching, would try to swap to a non-existent primary weapon and thereby stop the launch. This commit fixes that.


xezon (dismissed):

commy2:
This is great. Didn't know about that issue.

I also did the Scud-bug fix for the SP only SCUDs.

These can appear in scripted maps. So I think we should add the same fix to these objects:
GC_Chem_GLAScudStorm
GC_Slth_GLAScudStorm
Boss_ScudStorm

commy2:
Seems to re-introduce the Scud-bug.

xezon (commented):
Rereview.

PR #245, opened 2021-09-08
Add test map for all factions

MTKing4:
A map with all factions and sub-factions


xezon (changes requested):
Please add a better title for this change.
Also please use "Test" prefix for test map and pick a name that makes its intent descriptive. It is not clear what `JackOfAllTrades` means.

MTKing4:
i changed the title, i could make the map name with the same name? `MapWithAllFactions`

xezon:
How about `TestAllFactions` ?

MTKing4 (commented):
fixed

xezon (approved):

PR #255, opened 2021-09-09
Fix: Mixing Battle Masters From Different Sub-Factions Does Not Grant Horde Bonus

commy2:
ZH 1.04

- 4 Nuke Battle Masters and 1 Tank Battle Master do not count as a Horde, but 4 Nuke Tank Hunters and 1 Tank Tank Hunter do. 
- The unit type has to match exactly instead of having any of 5 hordable tanks\vehicles.

After ZH 1.04

- Any 5 Battle Masters will form a horde, no matter what sub-faction they are from.

I think this is a thing, because in CCG only one vehicle with horde existed, while there always were two different infantry units.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
@@ -2798,13 +2798,15 @@ Object Tank_ChinaTankBattleMaster
     ScienceRequired = SCIENCE_BattlemasterTraining
   End
 
+  ; Patch104p @bugfix commy2 09/09/2021 Fixed horde bonus required exact match when mixing sub-faction Battle Masters.
+
   Behavior = HordeUpdate ModuleTag_04
     RubOffRadius = 150    ; if I am this close to a real hordesman, I will get to be an honorary hordesman
     UpdateRate = 1000     ; how often to recheck horde status (msec)
     Radius = 75           ; how close other units must be to us to count towards our horde-ness (~30 feet or so)
     KindOf = VEHICLE      ; what KindOf's must match to count towards horde-ness
     AlliesOnly = Yes      ; do we only count allies towards horde status?
-    ExactMatch = Yes      ; do we only count units of our exact same type towards horde status? (overrides kindof)
+    ExactMatch = No       ; do we only count units of our exact same type towards horde status? (overrides kindof)
     Count = 5             ; how many units must be within Radius to grant us horde-ness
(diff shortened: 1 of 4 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
```diff
@@ -2798,13 +2798,15 @@ Object Tank_ChinaTankBattleMaster
...
+  ; Patch104p @bugfix commy2 09/09/2021 Fixed horde bonus required exact match when mixing sub-faction Battle Masters.
+
   Behavior = HordeUpdate ModuleTag_04
     RubOffRadius = 150    ; if I am this close to a real hordesman, I will get to be an honorary hordesman
     UpdateRate = 1000     ; how often to recheck horde status (msec)
     Radius = 75           ; how close other units must be to us to count towards our horde-ness (~30 feet or so)
     KindOf = VEHICLE      ; what KindOf's must match to count towards horde-ness
     AlliesOnly = Yes      ; do we only count allies towards horde status?
-    ExactMatch = Yes      ; do we only count units of our exact same type towards horde status? (overrides kindof)
+    ExactMatch = No       ; do we only count units of our exact same type towards horde status? (overrides kindof)
```
xezon, line 2809: How does this know that another faction Battlemaster is ok for horde but a Gattling Tank is not?

commy2: None of the Gattling tanks have a HordeUpdate module. They would need a HordeUpdate module, that is also set to `ExactMatch = No`.

commy2: For example, 5 Hackers and a Red Guard do not grant Horde for the Red Guard either.

xezon: Hmm. And Tank Hunters? Will 2 Battlemaster Horde with 3 Tank Hunters?

commy2: Nope. The HordeModule for the Red Guard and Tank Hunter only counts for `KindOf = INFANTRY`.
The HordeModule for the Battle Master(s) however only counts for `KindOf = VEHICLE`.
They do not count cross type.

xezon: Ok that sounds very reasonable then.

xezon (approved):

PR #256, opened 2021-09-09
Fix: Sentry Drone Lacks Voice Line When Leaving Factory

commy2:
ZH 1.04

- The Sentry Drone is so ashamed of its poor performance, it's the only unit in the game that does not announce itself when constructed.

After patch:

- beep beep beep boop
- The sound effect is only played for the player, not for the enemy or allies. It works like every other announcer voice line.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/SoundEffects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/SoundEffects.ini
@@ -6800,6 +6800,17 @@ End
 ;  Type = ui voice player
 ;End
 
+; Patch104p @bugfix commy2 09/09/2021 Sound used when Sentry Drone leaves factory.
+
+AudioEvent SentryDroneVoiceCreate
+  Sounds = vsenseld
+  Control = random
+  Volume  = 80
+  MinVolume = 70
+  Priority = high
+  Type = world global player
+End
+
 AudioEvent SentryDroneVoiceSelect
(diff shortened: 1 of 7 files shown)


Patch104pZH/GameFilesEdited/Data/INI/SoundEffects.ini
```diff
@@ -6800,6 +6800,17 @@ End
 ;  Type = ui voice player
 ;End
 
+; Patch104p @bugfix commy2 09/09/2021 Sound used when Sentry Drone leaves factory.
+
+AudioEvent SentryDroneVoiceCreate
+  Sounds = vsenseld
+  Control = random
+  Volume  = 80
+  MinVolume = 70
```
xezon, line 6809: How did you pick these volumes? I am asking because for `VoiceCreate` sounds of other units I mostly see these values:
```
  Volume  = 110
  MinVolume = 100
```

commy2: Most units use 90 db for "orders" and 100-110 db for "announcing". The Sentry Drone uses 60 db for orders, so I made creation also 30 db quiter. It's probably a loud sample or something. It feels jarring if the unit is much louder when created compared to when used.

xezon: Ok. If you say it sounds proper in game with these values, then its all good.

xezon (approved):

PR #260, opened 2021-09-09
Fix: Demo General Terrorist On Bike Missing Red Smoke Effect After Demolitions Upgrade

commy2:
ZH 1.04

- The Demo General's Terrorist does not use the red explosion effects when destroyed or suicided while riding a Bike after being upgraded with Demolitions, but the explosion effects of a normal Terrorist instead.

With patch:

- Correct effect is used. Same as when not riding the Bike.


commy2:
Note: This is purely cosmetical. No damage is changed.

xezon (approved):

PR #261, opened 2021-09-09
Fix: Listening Outpost Damaged Smoke Effect Sticks After Repair

commy2:
ZH 1.04

- Once a Listening Outpost was in its damaged visual state, it will continuously spawn smoke particles even after it has been fully repaired.
- Additionally, these smoke particles are visible to enemy players even while the Outpost is stealthed.

After patch:

- The smoke effect will work as intended: It will be removed once the Outpost is repaired. It will therefore no longer reveal hidden repaired Outposts with smoke columns coming out of nowhere.
- The smoke columns will still be visible while damaged. This happens for all units and buildings.


xezon (changes requested):
W3D files are no longer automatically added to a .big file by the scripts. We need to explicitly specify their target in either

`GeneralsGamePatch\Patch104pZH\Scripts\MAKE_Patch104pArtZH.bat`
or
`GeneralsGamePatch\Patch104pZH\Scripts\MAKE_Patch104pZH.bat` (Core)

If the *.W3D files are optional, then you can put it into the first script.

xezon:
What is changed in the W3D files?

commy2:
> What is changed in the W3D files?

SMOKExx changed to SMOKESxx

The "S" is for "small". The current system is normally used for buildings. It does not work properly for units: When repaired, the effect stays.

In the damaged model I also added SMOKES03 at the exact voxel position where it is used in the normal model. Makes no sense that less smoke is used in the damaged model.

I also changed FIES01 to FIRES01, because it's a typo. With that change, a small fire is added to the center of the bigger smoke column, which looks neat if you notice it.

xezon:
Hmm ok, I am not sure what these names refer to. Be aware that changing Bone names can or will cause version mismatch. W3D files that affect client compatibility need to go into the `Patch104pZH.big` (Core).

commy2:
> I am not sure what these names refer to.

SMOKExx is just a bone that can be used in the TransitionDamageFX module. Arbitrary name.

SMOKESxx, SMOKEMxx, SMOKELxx, FIRESxx, FIREMxx, FIRELxx are linked in [GameData.ini](https://github.com/xezon/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesOriginal/Data/INI/GameData.ini#L142-L162) to automatically apply smoke and fire effects to damaged units. They are the names used by every other unit model in the game.

> Be aware that changing Bone names can or will cause version mismatch.

This is true, however, this is not an optional model. Is this not the right .bat script? Because the Avenger fix should also be a mandatory model (I see no advantage in somebody staying with the broken canon-wheel thing).

xezon:
If it is not an optional Art change, it must go into `MAKE_Patch104pZH.bat`.
If Avengers are not optional either, then they also must go into `MAKE_Patch104pZH.bat`.

An optional change is if it is just visual or audio, that causes no version mismatch between clients and does not block any functionality.

xezon (approved):

xezon:
Removing Controversial tag because it is not.

Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
```diff
@@ -15595,7 +15595,6 @@ Object Infa_ChinaVehicleTroopCrawler
   End
 
   Behavior                 = TransitionDamageFX ModuleTag_08
-    ReallyDamagedParticleSystem1 = Bone:Smoke RandomBone:Yes PSys:SmallLightSmokeColumn
```
xezon, line 15598: @commy2
Here you removed it from Troop Crawler instead of Outpost. Was this mistake?

xezon: Follow up at #1774

PR #262, opened 2021-09-09
Fix: Nuke Cannons Start With Neutron Shells

commy2:
- There is an exploit that can be prevented by having the Nuke Cannons upgrade their weapon set with the Neutron Shells upgrade instead of just enabling the button.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -1560,8 +1560,15 @@ Object ChinaVehicleNukeLauncher
   Side = China
   EditorSorting   = VEHICLE
   TransportSlotCount = 10                 ;how many "slots" we take in a transport (0 == not transportable)
+
+  ; Patch104p @bugfix commy2 09/09/2021 Prevent exploit that allows Neutron Shells without upgrade.
+
   WeaponSet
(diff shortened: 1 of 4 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
```diff
@@ -1699,6 +1706,12 @@ Object ChinaVehicleNukeLauncher
     MaxQueueEntries = 1; So you can't build multiple upgrades in the same frame
   End
 
+  ; Patch104p @bugfix commy2 09/09/2021 Prevent exploit that allows Neutron Shells without upgrade.
+
+  Behavior = WeaponSetUpgrade ModuleTag_13
```
xezon, line 1711: I am always super worried when I see `ModuleTag_xx` because there is always the chance of making a mistake and using the same module name in an object, right?

I wonder if it would be beneficial to prefix these with `Patch104pModuleTag`, this way there is less chance of making mistake.

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
commy2: The game would crash when you start it, so I don't think there is much worry.

PR #267, opened 2021-09-10
Fix: Battle Bus With Passengers Freezes When Ordered To Attack Out Of Range Airborne Targets

commy2:
ZH 1.04

- The dummy weapon that allows the Battle Busses passangers to engage airborne targets also makes the Battle Bus freeze long before the Bus is in range when targeting airborne units.

After patch:

- When clicking on an enemy airborne target (e.g. Chinook), the Battle Bus will move into range, so that the troops inside can actually fire.

Note: `BattleBusDummyWeapon` is only used against airborne targets, despite the name being unclear about that.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -8467,7 +8467,7 @@ End
 Weapon BattleBusDummyWeapon
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 320.0
+  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range targets.
   DamageType = EXPLOSION
   DeathType = EXPLODED
   WeaponSpeed = 999999.0
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -8467,7 +8467,7 @@ End
 Weapon BattleBusDummyWeapon
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range airborne targets.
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8467,7 +8467,7 @@ End
 Weapon BattleBusDummyWeapon
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 320.0
+  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range targets.
```
commy2:
```suggestion
  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range airborne targets.
```

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8467,7 +8467,7 @@ End
 Weapon BattleBusDummyWeapon
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 320.0
+  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range airborne targets.
```
xezon: Same as the other change. How did this 150 value came to be?

commy2: See: https://github.com/xezon/GeneralsGamePatch/pull/270#discussion_r706317490

xezon: See: https://github.com/xezon/GeneralsGamePatch/pull/270#discussion_r706321588

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2:
```suggestion
  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Bus freezing when ordered to attack out of range airborne targets. 150.0 was determined to be the highest range where missile infantry passengers reliably engaged the target.
```

PR #268, opened 2021-09-10
Fix: Battle Bus Turns Chassis When Loaded Troops Are Ordered To Engage Airborne Targets

commy2:
ZH 1.04

- When the Battle Bus with infantry inside is ordered to attack a flying unit, the whole Bus turns towards the target needlessly.

After patch:

- The Bus will order passengers to attack the airborne target and not turn the chassis pointing towards it.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -8485,6 +8485,7 @@ Weapon BattleBusDummyWeapon
   AntiAirborneVehicle = Yes
   AntiAirborneInfantry = Yes
   AntiGround = No
+  AcceptableAimDelta = 180  ; Patch104p @bugfix commy2 10/09/2021 Fix Bus turning when ordered to attack airborne targets.
 End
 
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -690,6 +690,7 @@ Weapon HumveeMissileWeaponAirDummy
   AntiAirborneVehicle = Yes
   AntiAirborneInfantry = Yes
   AntiGround = No
+  AcceptableAimDelta = 180  ; Patch104p @bugfix commy2 10/09/2021 Fix Bus turning when ordered to attack airborne targets.
```
xezon: Is this meant to be under `HumveeMissileWeaponAirDummy` ?

commy2: Kind of, but no,. I think I made a boo boo. Will sort it out.

commy2: Fixed. Was error while cherry picking.

xezon (approved):

PR #270, opened 2021-09-10
Fix: Humvees Without TOW Missiles Upgrade Freeze When Attack Moving Into Airborne Targets

commy2:
ZH 1.04

- The dummy weapon that allows the Humvee's passangers to engage airborne targets also makes the Humvee freeze when attack moving into a group of units with airborne targets i.e. other Humvees with Scout Drones.
- This is basically the same issue as https://github.com/xezon/GeneralsGamePatch/pull/267

After patch:

- The Humvee will move into range when ordered to attack enemy airborne targets, such that the passenger can fire at the target.
- This does not change anything about the Humvee after the TOW-Missiles upgrade, as that one loses access to this dummy weapon.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -677,7 +677,7 @@ End
 Weapon HumveeMissileWeaponAirDummy
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Humvee freezing when ordered to attack out of range airborne targets.
+  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Humvee freezing when ordered to attack out of range airborne targets. 150.0 was determined to be the highest range where missile infantry passengers reliably engaged the target.
   DamageType = EXPLOSION
   DeathType = EXPLODED
   WeaponSpeed = 999999.0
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -677,7 +677,7 @@ End
 Weapon HumveeMissileWeaponAirDummy
   PrimaryDamage = 0.0001
   PrimaryDamageRadius = 0.0
-  AttackRange = 320.0
+  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Humvee freezing when ordered to attack out of range airborne targets.
```
xezon: How did this value of 150 came to be? Is there a rationale or just approximated by testing?

commy2: The range of missile infantry is 170 (Tunnel Defender, Missile Defender, Tank Hunter). But if you set the range at exactly 170 as well, the units may still not shoot, because the fire ports (memory points in the Humvee etc. models) are not exactly at the centrer of the model.

I first had this at 160 in my project, but I noticted that this still made not all units fire. 150 seemed to be the largest round number were all passengers reliably engage.

xezon: Maybe this information can be reflected in the comment. When someone else looks at this value, he would not understand if 150 made sense or not.

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2:
```suggestion
  AttackRange = 150.0 ; Patch104p @bugfix commy2 10/09/2021 Fix Humvee freezing when ordered to attack out of range airborne targets. 150.0 was determined to be the highest range where missile infantry passengers reliably engaged the target.
```

PR #271, opened 2021-09-10
Fix: Fake Command Centers Lose Sub-Faction Decal After Fortified Structures Upgrade

commy2:
ZH 1.04

- When upgraded with Fortified Structures, the Fake Command Centers of the Stealth, Tox and Demo General lose their sub-faction decal. They look like a regular GLA Command Center with Fortified Structures.

After patch:

- The Fake Command Centers will keep their sub-faction decal after being upgraded with Fortified Structures.


xezon (approved):
I hope you did not get a brain aneurysm when working on these names.

commy2:
lol
This is clearly written by the old guard that got used to 64 byte var names.

PR #276, opened 2021-09-10
Fix: GPS Scrambled Rangers And Red Guards Remain Stealthed While Capturing Buildings; Same For Non-Infantry General Hackers Disabling Buildings

commy2:
ZH 1.04

- When a Rebel is stealthed by the GPS Scrambler, he is revealed while capturing a building. Not so a Ranger or Red Guard (or Mini-Gunner).
- When a non-Infantry General Hacker is stealthed by the GPS Scrambler, he remains invisible while haking/disabling a building.

After patch:

- Rangers, Red Guards and Minigunners are revealed when capturing buildings, just like Rebels are.
- All Hackers are revealed when hacking a building while stealthed.

This does not affect internet hacking in any way.


Patch104pZH/GameFilesEdited/Data/INI/Default/Object.ini
```diff
@@ -105,11 +105,14 @@ Object DefaultThingTemplate
...
   End
+
+  ; Patch104p @bugfix commy2 10/09/2021 Fix GPS-scrambled non-Rebel infantry not being revealed while capturing buildings.
+
   ; OverrideableByLikeKind means "this default module instance will be replaced by any of the exact same class"
   OverrideableByLikeKind
     Behavior = StealthUpdate ModuleTag_DefaultStealthUpdate
       StealthDelay                = 2000 ; msec
-      StealthForbiddenConditions  = FIRING_PRIMARY FIRING_SECONDARY FIRING_TERTIARY RIDERS_ATTACKING
+      StealthForbiddenConditions  = FIRING_PRIMARY FIRING_SECONDARY FIRING_TERTIARY RIDERS_ATTACKING USING_ABILITY
```
xezon: I do wonder if this change may have unplanned side effects. For example, could a GPS'ed China Hacker suddenly become unstealthed if he uses internet hack ability?

commy2: No. If there is any StealthUpdate module on the unit, this ModuleTag_DefaultStealthUpdate module is overwritten. That's why it's encapsulated in OverrideableByLikeKind.

xezon: But the V China hacker has no stealth module. So GPS'ed hacker may be stealth, but with this change not anymore. I can try to test this later.

commy2: I vaguely remember testing this earlier in the year. I will double check though.

commy2: I double checked this. Hacking ability does not set the StatusBit `USING_ABILITY`. Therefore, GPS Scrambled Hackers will not be revealed.

xezon: I suspect we need to check all unit abilities though?

commy2: Laser Lock also sets FIRING_SECONDARY already. Hmm, Saboteur deletes himself. Mine clearing worker is FIRING_PRIMARY. Can't really think of anything else.

commy2:
Alternate approach where Rangers have their own StealthUpdate module that reveals them when capturing buildings.

Testing that. If it works, do the same for Red Guard and Hacker.

Jundiyy:
> Alternate approach where Rangers have their own StealthUpdate module that reveals them when capturing buildings.
> 
> Testing that. If it works, do the same for Red Guard and Hacker.

I think this would be easier and more simple to test, rather than having to go through all units.

commy2:
Done that.

Madrageeeeee:
this makes sense to change

xezon:
@commy2 
Are you confident this works correctly for the units and has no side effects, or would it be better to have MisterGerbol test it more closely?

commy2:
Since the current iteration of the fix is applied directly to Rangers, Red Guards and Hackers, I do not have any concerns. Internet Hack was tested and does not grant USING_ABILITY, thus is not a problem. No other abilities are used by the mentioned units.

The worst would be that we missed some unit with ability. I can't think of any though, and even if we did, we can fix them later as well.

xezon (approved):

PR #279, opened 2021-09-10
Fix: Dummy Weapons Cause Notifications And Hit Effects

commy2:
ZH 1.04:

- Objects with dummy weapons do a tiny amount of damage that triggers the "we're under attack!" notifications, and cause hit effects (sparks and impact sounds).

After patch:

- No more notifications or impact effects / sounds by dummy weapons.

Repro (one notorious example):

- Place a Sneak Attack. Force-fire with the Sneak Attack on a nearby unit.

Note:

- more thorough implementation of:
- close https://github.com/xezon/GeneralsGamePatch/pull/136

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -2860,3 +2860,46 @@ End
 
 
 
+;------------------------------------------------------------------------------
+Object DummyWeaponProjectile
+
+  ; ***DESIGN parameters ***
+  DisplayName      = OBJECT:RangerTeamMissile
+  EditorSorting   = SYSTEM
+  VisionRange = 0.0
+  ArmorSet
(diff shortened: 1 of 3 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -2860,3 +2860,46 @@ End
...
+    SubdualDamageCap = 200
+    SubdualDamageHealRate = 100000
+    SubdualDamageHealAmount = 50
+  End
+
+  Behavior = PhysicsBehavior ModuleTag_06
+    Mass = 1
+  End
+  Behavior = MissileAIUpdate ModuleTag_07
+    TryToFollowTarget               = Yes;;;;;;;;;No
```
xezon, line 2894: Can remove the comment.

xezon (approved):

PR #281, opened 2021-09-10
Add English Localization File

commy2:
This is the English 1.04 text. .STR file is easily editable and accepted by the game.

However we implement this, and changes in this file, it has to be optional, so we don't fuck up people's custom hotkeys.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/generals.str
+++ b/Patch104pZH/GameFilesEdited/Data/generals.str
@@ -1,0 +1,25688 @@
+GUI:GameOptions
+"GAME OPTIONS"
+END
+
+GUI:SinglePlayer
+"SOLO PLAY"
+END
+
+GUI:Network
+"NETWORK"
+END
+
+GUI:Options
+"OPTIONS"
+END
+
+GUI:Skirmish
+"SKIRMISH"
+END
+
+GUI:Back
+"BACK"
+END
+
+GUI:StartGame
+"PLAY GAME"
+END
+
+GUI:Color
+"Color"
+END
+
+GUI:Slower
+"Slower"
+END
+
+GUI:Faster
+"Faster"
+END
+
+GUI:Low
+"Low"
+END
+
+GUI:Medium
+"Medium"
+END
+
+GUI:High
+"High"
+END
+
+GUI:None
+"None"
+END
+
+GUI:Random
+"Random"
+END
+
+GUI:Cancel
+"CANCEL"
+END
+
+GUI:Ok
+"OK"
+END
+
+GUI:Delete
+"Remove"
+END
+
+GUI:Description
+"Description"
+END
+
+GUI:Load
+"LOAD"
+END
+
+GUI:Progress
+"Progress"
+END
+
+GUI:Assign
+"Assign"
+END
+
+GUI:CustomizeKeyboard
+"Customize Keyboard"
+END
+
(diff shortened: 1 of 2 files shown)


xezon:
What do you want to add that for? And does it load from inside .big file?

commy2:
> What do you want to add that for? 

So text edits can be made, which some of my tasks require.

> And does it load from inside .big file?

Pretty sure it does.

xezon:
How is this generals.str file generated? I am asking because typically we use generals.csf file for text edits. Also, we cannot just apply fixes to the English localization. Changes need to be applied to all language files. I can provide all .csf files except for Brazilian.

commy2:
> How is this generals.str file generated?

Exported it with FinalBig. Then used the CSFtoSTR(0.9.3).exe tool. Iirc I also had to clean up a few special characters. (Quote marks?)

> I can provide all .csf files except for Brazilian.

Would be awesome if you can taker over this and give contributors a way to make text edits.

xezon:
Tools here #306
I did not know STR was a thing. Usually game works with CSF file. There are editors for CSF file so we can use this native format. I don't know yet how to get hands on Brazilian which is a bit of a problem. How soon do you want to make edits to language files?

commy2:
> How soon do you want to make edits to language files?

Depends on how many tasks remain that do not involve text changes. I guess in two weeks would be nice, so no hurry.

> I did not know STR was a thing.

As far as I can tell, there is literally no reason to use CSF.

xezon:
There are a few considerations:
* Is STR format stable? No crashes on any language?
* Does STR format void any user CSF files? If yes that would be bad.
* Does STR load from within BIG file?
* Can STR contain optional strings that APPEND to the ones configured in CSF? If yes, then it would be detrimental to use STR format.

commy2:
> Does STR format void any user CSF files? If yes that would be bad.
> Can STR contain optional strings that APPEND to the ones configured in CSF? If yes, then it would be detrimental to use STR format.

Can CSF do that?

xezon:
Load order will pick the first copy of CSF it finds. In regards to STR, we would need to test and understand how it interacts with CSF file. I can do some tests and figure out what to do with it.

Generalcamo:
`Does STR format void any user CSF files? If yes that would be bad.`
It would. Mods use STR files as CSF files are very difficult to maintain long-term. But an STR file will override an existing CSF file.

xezon:
> It would. Mods use STR files as CSF files are very difficult to maintain long-term. But an STR file will override an existing CSF file.

Ok this on its own already voids STR approach.

commy2:
I was told this same problem applies to CSF files, because some people change their hotkeys in the CSF file inside the EnglishZH.big file. At least that was the reason given that no text changes at all are possible in the 1.04+ patch.

Is there a STR -> CSF compiler?

xezon:
We could write a STR to CSF command line tool and embed this within our build scripts, so that we can edit text files but it will be compiled into CSF automatically on build. This would be more comfortable for development. CSF file edited in EnglishZH.big is very uncommon I would say. The most common and likely scenario will be a extracted GENERALS.CSF, because that will take precedence over all other CSF files.

xezon:
I opened a task here and will take care of implementing it all.
* https://github.com/xezon/GeneralsGamePatch/issues/318

PR #285, opened 2021-09-11
Fix: Extended Range Exploit

commy2:
ZH 1.04

- It is possible to make Tow-Humvees, Quad Cannons and Gatling Tanks shoot with the extended range against air units while firing at ground units.

After patch:

- The AA weapons can no longer switch target to engage ground units.


xezon (approved):

MTKing4:
There's a similar situation that happens with inf helix, if garrisoned by a mini gunner and have the helix selected with another minigunner on the ground and shoot an air unit, the minigunner inside the lix will shoot the air unit with the same extended AA range

PR #286, opened 2021-09-11
Fix: Paradrop Ability Spawns Wrong Ranger Type

commy2:
- similar to https://github.com/xezon/GeneralsGamePatch/pull/73

ZH 1.04:

- Paradrop ability always spawns the Rangers of the normal USA faction, even as Air Force, Super Weapon or Laser General.

Repro:

- Build a Ranger as Air Force General. Use the Paradrop ability. Double click on a Ranger -> won't select both

After patch:

- The Rangers from the correct sub-faction are dropped via Paradrop, so they can be easily selected by double clicking when mixed with Barracks build Rangers or Rangers dropped from destroyed buildings.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
@@ -5301,6 +5301,341 @@ ObjectCreationList SUPERWEAPON_Paradrop3
   End
 End
 
+; Patch104p @bugfix commy2 11/09/2021 Lists used to spawn sub-faction specific Rangers via Paradrop ability.
+
+; -----------------------------------------------------------------------------
+; -----------------------------------------------------------------------------
+ObjectCreationList AirF_SUPERWEAPON_Paradrop1
+  DeliverPayload
+    Transport = AmericaJetCargoPlane
+    StartAtPreferredHeight = Yes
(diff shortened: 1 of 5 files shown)


Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
```diff
@@ -5301,6 +5301,341 @@ ObjectCreationList SUPERWEAPON_Paradrop3
...
+      Style             = SHADOW_ALPHA_DECAL
+      OpacityMin        = 25%
+      OpacityMax        = 50%
+      OpacityThrobTime  = 500
+      Color             = R:227 G:229 B:22 A:255
+      OnlyVisibleToOwningPlayer = Yes
+    End
+    ;RequiresLivePlayer = Yes ; I want to require a live player, but I'll settle for dropping neutral rangers instead
+  End
+  DeliverPayload
```
xezon, line 5391: Why does the Paradrop3 have 2 identical DeliverPayloads? It's 2 planes coming?

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
Generalcamo: Indeed, it's split between two planes.

PR #288, opened 2021-09-11
Fix: Super Weapon General Particle Cannon Uses Some Blue Effects

commy2:
ZH 1.04

- The Super Weapon General's Particle Cannon has some blue colored effects mixed in with the pink effects when ready to fire and when getting destroyed, when they all should be pink instead. This looks disgustingly unfinished and rushed.

After patch:

- Same effects are used, but they all are the correct color - the pink of the actual laser beam.


Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
```diff
@@ -55134,6 +55252,65 @@ ParticleSystem ParticleDeathFlare
   WindPingPongEndAngleMax = 0.000000
 End
 
+; Patch104p @bugfix commy2 11/09/2021 Recolored effect used when SWG's Particle Cannon is blowing up.
+
+ParticleSystem SupW_ParticleDeathFlare
```
xezon, line 55257: Where is this used? It's not used in this change.

Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
```diff
@@ -11846,6 +11846,64 @@ ParticleSystem ParticleDeathImplode
   WindPingPongEndAngleMax = 6.283185
 End
 
+; Patch104p @bugfix commy2 11/09/2021 Recolored effect used when SWG's Particle Cannon is blowing up.
+
+ParticleSystem SupW_ParticleDeathImplode
```
xezon, line 11851: Where is this used? It's not referenced in this change.

Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
```diff
@@ -9830,12 +9830,14 @@ Object SupW_AmericaParticleCannonUplink
...
 
+  ; Patch104p @bugfix commy2 11/09/2021 Use destruction effects tailored for the pink laser.
+
   Behavior = SlowDeathBehavior ModuleTag_18
     ; don't run this death if we are under construction...
     ExemptStatus   = UNDER_CONSTRUCTION
     DestructionDelay = 2000
-    FX             = INITIAL FX_ParticleUplinkDeathInitial
-    OCL            = INITIAL OCL_SDILinkLasers
+    FX             = INITIAL SupW_FX_ParticleUplinkDeathInitial
```
xezon, line 9845: This effect exists already? It's not added in this change.

xezon (commented):
Are the W3D and dds files new or modified assets? If they modify original ones, please add the corresponding original files to `GameFilesOriginal`.

commy2:
> Are the W3D and dds files new or modified assets?

Copy pasted from the original vanilla USA Particle Cannon, renamed and color edited.

Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
commy2: It's here: https://github.com/xezon/GeneralsGamePatch/pull/288/files#diff-758c48cb4371d9c579a06719a58844bfbcd625b174c0cd405768be6c94396c40R7370

It's a new effect that links the particle systems above.

Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
commy2: Used in `SupW_FX_ParticleUplinkDeathInitial`. See: https://github.com/xezon/GeneralsGamePatch/pull/288/files#diff-758c48cb4371d9c579a06719a58844bfbcd625b174c0cd405768be6c94396c40R7379

Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
commy2: Also used in `SupW_FX_ParticleUplinkDeathInitial` ^^

See: https://github.com/xezon/GeneralsGamePatch/pull/288/files#diff-758c48cb4371d9c579a06719a58844bfbcd625b174c0cd405768be6c94396c40R7375

xezon: Oh my mistake. Must have clicked the file away when reviewing.

xezon:
> Copy pasted from the original vanilla USA Particle Cannon, renamed and color edited.

Ok in that case we can add the original ones too for reference.

commy2:
Added the original files. Note that they are from the Generals folder, not ZH. The game loads assets from both, which is why you need to install CCG to install and play ZH.

xezon:
That's fine.

xezon:
You added `exlaser2.dds` as original and `exlaser3.dds` as edited. Is this intentional?

commy2:
The original is exlaser2.dds. That effect has to remain as it used for the non-Alex Particle Cannons. `exlaser3.dds` is an edited version that is pink instead of blue.

I cannot change the name too much, as it appears inside the model file. The model file uses animations, so it's very hard to import and export out of RenX correctly. Instead, I opted to hex editing the filenames of the textures in the W3D file.

xezon:
Sounds reasonable. Very resourceful.

xezon (approved):

Issue #283, opened 2021-09-11
Aurora Alpha Fuel Air Bomb detonation does not grant veterancy


Generalcamo:
As the title says, while the bombs themselves grant veterancy, the resulting explosion does not.

xezon:
Given the powerful destruction of the Alpha Aurora, I assume it would level up quite quickly. So this likely is a controversial change. On the other Hand this makes no sense. Maybe we can fix it and then nerf the veterancy requirements to balance it out.

Madrageeeeee:
Yes think you also need to consider the fact that auroras often get shot down on the way back. If they vet up even quicker they will also give considerable xp to the enemy. 

Think the solution of xezon makes sense in that way.

on the other hand dont think there have ever been issues with this and dont think it is really a necessary adjustment at the moment.

on this topic: think there are also different generals powers that do / do not give xp. Unsure which ones but think a10 does not give xp for things it destroys? Might be worth a separate topic

commy2:
^ A10 and Artillery Strike do not give XP to the player. Other powers do.

Jundiyy:
Vet fix could work but would still result in different outcomes. Currently you kill one Technical and blow another 9 away, no vet up.
You could now increase the amount needed to vet up but the same situation, kill one Technical and blow 9 away, you'd prob vet up.

It's not exactly the same but other units do have a similar design, Infernos with firestorms don't give vet, same for Migs, Dragon Tanks with firewall etc. If they original does not get the kill then the extra won't give vet.

RisingZH:
I think they all should be fixed, if a unit kills another unit/building it should get vet. Flame wall giving vet would be the most impactful. 

Am I right in thinking that you also don’t get vet for killing units which are inside units? (Eg a quad would get the same vet for killing a full Humvee as it would an empty one)

commy2:
> Am I right in thinking that you also don’t get vet for killing units which are inside units?

Yes. If you kill a Humvee, you get the XP awarded for the Humvee only, not the passengers. This is because the passengers die by the Humvee and only indirectly by whatever killed the Humvee.

RisingZH:
Is it possible to change that?

It’s one of the reasons USA is op.

commy2:
I don't know an easy solution to this, sadly.

xezon:
I created a separate issue for that
* TheSuperHackers/GeneralsGameCode#55

Madrageeeeee:
[...]

PR #292, opened 2021-09-11
Fix: Aurora Alpha Fuel Air Explosion May Trigger Before Bomb Hits

### LINKED ISSUE #15: Sometimes Aurora Alpha fuel bomb does not happen after initial bomb it dropped on target
Sometimes Aurora Alpha fuel bomb does not happen after initial bomb it dropped on target. Fixable?

Haubibban:
> Put a unit behind the tall structure and order Alpha to attack it from opposite side

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/15

ZH 1.04

- The Aurora Alpha secondary fuel air explosion is created where the bomb first collides with a building instead of where the bomb actually hits.

After change:

- The secondary explosion is triggered when the primary bomb explodes.

Note: This is related to https://github.com/TheSuperHackers/GeneralsGamePatch/issues/283. It does however not fix that issue. That issue is also still under discussion, since it has gameplay / balance considerations.


xezon:
From my side this is ok, but would prefer if @alanblack166, @ThePredatorBG or GeneralCamo look over this, because it is not clear to me if such a weapon change could have unintentional side effects.

MTKing4:
Wait does that mean that the fuel air explosion triggers instantly now?

commy2:
> Wait does that mean that the fuel air explosion triggers instantly now?

No. It triggers after the same delay. However, in 1.04, it actually triggers when the primary bomb dies. The primary bomb dies once it collides with a structure. However, the primary bomb dying does not start the primary explosion. The "dead" bomb continues towards the target.

This means that the secondary explosion happens where the bomb first collides, instead of where the projectile actually hits.

MTKing4:
So now when the bomb dies, nothing happens?

commy2:
After patch: When the bomb dies, i.e. phases through a building, it still continues until it hits its target. Once it hits the target, it will cause the primary explosion damage. After 1 second, the secondary explosion will trigger where the first one was.

MTKing4:
Hmmm, that's not good, gonna make alpha auroras more OP and easier to kill targets and nightmare for dozers (can't hide behind buildings anymore)

Maybe it will be better if the bomb dies completely if it hits another building

Nice sketch btw :)

commy2:
> can't hide behind buildings anymore

They already die to the primary hit in 1.04. This would only be an issue if you group multiple dozers behind the same building.

> Maybe it will be better if the bomb dies completely if it hits another building

Maybe, but then this would drastically expand the scope of this change. Then we'd have to change all Auroras, and probably a lot of other units with the MissileAIUpdate logic driving the projectile.

MTKing4:
> They already die to the primary hit in 1.04. This would only be an issue if you group multiple dozers behind the same building.

Not always, buildings save the dozers often, especially if high enough to detonate the bomb as far as possible from the dozer

> Maybe, but then this would drastically expand the scope of this change. Then we'd have to change all Auroras, and probably a lot of other units with the MissileAIUpdate logic driving the projectile.

Making the bomb phase through is what would drastically expand the scope of this change, all other auroras bombs just colide and explode on contact with any building, others have rockets not bombs, and still collide, only alpha aurora has this issue, so fixing it to be consistent with the other bombers the better approach imo

xezon:
It certainly is reasonable to suggest that bomb shell and bomb fuel may both explode at P1 instead of P2.

commy2:
> Not always, buildings save the dozers often, especially if high enough to detonate the bomb as far as possible from the dozer

Yeah, if the building is on the enemy side of the Aurora, the building will catch the bomb... including the secondary explosion. This change does not make the bomb phase through buildings. It makes the secondary explosion happen where the bomb hit in cases if the bomb already phased through the building.

> Making the bomb phase through is what would drastically expand the scope of this change, all other auroras bombs just colide and explode on contact with any building, others have rockets not bombs, and still collide, only alpha aurora has this issue, so fixing it to be consistent with the other bombers the better approach imo

That's not what this change does. Making bombs phase through more buildings would be a bad idea, we agree on that. No one wants that. However, it already does on neutral and friendly buildings, and I don't think changing that is in scope for this.

MTKing4:
Yeah i misunderstood my bad, i get it now

Then now we should fix the issue of when bombs phase through :D

alanblack166 (approved):

PR #297, opened 2021-09-11
Fix: Flashbangs Cannot Target Stinger Sites 

### LINKED ISSUE #302: GLA Stinger can be attacked by USA Sniper but not by USA Flashbang
GLA Stinger can be attacked by USA Sniper but not by USA Ranger + Flashbang. Once USA is level 3, GLA Stingers become mostly useless with this design. Discuss if anything should to be done about this.

commy2:
ZH 1.04

- Due to a bug, Rangers while equipped with Flashbangs cannot be ordered to attack Stinger Sites.
- Rangers will still target nearby Stinger Troopers in range when idle. 
- The player must click force fire next to the Stinger Site to target them manually. The cursor will show a forbidden sign while hovering over the Stinger Site.

After patch:

- The attacker can just click on the Stinger Site with Flashbang Rangers and damage the Stinger Troopers normally.


commy2:
- [ ] Rename commit when squashing.

xezon:
Thinking about this now, I think this needs to be carefully evaluated. I am not convinced this is a bug, because China bunkers can also not be flash banged, right? Looks more like Design question. It would certainly be a little buff for USA when Stinger soldiers can be targeted by Flashbang men. This would allow a Ranger inside a Humvee for example to target a Stinger comfortably which would not be possible before.

commy2:
What's weird is that they can already be hurt by Flashbangs. You just have to force-fire the ground next to the Site.

The Flashbangs damage type (SURRENDER) is also already set up to forward to the Trooper boys:

https://github.com/xezon/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesOriginal/Data/INI/Object/ChemicalGeneral.ini#L5050

We could alternatively make the Troopers immune to Flashbangs completely. The current state is very strange though.

MTKing4:
i think making them immune is the wisest call, it's a base defense, it shouldn't be taken out by a main infantry unit

RisingZH:
Interesting one, if it was fixed how many flashbang grenades would it take to clear the stinger?

xezon:
I tracked a related design question here for completeness sake: #302

commy2:
> how many flashbang grenades would it take to clear the stinger?

3

Edit: Sometimes 4. Depends on how the Flashbang scatters.

Madrageeeeee:
again, dont think this is a clear bug problem. Not sure if this requires patching / changing. Usa with humvees dont really have a problem with stinger sites anyway currently

xezon:
I think we should not buff USA with being able to kill Stinger Trooper with Flashbang. USA is good enough as is. Making the Stringer Troopers immune to flashbang should be ok however, because no one really force fires next to Stingers currently and USA does not need that ability to be strong.

commy2:
> Usa with humvees dont really have a problem with stinger sites anyway currently

Flashbangs are able to target Stinger Sites while inside Humvees regardless of this fix.

> I think we should not buff USA with being able to kill Stinger Trooper with Flashbang.

They already will target nearby Stinger Troopers if no other valid target is closer. This fix just enables the player to actually controll their units, which is better for gameplay.

> i think making them immune is the wisest call, it's a base defense, it shouldn't be taken out by a main infantry unit

I see no reason to buff Stinger Sites and nerf Flashbangs. Making the player able to properly control their units does not make USA OP suddenly. This is an extremely niche control bug. Fixing it will not suddenly make Stinger Sites worse against USA.

xezon:
If this is merged, then #636 needs to be reverted?

commy2:
Done.

commy2:
now works as intended

Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
```diff
@@ -747,6 +747,11 @@ Armor BaseDefenseArmor            ;used for anti-air base defenses.  Vulnerable
   Armor = SUBDUAL_BUILDING  100%
 End
 
+; Patch104p @bugfix commy2 16/07/2023 Fix Flashbang Rangers not being able to target Stinger Sites.
```
xezon: Please add (#297) at end of line. And to the other new added change comments too.

Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -6269,10 +6269,10 @@ Object Demo_GLAStingerSite
     InitialHealth      = 1000.0
     ;**Careful with these damage types -- because area damage types will already
     ;**damage slaves.
-    ; Patch104p @bugfix commy2 01/01/2022 Fix Stinger Troopers receive damage from Flashbangs despite the Stinger Site not being selectable as target by Flashbang-Rangers.
```
xezon, line 6272: Please make sure to update or delete file `Patch104pZH\Design\Changes\v1.0\636_stinger_flashbang_damage.yaml` and create a new one for `297_stinger_flashbang_attack.yaml`.

The change logs are generated from these yaml files.

commy2:
Like this?

xezon:
Yes, with the only exception that
```
  - fix: Removed immunity to SURRENDER in StingerSiteArmor
  - fix: Added SURRENDER to SwallowDamageTypesIfSlavesNotExisting in HiveStructureBody of StingerSite
```

Likely is too technical for the intended audience. It can be written in a way that regular players would understand it.

commy2:
Uhh, I suck at this. Let me think about it ...

xezon:
Documentation sometimes is harder than the actual implementation.

commy2:
Done I think?

Patch104pZH/Design/Changes/v1.0/297_stinger_flashbang_attack.yaml
```diff
@@ -4,8 +4,7 @@ date: 2023-07-16
 title: Fixes issue where Rangers equipped with Flashbang grenades could not be ordered to target Stinger Sites
 
 changes:
-  - fix: Removed immunity to SURRENDER in StingerSiteArmor
-  - fix: Added SURRENDER to SwallowDamageTypesIfSlavesNotExisting in HiveStructureBody of StingerSite
+  - fix: Attack cursor disappears when attempting to target Stinger Site with Flashbang Rangers
```
xezon: I recommend to write:

"The Attack cursor no longer disappears when attempting to target a Stinger Site with Flashbang Rangers."

or

"The USA Ranger with Flashbang can now properly attack the GLA Stinger Site."

xezon (approved):

ExiLe-ZH:
Think it makes sense if flashbangs do clear stingers, although I don't believe dragons do?

commy2:
Flashbangs don't clear Stinger Sites, they damage Stinger Soldiers. Same for Dragon tanks, they also don't clear Stinger Sites, but damage Stinger Soldiers via splash damage. The difference is only that you can click on Stinger Sites with Dragons, but not Flashbangs in 1.04. This is the fix for that. Rangers shoot and damage Stinger Troopers with or without this when idle.

PR #298, opened 2021-09-11
Fix: Stinger Site Does Not Benefit From AP Rockets When Engaging Airborne Targets

commy2:
ZH 1.04

- AP Rockets does not work on Stinger Sites when they are engaging airborne units. It works only against ground targets.

After patch:

- AP Rockets increases damage against air and ground targets.

Notes: 
- Was in a mood to finish all issues with the Stinger Site.
- I don't think (m)any peope even know about this.
- This needs to be discussed for balance, I think.

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -11,7 +11,7 @@ https://github.com/commy2/zerohour/issues/223 [DONE][NPROJECT]        Sentry Dro
 https://github.com/commy2/zerohour/issues/222 [MAYBE]                 Poison Fields Can Clear Poison Fields
 https://github.com/commy2/zerohour/issues/221 [IMPROVEMENT][NPROJECT] Units Sometimes Waste Shots On Overlord Debris
 https://github.com/commy2/zerohour/issues/220 [DONE][NPROJECT]        Units Sometimes Waste Shots On Poisoned Or Burned Infantry
-https://github.com/commy2/zerohour/issues/219 [MAYBE][NPROJECT]       Stinger Site Does Not Benefit From AP Rockets When Engaging Airborne Targets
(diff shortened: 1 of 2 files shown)


xezon (commented):
This will certainly help GLA bunker against Air units in late game. Not good for China. Acceptable for USA Airforce. We could have this submitted and tested.

commy2:
Yeah, we should determine how many Stingers MIGs can eat with and without that fix. Also with and without MIG Armor.

Jundiyy:
Hmmm, yea this is something that will help GLA bunker some more, You usually lose Migs if there are 2 or more Stinger Sites. Prob a bad thing since it might only help vs Air but bad for 4 Chinas.

Edit: Done some testing.
Stinger Site vs Mig - 6 shots
Stinger Site vs Armoured Mig - 7 shots
AP Stinger Site vs Mig - 5 shots
AP Stinger Site vs Armoured Mig - 6 shots

So yes one shot less, might not seem like a lot but could be the difference of losing a Mig or 2.

commy2:
7 shots vs 6 is a major difference, because the Stinger Site fires in salvos of 3.

To flip this around, currently the Stinger Site does +50% damage against air vs ground.
This decreases to +20% after AP Rockets, as vs ground damage catches up, while vs air damage does not change at all.

Maybe the Stinger Site should get less of a bonus vs air to begin with, which is then upgraded to what it is now after AP Rockets?

I don't feel very strongly about this change. It's just weird as it is right now.

Madrageeeeee:
would be afraid that toxin gamma AP stinger sites would become absolutely insane shooting down all general powers and all air units without any micro required.

never felt this was much of a problem. Wouldnt change.

xezon:
I am closing this for now. It is risky change. GLA as it stands is the best bunker faction in the game and it does not need this buff against air units (other than against Airforce General). China faction already has a number of difficulties against GLA, and this change would make it unnecessarily more difficult to engage GLA base defenses.

We may put this back up in a second iteration of patch changes when community is more friendly to bigger changes.

PR #308, opened 2021-09-11
Fix: Some Tech Buildings Reward XP When Killed

commy2:
ZH 1.04

- The Repair Bay and Reinforcement Pad grant 200 XP when destroyed if controlled by the enemy, while Oil Derricks, Hospitals and Artillery Platforms do not.

After patch:

- None of the enemy captured tech buildings reward experience when destroyed.


xezon (approved):

Jundiyy:
Hmmm, although it does make sense that they all do give xp or don't, but would this be something to change without more thought? It's a tactic people will use, allow the enemy to capture and then destroy.

Weren't these tech buildings added in ZH? If so, it seems they added the xp on purpose? And perhaps forgot to add it in for the old tech buildings. So perhaps it should be the other way, older tech buildings also give xp?

xezon:
No one kills tech buildings with the intention to gain 200 XP.

Jundiyy:
Xezon, you do play 1.04 right :) ? Of course this is a strat.

If someone doesn't know about it then that's different but a lot of people know about it. One of the reasons more know about it now is because Domi mentions it a lot on stream and he has one of the biggest following.

xezon:
Ok. I have no objection about streamlining this in a different way. Up to you. It's a design question now after this inconsistency has been fixed with this change. You can open a new issue report for further steps.

PR #312, opened 2021-09-11
Fix: Minigunner Has Broken Recoil Animation; Also Tracer Emerges From Below Unit When Aiming At Airborne Targets

commy2:
This is three seperate issues really.

ZH 1.04:

- The Minigunner uses the Red Guards rifle style recoil animation when firing.
- When firing at airborne targets only, the tracer emerges from the feet of the Minigunner instead his weapon.

After patch:

- Recoil animation is speed up 4 times, which makes it look pretty good (as far as ZH goes)
- The tracer emerges from the muzzle, even when aiming at air units.


xezon:
Could you separate the "When firing at airborne targets only, the Minigunner uses the weapon sound effect of the Gatling Tank instead of his normal effect" from this change? It is unclear if players will prefer the original design, because actually it helps readability if you know the Minigunner is shooting on air units. For example if you play as USA vs Infantry and you hear this sound, then you know your Chinook is likely in danger without even looking. So it is an audio change that actually affects gameplay.

commy2:
Removed that part and opened https://github.com/xezon/GeneralsGamePatch/pull/315 instead. Can talk about it there.

xezon (approved):

PR #315, opened 2021-09-11
Fix: Minigunner Uses Gatling Tank Sound When Aiming At Airborne Targets

commy2:
This is three seperate issues really.

ZH 1.04:

- When firing at airborne targets only, the Minigunner uses the weapon sound effect of the Gatling Tank instead of his normal effect.

After patch:

- The weapon sound does not change when aiming at air units.


xezon (dismissed):

xezon:
It is unclear if players will prefer the original design, because actually it helps readability if you know the Minigunner is shooting on air units. For example if you play as USA vs Infantry and you hear this sound, then you know your Chinook is likely in danger without even looking. So it is an audio change that actually affects gameplay.

Maybe we could also explore options to introduce a new modified minigunner sound for the Air Attack. So it would not be the Gattling Tank sound, but still distinguishable.

xezon (approved):
I would suggest we submit this now and rereview it if players send big complaints later.

MTKing4:
Maybe give an echo sound effect to the minigunner when shooting air? Nor sure it will be as effective as having a different sound, but could try that, i think if we don't find an alternative it's better to keep it as it is now

PR #316, opened 2021-09-11
Fix: Match Demo Terrorist suicide flags with flags of other faction Terrorist

Generalcamo:
The Demolition General's Terrorist, with this commit, will now explode when crushed, burned, or exploded like other faction's terrorists. This was a change done previously by 1.06, and their reasoning is quite succinct:
`One question: why didn't they explode when driven over? This weakness doesn't make any sense, especially since it's Juhziz's Terrorist, who is supposed to be better.`

This is now ported to this patch.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -13391,7 +13391,7 @@ Object Demo_GLAInfantryTerrorist
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_Death08
     DeathWeapon   = Demo_SuicideDynamitePack
     StartsActive  = Yes                      ; turned on by upgrade
-    DeathTypes = NONE +SUICIDED
+    DeathTypes = NONE +SUICIDED +CRUSHED +SPLATTED +LASERED +BURNED +EXPLODED
   End
 
 


Jundiyy:
Hmmm, I don't think having Terrorist acting differently is a problem and secondly they are already very powerful, so being able to deny them is a good thing.
For example, a common tactic to counter a TT is to run them over with a technical.

Stubbjax:
> 
> 
> Hmmm, I don't think having Terrorist acting differently is a problem and secondly they are already very powerful, so being able to deny them is a good thing.
> For example, a common tactic to counter a TT is to run them over with a technical.

Why do you believe it is not a problem? Would you intentionally design the game this way? It is a problem, as it's very confusing behaviour for players who are new or not familiar with the game. It does not make any sense for Demolition terrorists to have this arbitrary weakness, given they belong to the Demolition General who is meant to specialise in these units.

If there was any change that would make sense here, it would be that this behaviour was reversed so that only Demolition terrorists exploded when killed via the aforementioned means. But that is not a change I'd be advocating for.

MTKing4:
Even though it makes sense that way, it'd be too big of a change, we need experts' opinion on this

xezon:
Personally I would prefer all terrorists behaved the same, either all can be run over or not. It's certainly a very controversial change. It's difficult to run over terrorists successfully, This makes epic moments.

commy2:
To mix things up: Should any of the Terrorist explode when run over?

Stubbjax:
> 
> 
> Even though it makes sense that way, it'd be too big of a change, we need experts' opinion on this

Why do you feel like it'd be too big of a change? It would certainly be more in line with players' expectations.

> 
> 
> To mix things up: Should any of the Terrorist explode when run over?

I'd actually vote for no here (no explosion when crushed). The high risk / high reward dynamic that a change like this would bring could be very interesting.

MTKing4:
> Why do you feel like it'd be too big of a change? It would certainly be more in line with players' expectations.

because it's not a broken thing that needs fixing and it will affect balance quite a lot, I kinda like it that demo terrorists have their weakness in the early game as compensation for how strong they become after demo upgrade

i'd go with the same thing i said about toxin terrorists, add this as a description in their tooltip "don't explode when run over"

Jundiyy:
> > Hmmm, I don't think having Terrorist acting differently is a problem and secondly they are already very powerful, so being able to deny them is a good thing.
> > For example, a common tactic to counter a TT is to run them over with a technical.
> 
> Why do you believe it is not a problem? Would you intentionally design the game this way? It is a problem, as it's very confusing behaviour for players who are new or not familiar with the game. It does not make any sense for Demolition terrorists to have this arbitrary weakness, given they belong to the Demolition General who is meant to specialise in these units.
> 
> If there was any change that would make sense here, it would be that this behaviour was reversed so that only Demolition terrorists exploded when killed via the aforementioned means. But that is not a change I'd be advocating for.

Whilst the last part does make sense

I don't believe the first part would be confusing. Wouldn't you say Demo and Tox doing 700 damage instead of 500 is more confusing? But the fact that Demo has a sign (red explosion) and Tox has a sign (Anthrax) is more than enough to let players know they are different to regular Terrorists and will/may do something different.

Madrageeeeee:
It is one of the things that makes demo slightly worse at gettign a successful TT in gla mirrors. But you do get better terrorists so balances out i guess.

i wouldnt change this, interesting dynamic and makes for a nice little micro trick

also remember it can have benefits, as rpg + terrorist techs can explode if you run the terorists over for demo this doesnt happen.

Although this is perhaps a different thing, that demo terrorists dont do damage to their own units.

All in all i wouldnt really change this i think.

commy2:
> Although this is perhaps a different thing, that demo terrorists dont do damage to their own units.

This is correct. Even if the Demo Terrorist would explode when run over, it would not hurt allies, because Demo.

xezon:
Fixes #502
Fixes #503

PR #330, opened 2021-09-12
Fix: Aurora And Carpet Bomb Model Is Extremely Low Poly

commy2:
The Carpet Bomber and Aurora bomb model is bad. Really bad, even for 2005 standards.

<details><summary>old model</summary>

</details>

What I did is take the bomb model from the USA03 campaign (where Burton and Lotus race to destroy the lab with an avalanche) and adjust it for use as generic bomb.

<details><summary>new model</summary>

</details>

I am unsure if this should be a mandatory change or go with the other (potential) model changes. It really is that bad right now lol


xezon:
It touches INI files, so model must go in core, no?

commy2:
> It touches INI files, so model must go in core, no?

Yes. Done.

xezon (approved):

PR #333, opened 2021-09-12
Fix: Units Sometimes Waste Shots On Wrecks

commy2:
ZH 1.04:

- Units will attack certain wrecks.

After patch:

- All wrecks are left alone.

Related: https://github.com/xezon/GeneralsGamePatch/pull/257

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
@@ -1,4 +1,4 @@
-; Patch104p @bugfix commy2 09/09/2021 Fix units attacking wrecks.
+; Patch104p @bugfix commy2 09/09/2021 Add UNATTACKABLE to fix units attacking wrecks.
 
 
 ;------------------------------------------------------------------------------
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
@@ -1,3 +1,4 @@
+; Patch104p @bugfix commy2 09/09/2021 Fix units attacking wrecks.
 
 
 ;------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
```diff
@@ -1,3 +1,4 @@
+; Patch104p @bugfix commy2 09/09/2021 Fix units attacking wrecks.
```
xezon: Could say 
> Add UNATTACKABLE KindOf to debris object to avoid enemy units attacking them.

```suggestion
; Patch104p @bugfix commy2 09/09/2021 Add UNATTACKABLE to fix units attacking wrecks.
```

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
```diff
@@ -1,3 +1,5 @@
+; Patch104p @bugfix commy2 09/09/2021 Fix units attacking wrecks.
```
xezon:
```suggestion
; Patch104p @bugfix commy2 09/09/2021 Add UNATTACKABLE to fix units attacking wrecks.
```

xezon:
@commy2 
Is there a reason we did not apply `UNATTACKABLE` to all Vehicle Hulls?

commy2:
Iirc, only the ones with ActiveBody module can be shot at, so it was only added to those.

xezon:
Ok. Would it hurt to add it to all of them?

commy2:
_Probably_ not, but it would be pointless. Don't do pointless changes.

ExiLe-ZH:
Was it taken into consideration that this could alter the performance of attack move with certain units?

MTKing4:
In what way?

ExiLe-ZH:
no more wasting shots = faster follow up kills.

MTKing4:
Yes it will, but i think it's not only for attack move, Normal stance and guard mode will also be better

ExiLe-ZH:
That's true.

We gotta check though which units it makes better.

MTKing4:
According to the Code in commits, Any unit that attacks these units: 

- Rocket Buggy
- Combat Bike
- Microwave Tank
- Humvee
- Battlemaster Tank
- Overlord
- Listening Outpost
- Dragon Tank
- Troop Crawler
- Assault Troop Crawler

ExiLe-ZH:
So this needs to be tested and reviewed better. Especially seeing Vees and Buggies on that list.

MTKing4:
We can test

PR #334, opened 2021-09-12
Reorganize and add documents

xezon:


Documents/Wnd/WndReference.txt
```diff
@@ -0,0 +1,159 @@
...
+
+
+TEXTCOLOR =
+  ENABLED: R G B A,
+  ENABLEDBORDER: R G B A,
+  DISABLED: R G B A,
+  DISABLEDBORDER: R G B A,
+  HILITE: R G B A,
+  HILITEBORDER: R G B A
+
```
commy2: Consider changing the file extension of this to .md? Then we can have markdown formatting, e.g. those #### become headlines.

xezon: Yeah we can do that. Do you want that to happen for this submit?

commy2: Sure.

PR #335, opened 2021-09-12
Test all factions map

MTKing4:
Screenshot (127)

added
+ All factions buildings + WB Special Buildings and factions (Marked by a flag nearby) 
+ All factions Units
+ Granted Rank 5
+ Unlimited Power
+ Map always Revealed
+ Given some test targets to the AI

if you see this flag nearby a building that means this building is WB speical
Stupidf Flag

also bare in mind that i've reached a unit limit that would cause WB to crash if zoomed out too much, so i couldn't risk it and add more


xezon:
Could you add this map with a new name this time? Since this is a major change to the original.

TestAllFactionsAndUnits.map

MTKing4:
what about buildings/superweapons/promotions? name would be too long to describe everything imo
or we could just call it test everything! :) 
the old map is obselete, there's no need to go back to it anymore
here, i'll put some pictures so people know what's up

Screenshot (124)
Screenshot (125)
Screenshot (126)


also, i always put "!" in the beginning of the name and my version has this, it makes it show at the top of the unoffical maps list which is super handy and time saving as opposed to having to search for the map everytime time, but since you've removed it from my other submitted maps i assumed you didn't want that

xezon:
I would say we leave the previous map as is, because it allows to build things ourselves from the headquarters. This one is different as it has everything pre-placed. How do the test targets work?

We can name it `TestAllFactionsBuildingsUnitsPromo.map`

Regarding `!` name: I was not aware. We can rename maps afterwards yes.

MTKing4:
ok, i will make some adjustments tomorrow, bunch up the units to free up some space, see what i can do regarding spawning units that you requested, and rename the map to `!TestAllFactionsBuildingsUnitsPromo.map`
but if you want to rename it manually you'd need to rename all files + the folder so i think it would be best if i give it the final name so everyone can use it without the extra hassle

xezon:
I will take care of the renaming of the other maps.

PR #336, opened 2021-09-12
Change to cheaper texture for BinaryDataStream to potentially cut performance cost

xezon:
Change to cheaper texture for BinaryDataStream to potentially cut performance cost. On my machine I could not observe any performance difference.

Set Patriot stream segment count to 5.
Set Lotus and Hacker segment count to 10.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -297,19 +297,25 @@ End
 
 
 ;------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


xezon:
@commy2 
When you have time, can you please test whether or not this change improves performance? You can use

**Maps\SkirmishOnly\SkirmishTestPatriotBinaryDataStream**

commy2:
Will do.

xezon:
Increased from 3 to 10 segments for Lotus and Hackers.

shot_20211002_182717_2

xezon:
Original

shot_20220422_201557_2

Optimized & Refurbished

shot_20220422_202149_1

PR #340, opened 2021-09-13
Fixed up Quad Cannon draw code

Generalcamo:
- Fixed radar dish animation wasn't showing. ~commy2
- Level 1 salvaged quad cannon now properly uses damaged model
- Corrected dust trail to use the Quad Cannon's particle rather than the Rocket buggy's

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -16607,11 +16607,14 @@ Object Chem_GLAVehicleQuadCannon
   ;UpgradeCameo4 = XXX
   ;UpgradeCameo5 = XXX
 
+  ; Patch 1.04p @bugfix GeneralCamo 13/09/2021 - Fixed animations, set dust trail to the correct particle, and corrected salvage level 1 to properly use damaged model
   Draw = W3DTankTruckDraw ModuleTag_01
     OkToChangeModelColor = Yes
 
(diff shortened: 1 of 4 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
```diff
@@ -16607,11 +16607,14 @@ Object Chem_GLAVehicleQuadCannon
   ;UpgradeCameo4 = XXX
   ;UpgradeCameo5 = XXX
 
+  ; Patch 1.04p @bugfix GeneralCamo 13/09/2021 - Fixed animations, set dust trail to the correct particle, and corrected salvage level 1 to properly use damaged model
```
xezon, line 16610: Please use `Patch104p` as opposed to `Patch 1.04p` in your comments so it is consistent with the keyword of all other comments we added.

xezon: Other than that it looks all good.

xezon (changes requested):

commy2:
What does "animation playback" mean here?

This is missing the GC Quads.

Generalcamo:
@xezon Will commit a fix.

@commy2 I'll add the GC Quads and the CINE quads. Specifically, the Quad Cannons have an animation for their radar dishes. It currently does not play at all however, and this code fixes that.

commy2:
Nice, dishes. Yumm.

No need to fix the CINE ones.

PR #342, opened 2021-09-13

### LINKED ISSUE #300: Inf China Minigunner in Helix has extended firing range when attacking air unit together with a grounded Minigunner

commy2:
- fix https://github.com/xezon/GeneralsGamePatch/issues/300

ZH 1.04

After patch:

Drawback:


MTKing4:
this doesn't solve the issue, it just makes it easier for inf to exploit the range due to having a minigunner while other chinas don't

https://user-images.githubusercontent.com/83613147/133244018-052d5279-38aa-4877-b64e-370c6dc292bf.mp4

Jundiyy:
Ouch.

Cryoblooded:
> this doesn't solve the issue, it just makes it easier for inf to exploit the range due to having a minigunner while other chinas don't
> 
>  MiniHelixFix.mp4

how is this an expliot? I think it's perfectly fine. i don't think this will effect the game play too much after all
why would you make a helix against inf in the first place?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -6264,7 +6264,7 @@ Weapon Infa_MiniGunnerGun
   ContinuousFireCoast   = 1000 ; msec we can coast without firing before we lose Continuous Fire
   WeaponBonus           = CONTINUOUS_FIRE_MEAN RATE_OF_FIRE 200% ; When the object achieves this state, this weapon gets double the rate of fire
   WeaponBonus           = CONTINUOUS_FIRE_FAST RATE_OF_FIRE 300% ; This is not cumulative, so with Delay of 40, and values of 2 and 4 for these bonuses, you shoot at (40, 20, 10)
-  WeaponBonus           = PLAYER_UPGRADE DAMAGE 125%     ; ChainGun upgrade
+  ;WeaponBonus           = PLAYER_UPGRADE DAMAGE 125%     ; ChainGun upgrade
```
xezon: Did this chain gun upgrade buff work? Does this mean Minigunners are now weaker without the Chaingun upgrade? Does Infantry General even have this upgrade?

commy2: No, `PLAYER_UPGRADE` is never set for the Minigunner, because it has no `WeaponBonusUpgrade` module at all, let alone one that listens to `Upgrade_ChinaChainGuns`.

commy2: The actual drawback of making the Helix slightly reduce the range of the Minigunner is, that now various infantry weapons are upgraded by their `PLAYER_UPGRADE` bonus.

These include the Rebel and RPG-Soldier (AP Bullets, AP Rockets), but also the Missile Defender, which has a +25% damage bonus that is otherwise never triggered by any upgrade (also due to the lack of any `WeaponBonusUpgrade` module).

xezon: In the case of Missile Defender we could simply remove the +25% bonus if that is never used in the original game anyway right? In case of Rebel and RPG, what are the options?

commy2: > In the case of Missile Defender we could simply remove the +25% bonus if that is never used in the original game anyway right?

Yes. Same for Pathfinder and Jarmen Kell and possibly other units that I have missed.

> In case of Rebel and RPG, what are the options?

1. Ignore it, as GLA rarely has Helixes; or upgrades to AP Bullets / AP Rockets anyway. China capturing a GLA Barracks should also be rare. 
2. Change their upgrades so they don't trigger PLAYER_UPGRADE, but instead upgrade their weaponsets. This will however conflict with Tox General RPG-Soldiers (Anthrax Gamma upgrade), as well as Demo General Rebels and RPG-Soldiers (Demolitions upgrade).
3. Revert the Helix-Minigunner range nerf.

The worst abuse here is 2v2 with GLA surrendering to give Tunnel spawned RPG's to the allied China player, which then loads them into Helixes.

I'm somewhere between 1 and 3, as 2 has the mentioned problems.

xezon: Yes. 2 does not read good at all.

Madrageeeeee: would be nice to be able to target airunits with the china bunkerlix.

Will this not make red guard be able to shoot at air units though from a bunkerlix?

Also: china does sometimes make a helix versus infantry (see my challenge game versus excal)

It would be nice if the targeting works the same as the combat chinook with regards to targeting air units.

commy2: > Will this not make red guard be able to shoot at air units though from a bunkerlix?

No, only passengers that can engage air units already will accept the order. This is why a Combat Chinook does not enable Rangers to fire at planes either.

commy2:
"Experimental", because I'm still not sure about the Mingunner-In-Helix range hack.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -6766,6 +6767,27 @@ Weapon ListeningOutpostUpgradedDummyWeapon
   AntiAirborneVehicle   = Yes
 End
 
+; Patch104p @bugfix commy2 13/09/2021 Add weapon used by Helix with Bunker to order passengers to attack airborne targets.
+
+;------------------------------------------------------------------------------
+Weapon HelicopterUpgradedDummyWeapon
```
xezon, line 6773: We could call this "Helix", as it is not used for Chinook or Comanche, right?

commy2: Sure. I pulled this from a game build were I reused the same dummy for both Combat Chinook and Helix, because it makes sense to me that both behave the same in this regard.

Jundiyy:
Oh no :(

ExiLe-ZH:
MG range is ridiculous, this is not fun for USA.

Also, does this change Rocket range vs other Helixes? Where they track and hit evading Lixes?

> > this doesn't solve the issue, it just makes it easier for inf to exploit the range due to having a minigunner while other chinas don't
> > MiniHelixFix.mp4
> 
> how is this an expliot? I think it's perfectly fine. i don't think this will effect the game play too much after all why would you make a helix against inf in the first place?

Helix is useful vs Inf (dual WF build), it can crack Outposts and kill Dragons without receiving damage.

xezon:
Note that the long range fire is possible in 1.04 if you group Helix with ground unit. What we need to do here is find graceful consistent solution.

ExiLe-ZH:
> Note that the long range fire is possible in 1.04 if you group Helix with ground unit. What we need to do here is find graceful consistent solution.

Exploit tbf

PR #345, opened 2021-09-13
Fix: Demo General Combat Bike Suicide Button Does Not Work

commy2:
ZH 1.04

- If you have a Demo General Bike, and then upgrade Demolitions, it will gain a Suicide button.
- This button will always be greyed out and not work.
- If you switch Riders, the button will disappear.
- If you build a new bike, it will not have this button, only old ones will have it.
- If you have a Demo General Bike with Jarmen Kell, and then upgrade Demolitions, it will lose the Sniper Attack button...
- Getting Jarmen off the Bike and on again will make the Sniper Attack button show again.

After patch:

- The Suicide ability button will always show.
- It will stay disabled until the Demolitions upgrade has been bought. It will work correctly and suicide the Bike.
- The Sniper Attack button will not disappear.

Drawback:

- The button will show as disabled before the Demolitions upgrade. This is the same as the Worker in 1.04 (though fixed already in this patch by this commit https://github.com/TheSuperHackers/GeneralsGamePatch/pull/161). Best I can do with the interaction of Rider logic and command set upgrades.

Since this is essentially a new feature that changes micromanagement of the dreaded Demo Bikes, I'll mark this as controversial.

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -172,7 +172,7 @@ https://github.com/commy2/zerohour/issues/49  [DONE][NPROJECT]        Demo Gener
 https://github.com/commy2/zerohour/issues/48  [MAYBE][NPROJECT]       Demo General Scud Launcher With Demolitions Upgrade Deals Full Damage When Destroyed
 https://github.com/commy2/zerohour/issues/47  [DONE]                  Demo General Battle Bus Is Deleted Without Dealing Damage When Suicided
 https://github.com/commy2/zerohour/issues/46  [DONE][NPROJECT]        Technical Has To Recenter Turret To Suicide
-https://github.com/commy2/zerohour/issues/45  [IMPROVEMENT]           Combat Bike Has Disabled Suicide Ability Button
+https://github.com/commy2/zerohour/issues/45  [DONE]   	              Combat Bike Has Disabled Suicide Ability Button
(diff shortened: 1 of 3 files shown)


xezon (approved):

Jundiyy:
I know this has been added to main branch but I think it should be fixed the way I have mentioned in #614 
In short, the button is completely removed from the game. Already explained in #614.

ExiLe-ZH:
Agree with Jundiyy on how it should be fixed.

Btw, there shouldn't be any controversial changes like this in the main branch, not without testing.

xezon:
I suggest to test it in matches and then evaluate whether or not it should be accepted.

Jundiyy:
I do agree to testing but as I've said in the discord server before, bigger changes should first be tested internally and not pushed out till then. Or else ask some good players to test in private. 

Also, I don't see why it can't be the same as other changes we have, talked about and then when happy we add it into main, this is added first and then talked about.

MTKing4:
I'm neuteral towards this, but what internal testing? We don't have any. The only test we have is what we sponsor on gr.org

I suggest to make a round robin with multiple rounds each has its own mini patch with contraversal change/s to test with same players

For example

1. No Scrap Deny 
2. Gatlix + Minilix AA
3. Demo bikes with detonation
4. Quads Scrap Up fix

More or less can be batched into each category in a fitting manner to make the best out of the test and save much more testing time with this method

xezon:
We could revert this change on **main** and move it to the **main-balance-any** branch once we have. Then all testing can be performed from the other branch, whereas **main** can remain more pure.

ExiLe-ZH:
> 

Think things like these should be placed in a dedicated branch for controversial changes since they require special attention.

PR #347, opened 2021-09-13
Fix: Inconsistent Outpost Geometry

commy2:
ZH 1.04

- All sub-faction Outposts have different geometries.

||vChina | Tank/Nuke | Inf |
|:-|:-:|:-:|:-:|
| length | 20 | 20 | 22 |
| width | 10 | 9 | 9 |
| height | 22 | 13 | 13 |

After patch:

- All sub-faction Outposts have the same geometry.

I went with the Tank/Nuke one, since 50% were already using those, and they were the more similar to Inf.

I'm not entirely certain about the implications of this. Someone who knows please explain.


xezon:
Hmm interesting. What does the geometry do? Maybe it puts a bounding box that is used for physics? How do we know which sizes are appropriate for the given w3d model?

xezon (approved):
I think this is ok to submit for now. We can revisit this when we investigate #354

PR #349, opened 2021-09-13
Fix: Vanilla China Outposts Missing Detection Effect; Non-Vanilla China Outposts Missing Detection Sound Effect

commy2:
ZH 1.04

- The normal China Listening Outposts is not displaying the red grid below a detected object and the red ring around the unit that the other Listening Outposts have.

After patch:

- All Listening Outposts use the same effects.


Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
```diff
@@ -2656,16 +2656,19 @@ Object ChinaVehicleListeningOutpost
...
     DetectionRange            = 300 ;John, enable this for independant balancing!
     CanDetectWhileGarrisoned  = No ;Garrisoned means being in a structure that you units can shoot out of.
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
-    PingSound                 = IRPing
-    LoudPingSound             = IRPingLoud
-    ;;;IRParticleSysName         = IRDetectSonar
-    ;;;IRBrightParticleSysName   = IRDetectSonarBright
-    ;;;IRGridParticleSysName     = IRDetectGrid
+;    PingSound                 = IRPing
+;    LoudPingSound             = IRPingLoud
```
commy2: Note on these two. I made them consistent across all sub-factions with this change. As far as I can tell, this sound is never played.

xezon: Looks ok to me, but would prefer @alanblack166 look over this.

alanblack166 (approved):

PR #352, opened 2021-09-13
Fix: Chinook And Supply Center Swapped Supply Lines Upgrade Icon

commy2:
- The Supply Lines upgrade icon should be moved from the Supply Center to the Chinook, because the cash bonus is linked to the Chinook and not the Supply Center: A Chinook collecting on a GLA or China Supply Stash will drop the boosted $660, while a Supply Truck will drop the regular $300 on a USA Supply Center.


xezon (approved):
Your attention to detail is remarkable. o_O

commy2:
### But at what cost?

xezon:
Still laughing at image.

commy2:
But at what cost?

PR #356, opened 2021-09-14
Testing Minigunner Range Penalty

commy2:


MTKing4:
seems to be working correctly

https://user-images.githubusercontent.com/83613147/133302008-fe37347b-5c79-4b0b-a1c3-8a69fb0cc0d4.mp4

MTKing4:
TC Works as well

https://user-images.githubusercontent.com/83613147/133305068-3952fed8-4272-461f-8d50-ac2b69bf9206.mp4

Jundiyy:
What's the purpose of this? Sorry not following.

MTKing4:
Reducing minigunner AA range when in Helixes

Jundiyy:
Yes I can see that but why? Is it bugged?

MTKing4:
Because this is the only way to prevent the range exploit

Cryoblooded:
THIS IS NOT AN EXPLOIT
the Bunker upgrade gives extra range for troops inside the lix

Jundiyy:
> Because this is the only way to prevent the range exploit

Ah right, well does it have to have this less range? That new range is just bad, makes it kinda useless to use a Mini Gunner Lix unless you are 100% sure you will come out alive.

commy2:
The actual range can be adjusted in both versions. The 10% used here is just a rough estimation of what the vs ground range is.

Cryoblooded:
How does this effect game play ? i mean its usefull for aod but the only scenario i see it unfair is a lix battle in cashbounty map.

Jundiyy:
So does that mean we can fix the bug and keep original AA value? If so, that should be the way to go.

commy2:
>So does that mean we can fix the bug and keep original AA value?

Yes.

MTKing4:
> So does that mean we can fix the bug and keep original AA value? If so, that should be the way to go.

You mean extended range for minigunner in lix? That would defeat the purpose of all this change, might as well revert everything, at least then not a lot of people know of the exploit

Jundiyy:
> > So does that mean we can fix the bug and keep original AA value? If so, that should be the way to go.
> 
> You mean extended range for minigunner in lix? That would defeat the purpose of all this change, might as well revert everything, at least then not a lot of people know of the exploit

The extended range isn't the bug, it works perfectly fine. The auto AA was the bug which has been fixed?

MTKing4:
> The extended range isn't the bug, it works perfectly fine. The auto AA was the bug which has been fixed?

No, Auto AA was never an issue as it will likely never happen in games, the bug was shooting AA with helix by selecting a minigunner with it, and you just fixed it by legitimizing the bug and made it a feature, helix was never intended to shoot AA units directly. now if 0.1% of the people used that exploit to ruin the game, now everyone will, great.

I don't understand how this is fine to you, this is not fine at all. In fact it's game breaking

This is the best solution we've got and if you think it's bad then let's just put this in the survey and ask the experts provided they see the footage, if they think it's fine then i concede

xezon:
I find it hard to believe that the original Minigunner AA range in Helix is the best design there is. It looks like it can shoot further than anything else in the game. Could we adjust AA range in Helix to match the AA range from Troopcrawler? And I don't mean the absolute range: If Troopcrawler, Aircraft and ground below Aircraft form a triangle where the Hypothenuse is the firing range, then I would expect the firing range of Minigunner in Helix be the Adjacent of aforementioned triangle. Does that make sense?

Here is a visualization. I suspect Minigunner in Helix has firing range of Red line, but blue line range would be a lot more sane for gameplay.

minigunnerhelix

commy2:
> It looks like it can shoot further than anything else in the game.

The vs-air range is 350 times the 1.33 GARRISONED range bonus. This totals 465.5, which is larger than the Scud Launcher with scrap+1/scrap+2 (450.0).

It does in fact shoot further than every other unit in the game, aside from scrapped Scud Launcher with Search and Destroy.

As for the image, all ranges are calculated in 2D. The only effect z/height has is travel delay, and those are only relevant for projectile based weapons. (The Minigunner deals damage the same frame it shoots; the tracer is only visual fluff.)

xezon:
Ok excellent. Could we test with the 1.33 bonus removed, so Minigunner in Helix has same Range as in Outpost and Troopcrawler?

commy2:
We could. Multiplication makes a huge difference when one factor is already large (snowballing), which is why this is currently so broken and why that change would help a lot.

Implementation is done in https://github.com/xezon/GeneralsGamePatch/pull/357/files.

There's a comment that says PR-#357 breaks the ATC, but I don't think that is the case. I think it is just poorly understood how weapon changing works for passengers.

xezon:
I will test these 2 in the evening 👍

Jundiyy:
@MTKing4 

I think you misunderstood what I said. I know how to do the bug and what it is.

Once again,
The AA weapon itself is not bugged. It just has a large range, which is fine. (Of course, it uses the AA weapon once in range.)
The bug is using a ground Mini Gunner as you have said above, which has been fixed by commy.
Hence why I said it's fine to use original AA weapon. No more abuse. How is what I have mentioned game breaking? Or you just misunderstood?

Regarding Helix targeting air units, I didn't even mention anything about that?

Also, keeping the above changes would mean you would have to come really close to take out a Chinook when trying to take out eco, much harder to snipe down GP planes, very hard to take down Migs in a Mig spam game (one of the worst problems for Inf).

@xezon 

I don't think it's a valid argument to change something just because it is hard to believe? Although reducing it slightly compared to original value above as you has been mentioned doesn't sound as bad. Reasoning should be good though.

MTKing4:
> There's a comment that says PR-#357 breaks the ATC, but I don't think that is the case. I think it is just poorly understood how weapon changing works for passengers.

I only made the testing and in the test the ATC has to come real close to shoot, which wasn't the case before, i don't what else i need to know about

> The AA weapon itself is not bugged. It just has a large range, which is fine. (Of course, it uses the AA weapon once in range.)

I haven't misunderstood anything, but it seems you are, 
No one said anything about AA weapon being bugged, this was never about the AA weapon itself, its purely about using it inside of a helix

> The bug is using a ground Mini Gunner as you have said above, which has been fixed by commy.
Hence why I said it's fine to use original AA weapon. No more abuse.

This is what i have a problem with, how's that a 'no more abuse' if you're enabling a completely broken feature for everyone to use is fine?

> Regarding Helix targeting air units, I didn't even mention anything about that?

Again, this was purely about the helix Targeting air units with AA range, i don't know what else have you understood from that, minigunner AA range has not been touched nor in the ATC as I've demonstrated in my videos

> Also, keeping the above changes would mean you would have to come really close to take out a Chinook when trying to take out eco, much harder to snipe down GP planes, very hard to take down Migs in a Mig spam game (one of the worst problems for Inf).

Again, minigunner range has not been touched, this is purely for helix, and the range above can be increased a bit np

Jundiyy:
> its purely about using it inside of a helix

Using inside a Helix, it is given a bonus which isn't a bug as well, (not talking about the ground mixing MG). That's all I'm talking about. But perhaps I'm just not getting the point so I will leave it there.

commy2:
> I only made the testing and in the test the ATC has to come real close to shoot, which wasn't the case before

My understanding is, that the ATC has to get in range for the Minigunner vs ground weapon for the Minigunner to switch and use the extended range of the vs air weapon. Maybe test this again with and without that PR-#357? That PR  contains no changes that should affect the ATC.

MTKing4:
> Using inside a Helix, it is given a bonus which isn't a bug as well

Never said it was a bug, but it's a drastic game change and a bad one as well when the helix can target air units from afar

> My understanding is, that the ATC has to get in range for the Minigunner vs ground weapon for the Minigunner to switch and use the extended range of the vs air weapon. Maybe test this again with and without that PR-#357? That PR contains no changes that should affect the ATC.

That is probably correct, but the idea was that The ATC should always use the AA gun automatically, it shouldn't need to get close just to activate it is what I'm saying, pretty sure this is how it works in 1.04 and the plan is to maintain it that way, will get a test of that still and compare it with pr #357 just to be sure

commy2:
> the idea was that The ATC should always use the AA gun automatically, it shouldn't need to get close just to activate it is what I'm saying

I would like that too, but I don't know how you'd do that, because

> pretty sure this is how it works in 1.04 and the plan is to maintain it that way

this doesn't seem to be case to me. At least I see no reason how the changes in PR-#357 could've brought that result about.

MTKing4:
Ok, let me get the test first then and we can discuss what to do after

Cryoblooded:
> > Using inside a Helix, it is given a bonus which isn't a bug as well
> 
> Never said it was a bug, but it's a drastic game change and a bad one as well when the helix can target air units from afar
> 
> > My understanding is, that the ATC has to get in range for the Minigunner vs ground weapon for the Minigunner to switch and use the extended range of the vs air weapon. Maybe test this again with and without that PR-#357? That PR contains no changes that should affect the ATC.
> 
> That is probably correct, but the idea was that The ATC should always use the AA gun automatically, it shouldn't need to get close just to activate it is what I'm saying, pretty sure this is how it works in 1.04 and the plan is to maintain it that way, will get a test of that still and compare it with pr #357 just to be sure

So what you want is to lose a 3k+ cost helix for nothing with out killing a mig or a raptor?

Cryoblooded:
I'm confused as well i see no harm at all for the helix targeting air units it will only help in aod and vs airforce late ffa game. so why are u worried about a helix vs helix it makes no sense to build a helix against infantry general.
Am I making myself clear?

commy2:
> i see no harm at all for the helix targeting air units

Me neither, but it comes at the cost of easy to use Minigunners with a range larger than scrapped up Scud Launchers. Maybe we have to compensate some of the extra range coming from stacking vs air range AND the GARRISONED range bonus of the Helix Bunker. Evaluating the technical possibilities is the point of this branch.

Cryoblooded:
ok let's do some math. the helix alone with bunker cost 2k same as the avenger strongest aa in the game and every minigunner is 350. I think it worth have the range because of how expensive this is.

PR #371, opened 2021-09-16
Change: Enable Quad Cannon Radar Dish Animation

commy2:
ref: #340

There apparently is an animation for the radar dish on the Quad, which is not un-used.

Note:

- Should check if RUBBLE state should have this. When kaputt, it may be better to stop the animation?

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -16617,11 +16617,15 @@ Object Chem_GLAVehicleQuadCannon
   ;UpgradeCameo4 = XXX
   ;UpgradeCameo5 = XXX
 
+  ; Patch104p @bugfix commy2 13/09/2021 Add radar dish animation.
+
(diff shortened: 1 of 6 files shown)


commy2:
- [ ] rename when squash

commy2:
Apparently the animation only works on the scrap +2 Quad, as other Quads don't have a radar dish.

The radar dish seems to float above the Quad when rotating.

MTKing4:
https://user-images.githubusercontent.com/83613147/133691096-c7fa7e43-1c69-4382-81e5-a33641bb93e4.mp4

commy2:
Looks weird. It floats.

I'd say we do not merge this.

xezon:
Regardless of the floating, I don't like this. If base Quad has no radar, then I don't think Scrap 2 should have either. Quad does not gain any stealth detection capabilities with scrap that would warrant a radar dish.

commy2:
Well, scrap+2 always had a radar. It's just not animated.

xezon:
Oh. I never noticed.

commy2:
Close because this unused animation looks unfinished.

PR #372, opened 2021-09-16
Change: Use Unique Dust Effect For Quad Cannon

commy2:
ref: #340 

Apparently there is a unique dust effect for the Quad Cannon. In ZH 1.04, it shares the effect with the Rocket Buggy (and many other units).

I do not know if this effect looks better or not.


xezon:
I will take a look at it later and leave my 2 cents afterwards.

xezon:
1.04 Dust

https://user-images.githubusercontent.com/4720891/133883098-b86a9579-6b76-4bbd-8776-b5e5c985503a.mp4

Changed Dust

https://user-images.githubusercontent.com/4720891/133883110-0c079c2c-1440-4b2f-83d3-b9ff1f7a6964.mp4

It's quite a big visual difference.

xezon:
I would say we leave particle effect as Original. The particle change is quite obvious and potentially bothers people, because it is brighter and larger.

commy2:
Agreed.

PR #373, opened 2021-09-16
Fix: Laser Turret Has 1 Extra Shot When Assisting Or Engaging Airborne Targets Than Intended

commy2:
ZH 1.04

- The Laser Turret is intended to fire 3 shots instead of 4 rockets like the Patriot Battery. However, when engaging airborne units or when assisting other Laser Turrets, it still uses the original 4 shots.

After patch:

- The Laser Turret shoots in salvos of 3, no matter against what target and in what mode.

This makes it weaker against air targets and in general when assisting. Therefore, controversial?

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -6530,7 +6530,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon Lazr_PatriotMissileWeaponAir
-  PrimaryDamage               = 35.0
+  PrimaryDamage               = 40.0
   PrimaryDamageRadius         = 3.0
   AttackRange                 = 350.0
   DamageType                  = EXPLOSION
@@ -6556,7 +6556,7 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 Weapon Lazr_PatriotMissileAssistWeapon
-  PrimaryDamage               = 35.0
+  PrimaryDamage               = 40.0
   PrimaryDamageRadius         = 3.0
   ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -6526,6 +6526,7 @@ Weapon Lazr_PatriotMissileWeapon
 End
 
 ; Patch104p @bugfix commy2 13/09/2021 Fix sound and effect used against airborne targets and when assisting to match normal anti ground weapon.
+; Patch104p @bugfix commy2 13/09/2021 Reduce ammo against air and when assisting to match normal anti ground weapon.
 
 ;------------------------------------------------------------------------------
 Weapon Lazr_PatriotMissileWeaponAir
@@ -6540,7 +6541,7 @@ Weapon Lazr_PatriotMissileWeaponAir
   FireFX                  = Lazr_WeaponFX_LaserCrusader
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
   DelayBetweenShots           = 250                   ; time between shots, msec
-  ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
+  ClipSize                    = 3                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 2000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
   AntiAirborneVehicle         = Yes
@@ -6567,7 +6568,7 @@ Weapon Lazr_PatriotMissileAssistWeapon
   FireFX                  = Lazr_WeaponFX_LaserCrusader
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
   DelayBetweenShots           = 250                   ; time between shots, msec
-  ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
+  ClipSize                    = 3                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
   AntiAirborneVehicle         = Yes
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -34,7 +34,7 @@ https://github.com/commy2/zerohour/issues/196 [DONE][NPROJEC!]        Sub-Factio
 https://github.com/commy2/zerohour/issues/195 [DONE]                  Nuke Cannons Start With Neutron Shells
 https://github.com/commy2/zerohour/issues/194 [DONE][NPROJEC!]        Patriot Batteries And Laser Turrets Lag When Assisting
 https://github.com/commy2/zerohour/issues/193 [DONE][NPROJEC!]        Laser Turret Uses Different Laser Model And Sound When Assisting Or Engaging Airborne Targets
-https://github.com/commy2/zerohour/issues/192 [IMPROVEMENT][NPROJECT] Laser Turret Has 1 Extra Shot When Assisting Or Engaging Airborne Targets Than Intended
+https://github.com/commy2/zerohour/issues/192 [DONE][NPROJECT]        Laser Turret Has 1 Extra Shot When Assisting Or Engaging Airborne Targets Than Intended
 https://github.com/commy2/zerohour/issues/191 [IMPROVEMENT]           Patriot Battery And Laser Turret Extended Range Exploit
 https://github.com/commy2/zerohour/issues/190 [IMPROVEMENT]           Black Lotus Seleced By Q-Shortcut
 https://github.com/commy2/zerohour/issues/189 [DONE][NPROJEC!]        China Dozer Movement Behavior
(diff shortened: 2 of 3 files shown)


xezon:
If we don't want to change overall damage output, then we could increase Air Damage by factor 1.33 per clip, no?

commy2:
Yes.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -6430,6 +6430,8 @@ Weapon Lazr_PatriotMissileWeapon
   ProjectileCollidesWith      = STRUCTURES
 End
 
+; Patch104p @bugfix commy2 13/09/2021 Reduce ammo against air and when assisting to match normal anti ground weapon.
+
 ;------------------------------------------------------------------------------
 Weapon Lazr_PatriotMissileWeaponAir
   PrimaryDamage               = 35.0
```
commy2:
```suggestion
  PrimaryDamage               = 45.0
```
Keep damage per clip the same.

commy2: Dang, can't reach the assist with comments :/

xezon: 35 * 1.333 = 46.655

Jundiyy:
Does sound like a fix but I see a 'small' a problem with this fix, what do you guys think?

By increasing the damage per shot, it would mean a unit could die faster as that shot could be the killer, whereas having less damage would mean it needs 2 shots.

Just making an example up, a unit comes to attack with 45 hp left
With the original stats, the Laser Turret would need 2 shots to take it down.
With the new stats, it would only need 1.

Of course, this can be scaled up, the unit has 135 hp,
Original = 4 shots
New = 3 shots

Thoughts?

xezon:
Yeah but these clips are in quick succession, so change should have little to no tangible impact. But yea its more of a Design question than a real Bug I think. If we feel strongly about streamlining the clip sizes, which is reasonable, then its a fair change. Just because it always had 4 clips for Air does not mean it always needs to keep that. It is very likely that this was a mistake by EA Designers.

Are the 3 and 4 ammo clips spanned over equal time or do they have fixed delay between them?

commy2:
8 frames between shots. 61 frames to reload. So before it takes 4\*8+61 = 93 frames from clip to clip, while after it takes 3\*8+61 = 85 frames.

Consider also that 4 shots are more easily spread over multiple units than 3.

All in all, this thing shreads through everything either way. A possible scenario where this could affect the outcome is Comanche spam vs Laser turrets or Helix spam vs Laser turrets. Maybe an AFG flying over Laser turrets? Reaching here.

Edit:
Oh, the smaller the clip size, the less impactful the Patriot Extended Range bug (unintentional) / exploit (intentional) become.

xezon:
Aha ok. In that case distributing damage of 4 clips in 3 clips would be a buff to the Laser Turret, because it would take 8 frames less for shooting the next load. In that case correct damage distribution would need to account for that if I am not mistaken.

`DamagePerClip = (OriginalDamage - OriginalDamage * 8/93) * OriginalClipSize/NewClipSize`

`(35 - 3) * 4/3 = 42.6`

That correct?

commy2:
Sound about right, however, you always have to keep in mind that damage is quantized, not continuous for slow firing units, and that you are firing at multiple singular targets instead of one blob of army.

xezon:
Yes, 4 shots would be able to kill 4 targets max while 3 shots would be able to kill 3 max. But how much of a practical difference does it make it in the real world? I would assume its a non factor. I am ok with both streamlining clips or leaving as is (original).

commy2:
Yeah. Would streamline and keep damage at 45.0 vs air.

Jundiyy:
Yes Xezon, you're right that it won't always make the difference because many times you will kill a unit off with the amount of needed shots. But there will also be times where that extra shot is needed and won't be given now. I can't say what percentage as it needs testing, so can't say if it makes a big difference.

Commy if you can send me the code with the new weapons, I can add random units to a map and see what happens.
Or leave it without a test.

xezon:
Jundiyy, the code is here in GitHub on this very branch the Pull Request is based on. I highly recommend you learn how to use it, otherwise it will just slow you down massively over the long term.

commy2:
Set it to 40 for now, because that matches vs ground. At least now every line can be edited by ```suggestion.

Jundiyy:
Oh I didn't get back to this, sorry forgot after the busy weekend. Will try to, I was thinking just flying loads of Comanches over and see how quick they get shot down, comparing old to new (have both weapons on the map at the same time, 'carbon copy it in the INI')

And yes Xezon, you are right, I just need to sit and watch some stuff.

xezon:
I punched the numbers into our DPS Calculator.

The formulas become
```
35 * 4 / ((4 * 266 + 1000) / 1000) = 67.74 DPS
40 * 3 / ((3 * 266 + 1000) / 1000) = 66.66 DPS
```

This means setting the weapons to 40 damage and 3 clips is appropriate, because the damage is essentially the same. I will update the excel sheet with this data so we have it tracked.

xezon:
@commy2
In the first post you write it "is intended to fire 3 shots instead of 4 rockets". How did you come to this conclusion? I am asking because I see the mention of four clips 3 times and three clips 1 time. Wouldn't that suggest the intended clip size was 4?

commy2:
All Patriot turrets use three different weapons, one for ground attacks, one for anti-air attacks, and one for "assisting" nearby Patriots.

The CCG Patriot Battery (as well as the EMP Patriot) fire four rockets each salvo ("clip"). Therefore, all of their three (six) weapons have ClipSize set to 4.

The Laser Turret on the other hand does absurd amounts of damage each shot and has a ClipSize of only 3 for ground attacks. This is what makes the building different from the Patriots of the other factions.

However, the anti-air as well as the assist variants still have a ClipSize of 4. I believe this is a mistake by EA, because to me **it would make more sense if a Laser Turret salvo had the same amount of shots regardless of whether it is currently shooting at ground or air units and regardless of whether it is currently "assisting"**. 

**I believe this was an oversight during the design process when the original weapon code of the Patriot was copied: They made the change in the "main weapon" (the anti ground one), but forgot to adjust the number "4" to "3" in the other places below that.**

Why 3? Because that's what the anti ground weapon has, which is the first one that appears in the list and which the other weapons are derived from so to say. The normal Patriots obviously should have 4, because that is what their art implies.

Admittedly, this is a weak reason for making such a change, but inconsistencies like this keep me up at night /shrug.

MTKing4:
@commy2 so your suggestion is to make laser Patriot shoot 3 volleys across all three weapons and vanilla patriot shooting 4 instead of the current 3?

If so I'm up for such a Change, basically nerfing the strong and buffing the weak ones. In a way that makes the most sense

xezon:
I am still not convinced that 3 shots are more reasonable than 4. Right now the change has 5 edits to streamline the clip sizes. However, it is also possible to streamline the clips with just 2 edits on the primary weapon. And then all Patriots have the same clip size too.

Maybe the EA dev was like: "Oh let's set this one weapon to 3 and test how it looks in game compared to the 4. Oh it is not so great, let us revert it again". And then he was interruped for his lunch break. And when he returned to his workplace he already forgot about it because it was such an obsolete change.

As long as DPS is about the same, I do not think we need to worry about gameplay impact on this one. If any it will be negligible.

ExiLe-ZH:
Doesn't the rate of fire effectively get higher when making it 3 shots instead of 4 while keeping the cooldown between shots/bursts the same?

xezon:
3 clips shot means it waits less total time than clip of 4 until the next clip of 3. I think the most important values are the DPS (Damage Per Second). I posted them above.

ExiLe-ZH:
> 3 clips shot means it waits less total time than clip of 4 until the next clip of 3. I think the most important values are the DPS (Damage Per Second). I posted them above.

By clips you mean zaps?

If so, then let me try to illustrate the problem (imagine that "1" is a zap and "-" is cooldown):

3 zaps: 111-111-111-111-111
4 zaps: 1111-1111-1111-1111

With 3 zaps the total amount of zaps is 15.
With 4 zaps it's 16 zaps within the same time.

So if you increase damage for 3 zaps to 133% per shot it's a buff right? But not sure how accurate the illustration is though.

xezon:
It looks like so:
```
0s  1s  2s  3s  4s  5s  6s  7s
-   -   -   -   -   -   -   -
111----111----111----111----111-
1111----1111----1111----1111----
```

We can do it either way, but I think it is better to streamline on 4 zaps, because it is a smaller change.

commy2:
> So if you increase damage for 3 zaps to 133% per shot it's a buff right? But not sure how accurate the illustration is though.

There is a [DPS calculator](https://github.com/TheSuperHackers/GeneralsGamePatch/blob/787d735f2c8b961e4765770a3197db3ac1e209c4/Patch104pZH/Design/Balancing/weapon_dps_calculator.xlsx) that should take clipsizes into account. 

---

4 -> 3 may be more changes in code, but the anti ground weapon is what gets exposed in gameplay more frequently. So from an ingame perspective, 3 -> 4 may be the larger change overall.

Consider matches vs GLA. There're no air targets (aside from the anthrax bomber), so unless you build 2 Laser turrets nearby, this change does nothing at all.

xezon:
In Bunker matches, AOD, No Rush 10, etc you will mostly see Assisted Laser beams. A single Laser Patriot is mostly for early game. In competitive matches players often build Firebases instead of Laser Patriots, because Firebase requires no power and has good range vs Ground. I still believe 4 Clips are less of a change.

Jundiyy:
I don't think we should change to 4 clips, there isn't an actual need for it and it would look different when firing for no reason.
It's not like it has visuals the way a Patriot does.

MTKing4:
Agree with jundiyy, i think laser Patriot should be streamlined to 3, and vanila Patriots to be 4, increasing their effectiveness

xezon (approved):

PR #374, opened 2021-09-16
Fix: Laser Turret Uses Different Laser Model And Sound When Assisting Or Engaging Airborne Targets

commy2:
ZH 1.04

- The Laser Turret uses a different sound effect and laser beam when either assisting or when attacking airborne targets.
- The sound effect used is the PDL anti missile / infantry zap of the Paladin. (Different FXList, but same sound ultimately.)

After patch:

- The same laser model is used, regardless of the target.


xezon (approved):
Odd. I never noticed any visual inconsistencies with this.

PR #376, opened 2021-09-17
Fix: Tactical Nuke Mig Upgrade Not Buildable While Low On Power

commy2:
ZH

- The Tactical Nuke Mig upgrade cannot be purchased from the Nuclear Missile Silo while low on power. Neutron Shells, as well as Nuclear Tanks and Uranium Shells for other factions, can still be bought while low on power.
- If the upgrade has already been queued, but is not finished yet, it can still progress to completion even when low on power.
- When the building is disabled by an EMP strike however, none of the upgrades can be purchased and their progress will freeze... Except when the building is low on power AND disabled by an EMP strike: Then none of the upgrades can be purchased, but already queued upgrades will progress... The progress will freeze again once the building has enough power, but is still disabled by EMP... ...

After patch:

- The Tactical Nuke Mig upgrade behaves like every other upgrade on the Nuclear Missile Silo.

Note:

- Apparently, in an earlier build of the game, Tactical Nuke Mig was a Propaganda Center upgrade, while Isotope Stability was a Nuke Silo upgrade. I moved the flag away from the Isotope Stability, because a Propaganda Center, while it consumes power, cannot be low on power (not to be confused with "disabled" by EMP).

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -2515,6 +2515,7 @@ End
 CommandButton Command_UpgradeChinaTacticalNukeMig
   Command       = PLAYER_UPGRADE
   Upgrade       = Upgrade_ChinaTacticalNukeMig
+  Options       = IGNORES_UNDERPOWERED  ; @bugfix commy2 17/09/2021 Fix upgrade not buildable when low on power.
   TextLabel     = CONTROLBAR:UpgradeChinaTacticalNukeMig
   ButtonImage   = SSMigNuke
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -6521,7 +6522,6 @@ End
 CommandButton Nuke_Command_UpgradeChinaIsotopeStability
   Command       = PLAYER_UPGRADE
   Upgrade       = Upgrade_ChinaIsotopeStability
-  Options       = IGNORES_UNDERPOWERED
```
xezon, line 6524: Why is this removed?

commy2: This upgrade appears on the Propaganda Center. The Propaganda Center cannot be low on power: if the player is low on power, the Propaganda Center continues operation, so this flag does nothing.

The Propaganda Center can only be disabled by EMP strikes or Microwave, but that is not covered by this flag.

It does nothing, so I removed it.

xezon (approved):

PR #382, opened 2021-09-18

commy2:
ZH 1.04

patch:

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -77,7 +77,7 @@ https://github.com/commy2/zerohour/issues/145 [DONE][NPROJECT]        Toxin Gene
 https://github.com/commy2/zerohour/issues/144 [IMPROVEMENT]           Overlord Does Not Fit Into Chinook But Only Takes 3 Slots Inside Helix
 https://github.com/commy2/zerohour/issues/143 [IMPROVEMENT]           GPS Scrambled Rangers And Red Guards Remain Stealthed While Capturing Buildings
 https://github.com/commy2/zerohour/issues/142 [IMPROVEMENT][NPROJECT] Fire Base Shoots At Awkward Angle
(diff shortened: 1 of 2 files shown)


Patch104pZH/Design/Tasks/commy2_tasks.txt
```diff
@@ -77,7 +77,7 @@ https://github.com/commy2/zerohour/issues/145 [DONE][NPROJECT]        Toxin Gene
 https://github.com/commy2/zerohour/issues/144 [IMPROVEMENT]           Overlord Does Not Fit Into Chinook But Only Takes 3 Slots Inside Helix
 https://github.com/commy2/zerohour/issues/143 [IMPROVEMENT]           GPS Scrambled Rangers And Red Guards Remain Stealthed While Capturing Buildings
 https://github.com/commy2/zerohour/issues/142 [IMPROVEMENT][NPROJECT] Fire Base Shoots At Awkward Angle
-https://github.com/commy2/zerohour/issues/141 [IMPROVEMENT][NPROJECT] Toxin General RPG-Trooper Shoots Stinger Missiles
+https://github.com/commy2/zerohour/issues/141 [DONE][NPROJECT] 		  Toxin General RPG-Trooper Shoots Stinger Missiles
```
xezon: There are tabs instead of spaces.

xezon (approved):

xezon:
I was thinking about this today once more and remembered that Toxin RPG's have this Stringer tracking that other RPG's have not. I thought this would be a pure visual change but I suspect now this is no longer just a visual change, so we may have to revert it.

ImTimK:
This is definitely a controversial change, but I personally think it's for the better, only if the damage output remains the same though. 

Now China has atleast 1 counter to Gamma Busses, because losing all your migs everytime every time is no fun and isn't very balanced. Migs can sometimes be saved by ECM's but unreliably so.

commy2:
Commit has been reverted. Discuss this change here: Pr-#394

PR #73, opened 2021-08-27
Particle Cannon ranger spawn fix

alanblack166:
All Particle Cannons should now spawn rangers that belong to their respective generals.

PR #388, opened 2021-09-19
Fix: Boss Destroyed American Buildings Spawn Wrong Ranger Type

commy2:
ZH 1.04

- Boss General Fusion Reactor, Patriot Battery and Particle Cannon drop the normal USA Rangers when destroyed instead of the Boss General's Rangers.

After patch:

- The Boss General buildings drop the Boss General Rangers.

I also renamed hanfields object creation lists from PR-#73. {faction}_OCL_whatever is consistent with the OCL's EA already uses.

Note that `Boss_OCL_AmericanRangerDebris03` already exists (as well as the unused `Boss_OCL_AmericanRangerDebris04`). `Boss_OCL_AmericanRangerDebris02` however did not, so it had to be added.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -10723,7 +10723,7 @@ Object AirF_AmericaParticleCannonUplink
     FX             = INITIAL FX_ParticleUplinkDeathInitial
     OCL            = INITIAL OCL_SDILinkLasers
     FX             = FINAL   FX_StructureMediumDeath
-    OCL            = FINAL   OCL_AirF_ParticleUplinkDeathFinal ; Patch104p @bugfix hanfield 26/08/2021 Added proper death OCL to spawn Airforce General Rangers.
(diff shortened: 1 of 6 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -10723,7 +10723,7 @@ Object AirF_AmericaParticleCannonUplink
     FX             = INITIAL FX_ParticleUplinkDeathInitial
     OCL            = INITIAL OCL_SDILinkLasers
     FX             = FINAL   FX_StructureMediumDeath
-    OCL            = FINAL   OCL_AirF_ParticleUplinkDeathFinal ; Patch104p @bugfix hanfield 26/08/2021 Added proper death OCL to spawn Airforce General Rangers.
+    OCL            = FINAL   AirF_OCL_ParticleUplinkDeathFinal ; Patch104p @bugfix hanfield 26/08/2021 Added proper death OCL to spawn Airforce General Rangers.
```
xezon, line 10726: Hmm. When I search INI, I see this:
```
; Patch104p @bugfix hanfield 26/08/2021 Added death OCL for Airforce General Particle Cannon.

ObjectCreationList OCL_AirF_ParticleUplinkDeathFinal
```

xezon (changes requested):
Please double check OCL name changes.

xezon (approved):

PR #394, opened 2021-09-19

commy2:
ZH 1.04

patch:

old PR: #382


xezon:
The Stinger missile has tracking. It is much better than a regular RPG missile. Changing this will make Toxin RPG equal to regular RPG. Toxin General is quite OP, so it is reasonable to check whether or not this should be streamlined. It is controversial.

ExiLe-ZH:
This is definitely a controversial change, but I personally think it's for the better, if it ONLY affects the range.

Now China has atleast 1 counter to Gamma Busses, because losing all your Migs every time is no fun and isn't very balanced. Migs can sometimes be saved by ECM's but still unreliably so.

Jundiyy:
Quite a bit change to be honest, not sure I'd like it, not same example but it's like reducing the PDL speed or the amount of PDLs that are used in a certain time.
If they are normal rockets, that would mean they would not be good vs USA Air as well, currently they are very good vs King Raptors when used together. Taking down GP planes would also be much harder. Overall just fighting vs anything, this change would be a change vs every other army.

xezon:
Maybe give the same RPG Stinger Tracking to all other GLA's for consistency? It's odd that only Tox should have it, considering it is Over Powered anyway.

ExiLe-ZH:
> 
> 
> Quite a bit change to be honest, not sure I'd like it, not same example but it's like reducing the PDL speed or the amount of PDLs that are used in a certain time.
> If they are normal rockets, that would mean they would not be good vs USA Air as well, currently they are very good vs King Raptors when used together. Taking down GP planes would also be much harder. Overall just fighting vs anything, this change would be a change vs every other army.

Other GLA's never had this overpowered advantage either, so you know the exact result of this 'nerf'. 

'Very good' vs King Raptors is quite an overstatement tbh, I've never seen a King Raptor die to tracking RPG's ever, only upclose on the airfield. Tracking doesn't play a role here.

Overall, the negative impact of this mechanic (China having no counter) is just much larger than the positive impact (RPG's countering Kraptors) on balance.

I guess you could argue that removing or nerfing PDL's is the same thing, but one thing you would never do is giving all USA's PDL's. 

For these reasons it's a big no go to give all GLA's Stinger tracking.

Again, the thing that's most important here, like I've mentioned before, is that the damage output vs any unit needs to remain the same. I can only approve this if it purely removes the tracking mechanic.

Jundiyy:
Whilst I agree no doubt it's a nerf and will become  the same as other GLAs so we can see most of the difference it would make, this is changing a core mechanic, nearly every battle will have RPGs involved, we don't want Tox to feel different to how it plays. This is what the patch promises.

Funny you mentioned KRaptor actually, just watched an Air vs Tox stream a day or 2 ago, prob Size vs Fargo, and 2 KRaptors went to attack a Tunnel on the left, bunch of RPGs and I think 1 Quad, not sure. But 1 didn't survive on the way back and the second lost 3/4 health. But let's forget KRaptor, every other air unit would also get away just like they do from normal RPGs, support powers wouldn't get shot down at as much.

Also, having PDLs, why should Air have it when no other USAs do, it's the same here, why should Tox have it when others don't. It makes sense for Laser actually. But anyway, as mentioned above to your reply, it's changing a core mechanic, just like we won't remove PDLs to nerf Air. It's not a thing we should remove. 

We should also work the way we have been doing, buff over nerf priority, sometimes nerf but with a thing like this it has a much bigger impact. I'd prefer to buff other armies.

My thoughts anyway.

ExiLe-ZH:
> 
> 
> Whilst I agree no doubt it's a nerf and will become the same as other GLAs so we can see most of the difference it would make, this is changing a core mechanic, nearly every battle will have RPGs involved, we don't want Tox to feel different to how it plays. This is what the patch promises.
> 
> Funny you mentioned KRaptor actually, just watched an Air vs Tox stream a day or 2 ago, prob Size vs Fargo, and 2 KRaptors went to attack a Tunnel on the left, bunch of RPGs and I think 1 Quad, not sure. But 1 didn't survive on the way back and the second lost 3/4 health. But let's forget KRaptor, every other air unit would also get away just like they do from normal RPGs, support powers wouldn't get shot down at as much.
> 
> Also, having PDLs, why should Air have it when no other USAs do, it's the same here, why should Tox have it when others don't. It makes sense for Laser actually. But anyway, as mentioned above to your reply, it's changing a core mechanic, just like we won't remove PDLs to nerf Air. It's not a thing we should remove.
> 
> We should also work the way we have been doing, buff over nerf priority, sometimes nerf but with a thing like this it has a much bigger impact. I'd prefer to buff other armies.
> 
> My thoughts anyway.

While it's true that RPGs are involved in most battles, it's not true that the tracking mechanic affects all those battles. This mostly affects hit and run units like Migs, which happens to be the ONLY counter China has vs Battle Busses. While PDLs are unfair, it doesn't break the balance as much as this one does.

Regarding your example of the Kraptor, that's most likely another scenario where mass RPGs defeat the PDL upclose. Whilst possible, it's VERY rare for a KRaptor to die in the Stinger tracking range after it's already evading. The tracking property is most definitely not a meta counter in this case.

On the other hand, Tox RPGs are a meta counter vs Comanches. This definitely is a factor that makes Tox vs Air more balanced than Demo vs Air for example. That said, after the tracking removal, Gamma should still make RPGs much more effective vs Comanches, this also affects Stinger Sites.

Madrageeeeee:
gamma and this tracking ability is definitely what makes tox so strong lategame.

Agree in general with Jundiyy that you want to provide the same gameplay.

1.06 always felt a bit 'slow' compared to 1.04 due to the many decreases in killing power making the game feel a bit less exciting and on the edge of a knife to play

Also you dont want to make the different subfactions too similar as that takes away from the fun of having a different gaming experience with each one

I think the tracking is mostly a problem for migs and helixes as toxin becomes so insanely hard to play against for china late game if the toxin general builds some nests / busses / puts some rgs in a building. And the ground army doesn't really stand a chance anyway except for perhaps the inferno spam (but 1 flanking buss and gg for that too)

Perhaps a slight tweak might already do a lot: make toxin RPG troopers with gamma deal slightly less damage to aircraft but keep the tracking ability and keep the same damage to ground units?

ExiLe-ZH:
> 
> 
> gamma and this tracking ability is definitely what makes tox so strong lategame.
> 
> Agree in general with Jundiyy that you want to provide the same gameplay.
> 
> 1.06 always felt a bit 'slow' compared to 1.04 due to the many decreases in killing power making the game feel a bit less exciting and on the edge of a knife to play
> 
> Also you dont want to make the different subfactions too similar as that takes away from the fun of having a different gaming experience with each one
> 
> I think the tracking is mostly a problem for migs and helixes as toxin becomes so insanely hard to play against for china late game if the toxin general builds some nests / busses / puts some rgs in a building. And the ground army doesn't really stand a chance anyway except for perhaps the inferno spam (but 1 flanking buss and gg for that too)
> 
> Perhaps a slight tweak might already do a lot: make toxin RPG troopers with gamma deal slightly less damage to aircraft but keep the tracking ability and keep the same damage to ground units?

I also agree with keeping the core gameplay intact as much as possible, the battle of OP's is definitely one of the strengths that makes up for the pace and excitement in 1.04. Another small example is how Tox Terrorists behave differently, imo learning and applying these strengths is fun.

With these discussions however I just hope that we can come up with some middleground solutions for problems like these.

Everyone knows Tox vs China is extremely unbalanced and the tracking issue is a major cause of this. Ground cannot counter Gamma Busses, Lixes like you said are no option (which generally are very good vs GLA) and even Migs don't return home in 90% of the cases. 

Maybe your suggestion can work, or maybe shortening the tracking range could also work (without completely making it like other GLA's).

Hope to hear some more suggestions.

commy2:
Making a custom missile with tracking behaviour somewhere between RPG and Stingers is definitely possible.

RisingZH:
I would remove it. Makes no sense to me for tox RPG’s to be better than other gla’s.

ExiLe-ZH:
> I would remove it. Makes no sense to me for tox RPG’s to be better than other gla’s.

Yes right? I get it that different factions have their different strengths, but where is it described or what is the reasoning Tox gets this particular strength?

To me it doesn't make sense that Tox's standard Rocketmen get Stinger tracking technology, USA doesn't even have this kind of tech. 

Btw, the real reason it is the way it is:
Commy2: 
> They're supposed to have a coloured trail. Bluish with Anthrax Beta (Tox starts with that) and pink with Anthrax Gamma.
> The ProjectileExhaustEffect, as its called is copy pasted from the Stinger Site and color adjusted. For the RPG, EA was lazy and just copy pasted the projectile itself as well. That is why the Tox RPGs shoot stinger missiles instead of RPG rockets. It is totally possible to get the trail with the correct projectile though.
> It's not Gamma rockets. It's the Beta rockets as well.

> What was a flavour effect turned into a gameplay perk.
> They just didn't care or thought about it at all.

xezon:
I wonder if most players would actually notice that this behavior changed if the visuals remained the same.

Jundiyy:
It would be noticeable because you know that a unit trying to get away will usually be shot at, especially air units.
I'd still like to keep it the same, might be op but making them normal RPGs would take a lot of the fun out of Tox.
Buff other needed factions.

ExiLe-ZH:
Proposal for zh purists who defend everything in 1.04 till death lol: 

Keep tracking, reduce damage over distance, if even possible.

xezon:
I doubt that is possible. But maybe we can decrease tracking range.

xezon:
# RPG vs Toxin RPG Weapon

https://user-images.githubusercontent.com/4720891/184421124-0bb4e984-efda-4a53-8b02-50f3dd626aa9.mp4

https://user-images.githubusercontent.com/4720891/184421141-a8689573-d49b-4fe5-baf9-5d9e28f04e1d.mp4

# RPG and Toxin RPG vs Plane

https://user-images.githubusercontent.com/4720891/184421273-c55df056-0d36-48b2-9983-595e09854229.mp4

https://user-images.githubusercontent.com/4720891/184421289-1c848293-2aca-4617-951f-a6e52c436129.mp4

# RPG and Toxin RPG vs Helix

https://user-images.githubusercontent.com/4720891/184421816-bbe4d80c-22ee-43e0-b659-f7881d65d944.mp4

https://user-images.githubusercontent.com/4720891/184421840-6c31c812-6b36-4571-8414-bc71268e2e77.mp4

# Conclusion

Both RPG Troopers have same firing interval, but the Toxin Stinger RPG will always hit quicker. Against planes, the Toxin RPG will always hit twice, but RPG only ever once on a straight fly over. Against Helix, both RPG will deal about the same damage. In the video, the Toxin RPG shoots 1 more time, but this should be just coincidence depending on Helix route.

The advantage is 2 fold:
1. Toxin RPG Tracking follows planes far away
2. Toxin RPG flies faster and can kill targets quicker

commy2:
Killing your target quicker also means the RPGs get fired at less / vet up before the return fire hits them.

xezon:
Perhaps we can keep the quicker rocket speed and just remove the tracking. Then Toxin RPG are still superior, but Planes are not as vulnerable anymore.

Jundiyy:
Sorry to be the painful guy but I still prefer to leave as is. It's a thing of Toxins, even if it doesn't make sense to why Toxin should track. A bit like why air has PDL when Laser should have it.

I don't see any reason to remove it as well.

At most, the tracking could be reduced by x amount, slower units would still get shot at but faster units might evade. Although I'm still happy with how it currently is.

ExiLe-ZH:
> Perhaps we can keep the quicker rocket speed and just remove the tracking. Then Toxin RPG are still superior, but Planes are not as vulnerable anymore.

This is fair, the only real difference will be vs planes, every ground battle will remain exactly the same. I just can't imagine how anyone would defend the current "balance" vs planes... Jundiyy???!! :D

Btw, the difference between normal and tracking RPG's is probably bigger when planes and helis fly away diagonally.

Jundiyy:
I suppose ground battle wouldn't make a massive difference, other than vs Buggies and Technicals.

Can we agree to like half tracking? lol
I would still like the speed of the rocket to stay the same. Plus damage. Tracking/life time reduced by half?

PR #402, opened 2021-09-21
Fix: Raptor Has Misplaced Guard Air Button

commy2:
ZH 1.04

- The Guard Air button on the Raptor and King Raptor is placed differently than on the Migs. It appears that the developers forgot to move this button in the transition from the 12 button command sets in CCG to the 14 button command sets in ZH.

After patch:

- Raptor Guard Air button is placed in the free corner slot, same as Mig.

Dunno why it is tagged MAYBE. This improves the game by at least 200 % when playing any USA.


Jundiyy:
`Dunno why it is tagged MAYBE. This improves the game by at least 200 % when playing any USA.`

lol haha

xezon (approved):

PR #403, opened 2021-09-21
Fix: Air Force General Avenger Receives 30% More Damage From Jet Missiles Than Other Avengers

commy2:
ZH 1.04

- In EA's infinite wisdom, they made the Air Force General Avenger take 30% more damage from jet missiles than other Avengers.
- Or maybe they just fucked up when they specialized Avenger armor and forgot this one.

before patch it uses:
```
Armor AntiAirVehicle
  Armor = JET_MISSILES        30%
  Armor = STEALTHJET_MISSILES 30%
  Armor = CRUSH               50%
  Armor = SMALL_ARMS          50%
  Armor = GATTLING            50%    ;resistant to gattling tank
  Armor = COMANCHE_VULCAN     50%
  Armor = INFANTRY_MISSILE    50%
  Armor = POISON              50%
```

after patch it uses:
```
Armor AvengerArmor
  Armor = JET_MISSILES        23%
  Armor = STEALTHJET_MISSILES 30%
  Armor = CRUSH               50%
  Armor = SMALL_ARMS          50%
  Armor = GATTLING            50%    ;resistant to gattling tank
  Armor = COMANCHE_VULCAN     50%
  Armor = INFANTRY_MISSILE    50%
  Armor = POISON              50%
```

- JET_MISSILES damage is dealt by Raptors, King Raptors and Migs before the Black Napalm and Nuke Mig upgrades (both do EXPLOSION instead).

Controversial, because this ever so slightly affects balance, (even though this is one of the rarer units considering the other Arsenal AFG has).


xezon:
Design question.

Jundiyy:
I can only mainly think of a difference in Air mirrors, if it gets to that stage of using ground units.
Other USAs don't use Raptors, not vs Avengers anyway.

And also Air vs Chinas, it can happen where the China has stayed strong and gone for Mig spam, in such a case Avengers would be used with ground army.

Maybe good to keep so Air doesn't benefit even more? Even if it's not as common as other Avengers.

commy2:
Remember that upgraded Migs are not affected by this change.

The only Mig-spam that would realistically use un-upgraded Migs is with Nuke-Migs. I think if you're going un-upgraded Nuke-Mig-spam against Air Force General Avengers, you deserve to lose.

Jundiyy:
That's true,
So Chinas would only matter before Black Napalm, which is early game, and usually early game Air wouldn't make Avengers due to Strat Centre. Although there is still is the chance of using unupgraded Migs and Air doesn't need that benefit. 

So that leaves mainly Air mirror, which still isn't too common but would still change something.

Imo it's not too big of a deal but I'd prefer to keep air with less buff.

xezon:
The problem I have with this difference is, that it is not transparent to player. Let's say player owns 1 Airforce War Factory and 1 Superweapon Factory. Avenger has same build cost, but Superweapon Avenger is a bit better. Player assumes both are the same, so it should not matter where he builds it. But somehow it does and this is not cool.

If you compare this with Overlord vs Nuke Overlord, then you can clearly see they are different by their speed and radiation shells. This is better for user to understand.

Jundiyy:
That's true, makes sense.

Madrageeeeee:
this is a very niche question once again, but going from Jundiy's reasoning I'd say it doesnt matter too much. Not buffing Airforce makes sense.

Agree with Xezon too that you would like players to easily understand how units compare to one another and provide logical playing characteristics.

Dont have a clear preference here and dont think it matters too much.

RisingZH:
Agree with xezon, identical units should behave exactly the same across factions.

xezon (approved):

PR #405, opened 2021-09-22
Sync done tasks between commy2 and NProject task lists

xezon:


commy2:
Is the tag supposed to be NPROJEC! with an exclamation mark instead of a "t"?

xezon:
Yes. It is just an intermediate tag and I wanted to keep the same string length for simplicity sake.

commy2 (approved):

PR #407, opened 2021-09-23
Fix: Non-Vanilla USA Avengers Benefit From Composite Armor

commy2:
ZH 1.04

- The non-vUSA Avengers gain 33% extra health from the Composite Armor upgrade.

This patch:

- None of the Avengers gain extra health from Composite Armor.

Obviously controversial. Balance relevant is only Laser here, as SWG and AFG can't buy Composite Armor from their Strat Centers and vanilla USA does not get the bonus no matter what.

Alternatives here are:

1.) Giving vanilla USA Avengers the Composite Armor bonus (and add upgrade icons across all Avengers), but it's weird that what basically is a Humvee gets a upgrade for Crusaders, Paladin and Microwaves.

2.) Do nothing, but then why is the vanilla USA Avenger different from the others?


xezon:
Needs decision from Balance guys. Would be nice if behavior was consistent.

Jundiyy:
I wouldn't change the Laser one for sure, even if players don't know that it benefit, the reason being that it is part of how Laser works, it's in the blood and it will play it's role in it's balance.

The other 2 might as well stay if Laser is.

Regarding USA, sounds simple to say to add it, but that means a buff to an already good faction, which I don't know if it is safe to do. I'd prefer Rage/Rising/Dominator and other better players to get in on this one.
In 1v1s Crusaders are not spammed to the extent that they need Composite Armour unless vs Ovies/Emperors + ECMs (so that's where Avengers would benefit from it) and they'd be used when Migs are on the battlefield, although IF someone decided to get the upgrade vs GLAs then the upgrade would help vs Buggies.

Jundiyy:
Maybe this can be moved to the main so more people see it? Don't think everyone comes to this section.

Madrageeeeee:
Interesting never knew this!

I think making it consistent so that either Laser avengers do NOT get the buff or so that USAV avengers ALSO get the buff makes most sense.

I think this discussion is most relevant for 3v3/2v2v2/1v1v1v1v1v1v1 games where large groups of laser tanks and avengers are spammed late game by USA vs GLA mostly or in 3v3 where vee/avenger groups are spammed by Laser. I would argue that this buff is most relevant with regards to avengers being harder to kill by groups of migs/raptors because they will survive where you do not expect them to.

Removing it for Laser avengers wouldnt impact the game TOO much. And i think laser is already really powerfull in these spam situations due to the laser tanks/avenger combination being very strong. You could argue that the extra buff for avengers is not needed therefore.

On the other hand you could also argue that its not too big of a problem to give this buff to USAV as well, since usav is less strong in these spammy situations than Laser anyway and the avengers for usav cost more anyway so you will get less out of it anyway. 

In 1v1 it is as Jundiyy says, most relevant when facing tank with Laser and spamming crusaders/avengers. 

All in all I dont think either choice will impact gameplay too much.

Generalcamo:
Just a note:
EA 1.03 Patch Notes:
`Avenger:
[...]
Composite armor upgrade has been removed.
`
EA did intend to remove the composite armor upgrade from the Avenger. This PR as-is would correct this to what 1.03 intended.

The question later would be whether 1.03 was justified in this or not.

commy2:
Careful with those changelogs though. From my experience, what is written in the changelog is far from what was actually changed per version.

xezon:
If the vUSA Avengers without Composite Armor are fine, then so should be Laser Avengers without Composite Armor at a discount. I will merge this change as is. If in future we identify Avengers as being too weak, then we can revisit this and think about buffing them consistently. They should have no hidden perks across factions.

Issue #237, opened 2021-09-07
China Nuke Superweapon is ineffective against GLA buildings


xezon:
China Nuke Superweapon is ineffective against GLA buildings. The missile deals good damage on the center of the impact, but because the Nuke blast is singular, the GLA buildings will just blow down to their hole, regardless of the health they have remaining.

GLA SCUD Storm and USA Particle Cannon do not have this design issue.

See also #235

commy2:
The nuke damage logic is kind of complicated. There is an inner radius where some constant damage is dealt, and then there is an outer radius where damage gradiately decreases to 0 depending on the distance from the inner radius.

https://github.com/commy2/zerohour/commit/9d28d9a33df4b4bc98055cf3fb795183f7261256

With some math magic, one can use the 6 blast phases the engine already provides to graduately deal damage from blast center to blast circumference in steps, thus destroying holes (and also Battle Buses) - in such a way that the damage as function from the position stays the same.

ZekeDlyoung:
Based on my own experiments with my mods, I found that having a constant radius is better than a growing radius, but I agree with dividing the single blast into multiple blasts

MTKing4:
Divide 1 blast into multiple blasts

Madrageeeeee:
Think there is an issue with all factions. The damage is does seems very random. Sometimes it will destroy many buildings and sometimes it does hardly any damage. It often felt a bit like hit or miss. 

This isnt for gla structures only i feel. The damage dealt to the buildings always felt very random and not predictable

commy2:
That's because it decreases with range. There is an inner range where damage is flat. Outside the inner radius, the damage decreases to 0 proportionally.

xezon:
I reported a new issue for the comment provided by Rage
Issue #280

PR #410, opened 2021-09-24
Change: Make Nuke Missiles Destroy GLA Holes

commy2:
* As discussed in https://github.com/TheSuperHackers/GeneralsGamePatch/issues/237

ZH 1.04

- The Nuclear Missile is a single explosion, so even when overkilling GLA buildings, it always leaves behind a GLA hole.

After patch:

- The Nuclear Missile is divided up to 6 smaller explosions that total to the same amount of damage, but also may destroy GLA holes (and bunkered Battle Buses).

The PR also changes the worldbuilder only USA SWG Nuclear Missile explosion, the Nuke Truck from the Nuke General GC mission, and the unused Baikonour explosion dummy object from the CCG campaign, because all of these use the exact same logic and values for their mushroom cloud explosions.

damage before:

damage after:

distance ... distance from center of nuke to object
damage ... damage the object receives from nuke

<details><summary>script used to plot the damage</summary>

```python
import sys
from dataclasses import dataclass

import numpy as np
import matplotlib.pyplot as plt

@dataclass
class Explosion:
    delay: float
    dmg_min: float
    dmg_max: float
    r_inner: float
    r_outer: float

    def damage_at(self, x: float) -> float:
        if x < self.r_inner:
            return self.dmg_max

        if x > self.r_outer:
            return 0.0

        if self.r_outer <= self.r_inner:
            return 0.0

        return (
            self.dmg_min +
            (self.dmg_max - self.dmg_min) * (self.r_outer - x) /
            (self.r_outer - self.r_inner)
        )

@dataclass
class TotalDamage:
    explosions: list[Explosion]

    def __iter__(self):
        yield from self.explosions

    def time_steps(self):
        return np.array([ e.delay for e in self.explosions ])

    def total_time(self):
        return max(self.time_steps())

    def step_range(self):
        max_ = max(e.r_outer for e in self.explosions)
        return np.arange(0, round(max_) + 10, 0.001)

    def curve_at_time(self, time):
        explosions = [ e for e in self.explosions if e.delay <= time ]

        damage_values = []
        for x in self.step_range():
            total = sum(e.damage_at(x) for e in explosions)
            damage_values.append(total)

        return (np.array(self.step_range()), np.array(damage_values))


def main():
    delay = [580, 660, 720, 850, 1000, 1180]

    #title = "Nuke Missile 1.04"
    #r_inner = [60, 90, 120, 150, 180, 60]
    #r_outer = [60, 90, 120, 150, 180, 210]
    #dmg_min = [0, 0, 0, 0, 0, 300]
    #dmg_max = [0, 0, 0, 0, 0, 3500]
    title = "Nuke Missile patched"
    r_inner = [60, 90, 120, 150, 180, 210]
    r_outer = [90, 120, 150, 180, 210, 210]
    dmg_min = [0, 0, 0, 0, 0, 300]
    dmg_max = [640, 640, 640, 640, 640, 300]

    explosions = [ Explosion(*vals) for vals in zip(delay, dmg_min, dmg_max, r_inner, r_outer) ]
    nuke = TotalDamage(explosions)

    for explosion in nuke:
        time = explosion.delay
        x, y = nuke.curve_at_time(time)
        plt.plot(x, y, label=f"{time} ms")

    plt.title(title)
    plt.xlabel("distance")
    plt.ylabel("damage")
    plt.legend()
    plt.show()

    return 0


if __name__ == "__main__":
    sys.exit(main())
```

</details>

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -320,6 +320,8 @@ Object SupW_NeutronMissile
     SpecialPowerTemplate = SupW_SuperweaponNeutronMissile
   End
 
+  ; Patch104p @balance 05/09/2021 Change Nuke Missiles such that they destroy GLA holes.
+
   Behavior = NeutronMissileSlowDeathBehavior ModuleTag_06
     DestructionDelay    = 3501
     ScorchMarkSize      = 320
@@ -329,8 +331,8 @@ Object SupW_NeutronMissile
     Blast1Delay         = 580     ;in milliseconds
     Blast1ScorchDelay   = 100     ;in milliseconds
     Blast1InnerRadius   = 60.0    ;objects inside this get the full damage
-    Blast1OuterRadius   = 60.0    ;objects inside this get some of the full damage
-    Blast1MaxDamage     = 0.0     ;damage within inner radius of blast
+    Blast1OuterRadius   = 90.0    ;objects inside this get some of the full damage
+    Blast1MaxDamage     = 640.0   ;damage within inner radius of blast
     Blast1MinDamage     = 0.0     ;always do at least this much damage to objects
     Blast1ToppleSpeed   = 0.5     ;higher #'s topple faster
     Blast1PushForce     = 10.0    ;higher #'s push more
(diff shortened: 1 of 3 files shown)


xezon:
Took me a while to understand. But looks all correct. Well done Sir.

commy2:
Demo


https://user-images.githubusercontent.com/6576312/136666604-14eb771e-f1c2-473b-a1f6-5c3e88ba2c59.mp4

xezon (approved):

MTKing4:
Didn't understand anything from the graph, but it looks great

commy2:
The graph just means that the Palace in the video would've received exactly the same amount of damage without this change.

MTKing4:
I know what the change is, the same damage is split over time rather than delivered in one blow, graph just confused me with the distance thing
Looking at it again it all makes sense now

PR #412, opened 2021-09-25
Change: Reduce XP reward for regular ground tanks by 30%

### LINKED ISSUE #395: Killing Elite China Battlemaster grants a lot of XP
Killing Elite China Battlemaster grants a lot of XP. It is very common that any faction vs China Tank will progress much quicker with XP. Check if something needs to be done about this.

### Proposal 1
Reduce XP reward for China (Tank) Battlemasters by around 20%

commy2:
Fixes #395
Closes #779
Closes #780
Closes #781

## Original

* Tanks (Crusader, Paladin, Microwave, Battle Master, Marauder) grant 100% more experience than almost all other vehicles at 100 100 200 400
* This especially snowballs when combined with Vet 2 (e.g. Elite Battle Master Training), because Vet 2 vehicles grant double the amount of experience, making it 4 times as much XP per kill.

## Patched

* XP granted by killing regular Tanks is reduced by 30% to 70 70 140 280. This means that tank like units now grant only 40% more experience instead of 100% more experience when compared to other vehicles.

This change does not touch Overlords and Scorpion tanks. Scorpion tanks already only grant 20% more experience compared to other vehicles.

# Rationale

## Stats
In the following chart we see how many kills of different units it takes to reach a certain General Rank. Originally it would take 13 HERO tank kills to reach Rank 5. With this change it will take 18 HERO tank kills to reach Rank 5. The XP reward granted for these tanks is still significantly higher than light vehicles, long range vehicles and aircraft.

tanks_exp_edit

## Conclusion
The original kill rewards for tanks are high. Since tanks are typically not that difficult to take out with Rocket Soldiers, Long Range Vehicles and Aircraft, it appears unwarranted to award that much kill experience. A XP reduction by 30% is logical.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -20674,9 +20674,7 @@ Object Demo_GLATankMarauder
     Science = SCIENCE_MarauderTank
   End
 
-  ; Patch104p @balance commy2 25/09/2021 Reduce tank experience reward by 20%.
-
-  ExperienceValue = 80 80 160 320     ;Experience point value at each level
+  ExperienceValue = 70 70 140 280 ; Patch104p @balance commy2 xezon 02/08/2022 From 100 100 200 400
   ExperienceRequired = 0 200 300 600  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -20674,7 +20674,9 @@ Object Demo_GLATankMarauder
     Science = SCIENCE_MarauderTank
   End
 
-  ExperienceValue = 100 100 200 300    ;Experience point value at each level
+  ; Patch104p @balance commy2 25/09/2021 Reduce tank experience reward by 20%.
+
+  ExperienceValue = 80 80 160 240     ;Experience point value at each level
   ExperienceRequired = 0 200 300 600  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -20676,7 +20676,7 @@ Object Demo_GLATankMarauder
 
   ; Patch104p @balance commy2 25/09/2021 Reduce tank experience reward by 20%.
 
-  ExperienceValue = 80 80 160 240     ;Experience point value at each level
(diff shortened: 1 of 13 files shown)


Jundiyy:
This wasn't the proposed change, the change was just for Tank General since his tanks are used the most and so other Generals can easily rank up. This problem doesn't exist for the rest.

commy2:
I feel like the issue exists for all of these units, but is only apprent as Tank, because of Elite Training and the units you're pushed to use as that faction.

Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -20420,7 +20420,9 @@ Object Demo_GLATankMarauder
     Science = SCIENCE_MarauderTank
   End
 
-  ExperienceValue = 100 100 200 300    ;Experience point value at each level
+  ; Patch104p @balance commy2 25/09/2021 Reduce tank experience reward by 20%.
+
+  ExperienceValue = 80 80 160 240     ;Experience point value at each level
```
xezon: Why is this 240 as opposed to 320? Killing a 3 Star Marauder certainly deserved serious XP, because when it is double scrapped, it is a monster unit.

commy2: It currently grants 300 unlike other tanks, so 320 would in fact increase it for this unit at vet 3. Others have 100, 100, 200, 400 curve instead. Don't ask me why.

xezon: Increasing the Vet3 Kill XP by 20 points from 300 to 320 sounds reasonable and is consistent with the other tanks. I don't think we do China any service by decreasing the XP Reward for killing Vet3 Marauder.

Jundiyy:
> I feel like the issue exists for all of these units, but is only apprent as Tank, because of Elite Training and the units you're pushed to use as that faction.

The xp isn't a problem for them units since the units are not used as often and not in a vet 2 state when trained, the reason it is felt by Tank is because they are the main units. Hence why it helps with a reduction for that general.

Madrageeeeee:
Ageee it isnt an issue really for usa or other china factions.

specifically a problem for tank vet 2 battlemasters

xezon:
I added more information and stats to the opening post. I further tweaked tank XP reduction of 20% to 30%. This should solve most problems with tank kill experiences. Whether or not China Tank Battlemaster needs further adjustments is unclear to me, but this looks like reasonable starting point.

ExiLe-ZH:
@Jundiyy 
> The xp isn't a problem for them units since the units are not used as often and not in a vet 2 state when trained, the reason it is felt by Tank is because they are the main units. Hence why it helps with a reduction for that general.

@Madrageeeeee 
> Ageee it isnt an issue really for usa or other china factions.
> 
> specifically a problem for tank vet 2 battlemasters

It is an issue for other Tanks aswell, it makes Tank spam for all factions less valuable than Vet/Gat/Quad spam. Tanks give away 100% xp, it's a crazy punishment.

Here the stats for non-vetted units:
- Vee/Gat/Quad 50xp
- Scorp 60xp
- All other Tanks 100xp

As can be seen Scorps don't have the same problem.

Tank spam generally isn't that great in ZH with the many hardcounters, reducing xp would be a nice buff without actually increasing their effectiveness in the field.

xezon:
This change was merged into main branch. All tank kills reward 30% less experience accordingly.

ExiLe-ZH:
That said, -20 or -25% might be safer to test first.

PR #413, opened 2021-09-25
Change: Lotus XP Requirements And Rewards

### LINKED ISSUE #399: Ranger, Rebel and Red Guard get different amount of XP for capturing buildings
Ranger, Rebel and Red Guard get different amount of XP for capturing buildings.

USA Ranger +15 XP
GLA Rebel + 12 XP
China Red Guard + 4 XP
China Black Lotus + ? XP

Investigate.

### LINKED ISSUE #718: China Black Lotus requires and gives more EXP than other Heros
China Black Lotus requires and gives more EXP than other Heros. Does this need tweaking?

```
ChinaInfantryBlackLotus
  ExperienceValue       = 50 100 150 400 ;Experience point value at each level
  ExperienceRequired    = 0 150 450 900  ;Experience points needed to gain each level

AmericaInfantryColonelBurton
  ExperienceValue     = 50 100 100 150  ; Experience point value at each level
  ExperienceRequired  = 0 200 300 600   ; Experience points needed to gain each level

GLAInfantryJarmenKell
  ExperienceValue     = 50 50 100 150  ;Experience point value at each level
  ExperienceRequired  = 0 100 200 400  ;Experience points needed to gain each level
```

commy2:
Fixes #718

ZH 1.04

- Lotus requires to hack (steal or capture) 8, 23 and 45 buildings to vet up, which is impossible in any serious game beyond vet 1 in very rare cases.
- A vetted Lotus grants insane amounts of experience. At vet 3, it grants as much as [18 vet 0/1 Gatling tanks] / [9 vet 1/2 Battle Masters] / [180 vet 0/1 Red Guards] / [1.5 vet 3 Overlords].

After patch:

- Lotus requires 5, 10 and 20 building hacks to vet up.
- Lotus grants the same amount of experience when killed as Burton and Jarmen Kell (which is the same as almost all non-tank like vehicles). This is the same value at vet 0 before and after the patch, but raises significantly slower when leveled up.

Notes:
- This increases her HP and gives healing at vet 2. No hack ability is made faster or gets extended range or anything else similar to that by vetting up.
- Disabling vehicles does not grant any experience before or after this patch.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17756,8 +17756,10 @@ Object Boss_InfantryBlackLotus
   MaxSimultaneousOfType = 1
   CommandSet            = ChinaInfantryBlackLotusCommandSet
 
-  ExperienceValue       = 50 100 150 400    ;Experience point value at each level
-  ExperienceRequired    = 0 150 450 900  ;Experience points needed to gain each level
+  ; Patch104p @balance commy2 25/09/2021 Reduce XP required to level up and granted when killed.
+
+  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
+  ExperienceRequired    = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable           = Yes             ;Can gain experience
   CrushableLevel        = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet            = ChinaInfantryBlackLotusCommandSet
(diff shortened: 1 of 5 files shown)


xezon:
Some love for Mrs Lotus :-)

commy2:
> Mrs 

You mean I have no chance? :<

xezon:
Love for commy2 too ❤️

xezon:
Values look reasonable and much better than original. This is safe to merge into main.

xezon (approved):

xezon:
I messed up the main branch a bit yesterday and had to force push to it. We now lost this commit in main unfortunately.

xezon:
Remerged to main with #725

PR #414, opened 2021-09-25
Change: Make Hellfire Drone Kills Reward XP Like Battle Drones

commy2:
- close https://github.com/xezon/GeneralsGamePatch/issues/398

Hellfire gets it. Scout Drone and Spy Drone don't since they are unarmed. Keeps in line with the spirit of CCG.

The amount is tiny regardless (half a Ranger / MD, a fifth of a Humvee).


xezon:
How about we give it 12 to 15 XP? It is better than the Battle Drone, so it would be reasonable to give it more XP reward for kill, right?

Jundiyy:
Think the same would be fine, it is more expensive, so makes up for it. Also it was nothing and now something which is already against USAs.

xezon (approved):

PR #425, opened 2021-09-29
Reverting some changes (other than Camouflage upgrade).

Jundiyy:
Reverting some changes

Scud Storm revert to 5 mins:
Nerf was unnecessary. The main approach of 1.04p is to buff the weaknesses. A nerf is only used in cases where an unfair advantage is present relative to other factions (Air Force Chinook e.g.). With the Build Time included the balance is now like so: 
Particle Cannon 4+1 minutes, Nuclear Missile 5+1 minutes, Scud Storm 5+2 minutes.

Air/USA Patriots slight change to $750/25s:
BT reverted, the trade is a slight price increase again to make the overall change more minimalistic and more predictable for the user.

Inf Bunker revert build time and price ($700/8s):
Nerf was unnecessary (see Scud Storm motivation).

All Raptors to $1100:
To prevent unpredictable scenarios (potential over-buff), the Vanilla Raptor price reduction has been reverted to match the King Raptor. There should no reason why this inferior unit should be more expensive than the King Raptor.

All Stealth Fighters to $1250:
Same scenario, all Stealth Fighters have been matched to the Airforce price.

Crusader Tank price revert to 1.04 ($900):
Unnecessary buff that potentially can harm balance. Already an effective unit in specific situations. 

Paladin build time revert to 1.04 (12s):
Paladin drops have proved to be quite effective vs GLA's, a build time buff was not necessary. To see better quantities on the field in later game scenarios (relative to Crusader tank), the price reduction to $1000 is maintained.

Laser Crusader revert price ($900):
Same scenario as Vanilla Crusader.

China BM slight change to $700:
Same scenario as the Vanilla Raptor/Stealth Fighter. Price has been matched with Tank's Vetted Battle Master. 

Camouflage Upgrade to $800:
$1000 still seemed like an expensive upgrade for rebels in late game scenarios. Typically the ambush was most useful to capture sniped units like Overlords, killing Hackers or base scouting (to initiate Sneak Attacks), for which level 1 was sufficient without the upgrade. With the upgrade, it might now be viable in certain situations to pick higher level Ambushes and/or use them for different tactical purposes, like 'permanent' base scouting, or making base captures more efficient.

### DIFF
--- a/Patch104pZH/Design/Changes/game_balance.md
+++ b/Patch104pZH/Design/Changes/game_balance.md
@@ -35,7 +35,7 @@
 | Red Guard         |      |      | 10 |  9   |
 | Tank Hunter       |      |      |  5 |  4.5 |
 | Hacker            |  625 |  550 | 20 | 15   |
-| Battlemaster      |  800 |  650 |    |      |
+| Battlemaster      |  800 |  700 |    |      |
 | Gattling Tank     |  800 |  700 |    |      |
 | Dragon Tank       |  800 |  700 |    |      |
 | ECM Tank          |  800 |  750 |    |      |
@@ -56,7 +56,6 @@
 | Helix            |      |      |  20 |  25   |
 | Inferno Cannon   | 1100 | 1000 |  15 |  14   |
 | Nuke Cannon      | 1600 | 1400 |  20 |  18   |
-| Fortified Bunker |  700 |  750 |   8 |   9   |
 
 ### Nuke China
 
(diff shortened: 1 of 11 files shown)


xezon (changes requested):
Please reflect these changes also in the text document:

Patch104pZH/Design/Changes/game_balance.md

Jundiyy:
Done. I forgot to add in the Air Force Stealth Fighter, so have done so now.

xezon (approved):

PR #434, opened 2021-09-30
Change: Make Lotus Hack Abort When Out Of Range

commy2:
ref https://github.com/xezon/GeneralsGamePatch/issues/415#issuecomment-931352064

Demo:

https://user-images.githubusercontent.com/6576312/135530330-8b70110b-f7c0-45dd-973a-f4fc2f582e0c.mp4


xezon:
I take 2 issues with this
* Abort range is same as attack range, which likely will behave poorly on borderline attacks.
* 150 attack range is already bad enough, but now Lotus is even weaker here, which is not good for China. China does not need nerfs.

Jundiyy:
I don't like it, the same happens with Burton and Kell. They can still plant bombs on moving vehicles. Lotus hack feels more natural compared to them. And as Xezon said, it makes it worse for Chinas, especially when their late game is quite bad most of the time.

MTKing4:
Agreed with what was said above, this is bad for china, and will render the disable feature useless, since doing it too close will get your lotus killed most players give the enemy unit time to be far enough from killing her but not too far where she can't reach

commy2:
Implemented "non-Super Lotus range buff". See: https://github.com/xezon/GeneralsGamePatch/issues/415#issuecomment-931937357

Also made AbortRange 50 dst larger than Range to handle (literal) "edge-cases". +50 dst is what is used for the Missile Defender Laser Lock.

commy2:
> I don't like it, the same happens with Burton and Kell. They can still plant bombs on moving vehicles.

Demo Charges are also inherently melee weapons, so the comparison is a bit flawed.

Issue #416, opened 2021-09-25
China Black Lotus does not get XP for attacking vehicles


xezon:
China Black Lotus does not get XP for attacking vehicles. Lotus will get XP for Cash Hack and Building Capture however. Lotus practically never reaches Vet 1, 2 or 3 in any regular Multiplayer match.

Note: GLA Jarmen Kell gets no XP for Vehicle attack either.

See also:
* #399
* #415

commy2:
You could disable one vehicle out of reach of your enemy and grind on it infinitely. Would not change this.

commy2:
According to Jundiyy, Lotus passes XP to the player as well, meaning that the above tactic could be even worse.

Jundiyy:
Maybe a misunderstanding, she passes xp over for the ones she receives for only. So she gets it for cap building and cash hack and so does the player.

Coming to this issue itself, I don't think it's a needed one.
Would prefer to keep original if not needed. 
Although not totally against it as well.

-----
Edit: okay so you meant if we enable xp for vehicle hack then the player will get it.

Stubbjax:
> 
> 
> You could disable one vehicle out of reach of your enemy and grind on it infinitely.

Which would require a lot of micro / attention as well as an inattentive opponent. Players can already achieve the same effect with GLA buildings by continuously destroying them to a hole and letting them rebuild. Nobody really has time for these things at the highest levels of gameplay where a change like this really matters.

xezon:
That would be similar to abusing the GLA Sneak Attack for Neutron attacks. It's a valid point. See
* #361

Madrageeeeee:
Wouldnt change really, not a big problem of the game i feel

MTKing4:
Could we implement it for testing anyway to see how effective it is? With the jarmen addition as well

commy2:
Will do.

ExiLe-ZH:
What is even the advantage of a vetted Lotus? Only giving off more XP when shes killed?

xezon:
More health and self heal I think. Maybe faster capture too?

ZekeDlyoung:
Not faster capture, just more health and auto-heal

xezon:
Would it be possible to change capture times on veterancy?

ExiLe-ZH:
> Would it be possible to change capture times on veterancy?

That would make Super Lotus OP though if vet 2 & 3 make her even faster

xezon:
[...]

PR #437, opened 2021-10-01
Change: Make Lotus Earn XP When Hacking Vehicles

commy2:
Reference:
* #416

This change rewards China Black Lotus Vehicle Hack with 5 XP per hack.

A Building Hack gives 20 XP as per original.

The rewarded XP gives General XP, just like building captures and unit kills do.

In theory it can be abused to grind XP when an enemy unit has been trapped. Find below hack XP time estimations based on the proposed XP value of 5. A vehicle hack takes around 5 seconds to complete.

## Black Lotus vehicle hack XP times

| Veterancy | ExperienceRequired | req. Vehicle Hack count | req. Vehicle Hack time |
|-----------|--------------------|-------------------------|------------------------|
| 1         | 100                | 20                      | 100 seconds (01:40) |
| 2         | 200                | 40                      | 200 seconds (03:20) |
| 3         | 400                | 80                      | 400 seconds (06:40) |

| General Rank | SkillPointsNeeded | req. Vehicle Hack count | req. Vehicle Hack time |
|--------------|-------------------|-------------------------|------------------------|
| 2            | 800               | 160                     | 0800 seconds (13:20) |
| 3            | 1500              | 300                     | 1500 seconds (25:00) |
| 4            | 2500              | 500                     | 2500 seconds (41:40) |
| 5            | 5000              | 1000                    | 5000 seconds (83:20) |

Farming XP with Black Lotus is unpractical. Even just reaching Rank 2 would take over 13 minutes of non stop Lotus hacking. It is simply not worth the time. We can conclude, that a Vehicle Hack XP reward around 5 is low and safe to use.

https://user-images.githubusercontent.com/4720891/182186932-7dac4887-636e-4b19-8d69-84db425529b1.mp4

## Super Hacker

For Super Hacker the XP farming becomes a different story. Since a player can build unlimited amounts of Hackers and all of them can attack the same target vehicle at the same time over and over again, the rewards multiply depending on the number of Hackers. Therefore it is not safe to grant XP rewards on Super Hacker Vehicle Hack and Building Hack.

https://user-images.githubusercontent.com/4720891/182187063-faede770-fb90-4c87-8c0c-a978a6755279.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17933,7 +17933,7 @@ Object Boss_InfantryBlackLotus
     UnpackSound             = BlackLotusUnpack
     TriggerSound            = BlackLotusTrigger
     PrepSoundLoop           = BlackLotusPrepLoop
-    AwardXPForTriggering    = 0
+    AwardXPForTriggering    = 5  ; Patch104p @balance 01/10/2021 Make ability grant XP.
     ;SkillPointsForTriggering = ???  -- Dustin, fill me out if you want different balance values.
   End
   Behavior = SpecialAbility ModuleTag_12
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
@@ -760,7 +760,7 @@ Object ChinaInfantryBlackLotus
     UnpackSound             = BlackLotusUnpack
     TriggerSound            = BlackLotusTrigger
     PrepSoundLoop           = BlackLotusPrepLoop
-    AwardXPForTriggering    = 0
+    AwardXPForTriggering    = 5  ; Patch104p @balance 01/10/2021 Make ability grant XP.
     ;SkillPointsForTriggering = ???  -- Dustin, fill me out if you want different balance values.
   End
   Behavior = SpecialAbility ModuleTag_12
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
@@ -837,7 +837,7 @@ Object Infa_ChinaInfantryBlackLotus
     UnpackSound             = BlackLotusUnpack
     TriggerSound            = BlackLotusTrigger
     PrepSoundLoop           = BlackLotusPrepLoop
-    AwardXPForTriggering    = 0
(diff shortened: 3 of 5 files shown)


commy2:
I couldn't figure out a way to make Jarmen Kell Snipe Pilot grant XP, sadly.

xezon:
I certainly would like to see China Lotus have a chance to Vet up, but given the fact this XP is added to the General Promotion Rank, this might be too much.

commy2:
> the fact

I think Jundiyy said so, but someone should make sure it applies to this ability as well.

MTKing4:
Ok, can test, but need to know how many vehicle hacks needed to get to vet 1? That's the only way to know if she's benefitting from it or not, global xp should be observed easily with the the bottom right xp bar

commy2:
@MTKing4 
8, 23 and 45 hacks to vet up

MTKing4:
here it is
[Lotus XP Farm](https://youtu.be/0TFCJDHLVHk)

commy2:
Makes sense where you landed at. Lotus needs 900 XP to reach vet 3. Player Rank 2 is 800 XP, and Rank 2 to Rank 3 is 700 XP.

xezon:
I am ok with this change, but the XP Reward needs to be much lower.
5 XP per Vehicle could be acceptable I think.
Lotus [XP steps: **0 100 200 400**] would then need

* 20 Vehicle hacks for Level 1
* 40 Vehicle hacks for Level 2
* 80 Vehicle hacks for Level 3

Lotus gives 20 XP per Building Capture

* 5 Building captures for Level 1
* 10 Building captures for Level 2
* 20 Building captures for Level 3

Lotus could trap vehicle in home base after GLA Sneak Attack for example and then farm XP off of remaining vehicles. However, I doubt this is practical, as it requires attention that will be taken away as soon as the opponent player(s) launches new attacks.

xezon:
Changed XP reward from 20 to 5.

xezon (approved):

commy2:
I actually don't like it. EMP-Patriots, Hackers etc. don't get EXP for shutting down stuff either.

xezon:
Good point about Hackers. Could add the same there. As for EMP Patriots, they don't need veterancy, so that is ok.

xezon:
I tested this change and made some relevant calculations. All relevant information has been added to the opening post. From my side this change is ok and safe to go.

commy2:
Could this be abused? Example, chain disabling an enemy unit that got inside the base, but is now kept for xp farming?

xezon:
I am confused by this question. Did you look over the tables above?

commy2:
Don't underestimate my determination.

xezon:
Ok I am still confused. I tested XP farm times as outlined above and concluded that it is not worth exploiting, because the reward is so small. Do you have a different perspective on this?

commy2:
No.
Edit: count this as approve. Can't approve your own PR's apparently:
✔️

PR #442, opened 2021-10-01
Fix: ECM Tank Cannot Disable Heroic (Vet 3) Quad Cannons

commy2:
- fix https://github.com/xezon/GeneralsGamePatch/issues/396

600 comes from the health x 2 formula used by every other unit in the game. It was 440, because Quad Cannon health used to be 220.

This fix does not include the single player only GC Quads, because those have no logic at all for ECM (will probably fix that later when those factions are to be done).

I could not find any reason to set it to 450 as suggested. They wake up in the proper order for me (vet 3 first, no vet last etc.). Following the formula makes it consistent and we avoid possible similar bugs when health bonuses vs easy/hard AI, or Hold The Line are applied on the Quad.


commy2:
https://user-images.githubusercontent.com/6576312/135621048-b42a11d2-4c97-45a7-a28f-ead3e21edf4b.mp4

Jundiyy:
Hmmm, interesting, is the only change you've done here is the 600?
If so, perhaps there was a mistake on my end (although I only remember changing to 600),if 600 produces what you have shown then yes seems good.

commy2:
Yeah, the video is shot using this branch.

xezon (approved):

PR #443, opened 2021-10-01
Fix: Scaffolds Delete Money Crates

commy2:
- ref TheSuperHackers/GeneralsGameCode#23

This PR makes scaffolds not delete money crates.

This does not affect scrap boxes from GLA.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
@@ -220,6 +220,8 @@ Object SalvageCrate
 
 End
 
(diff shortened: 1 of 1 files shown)


xezon (approved):

ThePredatorBG:
Did you consider the drawback of crates being unable to collide with buildings? I have detailed explanations: https://github.com/xezon/GeneralsGamePatch/issues/151#issuecomment-912959003
Extra changes might be desirable.

commy2:
No, and in my tests, the crates collided with buildings just fine and were collected as expected.

ThePredatorBG:
> No, and in my tests, the crates collided with buildings just fine and were collected as expected.

That's surprising to me, but if you tested crates and they do indeed collide with buildings - okay and very nice!

PR #451, opened 2021-10-03
Fix: Boss General Base Defense Issues

commy2:
ZH 1.04:

- The Boss Generals Propaganda Tower has no upgrade icon for Subliminal Messaging.
- The Boss Generals Tunnel Network is constructed 6 times faster than GLA faction Tunnels (3 times less build-time plus power).
- The Boss Generals Bunker is immune to Microwave tanks.


xezon (approved):

Jundiyy:
Whilst I have done some of these changes myself in my Boss mod, and it's a no issue to do the changes, I think you could leave the unique things for Boss unless fixing bugs (even if some things were not changed by EA and they should have been)?
I only did changes because I wanted to make it a normal army, so it's not op and can be played in 1v1s.
But for the standard Boss, leaving it with op things is fine I think. There are maps where you can pick up or are given Boss Dozers. So they can take advantage of being op.

PR #453, opened 2021-10-03
Fix: Boss General Sentry Drone Issues

commy2:
ZH 1.04:

- The Sentry Drone works slightly different than the other Sentry Drones:
-- uses `HumveeArmor` type instead of the `SentryDroneArmor`
-- does not get extra health from `DroneArmor` upgrade
-- is immune to ECMs (missing jamming logic)
-- has 10% less vision range compared to all other Sentry Drones (smaller Guard Mode area)
-- uses default stealth detection range
-- automatically attacks even when stealthed


Jundiyy:
I find it more interesting since it becomes different and unique? (I know it looks the same, but fit the 'boss theme' as it's different, although I do know it's not the best Sentry Drone out there.)
Fix if you feel like it as it is not consistent but I kinda like that for the Boss.

xezon (approved):

PR #456, opened 2021-10-03
Fix: Boss General King Raptor Issues

commy2:
ZH 1.04:

- Unexplainable Boss King Raptor vs AFG King Raptor differences:
-- uses voice lines for normal Raptor exclusively, not King Raptor
-- does not get flares from `Countermeasures` upgrade
-- has 10% less vision range compared to all other (King) Raptors (smaller Guard Mode area)

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -15853,7 +15853,7 @@ End
 
 
 
-; Patch104p @bugfix commy2 3/10/2021 Fix the following issues with the Boss General Avenger:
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
```diff
@@ -15852,6 +15852,11 @@ End
 
 
 
+
+; Patch104p @bugfix commy2 3/10/2021 Fix the following issues with the Boss General Avenger:
```
commy2:
```suggestion
; Patch104p @bugfix commy2 3/10/2021 Fix the following issues with the Boss General King Raptor:
```

xezon (approved):

PR #484, opened 2021-10-05
Fix: Tank And Nuke Outpost Use Dummy Weapon When Empty

commy2:
main branch as of today

- The Tank and Nuke Outposts can be given attack orders when they are empty, while the China and Infantry Outposts can not.

after patch:

- None of the Outposts can be given an attack order while empty.

Continuation of #4


xezon (approved):

Jundiyy:
What if you have a bunch of Outposts, some empty and some full. If you attack, the empty ones will stop moving?
If so, better to give all the ability to attack?

commy2:
> What if you have a bunch of Outposts, some empty and some full. If you attack, the empty ones will stop moving?

They won't stop, they will complete the last move order. Same as China and Infantry Outposts, or even Combat Chinooks in 1.04.

Issue #463, opened 2021-10-04
Should China Helix with Gattling Cannon be able to attack air units?


xezon:
Should China Helix with Gattling Cannon be able to attack air units? In Zero Hour 1.04 it can attack and kill air units, but only when it is perfectly angled towards enemy air craft. This behaviour currently has been removed via:
* #284

https://user-images.githubusercontent.com/4720891/135819645-90fc754b-45b9-4830-a03c-818096bdb9c0.mp4

MTKing4:
It's also important to note that helix is shooting from the bottom of its hull not from the Gatling itself, if this change ever makes the cut, this should be fixed as well

ExiLe-ZH:
Didn't even know it could do this haha. I'm not sure if this will the hurt the balance though or whether it changes the game too much?

It would be nice to give the Gatlix a broader purpose. Right now it's only used for early game rushes and stealth detection. Would be funny if it could battle stealth Commanches :D (it detects stealthy air units too?), this kinda breaks the single unit spam meta as AFG needs Raps then aswell. That said, China's anti-air capability is already proficient. 

It might also be OP when it can target Chinooks, like I can imagine BO's where China goes WF+AF and harasses both supplies, which might be hard to counter before Tow-Missiles.

I'd like to test this though.

> It's also important to note that helix is shooting from the bottom of its hull not from the Gatling itself, if this change ever makes the cut, this should be fixed as well

Would like to see this fix aswell, but isn't it done to prevent targeting bugs? It already sometimes bugs out if a unit is right beneath the Lix.

commy2:
> it detects stealthy air units too?

Yes, Helix Gatling will detect any stealthed units, including Stealth Fighters and AFG Stealth Comanches.

MTKing4:
> Would like to see this fix aswell, but isn't it done to prevent targeting bugs? It already sometimes bugs out if a unit is right beneath the Lix.

wouldn't call it a bug, more like the unit is out of shooting angle of the Gatling cannon, also it only shoots like that when bugged like in the vid above, it shoots properly in normal conditions

MTKing4:
btw also i feel like this change would be consistent with the inf helix change, since then both will be able to target air units = more viable and less unfair for the other chinas vs inf

xezon:
@ImTimK
China needs buffs compared to 1.04.

@MTKing4 
Good point regarding making it able to compete with Infantry Helix.

[...]

PR #496, opened 2021-10-06
Change: Make Helix Gatling Target Airborne Units

commy2:
- ref #463


xezon:
Works. Has powerful range. Easy AI has no chance.

* Bullets and muzzle spawns at gun
* Gun rotates to targets correctly

shot_20211006_193053_12

PR #521, opened 2021-10-07
Fix: GLA02 Campaign Stealth General Vehicle Issues

commy2:
ZH 1.04

- GLA02 Stealth General vehicles generally behaves inconsistently when stealthed.
- Some GLA02 Stealth General's vehicles are immune to ECM tanks.
- The GLA02 Stealth General's Quad Cannon builds in 20 seconds.
- The GLA02 Stealth General's Quad Cannon has lower health and is vulnerable to fire.
- The GLA02 Stealth General's Quad Cannon does not engage when stealthed.
- The GLA02 Stealth General's Quad Cannon tracers emerge from below the vehicle when sniping infantry while scrapped up.
- The GLA02 Stealth General's Technical does not engage when stealthed.
- The GLA02 Stealth General's Radar Van has the default scan range.
- The GLA02 Stealth General's Battle Bus has no buttons to drop held infantry.
- The GLA02 Stealth General's Battle Bus has disguiser ability is not functional.
- The GLA02 Stealth General's Battle Bus has less health than other Battle Busses.
- The GLA02 Stealth General's Battle Bus cannot be ordered to target enemies.
- The GLA02 Stealth General's Battle Bus has a different effect when transitioning damaged states.
- The GLA02 Stealth General's Battle Bus has no sound effect for entering or leaving infantry.
- The GLA02 Stealth General's Battle Bus has no sound effect for "power sliding".
- The GLA02 Stealth General's Battle Bus is not classified as TRANSPORTER.
- The GLA02 Stealth General's Jarmen Kell on a Bike has a longer reload time for Sniper Attack than on foot.


commy2:
Battle Bus Disguiser is kind of interesting. Works like a Bomb Truck, but for the Bus. Unit will not reveal even when passengers are shooting... It was half implemented, so I did it completely. Could be useful for Casino like maps.

commy2:
https://user-images.githubusercontent.com/6576312/136462183-90e59927-856d-4f62-ab81-5c779371bef3.mp4

xezon (approved):
disguise ability looks funny.

Issue #482, opened 2021-10-05
Delayed kill of vehicle after napalm mig attack


Madrageeeeee:
Sometimes units remain alive after a napalm mig attack but then 2 seconds later they miraculously die anyway due to some sort of afterburning effect.

Unsure whether this is intented. Looks like weird game behavior though due to the delay and because no visual damage is being delivered anymore to the vehicle.

You often see this with either techs or vees. 

Not 100% sure but think this can also happen with flamer attacks.

Jundiyy:
This is the burn effect on units, they burn for n seconds.
```
Behavior = FlammableUpdate ModuleTag_09
AflameDuration = 5000         ; If I catch fire, I'll burn for this long...
AflameDamageAmount = 3       ; taking this much damage...
AflameDamageDelay = 500       ; this often.
End```

commy2:
Is there no visual effect for this? I know there're hooks that we could add additonal effects, but I always figured there already were visuals.

Madrageeeeee:
Perhaps my visual settings were / are too low but to me it always seems as if they randomly died. 

Verstuurd vanaf mijn iPhone

xezon:
If I recall correctly there are small fire effects on burning vehicles. Can be tested easily by shooting onto car with Mig Fighter.

Stubbjax:
https://user-images.githubusercontent.com/11547761/136356332-53a2ab03-ac21-4d00-89b3-5236dce0a4df.mp4

xezon:
Could you try the same on GLA Quad Cannon?

Stubbjax:
> 
> 
> Could you try the same on GLA Quad Cannon?

No fire effects there either.

xezon:
I suspect it is possible to add fire effects. But is it cheap?

Madrageeeeee:
Ahh so ragey not so crazy after all 😁 some small flames would already help people understand the prolonged effect of burning. And would look sickk (hopefully)

guess similar to the nuclear powerplant that slowly dies if it is turned on self-destruction mode. But then with flames?

commy2:
Testing possible visuals:


https://user-images.githubusercontent.com/6576312/136550275-0b7cbce5-a651-4bbc-ab47-f30f12b37a9b.mp4


https://user-images.githubusercontent.com/6576312/136550286-9beefb13-ae32-4acc-bda1-cb6864d39e06.mp4

commy2:
[...]

PR #526, opened 2021-10-08
Add: Fire Visuals For Aflame Units

commy2:
- ref #482

This adds visuals for the burn damage effect that happens in 1.04.

Only for vGLA Quad and vUSA Humvee / Ambo atm.


MTKing4:
group of aflame quads performance test

https://user-images.githubusercontent.com/83613147/136849408-17a321b2-4c63-4bf8-b2bc-f693efec330e.mp4

PR #538, opened 2021-10-09
Fix: Powered Arms Dealer Door Still Open When Chain Building Technicals and Combat Bikes

commy2:
- ref https://github.com/xezon/GeneralsGamePatch/issues/464#issuecomment-939258979

ZH 1.04:

https://user-images.githubusercontent.com/6576312/136654501-4f67f10d-6347-4ec0-91c1-8e45ddf80313.mp4

Patch:

https://user-images.githubusercontent.com/6576312/136654508-3c266b92-7962-4d4a-82c3-14f9fabd22a2.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLABuildings.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLABuildings.ini
@@ -4316,11 +4316,14 @@ Object GC_Chem_GLAArmsDealer
   Behavior             = FXListDie ModuleTag_09
     DeathFX       = FX_StructureMediumDeath
   End
+
+  ; Patch104p @bugfix commy2 09/10/2021 Fix door not closing in time when building Technicals or Combat Bikes with power.
+
   Behavior = ProductionUpdate ModuleTag_10
     NumDoorAnimations            = 1
     DoorOpeningTime              = 2000  ;in mSeconds
-    DoorWaitOpenTime             = 3000  ;in mSeconds
-    DoorCloseTime                = 2000  ;in mSeconds
+    DoorWaitOpenTime             = 2000  ;in mSeconds
+    DoorCloseTime                = 1800  ;in mSeconds
     ConstructionCompleteDuration = 1500  ;in mSeconds
   End
   Behavior = DefaultProductionExitUpdate ModuleTag_11
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -10420,11 +10420,14 @@ Object Chem_GLAArmsDealer
   Behavior             = FXListDie ModuleTag_09
     DeathFX       = FX_StructureMediumDeath
   End
+
+  ; Patch104p @bugfix commy2 09/10/2021 Fix door not closing in time when building Technicals or Combat Bikes with power.
+
   Behavior = ProductionUpdate ModuleTag_10
     NumDoorAnimations            = 1
     DoorOpeningTime              = 2000  ;in mSeconds
-    DoorWaitOpenTime             = 3000  ;in mSeconds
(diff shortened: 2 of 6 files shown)


commy2:
Aside from visuals, this means that Technical build time is 12 sec w/o power and 7 sec w/ power now always.

In 1.04, it is - with power - 7 seconds for the first Technical only, and 5 seconds starting from the second Technical.

Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLABuildings.ini
```diff
@@ -4316,11 +4316,14 @@ Object GC_Chem_GLAArmsDealer
...
   End
+
+  ; Patch104p @bugfix commy2 09/10/2021 Fix door not closing in time when building Technicals or Combat Bikes with power.
+
   Behavior = ProductionUpdate ModuleTag_10
     NumDoorAnimations            = 1
     DoorOpeningTime              = 2000  ;in mSeconds
-    DoorWaitOpenTime             = 3000  ;in mSeconds
-    DoorCloseTime                = 2000  ;in mSeconds
+    DoorWaitOpenTime             = 2000  ;in mSeconds
```
xezon, line 4325: Hmm. Wouldn't it have been enough to just decrease `DoorWaitOpenTime` by a few milliseconds? I don't understand why 2 variables had to be changed. Plus now Door close is quicker than door open. Why?

commy2: > Plus now Door close is quicker than door open. Why?

The time it takes for the door to close is animation based, so part of the model. `DoorCloseTime` is used to make the logic register that the door was already closed. You could set `DoorCloseTime` to 2 minutes and it wouldn't speed up / slow down the animation at all.

> Wouldn't it have been enough to just decrease `DoorWaitOpenTime` by a few milliseconds?

The game is not precise enough for a few msecs to mean anything. It runs at 30 FPS in multiplayer, so to make an actual difference, you need at least 33.33 msecs (= 1 frame).

commy2: This may help:



Edit: oops, replace "fps" with "milliseconds per frame". 1/30 FPS = 0.033 seconds per frame.

commy2: Condition for not bugging out door opening for fast building units:
DoorWaitOpenTime + DoorCloseTime < (fastest) BuildTime
2000 msec + 1800 msec < 4.0 sec (Combat Bike w/ power)
3800 msec < 4000 msec, check

Condition for not bugging out door animation:
\<animation frame count\> / 30 FPS < DoorCloseTime
50 (Arms Dealer) / 30 FPS < 1800 msec
1667 msec < 1800 msec, check

commy2: `DoorOpeningTime` and `DoorCloseTime` do not determine the speed of the door animation. To not bug out the visuals, these must be bigger than the actual door animation time though.

`DoorOpeningTime` is important in that it adds a constant to actual build-time of units.

`DoorCloseTime` and `DoorWaitOpenTime` are less important, but must together be smaller than the fastest `BuiltTime` of any unit, otherwise the `DoorOpeningTime` delay is skipped.

If necessary, the door animations can be adjusted by using `AnimationSpeedFactorRange` in the `W3DModelDraw` module. That parameter can be used indepentently for door opening and closing.

xezon: Ok now it makes sense. Bike build is 4000 ms. I was wondering why door close time is combined 3800 ms when Technical build is 5000 ms. What you refer to as "milliseconds per frame" is called "frame time" in programmer terms.

xezon (approved):
Thank you for detailed investigation and explanations!

PR #539, opened 2021-10-09
Fix: Grant Faction Inherent Upgrades By Dummy Instead Of By Building

commy2:
ZH 1.04

https://user-images.githubusercontent.com/6576312/136674410-5018de82-8676-4226-9e5c-ab0c7167615e.mp4

- Building an Infantry General Command Center, Barracks or War Factory grants Nationalism upgrade.
- Building a Tox Command Center grants Toxin Shells and Anthrax Beta upgrades.
- Building a Demo Command Center grants Booby Traps upgrade.
- Building a Stealthed Rebel grants Camouflage upgrade.

After patch:

- Only Infantry General spawns with Nationalism.
- Only Tox General spawns with Toxin Shells and Anthrax Beta.
- Only Demo General spawns with Booby Traps.
- Only Stealth General spawns with Camouflage upgrade.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
@@ -298,6 +298,7 @@ PlayerTemplate FactionChinaInfantryGeneral
   DisplayName       = INI:FactionChinaInfantryGeneral
   StartingBuilding  = Infa_ChinaCommandCenter
   StartingUnit0     = Infa_ChinaVehicleDozer
+  StartingUnit1     = ChinaNationalism_UpgradeDummy
   ScoreScreenImage  = China_ScoreScreen
   LoadScreenImage   = SNFactionLogoPage_China
   LoadScreenMusic   = Load_China
@@ -368,6 +369,8 @@ PlayerTemplate FactionGLAToxinGeneral
   DisplayName       = INI:FactionGLAToxinGeneral
   StartingBuilding  = Chem_GLACommandCenter
   StartingUnit0     = Chem_GLAInfantryWorker
+  StartingUnit1     = GLAToxinShells_UpgradeDummy
+  StartingUnit2     = GLAAnthraxBeta_UpgradeDummy
   ScoreScreenImage  = GLA_ScoreScreen
   LoadScreenImage   = SUFactionLogoPage_GLA
   LoadScreenMusic   = Load_GLA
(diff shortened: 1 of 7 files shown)


xezon:
Good from my side. Would prefer @alanblack166 looks over.

commy2:
hanfi:
> add at least lifetime 100 to the objects and NO_COLLIDE KINDOF

I just wonder why.

Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
```diff
@@ -368,6 +369,8 @@ PlayerTemplate FactionGLAToxinGeneral
   DisplayName       = INI:FactionGLAToxinGeneral
   StartingBuilding  = Chem_GLACommandCenter
   StartingUnit0     = Chem_GLAInfantryWorker
+  StartingUnit1     = GLAToxinShells_UpgradeDummy
+  StartingUnit2     = GLAAnthraxBeta_UpgradeDummy
```
xezon, line 373: On second inspection, it looks odd. Upgrades are not really units, right? So I assume we abuse a system here that was originally not intended to do that. Or does retail game do a similar thing somewhere?

Is there a limit to `StartingUnitX` ? If yes, is it a problem to add a number of new starting units here? If not, should we use a higher starting number?

Can `StartingUnitX` be used from map scripts? If yes, would a `StartingUnit1` on a script map now collide with this new entry? Would it crash or break?

commy2: We abuse the system.

There is a limit to StartingUnit#, but I do not know what it is.

> Can StartingUnitX be used from map scripts? If yes, would a StartingUnit1 on a script map now collide with this new entry?

Yes to both.

> Would it crash or break?

It would make the faction not get the upgrade.

xezon: Ok this needs more consideration then. We potentially break behavior on custom script maps here. Is there another way to do this without such side effect? Hanfield says StartingUnit goes from 1 to 9.

commy2: Maybe replace the CommandCenter (StartBuilding) with a dummy that replaces itself to the coresponding Command Center using `ReplaceObjectUpgrade` and also grants the upgrades. That would likely block building Workers / Dozers from the loading screen though.

Maybe a copy paste of the whole Command Center that gives the upgrades, but is different from the one you build by Dozer.

Maybe just using StartingUnit8 and StartingUnit9.

xezon: Is it possible to bundle multiple units into `StartingUnit1`, so we just use this one slot?

commy2: Sure. I would've prefered to keep this modular though. One dummy, one upgrade. No problem though.

commy2: Point against anyone using these in a map.ini is, that you may as well just place units next to the Command Center and transfer them by Worldbuilder script. I think the vast majority goes that route.

xezon: Is it also possible to bundle all in `StartingUnit0` ? So the worker unit + the upgrades?

PR #542, opened 2021-10-10
Fix: Inconsistent Black Market Particle Effects Between Factions And Fake Buildings

### LINKED ISSUE #469: GLA Fake Buildings use slightly wrong models when in damaged state
GLA Fake Buildings use slightly wrong models when in damaged state. This has been fixed in NProject Mod.

commy2:
- ref #469


xezon (approved):
Jesus Christ. I hope there are not much more cases like that.

Issue #455, opened 2021-10-03
GLA Saboteur can be used to detect Fake GLA Markets


xezon:
GLA Saboteur can be used to detect Fake GLA Markets. Should GLA Saboteur be able to steal cash from GLA Black Market and USA Supply Drop Zone?

commy2:
This does not work with the Stealth General Saboteur, since he cannot enter any Fake buildings. As reverse though, that is how you can tell apart fake and real factories, supplies and command centers as Stealth General.

Madrageeeeee:
think that sounds like a nice trick to tell them apart. Not really abusable in competitive matchplay id say. and if you do take the effort its a nice trick.

allowing them to steal cash from secondary eco kinda makes sense and increases their usefulness

commy2:
Fixed by the combination of https://github.com/TheSuperHackers/GeneralsGamePatch/pull/547 and https://github.com/TheSuperHackers/GeneralsGamePatch/pull/544.

ExiLe-ZH:
> think that sounds like a nice trick to tell them apart. Not really abusable in competitive matchplay id say. and if you do take the effort its a nice trick.
> 
> allowing them to steal cash from secondary eco kinda makes sense and increases their usefulness

In team games it could be quite useable I would say, it just costs $800 to do so, but it would help your USA (or China) mate alot to determine where Genpowers should be used.

PR #544, opened 2021-10-10
Fix: Stealth General Saboteur Cannot Disable Internet Centers And Destroy Fake Buildings

commy2:
- ref #455

ZH 1.04

- The Stealth Generals Saboteur cannot enter Internet Centers. The normal GLA Saboteur can enter Internet Centers and will disable them by doing so for 15 seconds.
- The Stealth Generals Saboteur cannot enter Fake Buildings. The normal GLA Saboteur however can: the fake building will be destroyed by that.
- Since the Stealth Generals Saboteur cannot enter Fake Buildings, the cursor that appears when hovering above an enemies GLA Barracks, Arms Dealer, Supply Stash or Command Center can be abused to tell apart real and fake buildings.

After patch:

- Both the normal and the Stealth Saboteur can enter Internet Centers and Fake buildings.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -16165,6 +16165,9 @@ Object Demo_GLAInfantrySaboteur
     BuildingPickup = Yes
     SabotageDuration = 30000
   End
+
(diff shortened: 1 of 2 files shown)


xezon (approved):
Nice.

commy2:
This enables the Stealth Saboteur to scout fake/real Black Markets by cursor like the vanilla GLA Saboteur can. That'd be fixed by #547 though.

PR #547, opened 2021-10-10
Change: Let Saboteurs Enter Supply Drop Zones And Black Markets (Steal $1000)

commy2:
- ref #455
- works best with: #544

ZH 1.04

- Saboteurs cannot enter Supply Drop Zones and Black Markets (among other buildings).
- Since vanilla GLA Saboteurs can enter Fake buildings, including Fake Black Markets, but cannot enter Black Markets, hovering with the cursor above enemy Black Markets can be used to tell fake and real ones apart.

After patch:

- All Saboteurs can enter Supply Drop Zones and Black Markets. They steal $1000 if they do so, similar to how they steal $1000 when entering a Supply Center or Supply Stash.
- The cursor can no longer be abused to tell apart real and fake Black Markets.

Sadly, the best way to fix the Saboteur cursor scouting bug is by doing this game design change. I think it's fine in this case, because Saboteurs are kind of lame, and them not being able to enter secondary eco stinks. I realize however that this is a highly controversial change.

Technical aspects: `FS_SUPPLY_DROPZONE` KindOf is safe to use here. It is only used by the engine to validate targets for the `SabotageSupplyDropzoneCrateCollide` module. 

<details><summary>chatlog with hanfi about FS_SUPPLY_DROPZONE</summary>

```
commy2
Where in engine is FS_SUPPLY_DROPZONE used?
I assume it is related to SabotageSupplyDropzoneCrateCollide. However, my question is if there're any other things that this is used for?

hanfield
only there

commy2
Splendid.
Where is FS_BLACK_MARKET used then? There does not seem to be a related SabotageXXX module.

hanfield
stealthupdate
BlackMarketCheckDelay
combined with NO_BLACK_MARKET forbiddenstealthcondition

commy2
BlackMarketCheckDelay does not seem to be used in 1.04.

hanfield
it works
but it's not used

commy2
Thank you.
Seems interesting. NO_BLACK_MARKET and FS_BLACK_MARKET.
```

</details>

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -2403,7 +2403,10 @@ Object Chem_GLABlackMarket
 
   ; *** ENGINEERING Parameters ***
   RadarPriority       = STRUCTURE
-  KindOf              = PRELOAD STRUCTURE SELECTABLE IMMOBILE CAPTURABLE FS_TECHNOLOGY MP_COUNT_FOR_VICTORY SCORE_CREATE FS_BLACK_MARKET
+
+  ; Patch104p @bugfix 10/10/2021 Add FS_SUPPLY_DROPZONE to let Saboteurs enter the building and steal cash.
+
+  KindOf              = PRELOAD STRUCTURE SELECTABLE IMMOBILE CAPTURABLE FS_TECHNOLOGY MP_COUNT_FOR_VICTORY SCORE_CREATE FS_BLACK_MARKET FS_SUPPLY_DROPZONE
   Body                = StructureBody ModuleTag_04
     MaxHealth         = 1000.0
     InitialHealth     = 1000.0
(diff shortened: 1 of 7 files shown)


commy2:
Demo:


https://user-images.githubusercontent.com/6576312/136712165-c80882df-bba4-452d-98cd-78a13b177863.mp4


https://user-images.githubusercontent.com/6576312/136712168-45b02a79-f62d-4833-9631-c29398d95bd3.mp4

xezon (approved):

Jundiyy:
Looks good, RIP my secret Black Market trick. 😭 

I don't think it should be a problem, makes Sabo's a bit more versatile but let's see what others say.

Jundiyy:
Just thought of a problem. By doing this, it gives GLAs a 50/50 chance, they might want to steal money but instead end up blowing up a Fake Market.

xezon:
I think that is ok. It is a fake market or real market. Attacking player is not supposed to know what he will get. It's the same when you shoot with a SCUD Storm on markets. You will only know afterwards if it was worth the attack.

Jundiyy:
Yes that is true, didn't think of that. Then I suppose it's fine.

PR #551, opened 2021-10-11
performance test map for flame particles

MTKing4:
added test map for performance test on flame particles from Commit #526
25 Group of Quads burning endlessly, i recommend testing on other hardware as well.

here's my test footage

https://user-images.githubusercontent.com/83613147/136850458-c312c764-e7ac-4fcf-8cd6-2be1b34382e2.mp4


commy2:
Can you change the base branch to `burn`? So that the effects are actually in this.

MTKing4:
You mean i push the commit to your branch?

xezon:
I think it's fine to have it separate. Afterall, map needs to be tested without flame particles too. When testing flames in **burn** branch, we can just cherry pick the map change.

xezon (approved):

PR #552, opened 2021-10-12
Remove LICENSE

xezon:
Let's remove LICENSE file until we figure out what to do with it.


commy2 (approved):

commy2:
GPLv2 is a popular one. I do not know if that would apply for this though, since we're basing this off of proprietary game files from 1.04.

As far as I am concerned, my stuff is free.

PR #568, opened 2021-10-16
Fix: Reinforcement Pad Is Deleted When Destroyed; It And Repair Bay Do Not Leave Ruins Like Other Tech Buildings

commy2:
ZH 1.04

https://user-images.githubusercontent.com/6576312/137589619-70d8d766-8cd6-4f1e-91bd-de8a5875df6d.mp4

After patch:

https://user-images.githubusercontent.com/6576312/137589641-214469a1-08aa-471a-91bd-892b0bc27c76.mp4

---

This happens because while all tech buildings have a `KeepObjectDie` module that is supposed to spawn a ruin, the Reinforcement Pad only has a conflicting `DestroyDie` module as well that deletes the building when destroyed.

This also applies to the unused Repair Pad (not to be confused with the Repair Station), which is also fixed by this PR.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/TechBuildings.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/TechBuildings.ini
@@ -447,8 +447,10 @@ Object TechReinforcementPad
     FactionOCL   = Faction:FactionBossGeneral        OCL:OCL_BossGen_ReinforcementPadCHIVehicle
   End
 
-  Behavior = DestroyDie ModuleTag_06
-  End
+  ; Patch104p @bugfix commy2 16/10/2021 Fix building disappears when destroyed instead of leaving ruin.
+
+  ;Behavior = DestroyDie ModuleTag_06
+  ;End
 
   Behavior            = DefaultProductionExitUpdate ModuleTag_07
     NaturalRallyPoint = X: 0.0  Y:-60.0  Z:0.0 ;NaturalRallyPointX must always match GeometryMajorRadius! -ML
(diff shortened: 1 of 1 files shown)


Jundiyy:
👍

xezon:
Hmm. This does not look proper. Killing Tech building leaves a relatively flat model. Here it looks almost as if it is still alive and could be repaired. It can cause user confusion. It would fine if it looked like the other tech buildings.

shot_20211016_154713_3

commy2:
Compare it with the Repair Bay.

xezon:
Which map is that?

commy2:
Cairo Commandos (3) comes to mind. Has both in center (as well as some oils).

xezon:
I see. The Repair Bay has the same behavior as this change. By this standard the change would be ok. I wonder if it was possible to make a follow up change where these building would spawn a different model, similar style to Oil Refinery. Maybe Enlima29 already has something like that in NProject Mod ready for grabs.

shot_20211016_172549_1
.

commy2:
That would be preferable, since the destroyed Repair Bay can indeed be confusing. We'd need the assets for that though, no such model exists.

alanblack166:
Shockwave/Rise of the Reds has a rubble model:
image

alanblack166:
Also for Reinforcement Pad:
image

xezon:
That looks much better!

commy2:
Very good. But who made them and are we allowed to use them?

alanblack166:
Those exist in the game already:
ZBLndBay_R
ZBLndBay_RS
ZBRprBay_R
ZBRprBay_RS

commy2:
Thank you for the tip. I implemented those models in this branch, and I think they look awesome.

commy2:
- [ ] don't squash

xezon (approved):

Jundiyy:
Looks great.

PR #570, opened 2021-10-16
Experimental: Let Pilots Enter Comanches

commy2:
ZH 1.04:

Pilots can enter Comanches, but only while they are landed for repairs.

Shockwave:

Pilots can be ordered to enter Comanches. The Comanche will land. If the Pilot walks into the landed Comanche, it will be promoted.

However if the Comanche lands on the Pilot standing below the Comanche, the Pilot will enter the Comanche without promoting it.

There is a fix in SW called `PilotNotShowOnVehicleWhenEnterAttackingVehicle` which in this case prevents the Pilot from being drawn inside the Comanche. However, it will not fix the underlying issue.

I don't like the fix in it's current state. To get this to work properly, we likely need Thyme. Putting this here anyway, as maybe someone smarter can come up with a good solution.

The `PilotNotShowOnVehicleWhenEnterAttackingVehicle` hackfix is commented out for easier debugging.


xezon:
Branch rebased. Yes fix does not look good in its current form. Is there no other way to put pilots gracefully into helicopters?

commy2:
I don't think so. This is the closest I've managed and have seen from others, and it's not good enough. Would require engine changes to fix it / add this feature properly.

PR #572, opened 2021-10-16
Add: Mammoth Tank Default Implementation With Textures From CCG Alpha Version

commy2:
Zero Hour has the model for a Mammoth tank, which emerged somewhere in the Alpha version of the game. The model can be made usable with map.ini scripting, however, it will lack textures and therefore glow pink.

<details><summary>vZH Mammoth tank model</summary>

mamm0

</details>

The textures are still in the alpha builds of the game. Maybe we could add them to the patch to make the model look not quite as out of place?

<details><summary>with alpha texture</summary>

</details>


xezon:
How much size do the textures add in a normal ZIP Archive?

commy2:
I guess like 300 kiBi?

xezon:
Zip: 204 kb
7z: 168 kb

I think it is fine.

Why is this change controversial?

commy2:
> Why is this change controversial?

Idk, filesize I guess and some may wonder what the point is.

xezon:
Is the tank fully functional in the game and can be used in script maps?

commy2:
> Is the tank fully functional in the game

No.

> and can be used in script maps?

It can be implemented via map.ini. I have seen Legionnaire play on fun maps with it, but it sadly was all pink there.

Edit: with this change, it would automatically get the texture, since the model points to these texture paths.

xezon:
Maybe it would be worthwhile to give the tank some common tank behavior as well then so it is usable from the get go?

commy2:
Done. Gave it Overlord weapon and voice. Also made it crush tanks like an Overlord. Ingame name is "Militia Tank" until we can give it the proper name.

xezon (approved):

Jundiyy:
Looks good, I see you have pilot eject sounds but no pilot eject module?

I think it would be nice if the stats are a bit different to the Overlord. 
See if you like anything from here, there is one thing I would really like to change though, the build time. Adding this unit to a fun map would mean it would take a very long time for GLA to get the unit. I think for fun units there should be a middle ground for build time, otherwise GLA players would always find it hard to use them due to no power.
Perhaps 25% off the would-to-be-normal time. So in this case, 15s. (of course it means other armies get it faster but at least GLA doesn't have to wait so long to use them.)

    Weapon               = PRIMARY MammothTankGun

  BuildTime              = 15.0          ;in seconds

    MaxHealth       = 900
    InitialHealth   = 900

      TurretTurnRate       = 90   // turn rate, in degrees per sec

Locomotor MammothTankLocomotor

Weapon MammothTankGun
  PrimaryDamage = 60 ;100.0
  PrimaryDamageRadius = 5.0
  ScatterRadiusVsInfantry = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
  SecondaryDamage = 20.0
  SecondaryDamageRadius = 10.0
  AttackRange = 170.0
  MinTargetPitch = -15                         ; we may not target anything outside of this pitch range
  MaxTargetPitch = 15                          ; ditto
  DamageType = ARMOR_PIERCING
  DeathType = NORMAL
  WeaponSpeed = 300                           ; dist/sec 
  WeaponRecoil = 5                            ; angle to deflect the model when firing
  ProjectileObject = OverlordTankShell
  FireFX            = WeaponFX_GenericTankGunNoTracer
  VeterancyFireFX   = HEROIC WeaponFX_HeroicGenericTankGunNoTracer
  ProjectileDetonationFX = WeaponFX_GenericTankShellDetonation
  FireSound = OverlordTankWeapon
  RadiusDamageAffects = ALLIES ENEMIES NEUTRALS
  DelayBetweenShots = 300              ; time between shots, msec
  ShotsPerBarrel = 1                   ; By default, shoot one shot per barrel
  ClipSize = 2                    ; how many shots in a Clip (0 == infinite)
  ClipReloadTime = 2000              ; how long to reload a Clip, msec
  WeaponBonus = PLAYER_UPGRADE DAMAGE 125% ; UraniumShells

  ; note, these only apply to units that aren't the explicit target 
  ; (ie, units that just happen to "get in the way"... projectiles
  ; always collide with the Designated Target, regardless of these flags
  ProjectileCollidesWith = STRUCTURES WALLS 
End

Locomotor MammothTankLocomotor
  Surfaces            = GROUND
  Speed               = 23   ; in dist/sec
  SpeedDamaged        = 20   ; in dist/sec
  TurnRate            = 90   ;25   ; in degrees/sec
  TurnRateDamaged     = 90   ;20   ; in degrees/sec
  Acceleration        = 150   ; in dist/(sec^2)
  AccelerationDamaged = 150   ; in dist/(sec^2)
  Braking             = 150   ; in dist/(sec^2)
  MinTurnSpeed        = 0    ; in dist/sec
  ZAxisBehavior       = NO_Z_MOTIVE_FORCE
  Appearance          = TREADS

  AccelerationPitchLimit = 2              ; Angle limit how far chassis will lift or roll from acceleration.
  DecelerationPitchLimit = 2              ; Angle limit how far chassis will dip from deceleration.
  PitchStiffness = 0.05                   ;  stiffness of the "springs" in the suspension forward & back.
  RollStiffness = 0.05                    ;  stiffness of the "springs" in the suspension side to side.
  PitchDamping = 0.8                      ;  How fast it damps.  0=perfect spring, bounces forever.  1=glued to terrain.
  RollDamping = 0.3                       ;  How fast it damps.  0=perfect spring, bounces forever.  1=glued to terrain.
  ForwardAccelerationPitchFactor = 0.5    ; How much acceleration will cause the front to lift, or dip for stops.
  LateralAccelerationRollFactor = 0.33    ;  How much cornering will cause the chassis to roll.
End

PR #575, opened 2021-10-17
Change: Tank Nuke Cannon and Inferno Cannon

Jundiyy:
Mappers can now add Nuke Cannons and Inferno Cannons to the Tank General War Factory.

First they must use the CommandButton script to add these units to the War Factory, then they must also place down a TechWarFactory per player (off map is fine) and script them so they are given to each player. 

The second route is to add the CommandButtons scripts then add scripts to ignore prerequisites.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
@@ -16640,3 +16640,484 @@ Object Tank_ChinaCommandCenter
 
 End
 
+;------------------------------------------------------------------------------
+;Also called NukeCannon
+Object Tank_ChinaVehicleNukeLauncher
+
+  ; *** ART Parameters ***
+  SelectPortrait         = SNNukeCannon_L
+  ButtonImage            = SNNukeCannon
+
+  UpgradeCameo1 = Upgrade_ChinaNeutronShells
+  ;UpgradeCameo2 = NONE
+  ;UpgradeCameo3 = NONE
+  ;UpgradeCameo4 = NONE
+  ;UpgradeCameo5 = NONE
+
+  Draw = W3DTankDraw ModuleTag_01
+
+    InitialRecoilSpeed   = 120
+    MaxRecoilDistance    = 8
+    RecoilSettleSpeed    = 6
+
+    OkToChangeModelColor = Yes
+
+    ExtraPublicBone = Turret01
+
+    DefaultConditionState
+      Model                           = NVNukeCn
+      WeaponLaunchBone                = PRIMARY Muzzle
+      WeaponMuzzleFlash               = PRIMARY MuzzleFX
+      WeaponFireFXBone                = PRIMARY Muzzle
+      WeaponRecoilBone                = PRIMARY Barrel
+      WeaponLaunchBone                = SECONDARY Muzzle
+      WeaponMuzzleFlash               = SECONDARY MuzzleFX
+      WeaponFireFXBone                = SECONDARY Muzzle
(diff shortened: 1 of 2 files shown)


commy2:
Couldn't the map maker not just do:
```map.ini
CommandSet Tank_SCIENCE_CHINA_CommandSetRank1
  3 = Command_PurchaseScienceNukeLauncher
End

CommandSet Tank_ChinaWarFactoryCommandSet
  9  = Command_ConstructChinaVehicleInfernoCannon
  10 = Command_ConstructChinaVehicleNukeLauncher
End

Object ChinaVehicleInfernoCannon
  Prerequisites
    Object = ChinaPropagandaCenter Tank_ChinaPropagandaCenter
  End
End

Object ChinaVehicleNukeLauncher
  Prerequisites
    Object = ChinaPropagandaCenter Tank_ChinaPropagandaCenter
    Science = SCIENCE_NukeLauncher
  End
End
```
That should work in 1.04 already.

Jundiyy:
This would require map.ini, which not everyone knows or a mapper may not want due to having to restart on transfer and after game.

Map makers can currently add any unit but have to remove the prerequisites through scripts, which means units can be built without tech.

The purpose of these units and some others is to allow a simple 'drag and drop', otherwise we could just ask all mappers to learn/use map.ini and no need to do changes. It's more of ease of access for them, find the units in WB and use them.

xezon:
With current map.ini tech it is desirable to reduce map.ini use cases, because it can cause mismatching in successive map loads.

commy2:
And doesn't work immediately on transferred maps.

I think that is the underlying issue here and should be addressed, possibly by Thyme.

xezon:
Yes. For current developments however I believe we should not rely on Thyme and cater to original game executable as usual.

xezon (changes requested):
Add Linebreak at end of file(s) and add comments to new added features.

Jundiyy:
Done.

Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
```diff
@@ -16640,3 +16640,486 @@ Object Tank_ChinaCommandCenter
...
+    TreadAnimationRate      = 4.0  ; amount of tread texture to move per second
+  End
+
+  ; ***DESIGN parameters ***
+  DisplayName      = OBJECT:NukeLauncher
+  Side = ChinaTankGeneral
+  EditorSorting   = VEHICLE
+  TransportSlotCount = 10                 ;how many "slots" we take in a transport (0 == not transportable)
+
+  ; Patch104p @bugfix commy2 09/09/2021 Prevent exploit that allows Neutron Shells without upgrade.
```
commy2: Would remove this, as that change is not made by this patch.

Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
```diff
@@ -16640,3 +16640,486 @@ Object Tank_ChinaCommandCenter
...
+    AflameDuration = 5000         ; If I catch fire, I'll burn for this long...
+    AflameDamageAmount = 3       ; taking this much damage...
+    AflameDamageDelay = 500       ; this often.
+  End
+
+  Behavior = ProductionUpdate ModuleTag_12
+    MaxQueueEntries = 1; So you can't build multiple upgrades in the same frame
+  End
+
+  ; Patch104p @bugfix commy2 09/09/2021 Prevent exploit that allows Neutron Shells without upgrade.
```
commy2: ditto

xezon (approved):

PR #581, opened 2021-10-20
Repair Drone edits and fixes.

Jundiyy:
Repair Drone now repairs units and can be built using a Command Button on fun maps.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
@@ -1174,10 +1174,15 @@ End
 
 
 ;------------------------------------------------------------------------------
+; Patch104p Jundiyy 20.10.2021 Edited to make the unit more useful.
 ; **data here is a copy and paste of AmericaVehicleBattleDrone**
 Object AmericaVehicleRepairDrone
 
   ; *** ART Parameters ***
+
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
```diff
@@ -1192,54 +1197,64 @@ Object AmericaVehicleRepairDrone
...
+    SubdualDamageHealAmount = 50
   End
 
-  Behavior = AIUpdateInterface ModuleTag_03
-    Turret
-      TurretTurnRate = 60   // turn rate, in degrees per sec
-      ControlledWeaponSlots = PRIMARY
-    End
-    AutoAcquireEnemiesWhenIdle = Yes
+  Behavior = AIUpdateInterface ModuleTag_03 ;Doesn't need a turret.
```
xezon, line 1252: I think we should prefix all our own comments with `Patch104p` keyword, so we know for sure they are ours. Someone may wonder if "Doesn't need a turret." is a comment from EA later on.

Jundiyy: Ah okay, yea sure will do, thought since I titled it with the keyword it was enough.

Jundiyy:
I think in the case of this unit, I will just make a carbon copy of it and have the edits on that, just in case the unit is being used on some maps and it breaks them.

PR #582, opened 2021-10-20
Amphibious Transport edits and fixes.

Jundiyy:
Amphibious Transport can now be built on fun maps, some edits to make it better.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -8297,3 +8297,13 @@ CommandButton Command_ConstructTechRadioStation
   ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
   DescriptLabel           = OBJECT:TechRadioStation
 End
+
+CommandButton Command_ConstructAmphibiousTransport ; Patch104p Jundiyy 20.10.2021 Allow to train this unit.
+  Command       = UNIT_BUILD
+  Object        = AmphibiousTransport
+  TextLabel     = CONTROLBAR:ConstructAmericaVehicleMedic ;Could be changed if we have something else, uses the A button as a shortcut.
+  ButtonImage   = SSFerry_L
+  ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
+  DescriptLabel           = CONTROLBAR:ToolTipUSABuildMedic
+End
+
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -7766,10 +7766,16 @@ Object Cin_CombatCycleByElements
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p Jundiyy 20.10.2021 Edited to make it more useful.
 Object AmphibiousTransport
 
   ; *** ART Parameters ***
+
+  SelectPortrait         = SSFerry_L ;Not exactly the same thing.
+  ButtonImage            = SSFerry_L
+
   Draw              = W3DModelDraw ModuleTag_01
+    OkToChangeModelColor = Yes
     ConditionState  = NONE
       Model         = AVAMPHIB
 
@@ -7840,6 +7846,7 @@ Object AmphibiousTransport
 
   ; ***DESIGN parameters ***
   DisplayName       = OBJECT:AmphibiousTransport
+  Side = Civilian
   EditorSorting     = VEHICLE
   CommandSet        = CivilianTransportCommandSet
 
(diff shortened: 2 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -8297,3 +8297,13 @@ CommandButton Command_ConstructTechRadioStation
   ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
   DescriptLabel           = OBJECT:TechRadioStation
 End
+
+CommandButton Command_ConstructAmphibiousTransport ; Patch104p Jundiyy 20.10.2021 Allow to train this unit.
+  Command       = UNIT_BUILD
+  Object        = AmphibiousTransport
+  TextLabel     = CONTROLBAR:ConstructAmericaVehicleMedic ;Could be changed if we have something else, uses the A button as a shortcut.
```
xezon, line 8304: Comment should not refer to data setting of generals.csf. If generals.csf changes, this comment is no longer accurate.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -8297,3 +8297,13 @@ CommandButton Command_ConstructTechRadioStation
   ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
   DescriptLabel           = OBJECT:TechRadioStation
 End
+
+CommandButton Command_ConstructAmphibiousTransport ; Patch104p Jundiyy 20.10.2021 Allow to train this unit.
```
commy2, line 8301: Should not inline this imo. The comment is for a whole section of code.

Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
```diff
@@ -7854,15 +7861,29 @@ Object AmphibiousTransport
     Weapon          = PRIMARY None
   End
 
+  BuildCost       = 400 ;500 ;A weak unit hence the low price.
+  BuildTime       = 5 ;10.0          ;in seconds ;Same as above.
```
commy2, line 7865: Disagree with 5 seconds. That's comically low like a Bike or Technical. Doesn't fit a massive unit like this at all. We should go for flavour as well here since this is no regular unit.

commy2:
I usally use this unit with `Scale = 1.1` or 1.2 below the geometry parameters. This resizes the model such that the hatch actually becomes big enough that a Crusader tank could fit inside.

I am not too happy with changing all these cinematic units. This will affect of cinematics look. We should instead consider making new units with different type names.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
Jundiyy: True, was just thinking of the general players that don't but will remove.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
Jundiyy: Didn't understand, you mean add the note above the line?

commy2: Yes. Inline comment refers to comment at end of line here.

Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
Jundiyy: Yes but the unit is also a weak unit? It dies in like 2 shots to Tanks. So wouldn't get much out of it.

commy2: If it's too fragile, make it less fragile, not build comically fast.

Jundiyy:
> I usally use this unit with `Scale = 1.1` or 1.2 below the geometry parameters. This resizes the model such that the hatch actually becomes big enough that a Crusader tank could fit inside.
> 
> I am not too happy with changing all these cinematic units. This will affect of cinematics look. We should instead consider making new units with different type names.

Yep I got you, I didn't make it bigger cos then it looks different to how it is already used in maps.

And I agree about causing problems to maps or intros and so on, hence why I haven't changed anything that would cause a problem. The price wouldn't matter as originally it had none and the salvage crate, I don't think so.
The only unit which was changed a bit was the Repair Drone, giving it the Repair ability and removing the Crusader weapon, but I did ask about it on the discord server first.

I would also not mind making a copy of units that are unfinished but would change the way they work if fixed.

Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
Jundiyy: I can easily do that but then I thought it might break maps that already use it? Or cinematic uses? I will just change the build time, does same as 1.04+ Sentry Drone time seem okay? 8s. That's also enough time to allow the door to close.

Jundiyy:
Can be closed as have made another PR at #601

PR #593, opened 2021-10-23
Fix: Frenzy-Scan Bug

### LINKED ISSUE #591: Should Frenzy reveal the map?
Frenzy reveals map under fog for a certain duration (longer than satellite scan even I think), should this be the case?

commy2:
ZH 1.04:

- Frenzy can be used to reveal the map. Especially relevant to Infantry General, since he gets Frenzy 1 at rank 1.

After this patch:

- Frenzy will no longer reveal the map where used.

See discussion in close #591


xezon:
Do we want to merge this finally?

xezon (approved):

PR #594, opened 2021-10-23
Fix: Cash Bounty Granted By Command Center Scaffold

### LINKED ISSUE #590: GLA cash bounty activated with just a CC scaffold

commy2:
- may fix #590

This needs to be thoroughly tested first. It would be very sweet if this worked, since this is the least hacky implementation.


xezon:
Looks clean to the eye at least :-)

commy2:
This change does not work. You will still get Cash Bounty from a scaffold.

PR #595, opened 2021-10-24
Fix: Cash Bounty Granted By Command Center Scaffold, Method 2

### LINKED ISSUE #590: GLA cash bounty activated with just a CC scaffold

commy2:
- fix #590
- replaces close #596

Seems to work from first trials with it. More testing required. Idk if a better method can be found.

- [x] needs comments
- [x] needs implementation for non GLA

# Reverted

This change and its follow up changes #1764, #2253 have been replaced by #2263.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -8638,16 +8638,10 @@ Weapon Demo_ScudStormDeathWeapon
   DamageDealtAtSelfPosition   = Yes
 End
 
+; Patch104p @bugfix commy2 24/10/2021 Spawns temporary object with cash bounty logic over and over.
+
 ;------------------------------------------------------------------------------
 Weapon CashBountyDummyWeapon
-  PrimaryDamage = 0.0
-  PrimaryDamageRadius = 0.0
-  DamageDealtAtSelfPosition = Yes
-  AttackRange = 100.0
-  DamageType = SMALL_ARMS
-  DeathType = NORMAL
-  WeaponSpeed = 99999.0
-  RadiusDamageAffects = ALLIES
   DelayBetweenShots = 250                ; time between shots, msec
   FireOCL = OCL_CashBountyDummy
 End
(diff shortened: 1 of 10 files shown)


commy2:
This change makes it such that the player needs to own a completed and non-disabled (e.g. microwaved) xGLA Command Center (or Boss General Command Center) to enable Cash Bounty after picking the tech from the generals promotion menu. A scaffold, like in 1.04, will no longer suffice.

xezon:
Reads like an appropriate solution if this is what design wants.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8637,3 +8637,11 @@ Weapon Demo_ScudStormDeathWeapon
   AntiGround                  = Yes
   DamageDealtAtSelfPosition   = Yes
 End
+
+; Patch104p @bugfix commy2 24/10/2021 Spawns temporary object with cash bounty logic over and over.
+
+;------------------------------------------------------------------------------
+Weapon CashBountyDummyWeapon
+  DelayBetweenShots = 250                ; time between shots, msec
```
xezon, line 8645: Is this something that could be costly for performance? Or is this regularly used in the game?

commy2: It's used every 8 frames. Of course it will take resources. I doubt it will be measureable at all though. I'd perhaps be worried if a logic like this were used on a unit or a base defense, but Command Centers tend to be one offs per player in the vast majority of games.

commy2: Take a Mircowave tank or ECM tank for example. They shoot their weapons (infantry killer and missile jammer respectively) every 250 msec too (also using the `FireWeaponUpdate` module), and no one ever complained.

EnigmaZift: Mods tend to use effects like this all the time. It has no performance cost if done correctly.

xezon (approved):

PR #596, opened 2021-10-24
Fix: Cash Bounty Granted By Command Center Scaffold, Method 1

### LINKED ISSUE #590: GLA cash bounty activated with just a CC scaffold

commy2:
- fix #590
- alternative to close #595

This change makes it such that an xGLA (or Boss General) player does not need a Command Center (scaffold) at all to unlock Cash Bounty. The Cash Bounty will be available as soon as the generals promotion is picked.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
@@ -127,6 +127,7 @@ PlayerTemplate FactionGLA
   DisplayName       = INI:FactionGLA
   StartingBuilding  = GLACommandCenter
   StartingUnit0     = GLAInfantryWorker
+  StartingUnit1     = CashBountyDummy
   ScoreScreenImage  = GLA_ScoreScreen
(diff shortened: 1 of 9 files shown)


Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
```diff
@@ -472,6 +476,7 @@ PlayerTemplate FactionBossGeneral
   DisplayName       = INI:FactionBossGeneral
   StartingBuilding  = Boss_CommandCenter
   StartingUnit0     = Boss_VehicleDozer
+  StartingUnit1     = CashBountyDummy
```
xezon, line 479: On mod maps with custom Starting Units this would break, right? If yes, then this is not good.

commy2: This is true for maps that use `StartingUnit1`. However, wouldn't the most obvious implementation of additional starting units on a scripted map be just placing them down via editor?

PR #601, opened 2021-10-28
Add: Amphibious Transport Implementation

Jundiyy:
Made a carbon copy of the unit instead, to give it more use in fun maps, old unit was really weak but didn't want to change stats as it could cause problems to maps that already use it.


commy2:
Uploading disable-transport.mkv.mp4…

commy2:
https://user-images.githubusercontent.com/6576312/139530289-21c9a5dc-f551-4749-b0e4-ff0f0977b829.mp4

commy2:
https://user-images.githubusercontent.com/6576312/139534974-549c1674-81eb-4e09-b794-dc74e199902f.mp4

Jundiyy:
Good work on it so far 👍 I still think a small reduction in size is better, 0.85 makes it a bit too big for the WFs/ADs. Also for the new TechWF
AT + WF

Top Right is China WF
Bottom Right is USA WF
Top Left is TechWF
Bottom Left is GLA AD

Jundiyy:
Good work on it so far 👍 I still think a small reduction in size is better, 0.85 makes it a bit too big for the WFs/ADs. Also for the new TechWF
AT + WF

Top Right is China WF
Bottom Right is USA WF
Top Left is TechWF
Bottom Left is GLA AD

commy2:
Uploading death.mp4…

commy2:
https://user-images.githubusercontent.com/6576312/142725760-1990e7c6-bb20-46c1-992b-3785807b5981.mp4

xezon (commented):
Where are these textures and models from? Do any files need to go into `GameFilesOriginal` folder for original reference? Were any DDS or TGA textures edited? If yes, then we should add the PSD source files.

commy2:
Models are edited from the base game AVAmbib.w3d / AVAmbib_A.w3d (door) model. Textures are from base game avambhib.dss, converted to .tga, slightly edited, and exported as .dds again.

xezon:
Ok for the texture I highly recommend to add the TGA source. It is better than DDS, because DDS is lossy format while TGA is lossless and thus can be re-edited in the future. We will then compress TGA (and PSD) to DDS in build step.

commy2:
They're here:
[avamphib_d.tga](https://github.com/xezon/GeneralsGamePatch/pull/601/files#diff-1fecbc381aaffb794f25ba073ca2834c4bedb45357370be3e3fb804b507a6797)
and
[avamphib_d1.tga](https://github.com/xezon/GeneralsGamePatch/pull/601/files#diff-50a1792a5a1eda445ca62a0b776ced039fce9b7e21f54c9759ccbaf40b54e492)

commy2:
Though they're made from the base game `avamphib.dds`. I don't think the source for that is around anywhere.

xezon:
Ok you added the source and compiled textures. That should be ok for now. We can remove the DDS later then.

Patch104pZH/Scripts/MAKE_Patch104pZH.bat
```diff
@@ -18,10 +18,19 @@ if "%ThisArg1%"=="build" (
     :: Optional files should be moved into one of the other scripts.
     xcopy /y /s "%GameFilesDir%\*.ini"                            "%GeneratedBigFilesUnpackedDir%\%BigName%\"
     xcopy /y /s "%GameFilesDir%\*.wnd"                            "%GeneratedBigFilesUnpackedDir%\%BigName%\"
+    xcopy /y    "%GameFilesDir%\Art\Textures\avamphib_d.dds"              "%GeneratedBigFilesUnpackedDir%\%BigName%\Art\Textures\"
+    xcopy /y    "%GameFilesDir%\Art\Textures\avamphib_d1.dds"             "%GeneratedBigFilesUnpackedDir%\%BigName%\Art\Textures\"
+    xcopy /y    "%GameFilesDir%\Art\Textures\SAAmphibiousTransport.dds"   "%GeneratedBigFilesUnpackedDir%\%BigName%\Art\Textures\"
+    xcopy /y    "%GameFilesDir%\Art\Textures\SAAmphibiousTransport_L.dds" "%GeneratedBigFilesUnpackedDir%\%BigName%\Art\Textures\"
```
xezon: SAAmphibiousTransport.dds and SAAmphibiousTransport_L.dds appear to be not be part of this change list. If game can load the tga's, then you could link to them here. I suspect game cannot load layered TGA. So if TGA is layered (which it should be), then must add DDS files for now.

commy2: These don't have DDS. All buttons/portraits are TGA. I just messed up the file extension.

xezon (approved):

commy2:
- [ ] geometry too small
- [ ] more parts coming apart when destroyed
- [ ] damage texture improvements
- [ ] smoke columns on badly damaged model
- [ ] possibly fire on badly damaged module or at least on transition between states
- [ ] prevent the thing from disappearing for a frame before spawning wreckage
- [ ] button doesn't use button art yet
- [ ] text

xezon:
shot_20211121_122440_2

xezon:
It is not so easy to replicate original damage look. Needs very dark patches on texture.
shot_20211121_123244_3

xezon (changes requested):
New unit needs several more fixes as illustrated above.

PR #622, opened 2021-11-08
Mammoth Tank Proposed Changes

Jundiyy:
Proposing changes for the Mammoth Tank to make it unique rather than being the same as the Overlord Tank.

Maybe commy can add a small flame to the shell? It feels like it could have something like that, shouldn't do insane damage but just burn a bit.


Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
```diff
@@ -269,8 +269,8 @@ Object MammothTank
   KindOf                 = PRELOAD SELECTABLE CAN_ATTACK ATTACK_NEEDS_LINE_OF_SIGHT CAN_CAST_REFLECTIONS VEHICLE SCORE
 
   Body                   = ActiveBody ModuleTag_02
-    MaxHealth       = 500
-    InitialHealth   = 500
+    MaxHealth       = 900
+    InitialHealth   = 900
```
commy2: SubdualDamageCap to 1800

Jundiyy: Oh yea, good catch.

Jundiyy:
Done.

xezon (approved):

xezon:
One more question: Are all loca strings for Mammoth Tank already in place?

PR #623, opened 2021-11-08
Tech Radio Station Proposal

Jundiyy:
Think it could have a bit more health.
I know it wasn't given any modules in the end but I think it's good to have something as a starting ground, so it can at least be placed and used by the simple mappers, if anyone wants to give it some extra abilities and have map.ini knowledge, they can do what they want.

I thought having a Speaker Tower styled Tech building seemed quite cool. Has a larger radius than the Speaker Tower, allowing players to use the Tech as a key fighting area.

(Of course it's not going to be built by the USA Dozer but just to show how it works.)

https://youtu.be/eNZvtTivvqQ


Patch104pZH/GameFilesEdited/Data/INI/Object/TechBuildings.ini
```diff
@@ -1143,8 +1143,8 @@ Object TechRadioStation
   RadarPriority       = STRUCTURE
 
   Body                = ActiveBody ModuleTag_04
-    MaxHealth         = 1000.0
-    InitialHealth     = 1000.0
+    MaxHealth         = 1500.0
+    InitialHealth     = 1500.0
```
commy2:
```suggestion
    MaxHealth         = 1000.0
    InitialHealth     = 1000.0
```
keep tiny building as is

commy2:
Should probably be a variant. To keep one without any effects for map makers.

Alternatively, make speakertower-effect turn on by upgrade. Give upgrade by either map.ini or without map.ini via worldbuilder script.

xezon:
Can Radio Station do this?
https://youtu.be/BwjNn-xNdUk?t=202

alanblack166:
The depicted effect is similar to the current Leaflet Drop behavior, except it also affects buildings. The code for this specific power is completely cut from the engine, but it should be possible to implement something similar using what already exists.

PR #625, opened 2021-11-20
Add: Amphibious Transport Voice Lines From Red Alert 2

commy2:
As suggested.

HOWEVER

This does need localization. And I'm not sure RA exists for all languages that ZH supports. Also, I do not have any files besides EN and DE in any case.


xezon:
Nice. In what way does it need localization? Do you mean tool tip for the unit? What are these original sounds? Unrelated unit voices?

commy2:
She speaks English. We'd need international voice lines.

xezon:
Oh right. That will be difficult to achieve I expect. However, we would only need German and French, because all other languages are English Spoken Language.

xezon:
You write you have German + English. So we would only need French on top. Any idea where to source that from?

commy2:
From someone that has Red Alert 2 in French I suppose.

Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
```diff
@@ -5621,41 +5621,33 @@ End
 ;===================================================
 
 AudioEvent AmphibiousTransportVoiceSelect
-  Sounds = vchisea vchiseb vchisec vchised vchisee vchisef
+  Sounds = vampsea vampseb vampsec vampsed ;vampsee
```
xezon, line 5624: I had to listen to **vampsee** to understand why you commented it :-)

xezon (approved):
Change is ok as is. This means all languages will have English audio for now. This is ok for all languages except German and French. Can make a second pass to add English and German.

**Squash when commit.**

PR #626, opened 2021-12-06
Add language files

xezon:
This is the language base line. From here on we can add all fixes on top. The tool is still yet missing and will be added as soon as available.

**Don't squash when commit!**


xezon:
Need to take another look at "Auto fill blank and starred texts with English text" whether or not it is ok. Perhaps revert it.

PR #630, opened 2021-12-24
Fix: PORTABLE_STRUCTURE objects no longer block building construction

alanblack166:
Adding ``INERT`` KINDOF to these objects ensures they no longer block scaffold placement.


commy2 (approved):
Tested and works.

Skyaero42:
I've retested in 1.04 + genpatcher and the bug with the avenger is still present

Issue #52, opened 2021-08-27
When Rocket Buggies are ordered to attack an enemy unit, they sometimes drive very close instead of attacking at the first possible chance


xezon:
When Rocket Buggies are ordered to attack an enemy unit, they sometimes drive very close instead of attacking at the first possible chance.

Jundiyy:
Is this not a terrain issue? As I mentioned today in the pitch discussion we had, they need a good line of sight otherwise they will go close to the enemy.

Stubbjax:
This happens with more than just Rocket Buggies and is likely due to the terrain.

As seen in the video below, giving this Technical an attack command on a worker in a terrain indent sees it get right up close and personal before firing:

https://user-images.githubusercontent.com/11547761/131472790-f4ac6dbd-38be-4921-8efa-70f19e70b181.mp4

xezon:
Nice video. Thank you. Yes this likely is related to angle limits. See pull request #111

xezon:
I suspect this is a code issue and caused by the object being attacked at 0,0,0 local space. Maybe setting the Z a bit higher above the ground avoids this problem.

PR #635, opened 2022-01-01
Increase Min and MaxTargetPitch of select vehicles

alanblack166:


xezon:
Lighter alternative to
* #111

Closes #52

PR #636, opened 2022-01-01
Fix: Flashbangs Can Damage Stinger Troopers

commy2:
- alternative to pull #297

Instead of making Stinger Sites an eligible target for Rangers in Flashbang mode, this makes the Stinger Troopers immune to Flashbangs.

### Original

- Stinger Sites cannot be targeted by Flashbangs. However, Flashbangs force-fired on the ground next to the Stinger Site can damage Stinger Troopers.

### Patched

- Flashbang grenades from USA Rangers no longer damage Stinger Soldiers.


xezon (approved):
If SURRENDER applies to flash bang only, then this looks good.

Weapon.ini

```
Weapon RangerFlashBangGrenadeWeapon
...
  DamageType              = SURRENDER
```

Appears to be the only one. Odd name.

commy2:
`RangerFlashBangGrenadeWeapon` is the only weapon with SURRENDER damage type, but I guess this damage type was also planned to be used for the cut stun bullets (China) and tranquilizer darts (GLA/Fundamentalists).

Same issue would apply to those.

commy2:
Doesn't work. Still can kill Troopers with force fire. Still cannot target in Stinger Site directly.

Issue #56, opened 2021-08-27
GLA Quad Cannon doubles DPS when vet 1 and becomes weaker when scrapped and vetted


commy2:
The current behaviour of the Quad is best summed up in this picture:

![https://i.imgur.com/DSOJdrb.png](https://i.imgur.com/DSOJdrb.png)

I propse we change the veterancy bonuses of the Quad to make them align with all other units, and make scrapping them up a small bonus on every veterancy level.

xezon:
Please investigate the damage values across veterancy, scrap & co. Maybe this can be fixed and tweaked in a meaningful way where Quad overall keeps same strength in early and late game.

ReLaX82:
Nice Summery, horrible to See.

Shall I make a Proposal?

commy2:
One "easy" to implement fix is to overwrite the Rate Of Fire veterancy bonuses for the Quad Cannon and to instead slightly increase the Damage veterancy bonuses to roughly match what the ROF would do in a game with more simulation steps per time unit.

See: https://github.com/commy2/zerohour/commit/b253b897c26f7fd71978e5478d59a594c56189e3
Also see: https://github.com/commy2/zerohour/commit/e7a0d3f1a27bea4c4bb9c5b19ac74c05fcec7802

Generalcamo:
Be extremely careful with that proposal. Both the mods C&C Shockwave and ROTR did fix this issue by using the "intended" values as you propose, however it turned out to radically boost the quad's damage to insane levels. We ended up having to perform an approximately 25% cut on the salvage damage levels as compared to the ZH "intended" level. Likely this was never caught during quality assurance testing, therefore the balance was never actually properly finished here. 

For reference for those not in the know: the bug itself is caused by the rate of fire being tied to frame rate; in a 30FPS match (the standard for multiplayer), it is technically impossible to have a delay between shots (DBS) that is less than 33.333 (which is rounded up to 66ms). The first salvage upgrade brings the quad cannon to a DBS of 50, which is rounded to 66; this is the absolute minimum value that is possible, therefore the second salvage level does nothing (and due to the first salvage level having less damage than no salvage at all, a single veterancy level that brings the quad to the absolute minimum DBS means that the salvage level would cause it to do less DPS)

Jundiyy:
[...]

PR #692, opened 2022-07-15
Balance: Invert Scale and Buff Quad Cannon Guns DPS

xezon:
Reference issue:
* #56

## Original
before

## Patched
quad_3_bonus

## Original
https://user-images.githubusercontent.com/4720891/182035421-266e6e12-fce2-44ca-a4f5-2abe67ba91e7.mp4

## Patched
https://user-images.githubusercontent.com/4720891/182035466-b0b29c02-b4e4-4677-bc95-212a2c8a151e.mp4


xezon:
@Jundiyy Can you upload you Quad Damage test map please?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -2699,6 +2708,15 @@ Weapon QuadCannonGunUpgradeOne
   AntiAirborneVehicle   = No
   AntiAirborneInfantry  = No
   WeaponBonus           = PLAYER_UPGRADE DAMAGE 125% ; APBullets
+  WeaponBonus           = HORDE              DAMAGE 112.5% ; Patch104p @balance Total DPS Bonus  60% to  80%
```
commy2: HORDE ? Quad cannot horde.

commy2:
How did you arrive at these values for NATIONALISM, FANATICISM, ENTHUSIASTIC (China) and TARGET_FAERIE_FIRE (USA)?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Yeah is not used, but does not hurt to have it here, just in case someone decides to create a Mod based on our Patch to add Quad Hordes :-)

xezon:
In first table above, you see see damage 100 60 60, the damage % tweak redistributes it to 60 80 80. The calculation are done via weapon_dps_calculator in https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Patch104pZH/Design/Balancing

commy2:
Wouldn't it be easier to lock the Quad at one shot per 2 frames (unscrapped) and one shot per 1 frame (scrapped) and then give reasonable damage buffs?

xezon:
From what I gathered weapon tick is fixed at every 2 frames, respectively 1/15 of a second or 66.66 ms. It cannot go lower. Quad already shoots at this lowest rate with the unscrapped weapon.

TLDR
We can't do that.

commy2:
Unscrapped without vet actually shoots every 2 frames. Any ROF bonus and it goes to every 1 frame.

My idea would be to invert all ROF bonuses on the unscrapped quad, and then compensate with reasonable damage buffs.

xezon:
Unscrapped without vet shoots every 4 frames. With ROF bonus it goes to 2 frames. I am all ears for new suggestions, because this proposal is still not ideal, when certain bonuses stack. For example check HERO+BATTLEPLAN_BOMBARDMENT, QuadCannonGun deals more damage than QuadCannonGunUpgradeTwo.

alanblack166:
Better to just make the quad cannon fire every 2 frames and rescale the damage accordingly. Veterancy can be compensated with extra damage per shot.

commy2:
> Unscrapped without vet shoots every 4 frames. With ROF bonus it goes to 2 frames

I was talking about simulation frames. There are twice as many draw frames. (30 vs 15 per second) So it was a misunderstanding.

xezon:
> Better to just make the quad cannon fire every 2 frames and rescale the damage accordingly. Veterancy can be compensated with extra damage per shot.

It is not clear how that would work. We need very accurate values for this. Reducing ROF changes a lot of parameters.

xezon:
I looked at the Excel sheet for another hour and tried tweaking some values, but its madness. Changing as little as possible of the base values is the most sane approach I think. Plus the differences on some stacked bonuses are not that big, so overall this change looks acceptable to me.

Here is diff from the new sheet.

(Redacted sheets. Contained Errors.)

commy2:
Why don't you start with the RATE_OF_FIRE changes from this PR:

https://github.com/TheSuperHackers/GeneralsGamePatch/pull/359/files

This locks the Quad to 1 shot per 2 logic frames without scrap and to 1 shot per 1 logic frame with one or two scraps, regradless of any ROF upgrade.

You can then give it a DAMAGE buff to whatever DPS increase you want, without any math required at all.

xezon:
I do not understand why you suggest that. To follow your inquiry, I punched the numbers of your "Generic Vet" into the DPS calculator, and this is the result:

(Redacted. Wrong data.)

~~This is not good distribution. Touching Fire Rate makes it an absolute mess.~~

~~Also: Changing fire rate will also affect the sound playback interval. And the bullet tracers. And muzzle flashes. I do not think Fire Rate should be touched at all here. At least not for the values as this change intended to tweak to (which it does).~~

xezon:
Additionally, I think this change is better on paper:
* #728

xezon:
I fixed the logic issue in the DPS calculator, and recalculated the correct modifiers for the intended target values. New comparison videos have been added to first post.

xezon:
@commy2 
If you want to change ROF, I suggest to create new Pull Request. I don't know how to do it gracefully. That said, I am happy with the numbers here as is now, albeit not perfect.

PR #694, opened 2022-07-15
xamo fix missing cameos

xezon:
Recovered branch. Not sure if it was meant to be merged?


Patch104pZH/GameFilesEdited/Data/INI/MappedImages/TextureSize_512/SAUserInterface512.INI
```diff
@@ -1386,3 +1386,23 @@ MappedImage SAStealth
...
 End
 
+; BEGIN Patch104p (xamorish) 03.09.2021:
+; Added selection cameos for Spectre Gunship.
+
+MappedImage SASpGunshipC_L
+  Texture = SAUserInterface512_001.tga
+  TextureWidth = 512
+  TextureHeight = 512
+  Coords = Left:1 Top:295 Right:121 Bottom:380
```
commy2, line 1396: These are stretched? 85 px height. The correct height is 96 px. I think this is to get rid of the bar below. 

Haven't we already fixed this?

xezon: I just added the original MappedImages INI files yesterday, so we have not touched them yet.

Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
```diff
@@ -2,9 +2,11 @@
 Object AmericaJetSpectreGunship
 
   ; Patch104p @bugfix commy2 11/09/2021 Add portrait and Countermeasures upgrade icon.
+  ; Patch104p @bugfix xamorish 03/09/2021 Add selection cameos for Spectre Gunship.
 
   ; *** ART Parameters ***
   SelectPortrait = SASpGunship2_L
+  ButtonImage = SASpGunshipC
```
commy2, line 9: The unit is not buildable, so this button is never seen. I don't think adding this is necessary. If it is added, I wouldn't added the cropped and streched image, but just `SASpGunship` (which is the image used by the commander ability button in 1.04).
If added, it should be added for every other iteration of the Spectre too.

Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
```diff
@@ -274,11 +274,13 @@ Object Nuke_ChinaVehicleHelix
...
   SelectPortrait         = SNHelix_L
   ButtonImage            = SNHelix
-
+  
   ; @bugfix commy2 19/09/2021 Moved upgrade icon for Nuke Bomb for cross faction consistency.
+  ; Do not show Upgrade_ChinaChainGuns to be consistent with other Helix variants.
 
   UpgradeCameo1 = Nuke_Upgrade_HelixNukeBomb
   ;UpgradeCameo2 = Upgrade_ChinaBlackNapalm
+  ;UpgradeCameo2 = Upgrade_ChinaChainGuns
```
commy2, line 283: Better argument for not adding the Upgrade_ChinaChainGuns upgrade cameo is the fact that the Helix only indirectly benefits from the upgrade. A Proplix and a Bunkerlix will never gain any benefit from the upgrade. (the Minigun is not affected, and Helix upgrades are exclusive). 
This is the same reason one could use on the Overlord. It possibly does not benefit from Chainguns if it's a Bunkerlord or Proplord.
You would have to treat Subliminal Messaging the same way. That cameo is also not there.
I would just go with: ignore all upgrades of turrets.

xezon:
commy2:
> This should be 3 PRs:
> 1. doc for SWG Aurora button / portrait art change
> I would approve.

> 2. Spectre button art change
> Uses a stretched image, because the image aspect ratio is wrong. This was presumably done to cut out the ammo bar below the Spectre in the image. I don't like this, it looks bad.
Also, a Spectre can not be build from any factory anyway, making this a meaningless change. If a button is really needed for the Spectre (I don't think it is), then the whole button should be used, with the ammo bar. Also, if a button is added to the Spectre, it should be added to all 5 (or so) Spectre sub-faction copies, not just the vUSA one.

> 3. Helix Chaingun upgrade cameo change
> Would reject this too. First of all, any upgrade cameo introduced here is commented out, so this doesn't do anything besides documentation. But it leaves out the Overlord entirely, which has the exact same behaviour. I argue that the Chainguns upgrade cameo is already on the turret. As far as in code documentation goes, that should be enough.

xezon:
Closing because the changes have no practical use.

Issue #79, opened 2021-08-28
USA Supply Drop Zone Cargo Planes are attackable


xezon:
Supply Drop Zone Cargo Planes are attackable. This is annoying in late game where many cargo planes fly to enemy base and trigger radar warnings and sound messages. NProject addresses this by making planes not attackable. This has implications on gameplay however, because plane can no longer be shot down and also units no longer become occupied with these practice targets which weakens USA.

A more sensible approach would be to simply disable attack warnings for these planes, if possible.

commy2:
Another argument in favour of this change is late game performance in spammy games like 6 player FFAs. Making these planes not attackables drastically reduces the number of particles created: No flares, no stray stingers and their trail particles, no exploding cargo planes, no Quad and Gat tracers, no impact particles on the plane.

Basically, this is a potential performance boost in games that tend to be laggy.

xezon:
That is a really good point. Would it be possible to remove particle effects from hits on these Cargo Planes? Or are hit effects solely bound to the attacking projectile/weapon?

commy2:
They come from the object. The biggest impact is probably from flares, missiles and AA gunfire though.

xezon:
Would it be possible to make planes manually attackable, but not auto attackable? Then they could still be shot down at least. But still it would weaken USA in late game to not have this extra cannon fodder unit. On the other hand, USA is very strong in late game with mass cash. USA is very likely to win in mass army spam games. Maybe it does not hurt USA overall.

ExiLe-ZH:
There's a few things to consider with this issue though:

- These planes are used tactically to soak up damage to time certain attack moves.
- It would be weird an enemy plane just flies over an enemy base without getting fired at.
- It's already weird the plane flies over the enemy base in the first place. 

Another idea/suggestion:
Make the plane curve off to the side of the map.

alanblack166:
Curving off is possible and has been done in Rise of the Reds. Making planes force attackable only can be handled by a FORCEATTACKABLE KINDOF.

Jundiyy:
Same points as Exile. Could curve it off map but then you can't use it to soak damage which does help a lot.

Stubbjax:
This is actually a serious quality-of-life issue for anyone playing with aircraft in guard mode.

[...]

PR #713, opened 2022-07-19
Change: USA Drop Zone Cargo Planes are no longer shot at automatically

xezon:
A solution for
* #79

Cargo Planes from Drop Zones are no longer auto attacked. Player must Force Fire on the plane to attack it. Consequences:

* Gets rid of attack notification spams on Radar and Voice
* Improves performance in spam/AOD games where many planes no longer draw fire
* Mig, Raptor, etc will no longer auto attack Cargo Planes on Guard Mode
* Cargo Planes will fly the same route, so enemy player knows USA has a Drop Zone where it comes from
* Cargo Planes can be attacked manually to prevent the money drop
* USA Faction loses its free cannon fodder unit

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
@@ -639,11 +639,14 @@ Upgrade Upgrade_ChinaOverlordBattleBunker
 End
 
 ;Even tho there is no radar upgrade, this is necessary to give the player free radar, or so I have gathered.  - CLH
-Upgrade Upgrade_CostReduction
-  DisplayName        = UPGRADE:CostReduction
-  ;Type               = OBJECT ; Patch104p Upgrade is repurposed as global upgrade for other unrelated functionalities
-  BuildTime          = 10.0
-  BuildCost          = 500
+; Patch104p @tweak former Upgrade_CostReduction is repurposed as global upgrade for more functionalities
+Upgrade Upgrade_GlobalDummy
+  ;DisplayName        = UPGRADE:CostReduction
+  ;Type               = OBJECT 
+  ;BuildTime          = 10.0
+  ;BuildCost          = 500
+  BuildTime          = 0.0
+  BuildCost          = 0
 End
 
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
@@ -641,7 +641,7 @@ End
 ;Even tho there is no radar upgrade, this is necessary to give the player free radar, or so I have gathered.  - CLH
 Upgrade Upgrade_CostReduction
   DisplayName        = UPGRADE:CostReduction
-  Type               = OBJECT
+  ;Type               = OBJECT ; Patch104p Upgrade is repurposed as global upgrade for other unrelated functionalities
   BuildTime          = 10.0
   BuildCost          = 500
 End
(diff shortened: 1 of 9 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
```diff
@@ -641,7 +641,7 @@ End
 ;Even tho there is no radar upgrade, this is necessary to give the player free radar, or so I have gathered.  - CLH
 Upgrade Upgrade_CostReduction
   DisplayName        = UPGRADE:CostReduction
-  Type               = OBJECT
+  ;Type               = OBJECT ; Patch104p Upgrade is repurposed as global upgrade for other unrelated functionalities
```
commy2: Are we sure this doesn't break anything?

commy2:
This is awesome, especially because you can still manually target SDZ planes if that is advantageous.

Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
xezon: Yeah I was wondering the same. hanfield says it does not break anything. I tested if building a Supply Drop Zone would grant cost reduction benefits, but it did not. So that is good.

We need to test if the actual Oil Refinery still works correctly:
* Does capture grant Cost Reduction?
* Does destruction remove Cost Reduction?

ZekeDlyoung: It shouldn't break anything as it is the refinery that gives the cost reduction not the upgrade.

commy2: So apparently Upgrade_CostReduction is superfluous in how 1.04 uses it?

We already have a unit dummy upgrade. Perhaps we should introduce a global dummy upgrade.

xezon: The rationale of repurposing the `Upgrade_CostReduction` was to keep the upgrade count low, because there is a tight count limit to upgrades.

commy2:
If the upgrade is granted by the SDZ, and the SDZ is capturable, I suspect that planes spawned by a captured SDZ are still automatically attacked with this implementation, unless the player that captured the SDZ has built an SDZ themselves.

Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
xezon: I tested Oil Refinery, and it worked normally.

shot_20220721_182101_2

shot_20220721_182105_3

xezon:
> If the upgrade is granted by the SDZ, and the SDZ is capturable, I suspect that planes spawned by a captured SDZ are still automatically attacked with this implementation, unless the player that captured the SDZ has built an SDZ themselves.

I think I need human player to test that.

MTKing4:
> > If the upgrade is granted by the SDZ, and the SDZ is capturable, I suspect that planes spawned by a captured SDZ are still automatically attacked with this implementation, unless the player that captured the SDZ has built an SDZ themselves.
> 
> I think I need human player to test that.

I can help with that, I'll let you know when

xezon:
Tested. After enemy capture, the USA Cargo Plane is auto attacked as usual.

How can we fix this? Is it possible to grant the Upgrade_CostReduction on Supply Drop Zone capture with some module?

shot_20220804_192909_2

xezon:
commy2
> 1. merge #539
> 2. use it to add a global dummy upgrade
> 3. make that upgrade trigger the flag

PR #714, opened 2022-07-20
Remove SCORE KindOf and UNIT RadarPriority from harmless Cargo Planes

xezon:
This change
1) removes the score from harmless Cargo Planes
2) hides Cargo Plane on Radar for enemy player

I tested Score in Skirmish but was unable to reproduce Score points given for killing Cargo Planes. But according to the INI code it would give Score. I don't understand.

I am not convinced this change is necessary at all, but I put it here for further discussion.


commy2:
What makes one plane that drops a Battlemaster harmless and the other not? I get this for the SDZ, but not so much the Reinforcement Pad.

xezon:
My rationale here was separating Automatically Spawned Planes and Generals Powers planes.

xezon:
Probably can close. There is also no bug report for it.

Issue #719, opened 2022-07-20
GLA Rocket Buggy has large discrepancy between EXP given and required


xezon:
GLA Rocket Buggy has unusual large discrepancy between EXP given (150) and required (800). Does this need tweaking?

Perhaps scraps level ups play into this.

```
Demo_GLAVehicleRocketBuggy
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800    ;Experience points needed to gain each level

Demo_GLAVehicleScudLauncher
  ExperienceValue    = 50 50 100 150  ;Experience point value at each level
  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level

ChinaVehicleInfernoCannon
  ExperienceValue    = 50 50 100 150  ;Experience point value at each level
  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level

ChinaTankBattleMaster
  ExperienceValue    = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600   ;Experience points needed to gain each level

GLATankMarauder
  ExperienceValue    = 100 100 200 300 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600   ;Experience points needed to gain each level

AmericaTankPaladin
  ExperienceValue        = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired     = 0 200 300 600 ;Experience points needed to gain each level

AmericaVehicleTomahawk
  ExperienceValue    = 50 50 100 150  ;Experience point value at each level
  ExperienceRequired = 0 200 400 800  ;Experience points needed to gain each level

ChinaVehicleNukeCannon
  ExperienceValue    = 50 100 200 400  ;Experience point value at each level
  ExperienceRequired = 0 400 600 1000  ;Experience points needed to gain each level

ChinaTankOverlord
  ExperienceValue    = 200 200 400 600 ;Experience point value at each level
  ExperienceRequired = 0 400 600 1200  ;Experience points needed to gain each level
```

xezon:
To me it looks as if GLA Rocket Buggy could be reduced a bit and GLA Scud Launcher increased a bit.

commy2:
My gut feeling is that Buggies work fine as they are. We would have to check how veterancy affects the counters to buggies. E.g. how much easier is it for a flock of  vetted vs unvetted Buggies to survive a Mig strike.

[...]

PR #726, opened 2022-07-21
Change: Streamline GLA Marauder EXP with China Battlemaster and USA Paladin

xezon:
Reference:
* #719

### Before
```
GLATankMarauder
  ExperienceValue    = 100 100 200 300 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600   ;Experience points needed to gain each level
```
### After
```
GLATankMarauder
  ExperienceValue    = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600   ;Experience points needed to gain each level
```
### Similar
```
ChinaTankBattleMaster
  ExperienceValue    = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600   ;Experience points needed to gain each level

AmericaTankPaladin
  ExperienceValue    = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired = 0 200 300 600 ;Experience points needed to gain each level
```

3 Star Marauder, especially with 2 Scrap, is very strong. Killing it deserves no less reward than killing Paladin and Battlemaster.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -19186,7 +19186,9 @@ Object Chem_GLATankMarauder
     Science = SCIENCE_MarauderTank
   End
 
-  ExperienceValue = 100 100 200 300    ;Experience point value at each level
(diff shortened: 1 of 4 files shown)


xezon:
Obsolete by #412

PR #727, opened 2022-07-21
Change: Increase required XP of GLA SCUD Launcher by 50% and decrease required XP of GLA Rocket Buggy by 25%

### LINKED ISSUE #719: GLA Rocket Buggy has large discrepancy between EXP given and required

xezon:
Closes #719

## Original
```
Demo_GLAVehicleRocketBuggy
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 200 400 800

Demo_GLAVehicleScudLauncher
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 100 200 400
```

## Patched
```
Demo_GLAVehicleRocketBuggy
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 150 300 600

Demo_GLAVehicleScudLauncher
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 150 300 600
```

## Other long range units for reference
```
ChinaVehicleInfernoCannon
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 100 200 400

ChinaVehicleNukeCannon
  ExperienceValue    = 50 100 200 400
  ExperienceRequired = 0 400 600 1000

AmericaVehicleTomahawk
  ExperienceValue    = 50 50 100 150
  ExperienceRequired = 0 200 400 800
```

# Rationale

There is no reason that justifies such a relatively small XP requirement by SCUD Launcher and such a relatively large XP requirement by Rocket Buggy. SCUD Launcher can kill alot of units with one hit and level up relatively quickly. It is no rare occurance to have multiple 3 Star SCUD Launchers sitting in the Tunnel Network. Meanwhite, Rocket Buggys attack single targets only, so will require a lot more time to level up compared to SCUD Launcher. There is no justification for them to require 2 times as much XP. This change consolidates the XP by subtracting XP requirement away from Rocket Buggy and adding it to the SCUD Launcher. Therefore the total sum of Experience Required across both units is identical to the original distribution.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -21269,7 +21269,7 @@ Object Boss_VehicleRocketBuggy
   End
 
   ExperienceValue       = 50 50 100 150    ;Experience point value at each level
-  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level
+  ExperienceRequired    = 0 150 300 600 ; Patch104p @balance xezon 21/07/2022 From 0 200 400 800 to fit better into comparable setups in GLAVehicleScudLauncher, ChinaVehicleNukeCannon, ChinaVehicleInfernoCannon
   IsTrainable           = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -15682,7 +15682,7 @@ Object Chem_GLAVehicleRocketBuggy
   End
 
   ExperienceValue       = 50 50 100 150    ;Experience point value at each level
-  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level
+  ExperienceRequired    = 0 150 300 600 ; Patch104p @balance xezon 21/07/2022 From 0 200 400 800 to fit better into comparable setups in GLAVehicleScudLauncher, ChinaVehicleNukeCannon, ChinaVehicleInfernoCannon
   IsTrainable           = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
@@ -17952,7 +17952,7 @@ Object Chem_GLAVehicleScudLauncher
   End
 
   ExperienceValue = 50 50 100 150   ;Experience point value at each level
-  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
+  ExperienceRequired = 0 150 300 600 ; Patch104p @balance xezon 21/07/2022 From 0 100 200 400 to fit better into comparable setups in GLAVehicleRocketBuggy, ChinaVehicleNukeCannon, ChinaVehicleInfernoCannon
   IsTrainable = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -16897,7 +16897,7 @@ Object Demo_GLAVehicleRocketBuggy
   End
 
   ExperienceValue       = 50 50 100 150    ;Experience point value at each level
-  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level
+  ExperienceRequired    = 0 150 300 600 ; Patch104p @balance xezon 21/07/2022 From 0 200 400 800 to fit better into comparable setups in GLAVehicleScudLauncher, ChinaVehicleNukeCannon, ChinaVehicleInfernoCannon
   IsTrainable           = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
@@ -19336,7 +19336,7 @@ Object Demo_GLAVehicleScudLauncher
   End
 
   ExperienceValue = 50 50 100 150   ;Experience point value at each level
-  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
+  ExperienceRequired = 0 150 300 600 ; Patch104p @balance xezon 21/07/2022 From 0 100 200 400 to fit better into comparable setups in GLAVehicleRocketBuggy, ChinaVehicleNukeCannon, ChinaVehicleInfernoCannon
   IsTrainable = Yes             ;Can gain experience
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
@@ -2899,7 +2899,7 @@ Object GC_Chem_GLAVehicleScudLauncher
   End
 
   ExperienceValue = 50 50 100 150   ;Experience point value at each level
-  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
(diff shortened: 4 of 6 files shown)


xezon:
I would like to go ahead and submit this. It makes most sense to me.

commy2 (approved):

xezon:
Dominator
> It makes sense to adjust I think.
> Except buggies already own in GLA late game mirrors and also in GLA vs USA.
> It will make late game GLA vs USA more OP than before. Buggies quads own all.

ExiLe-ZH:
> Dominator
> 
> > It makes sense to adjust I think.
> > Except buggies already own in GLA late game mirrors and also in GLA vs USA.
> > It will make late game GLA vs USA more OP than before. Buggies quads own all.

For this reason I think Buggies shouldn't be touched.

Scuds change does sound fine, in FFA or bunker games they can cause quite some mass destruction and vet up too fast.

ExiLe-ZH:
Excal:

> I would like to express some major concerns with some of the proposed changes.
> 
> The buggy less XP to level up change
> 
> This makes a massive difference. A heroic buggy one shots humvees very easily. What is laser supposed to do for example in this situation? Quad + buggy as a composition essentially owns everything in the game, with gla being the most powerful in the late game anyway.
> 
> I have never seen anyone complain about buggies not levelling up fast enough. Where are these changes coming from and for what reason?

Stubbjax:
> Excal:
> 
> > I would like to express some major concerns with some of the proposed changes.
> > The buggy less XP to level up change
> > This makes a massive difference. A heroic buggy one shots humvees very easily. What is laser supposed to do for example in this situation? Quad + buggy as a composition essentially owns everything in the game, with gla being the most powerful in the late game anyway.
> > I have never seen anyone complain about buggies not levelling up fast enough. Where are these changes coming from and for what reason?

I don't agree that it would make a massive (detrimental) difference. Firstly, Rocket Buggies are social vehicles that typically hunt in packs, which reduces the likelihood of an alpha buggy emerging with the majority of the kills / experience as it is spread out across the group.

Secondly, buggies are short-lived, and most don't even end up with 100xp, let alone 600xp or the original 800xp (and even then, salvage crates often do a lot of the work). Instead of destroying 16 Humvees in 1.04, a single buggy would now have to destroy 12, which is a reasonable (and very conservative) change considering how drastically high of a requirement it remains in relation to other units of similar capability.

And thirdly, the opponent shares responsibility in allowing a single buggy to destroy 600xp worth of stuff. It would take a lot of effort contrasted with relative negligence from the opponent to allow any meaningful number of buggies to rank up with the adjusted values. If a single buggy manages to kill that much stuff, then that is arguably the game-changer, not the fact that the buggy is now heroic and does a better job.

ExiLe-ZH:
> I don't agree that it would make a massive (detrimental) difference. Firstly, Rocket Buggies are social vehicles that typically hunt in packs, which reduces the likelihood of an alpha buggy emerging with the majority of the kills / experience as it is spread out across the group.
> 
> 
> 
> Secondly, buggies are short-lived, and most don't even end up with 100xp, let alone 600xp or the original 800xp (and even then, salvage crates often do a lot of the work). Instead of destroying 16 Humvees in 1.04, a single buggy would now have to destroy 12, which is a reasonable (and very conservative) change considering how drastically high of a requirement it remains in relation to other units of similar capability.
> 
> 
> 
> And thirdly, the opponent shares responsibility in allowing a single buggy to destroy 600xp worth of stuff. It would take a lot of effort contrasted with relative negligence from the opponent to allow any meaningful number of buggies to rank up with the adjusted values. If a single buggy manages to kill that much stuff, then that is arguably the game-changer, not the fact that the buggy is now heroic and does a better job.

Buggy will vet up faster by taking out buildings, defenses, vetted units, collecting scraps etc.

The main question is, does it need to vet up ANY faster? Seeing how they perform ingame I don't think so. 

Sometimes it's better to leave things be, even though it doesn't make sense to you when comparing things or when looking for logic.

xezon:
There is no need to quote all the preceding posts.

Stubbjax:
> 
> Buggy will vet up faster by taking out buildings, defenses, vetted units, collecting scraps etc.

It's not really a case of whether they rank up faster, but whether they rank up at all - which would remain rare, despite the reduced requirements. A _single buggy_ would still have to destroy several buildings and/or many units to achieve heroic status after the change. The likelihood of this occurring in competitive matches, particularly in a game-changing way, would remain astronomically low. (Also, salvage crates give ranks rather than experience, so this change has no affect there.)

>
> The main question is, does it need to vet up ANY faster? Seeing how they perform ingame I don't think so.

The primary justification is that it would be more rewarding for players if their buggies ranked up faster (or ranked up at all), and achieve a greater consistency with units of a similar calibre. Unit experience is a desirable mechanic that provides strategic depth and rewards good play (and punishes bad play), and it can clearly be utilised better in this case. The original requirements (and adjusted, to an extent) are so ridiculously high that the overwhelming majority of buggies never naturally rank up. It is not ideal that the mechanic is almost absent for buggies, and a 25% reduction would help alleviate the issue without affecting the outcomes of matches in any noticeable way.

> 
> Sometimes it's better to leave things be, even though it doesn't make sense ~to you~ when comparing things or when looking for logic.

Sometimes it is. Most times, even. But why is it better in this case? There is a clear and reasonable argument for why the 1.04 experience requirements are suboptimal.

The only argument that has been presented against this change as far as I can tell is that buggies are (marginally) buffed as a result, which is bad because they are already powerful. While I tend to agree with this argument as a whole, it is also very one-dimensional and seemingly ignores a lot of the nuance and implications of such a change - many of which I addressed in my previous comment.

ExiLe-ZH:
Saying they need to kill 16 Humvees is portraying an unrealistic state of affairs. They do vet up for reasons mentioned. Sure, it will always be a few of the pack, but it is entirely unnecessary from a balance perspective to increase the potential.

The balance argument might be 'one-dimensional' but it's the only factor important for gameplay.

Even if impact is none, why push so much for a change? Just because gamecode doesn't line up?

Shouldn't look so much into gamecode and numbers or theoretically trying to predict stuff > the field experience is there.

There simply is no current issue.

Stubbjax:
>
> Saying they need to kill 16 Humvees is portraying an unrealistic state of affairs. They do vet up for reasons mentioned. Sure, it will always be a few of the pack, but it is entirely unnecessary from a balance perspective to increase the potential.

What is unrealistic? That is an example of a direct experience equivalent. The point is that buggies have to kill an unreasonable number of units to rank up. A common pack of four buggies would have to take down a total of 64 Humvees for them to all achieve heroic status. If anything, _that_ is unrealistic.

> 
> The balance argument might be 'one-dimensional' but it's the only factor important for gameplay.

That is simply untrue. Fun, as quantifiable as it is, is just as important, if not more. If balance was the only important factor, nobody would enjoy the game in its current state.

> 
> Even if impact is none, why push so much for a change? Just because gamecode doesn't line up?

I'm not pushing for the change at all. I don't care either way. It's also a relatively minor change. I'm merely highlighting why it's a reasonable change and how the arguments against it are vastly overstated. Nothing to do with game code - just the experience of playing the game and observing that buggies rarely rank up.

> 
> Shouldn't look so much into gamecode and numbers or theoretically trying to predict stuff > the field experience is there.
> 
> There simply is no current issue.

I don't understand what you mean by this. My opinions are based on experience (of a heroic magnitude!) and an intrinsic understanding of the game and its mechanics, not game code.

ExiLe-ZH:
@Stubbjax long story short, you would like to see vetted buggies more often (be it ever so slightly), that's the only intention can I see here (besides aligning values) and I think it's not a good idea.

More reasons being that Buggies are a one of a kind unit, which are very mobile and very good both offensively and defensively, gameplay wise they're not really comparable with Tomahawks or Infernos. That's why they should be treated like their own unit. 

I would agree you don't see veterancy as often as with other units, but we wouldn't want that either seeing how effective they already are.

ZH is a very snowbally game in which veterancy plays a big role. So we have to be careful with these kind of things.

ExiLe-ZH:
Don't agree with this merge, first of all Scuds and Buggies should be separate PR's.

With Scuds I can agree with, but with the Buggies I still don't, it's a buff regardless and an unnecessary one.

Issue #707, opened 2022-07-18
Emergency Repair Generals Power is very rarely used


xezon:
Emergency Repair Generals Power is very rarely used across all 3 factions. It repairs vehicles in a selected area. Check if there can be anything done to make it more attractive to use.

## Stats
```
SpecialPower SuperweaponEmergencyRepair
  ReloadTime = 240000

Object RepairVehiclesInArea_InvisibleMarker_Level1
  HealingAmount = 100
  Radius = 100
  KindOf = VEHICLE

Object RepairVehiclesInArea_InvisibleMarker_Level2
  HealingAmount = 200
  Radius = 100
  KindOf = VEHICLE

Object RepairVehiclesInArea_InvisibleMarker_Level3
  HealingAmount = 300
  Radius = 100
  KindOf = VEHICLE
```

### Proposal 1
Decrease reload time.

### Proposal 2
Increase healing amount.

### Proposal 3
Increase radius.

### Proposal 4
Repair more kinds of things, such as STRUCTURE or INFANTRY.

### Proposal 5
Increase the radius. Make the repair effect linger for 10 seconds (with visible decal). Repair the units over time instead of in one burst. Upgrading the ability increases how long the effect lasts (20, 30 seconds?).
Example implementation: https://github.com/commy2/zerohour/commit/ea284287498fd7733bcaeefca2f099713bb65e4d

commy2:
Proposal 5

Increase the radius. Make the repair effect linger for 10 seconds (with visible decal). Repair the units over time instead of in one burst. Upgrading the ability increases how long the effect lasts (20,30 seconds?).

Example implementation is here: https://github.com/commy2/zerohour/commit/ea284287498fd7733bcaeefca2f099713bb65e4d

ZekeDlyoung:
Maybe make it so that it heals/repairs everything, not just vehicles, so it can be used in more situations and not just tank spams.

xezon:
One problem I have with this ability is the lack of immediate usefulness after cooldown. Unlike other abilities such as GPS Scrambler, Artillery Barrage, Spy Drone, et al, it may not be suitable to use straight away after power is ready to use. There may be no important battles going on yet. This will further increase its usage frequency and therefore decrease its impact on the match.

The same applies to the China Frenzy power.

MTKing4:
[...]

PR #740, opened 2022-07-23
Emergency Repair Proposal 1, Shorter Recharge Time (4min -> 2min)

commy2:
Implementation of proposal 1 for #707.

After this pull request, the recharge time of Emergency Repair is halfed. The player needs to wait 2 minutes instead of 4 to use the ability again.


Jundiyy:
I like this change the most out of the ones proposed.
Can you also make a pull for:

Reload is 3 mins but heal is 150/300/450.

PR #743, opened 2022-07-23
Emergency Repair Proposal 4, Heal Infantry And Buildings Too

commy2:
Implementation of proposal 4 for #707.

After this pull request, the Emergency Repair heals infantry as well as buildings in addition to just vehicles.


xezon:
What I like about this proposal in principle is that it adds more versatility to the power. For example you might want to heal your poor hero as he escapes enemy fire, or you want to heal your Superweapon as it gets rampaged by some Auroras.

commy2:
Repairing infantry sounds weird. Could always make it so it's only vehicles + buildings.

xezon:
You are right. It would then need new art and text. Adding a red cross next to the repair wrench would clarify this visually.

Jundiyy:
I don't think healing infantry is a bad thing for this GP (even though I currently prefer it to stick to vehicles only). But if we did decide to add more to it, I don't see why infantry should be left out. Could heal everything.

PR #744, opened 2022-07-23
Emergency Repair Proposal 5, Repair Vehicles In Area Over Time

commy2:
Implementation of proposal 4 for #707.

After this pull request, the repair effect lingers for 10 seconds (with visible decal). The vehicles are repaired over time instead of in one burst. Upgrading the ability increases how long the effect lasts (to 20 and 30 seconds respectively). Upgrading the ability does not increase the healing speed. The total healing amount only differs due to the time the ability lasts. The radius is increased by +50% (which equals +125% area).

Vehicles can move in and out of the repair area. They are only repaired while inside the area.

https://user-images.githubusercontent.com/6576312/180599717-03c0d0d9-29dd-43f8-bc17-97725ab3f7c3.mp4


xezon:
This is a really cool implementation, because it adds more depth to the power. Player needs to plan ahead where placing this makes most sense on the offense and defense. I can really see this being very useful.

Jundiyy:
Think it's quite cool but I prefer one of the changes I proposed.

Issue #747, opened 2022-07-23
China Mig has relatively long reload time when parked on Airfield


xezon:
China Mig has relatively long reload time when parked on Airfield. Is this ok?

Aircraft reload time comparison:
```
Raptor             : 8 seconds
Mig                : 8 seconds
Mig + Black Napalm : 2 seconds
Mig + Nuke Missile : 2 seconds
King Raptor        : 2 seconds
Stealth Fighter    : 8 seconds (no change with Bunker Buster)
Aurora             : 5 seconds (including Aurora Alpha)
```

Jundiyy:
I don't have a problem with keeping the same or reducing a bit. But my fear is, if we change other bugs which make China better then this one shouldn't be touched till we test to see the results. Just in case it stacks up too much.

Otherwise can reduce it to 6 secs.

commy2:
Added more info to OP.

xezon:
Looks to me a slight buff from 8 to 6 or 5 seconds looks totally reasonable. It is nothing that will turn the tides. Just makes it a bit more convenient.

RisingZH:
I think even going to 4 seconds would be reasonable, so that it gets halved by black napalm and is half that of a normal raptor (on the basis it only has 2 missiles compared to 4).

Would be an interesting one to test, at the moment wf/af isn't very good vs usa or gla as the migs aren't good enough to stop techs/vees pre-bn and researching bn ties up your wf for too long. If this is enough to allow you to get a couple more gats/flamers out before researching bn, could make for some interesting strategies and improve china's options.

ExiLe-ZH:
> I think even going to 4 seconds would be reasonable, so that it gets halved by black napalm and is half that of a normal raptor (on the basis it only has 2 missiles compared to 4).
> 
> 
> 
> Would be an interesting one to test, at the moment wf/af isn't very good vs usa or gla as the migs aren't good enough to stop techs/vees pre-bn and researching bn ties up your wf for too long. If this is enough to allow you to get a couple more gats/flamers out before researching bn, could make for some interesting strategies and improve china's options.

Would also improve chinas vs inf alot, because they're actually good counter vs dual wf outpost and flamer spam, but it's often hard or even impossible to stop the quantities due to reload time. When Inf reaches the airfield it's instantly gameover.

2 secs would be too effective though since black napalm isn't much needed like it is vs Vees.

MTKing4:
4s seem reasonable, and if proven too strong make it 6

PR #751, opened 2022-07-24
Change: Decrease China Mig reload time by 50% before Upgrade

commy2:
Implementation of proposal made in https://github.com/TheSuperHackers/GeneralsGamePatch/issues/747.

This pull request decreases the reload time of non-Black Napalm MIGs from 8 seconds to 4 seconds. This also applies to Nuke MIGs before the Nuke Mig Upgrade. The reload time after both upgrades remains at 2 seconds.

## Plane Reload Times
| Object                                       | ReloadTime |
|----------------------------------------------|------------|
| Original China Mig                           | 8 seconds  |
| Patched China Mig (this)                     | 4 seconds  |
| Original China Mig + Black Napalm            | 2 seconds  |
| Original China Mig + Nuke Missile            | 2 seconds  |
| Original USA Raptor                          | 8 seconds  |
| Original USA Stealth Fighter                 | 8 seconds  |
| Original USA Stealth Fighter + Bunker Buster | 8 seconds  |
| Original USA Aurora                          | 5 seconds  |
| Original USA Alpha Aurora                    | 5 seconds  |
| Original USA King Raptor                     | 2 seconds  |

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -4279,6 +4279,7 @@ Weapon MiGFirestormCreationWeapon
   FireOCL = OCL_MiGFirestorm
 End
 
+; Patch104p @balance commy2 24/07/2022 Decrease MIG reload time before upgrade.
 ;------------------------------------------------------------------------------
 Weapon NapalmMissileWeapon
   PrimaryDamage               = 75.0
@@ -4298,7 +4299,7 @@ Weapon NapalmMissileWeapon
   AcceptableAimDelta          = 30
   DelayBetweenShots           = 300             ; time between shots, msec
   ClipSize                    = 2                        ; how many shots in a Clip (0 == infinite)
-  ClipReloadTime              = 8000               ; how long to reload a Clip, msec
+  ClipReloadTime              = 4000               ; how long to reload a Clip, msec
   AutoReloadsClip             = RETURN_TO_BASE                 ; must return to base to reload this weapon
   AntiGround                  = Yes
   ProjectileDetonationFX      = WeaponFX_NapalmMissileDetonation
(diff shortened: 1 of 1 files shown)


xezon:
Can we write some rationale here please why 4 seconds and not 6 or 8?

commy2:
Well, 8 is the original and makes the Mig pretty bad before Black Napalm (which is a 400% reload speed upgrade in 1.04: 8 -> 2).

4 is half of what the vRaptor has (8). It also has half the missiles so it makes sense? Also, now Black Napalm is """only""" a 200% reload speed upgrade (8 -> 4).

xezon:
We also need a rationale from gameplay perspective, to explain how this helps China in match situations vs USA and GLA and other Chinas.

commy2:
vs USA not at all, because unupgraded Migs are resisted by Humvees. vs Infa it helps pop up more than one LOutpost before they camp your airfield. vs GLA, it helps hunting early Technicals.

xezon (approved):
Looks good.

xezon:
Jundiyy:
> Seems okay, can always change to 6 if we feel it's too good.

penfriendz:
Maybe make tech structures not give any exp when captured (and especially when destroyed)?

xezon:
scuba:
> So in 1v1, migs could become more viable against GLA earlier game to hunt workers / rebel pushes (similar to CCG). However it could create problems with USA vs China balance, which as shown is quite map dependant, so certain maps could be swung too much in favour of China, especially against SW. It could also lead to more campy style play for China against USA as they can reload quicker to hold against the same army more.

> In 2v2 and 3v3 however you have probably now caused imbalance in lots of matchups, where migs reloading fast supported by team mate could be too OP. Everyone knows how OP big groups of napalm migs can get in team games, and now you're buffing them. China combined with USA/GLAs vs multiple USA's is now messed up. China tank + ChinaV vs 2 USA's could now be difficult to stop in what is already quite balanced (2 wf Gat spam + napalm migs).

PR #759, opened 2022-07-26
Remove obsolete Chem_PoisonFieldGammaLarge OCL from Chem_GLAInfantryTerrorist

xezon:
The OCL looks obsolete. Tests with and without this OCL yielded identical visual result.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -13029,7 +13029,6 @@ Object Chem_GLAInfantryTerrorist
     FlingForceVariance  = 3
     FlingPitch          = 60
     FlingPitchVariance  = 10
-    OCL                 = INITIAL Chem_PoisonFieldGammaLarge
   End
   
   ; Patch104p @bugfix xezon 16/07/2022 Paste death types from other Terrorists to avoid Toxin Terrorists exploding by all kinds of deaths.


commy2:
`Chem_PoisonFieldGammaLarge` is not an OCL, thus this entry does nothing and can (and should) be removed.

Btw, the object `Chem_PoisonFieldGammaLarge` itself, which does exist, is never created in a normal game either. `OCL_PoisonFieldGammaLarge` points to `GC_Chem_PoisonFieldGammaLarge` with the GC tag. Same for Chem_PoisonFieldGammaMedium` and `Chem_PoisonFieldGammaSmall`. I would keep these though for backwards compatibility.

xezon:
Might as well just remove it entirely then to avoid clutter.

commy2:
There's a lot of clutter, but you can never be sure what is used on custom maps.

xezon:
I mean the commented code, not the OCLs.

commy2:
Commented code can be cleaned up entirely, though idk where to put the comment then. We always put comments on every change. If that line is added anyway, may as well leave what stood there originally, which serves as the best documentation I can think of.

xezon:
It appears unimportant enough to just remove it.

commy2:
Yeah, that is what we did before with similar entries, see ["Hallfire Drone"](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/319/files#diff-13d74f2143fa2fd64ca63f22f386d2ab2e8af36dbdb3cc81b3f4cc27ce0b1c30L2003).

The ini configs are full of this stuff.  This config format is pretty lenient with dead code inside the modules.

xezon:
> Commented code can be cleaned up entirely, though idk where to put the comment then. We always put comments on every change. If that line is added anyway, may as well leave what stood there originally, which serves as the best documentation I can think of.

I will add a feature to Mod Builder to have it strip all obsolete whitespaces and comments from INI files before BIG file generation. This should get rid of a few uncompressed megabytes.

PR #764, opened 2022-07-27
Fix: Adds status requirement to Terrorist death weapons

alanblack166:
- MASKED status is applied whenever a unit enters a container. With this status, the ``FireWeaponWhenDead`` modules will not activated whenever the MASKED status is active (inside a container/transport).

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -13032,17 +13032,20 @@ Object Chem_GLAInfantryTerrorist
   End
   
   ; Patch104p @bugfix xezon 16/07/2022 Paste death types from other Terrorists to avoid Toxin Terrorists exploding by all kinds of deaths.
+  ; Patch104p @bugfix hanfield 27/07/2022 Exempt MASKED status to postpone death explosion until after passenger exit, if ever.
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_Death04
     DeathWeapon   = SuicideDynamitePack
     StartsActive  = Yes                      ; turned on by upgrade
     DeathTypes    = NONE +SUICIDED +BURNED +EXPLODED
+    ExemptStatus  = MASKED
   End
   
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_CrushedDeath_Patch104p
     DeathWeapon   = CrushedDynamitePack_Patch104p
     StartsActive  = Yes
     DeathTypes    = NONE +CRUSHED +SPLATTED +LASERED
+    ExemptStatus  = MASKED
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag91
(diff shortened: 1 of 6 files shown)


xezon:
Would this change make #208 obsolete?

alanblack166:
Yes

xezon:
Fix confirmed.

xezon:
We fix the indentation of the new lines and add some comments to explain what this does. Then fix is good to go from my side.

commy2:
Lgtm too. Presumably, the Fire Base does not apply MASKED. Is this a problem?

xezon:
I can test if there is any difference of Terrorist in Firebase. But even if there is, I much rather have the fix for Bus. There is practically no chance that a terrorist sits in a Firebase in any serious match.

Jundiyy:
Lol haha. Terrorist in a Firebase!

commy2:
Big tactics.

ExiLe-ZH:
> Lol haha. Terrorist in a Firebase! 

It's a sneaky MTKing Demotrap (he'll understand lol) 😄

MTKing4:
> > Lol haha. Terrorist in a Firebase!
> 
> It's a sneaky MTKing Demotrap (he'll understand lol) 😄

Making jokes about topics you don't understand is hardly helping anyone here. So i suggest to stay on topic instead.

xezon:
|  | Original | Changed (this) |
|--|----------|----------------|
| Killed Terror in USA Firebase | Dies and explodes | Dies and explodes |
| Killed Terror in China Bunker | Survives with full health | Survives with full health |
| Killed Terror in GLA Palace | Survives with full health | Survives with full health |

_"Big tactics."_
shot_20220728_162926_1

commy2:
The "not leaving toxin" part could be because of https://github.com/TheSuperHackers/GeneralsGamePatch/pull/401 not being in 1.04.

xezon:
I checked main branch and they do leave toxins indeed.

commy2:
Okay, this just means Fire Base does not apply MASKED, which is fine.

Could you also check shooting down a Chinook with Terrorists inside over a building? I know Bomb Trucks explode that way, making for funny suicide Chinooks.

xezon:
Comments added and indentation fixed.

commy2 (approved):

xezon:
Chinook

## Original Terror in Killed Chinook 
shot_20220728_172947_3

## Changed Terror in Killed Chinook
shot_20220728_173111_1

commy2:
Figures.

xezon:
Changes behavior in Chinook, but still better than crash.

ExiLe-ZH:
> Chinook

Nook terrors are so pro and creative. Marakar actually pulled this off vs Excal ingame and won in the end.

Well I guess this doesn't change anything, just gives the enemy a chance to counter when it already reached it's target.

commy2:
We could undo this change later if a solution in Thyme is found and can be distributed.

xezon:
Actually I prefer the changed behavior. It explodes in the air and applies damage to the ground. Makes no sense.

Issue #754, opened 2022-07-24
China Hacker is vulnerable and provides insufficient income


xezon:
China Hacker is vulnerable. Toxins, Snipers, Bullets and crushes take them out in no time. There can only be one Internet Center to protect 8 of them. On top of that China Hacker appears to not give sufficient income in late game. China seems to struggle making ends meet, especially for China Tank General.

## Some Information on USA, China, GLA economy
https://www.gamereplays.org/cnczerohour/portals.php?show=page&name=tip-of-the-week-105-secondary-economy

## Related issues
* #223
* #689
* #753

## Stats
```
- Hackers cost $625, build-time 20 sec (Black Market: $2500, 60 sec; SDZ: $2500 45 sec, 4 power)
- Tank Hackers cost $780 instead of $625 (25% more expensive)
- Inf Hackers are stealthed and spawn at vet 1, have 40% armor vs poison and radiation

- No XP awarded for: hacking vehicles, hacking buildings
- XP awarded for every cash hack: 1 point
- cash hack XP does NOT count towards the player XP
- XP required for vet1/vet2/vet3: 100 300 500
- cash per cash hack vet0/vet1/vet2/vet3: $5, $6, $8, $10
- time between cash hacks: 2 seconds
- 1.8 seconds while in Internet Center (10% faster)
- 1.6 seconds while in Internet Center for Super Hackers (20% faster)
- this means time required for:
        Hacker          Super Hacker    Hacker in IC    Super Hacker in IC
vet1:   3 min 20 sec    0 min           3 min           0 min
vet2:   10 min          6 min 40 sec    9 min           5 min 20 sec
vet3:   16 min 40 sec   13 min 20 sec   15 min          10 min 40 sec
```

### Proposal 1
Decrease Hacker build cost.

### Proposal 2
Decrease China Tank Hacker cost only, for example from $780 to $625.

### Proposal 3
Decrease China Hacker build time, for example from 20s to 15s.

### Proposal 4
Add stealth for China Hackers while hacking.

### Proposal 5
Add stealth for China Hackers while hacking, beginning at Vet 1.

### Proposal 6
Add stealth for China Hackers always.

### Proposal 7
Add stealth for China Hackers always, beginning at Vet 1.

### Proposal 8
Increase Hacker cash rewards.

### Proposal 9
Increase Hacker robustness against attacks.

### Proposal 10
Increase Hacker movement speed.

### Proposal 11
Reduce Hacker noise.

[...]

PR #773, opened 2022-07-28
Change: Increase China Hacker movement speed by +25%

xezon:
* Relates to #754

All China Hackers move as quick as a China Red Guard now. Black Lotus walks quicker than Hackers. Tank Hunter walks slower than Hackers.

## Original

https://user-images.githubusercontent.com/4720891/181492505-e361d842-a2db-480d-ae97-0141a51c9507.mp4

## Changed

https://user-images.githubusercontent.com/4720891/181492628-5651909a-533e-480a-b458-ee39c137052d.mp4

# Rationale

Originally the China Hacker walks as slow as the Tank Hunter, despite just carrying a military grade laptop with him. When built, he typically needs to move a bit first to reach his optimal hack location, unlike the GLA Black Market and USA Supply Drop Zone, which begin collecting money passively straight away. The increased movement speed allows hackers to move quicker into positon and also flee the scene quicker in case of danger.


commy2:
Besides the balance considerations, this even looks better. Could likely be improved visually even more by slowing down the animation.

PR #777, opened 2022-07-28
Change: Increase USA Paladin Composite Armor bonus to match with USA Crusader Tank bonus (200)

### LINKED ISSUE #645: USA Paladin Tank requires a General Promotion, but is not better than Crusader Tank
Related:
* #409

USA Paladin Tank is not used much. It requires using a General Promotion at Rank 1. Should this be changed?

### Proposal 1
Match the armor buff of Paladin with the armor buff of the Crusader, so the Paladin remains the better tank after the upgrade.

xezon:
* Fixes #645

The USA Paladin Tank now has the same Composite Armor bonus as Crusader Tank does.

## Stats

| Unit                                | Health | Composite Armor Bonus |
|-------------------------------------|--------|-----------------------|
| Original USA Crusader Tank          | 480    | 200 |
| Original USA Paladin Tank           | 500    | 100 |
| **Patched USA Paladin Tank (this)** | 500    | 200 |

### USA Paladin advantages over USA Crusader (in Patch):

* +20 Health (500 vs 480)
* Has low range Point Defense Laser gun

### USA Paladin disadvantages over USA Crusader (in Patch):

* Requires 1x Level 1 Generals Promotion
* +200 cost (1100 vs 900)
* +2s build time (15.25 vs 13.25)

# Rationale

The Paladin tank is better than a Crusader in that it starts with 20 extra health points (HP) (+4.2%) and a slow firing Point Defense Laser (PDL) that can target infantry besides neutralizing missiles.

In general, it is not worth the extra cost of 200 and build time of 2 seconds. After Composite Armor, it has much less HP than a Crusader to the point where the Crusader survives longer against 2 rocket infantries than a Paladin does despite the PDL.

To keep the Paladin worth the investment after Composite Armor upgrade, its Armor bonus is now equalized with the Armor bonus of the Crusader Tank.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
@@ -1823,6 +1823,15 @@ End
 
 
 ;------------------------------------------------------------------------------
+; Paladin advantages over Crusader:
+;   +20 Health (500 vs 480)
+;   Low range Laser gun
+
+; Paladin disadvantages over Crusader:
+;   Requires 1x Level 1 Generals Promotion
+;   +200 cost (1100 vs 900)
+;   +2s build time (15.25 vs 13.25)
+
 Object AmericaTankPaladin
 
   ; *** ART Parameters ***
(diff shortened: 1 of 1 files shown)


xezon:
Dominator:
> Make composite armour double up on the paladin like on crusader

commy2:
So, what about the Microwave? It has the same health and armor as a Crusader, but only 100 buff.

xezon:
I think this ticket is not the right place to discuss this. You can open a new report for that if there is a rationale to why this is a problem.

PR #778, opened 2022-07-28
Change: Add XP reward of 5 to China Outpost and China Troopcrawler

### LINKED ISSUE #549: China Troop Crawler and China Outpost do not grant XP on kill
Have not checked this with ALL troop crawler variants, but the default vanilla China Troop Crawler has no XP values attached whatsoever. Killing it does not provide XP, which is probably unintentional.

xezon:
Fixes #549

## Original
China Outpost and Troopcrawler grant 0 XP points for kill.

## Patched
China Outpost and Troopcrawler grant 5 XP points for kill.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaMiscUnit.ini
@@ -206,6 +206,10 @@ Object ChinaVehicleTroopCrawlerEmpty
   Prerequisites
     Object = ChinaWarFactory
   End
+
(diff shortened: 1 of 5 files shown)


Jundiyy:
Not the biggest of changes, but not sure. Does weaken China a bit, Outposts are bad quite a bit. So are Troop Crawlers. 
Although, personally I would like everything to give xp tbh.

xezon:
Change looks safe to me. 5 XP is nothing really. You would have to kill 1000 empty Outposts to reach Generals Rank 5 :-)

ExiLe-ZH:
Idk, 5xp is weirder than no xp, i don't see the sense really?

Issue #464, opened 2021-10-04
Review factory door times


xezon:
Streamline all factory door opening times. To compensate changed build times, unit build times can be offset to door time changes accordingly.

Jundiyy:
Not sure about streamline, even if we set the unit build times to match the times they are now, it would mean a slower build time for units on low power as the door animation is set to a certain amount.
China perhaps could have a buff alone with the door, allowing units out a bit faster without changing the unit time.

Jundiyy:
Thinking over if a bit more. I think it's best to leave each army with their own times. All GLAs follow one rule, Chinas their own and USAs theirs.

Firstly there is the problem of low power, explain it more below. And secondly tied to this is we would need to do a lot more balance testing to see how much of a difference this actually makes, taking us back to square one of how much balance we have already achieved through changing unit times alone, as that has had a lot of testing done already, many fun games, clan wars and tournaments. 

By increasing door time for GLA, it only makes them more powerful with power, as units will be spammed out faster, and possibly too fast that the door doesn't even close, so that means even faster units that the time given.

For Chinas and USAs, attacking their power is not uncommon, so on low power they would suffer a lot more, which I don't think is a good idea.
The only faction I see which seems a reason to do is China, but without changing the original times of the common units.
This means, units on low power still come out a bit faster than in 1.04 and the less common units will take a bit longer.

xezon:
According to commy2 these are the door times:

| Object            | Door time (s) |
|-------------------|--------------:|
| USA CC            |          2.00 |
| China CC          |          3.00 |
| Airfields (both)  |          2.00 |
| Arms Dealer       |          2.00 |
| USA War Factory   |          3.25 |
| China War Factory |          4.00 |

Maybe bring down China CC to 2.00, USA War Factory to 3.00 and China War Factory to 3.00? It would already be more consistent then. It helps player to understand how long it will take unit from Unit Queue Icon completion to final production from factory.

Keep in mind we are 1 to 2 years away from any final release. You have enough time to design and test.

Jundiyy:
I don't think this part matters, `It helps player to understand how long it will take unit from Unit Queue Icon completion to final production from factory`, there are only a handful of players that know the times of units and will use them. Everybody else just watched the unit come out of the wf after it has been produced.

I'd prefer keeping USA at 3.25, so China could also have 3.25. Stays consistent and less balancing problems to think over, test and worry about as USA isn't being touched.

[...]

PR #785, opened 2022-07-29
Change: China Command Center door time from 3000 to 1500 ms

xezon:
Relates to
* #464

This change reduces China Command Center door times. It equals USA Command Center door times.

| Object | DoorOpeningTime | DoorWaitOpenTime | DoorCloseTime |
|--------|-----------------|------------------|---------------|
| Original USA CC         | 1500 | 3000 | 1500 |
| Original China CC       | 3000 | 3000 | 3000 |
| Patched China CC (this) | 1500 | 3000 | 1500 |

# Rationale

This change achieves multiple things.

1. Originally China Command Center door is 2 times slower than USA Command Center door. This means player needs to wait twice as long until Dozer spawns, after it shows completed in contruction queue. With the halved delay, this disadvantage over USA is removed.
2. Consistency makes game simpler to anticipate and learn.
3. This solves the issue where a chained Dozer construction glitches through the door (see first video below). This happens because Dozer spawns while door still closes. Originally DoorWaitOpenTime + DoorCloseTime equals 6 seconds, but Dozer construction is 5 seconds. So instead of 9 seconds, the second Dozer spawns in 6 seconds. With this change this bug no longer occurs and the second Dozer spawns in 6.5 seconds after the first one as one would expect.

| Object | First Dozer Spawn time | Chained second Dozer spawn time |
|--------|------------------------|---------------------------------|
| Original USA CC         | 6500 | 6500 |
| Original China CC       | 8000 | 6000 |
| Patched China CC (this) | 6500 | 6500 |

Regardless of the fixed bug, this change is controversial, because all initial China build orders can be 1.5 seconds quicker than before. Helix can be 1.5 seconds quicker in the air. Tank from one Warfactory can be attacking 1.5 seconds earlier. This can potentially put too much pressure on USA and GLA. It needs to be closely monitored.

## Original
https://user-images.githubusercontent.com/4720891/181831765-531bfc84-45a6-4244-af2a-c99e2b7d8fcb.mp4

## Patched
https://user-images.githubusercontent.com/4720891/181831807-a27656ae-5cd8-4db9-8bfd-97de26a32a3e.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -1040,6 +1040,7 @@ Object Boss_CommandCenter
       Animation       = NBConYard_A7.NBConYard_A7
       AnimationMode   = ONCE
       Flags           = START_FRAME_FIRST
+      AnimationSpeedFactorRange = 2.0 2.0 ; Patch104p @tweak Match animation with DoorOpeningTime
     End
     AliasConditionState = NIGHT DOOR_1_OPENING
     AliasConditionState = SNOW DOOR_1_OPENING
@@ -1050,6 +1051,7 @@ Object Boss_CommandCenter
       Animation       = NBConYard_A7D.NBConYard_A7D
       AnimationMode   = ONCE
       Flags           = START_FRAME_FIRST
+      AnimationSpeedFactorRange = 2.0 2.0 ; Patch104p @tweak Match animation with DoorOpeningTime
     End
     AliasConditionState = NIGHT DOOR_1_OPENING DAMAGED
     AliasConditionState = SNOW DOOR_1_OPENING DAMAGED
@@ -1060,6 +1062,7 @@ Object Boss_CommandCenter
       Animation       = NBConYard_A7D.NBConYard_A7D
       AnimationMode   = ONCE
       Flags           = START_FRAME_FIRST
+      AnimationSpeedFactorRange = 2.0 2.0 ; Patch104p @tweak Match animation with DoorOpeningTime
     End
     AliasConditionState = NIGHT DOOR_1_OPENING REALLYDAMAGED RUBBLE
     AliasConditionState = SNOW DOOR_1_OPENING REALLYDAMAGED RUBBLE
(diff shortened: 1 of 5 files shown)


Jundiyy:
Yes, closely monitored for sure. It means everything is faster. Already mentioned it in the WF topic.

xezon:
What we could do is increase China Dozer build time from 5 seconds to 5.5 or 6 seconds to reduce impact.

ExiLe-ZH:
> What we could do is increase China Dozer build time from 5 seconds to 5.5 or 6 seconds to reduce impact.

Well China Dozers are slow and clunky and need to drive all the way to supply unlike USA dozers, half a second buff might be justified. (Though new locomotor does improve this already?)

Anyone wants to playtest 1v1?

Stubbjax:
> What we could do is increase China Dozer build time from 5 seconds to 5.5 or 6 seconds to reduce impact.

6 seconds would be too much. A 1s build time increase translates to 2s without power, which would negate the 1.5s reduction and result in a 0.5s increase for the initial dozer.

xezon:
5.5 would be ok then.

That way, without Power China Dozer builds 0.5 seconds faster than original and 1 second slower than USA.

And with Power China Dozer builds 1 seconds faster than original and 0.5 seconds slower than USA.

That way this change would not be as drastic, but would be a buff for China, but not as good as USA is. Sounds good?

ExiLe-ZH:
> 5.5 would be ok then.
> 
> That way, without Power China Dozer builds 0.5 seconds faster than original and 1 second slower than USA.
> 
> And with Power China Dozer builds 1 seconds faster than original and 0.5 seconds slower than USA.
> 
> That way this change would not be as drastic, but would be a buff for China, but not as good as USA is. Sounds good?

Test normal time first?

xezon:
Can do. Will be separate change after this.

Jundiyy:
I think leave Dozer time for now, let's see how much of a difference the door makes.

MTKing4:
> > What we could do is increase China Dozer build time from 5 seconds to 5.5 or 6 seconds to reduce impact.
> 
> Well China Dozers are slow and clunky and need to drive all the way to supply unlike USA dozers, half a second buff might be justified. (Though new locomotor does improve this already?)
> 
> Anyone wants to playtest 1v1?

I could yeah, but I don't think it'll be much of a challenge since I'm way too slow :p

PR #786, opened 2022-07-29
Change: China War Factory door time from 4000 to 3300 ms

xezon:
Relates to
* #464

This change reduces China Warfactory door times. It equals USA Warfactory door times.

| Object | DoorOpeningTime | DoorWaitOpenTime | DoorCloseTime |
|--------|-----------------|------------------|---------------|
| Original USA Warfactory         | 3250 | 3000 | 4000 |
| Patched USA Warfactory          | 3300 | 3000 | 3300 |
| Original China Warfactory       | 4000 | 2000 | 5000 |
| Patched China Warfactory (this) | 3300 | 3000 | 3300 |

This enables China to build all vehicles 700 ms, 21 frames, quicker.
The door animation duration has been set to 99 frames to match the 3300 ms properly.

# Rationale

Factory door time is added on top of production time. This means when a vehicle shows ready in the production queue, it takes the door to fully open until the vehicle can spawn. The original 4 seconds make this an awfully long experience compared to USA and GLA. On top of that the visuals show the door open for 2 seconds until the vehicle finally spawns, which makes the wait more painful to anticipate, as the vehicle does not spawn after the door stopped moving, unlike it does with USA and GLA factories (See first video below).

Reducing the door time and adjusting the animation for the duration of new time will make the behaviour consistent with USA. It makes the machanic easier to anticipate and learn cross factions, and a more pleasant experience in practice.

Inherently this makes all China factions produce vehicles quicker. This can be desirable, because China is often the faction under early pressure against both GLA and USA. If vehicles produce too quickly because of it, it is advised to revisit vehicle build times.

## Original
https://user-images.githubusercontent.com/4720891/181832813-dc3a087d-d00c-4a3d-861a-fe65a8ed4174.mp4

## Patched
https://user-images.githubusercontent.com/4720891/181832882-c99d0a77-61f1-46c9-be0d-f34731e8ef6e.mp4


Jundiyy:
Seems good.

ExiLe-ZH:
Revisiting this, to start with, could we compensate China WF production times, so it doesn't affect balance? It just makes things less much less complicated.

xezon:
I read the opening rationale text and I still agree with what I wrote back then.

> Inherently this makes all China factions produce vehicles quicker. This can be desirable, because China is often the faction under early pressure against both GLA and USA. If vehicles produce too quickly because of it, it is advised to revisit vehicle build times.

I would recommend to slap + 700 ms on every vehicle instead of making the doors slower again. This way door times can keep this setup (and be consistent with USA) but vehicles practically build in same time.

ExiLe-ZH:
Yes, that was the suggestion indeed.

ExiLe-ZH:
I will go through the ini files later and upload it here

xezon:
You need to create a pull request. We cannot work with zip files.

ExiLe-ZH:
Oh like a separate one? Thought it'd be better to combine the files/edits with this commit.

xezon:
This change is already merged into main branch. You need to create a new branch with your new changes.

ExiLe-ZH:
Ok, will do, I didn't know you couldn't edit merged commits.

xezon:
Jundiyy rightfully raised that adding 700 ms to unit production times is not equivalent to counter this change, because the production time is affected by Power whereas the door time is not.

This means the door time would have to be increased again to restore the status quo.

PR #788, opened 2022-07-30
Fix broken binary stream effect

xezon:
#336 broke binary effect on some installs.

Appears EXBinaryStream.tga does not exist on CD version. On First Decade it does.

shot_20220730_093819_1


commy2:
Why not provide a copy of EXBinaryStream then?

xezon:
Could do. Would you prefer that?

commy2:
I don't know which one is better. We should use the low res one (from CCG), because it is less performance demanding and looks identical for all intends and purposes. I cannot tell you which one that is though.

xezon:
That would be the one without the 32 then. I will check.

xezon:
Both textures have the same pixel size and color bits. They look a bit different. The EXBinaryStream.dds texture is present in the Install I had the issue previously. With issue I mean the binary stream was not rendered ingame. Only after switching back to the original EXBinaryStream32.dds it rendered again. I recommend to keep it using the EXBinaryStream32.dds.

binarystream_tex

PR #790, opened 2022-07-30
Change: Increase Super Hacker Vehicle Hack range by +3%

xezon:
Increase Super Hacker Vehicle Hack range by +3%.

3% does not sound like much, but this is the difference between life and death vs idle units. With this change Super Hacker can try engage idle tanks. When idle, meaning no guard or retaliation, Super Hacker can successfully disable Tank without getting shot at.

Most Tanks Attack Range: 150
Original Super Hacker Attack Range: 150
Patched Super Hacker Attack Range: 155

This change needs to be closely monitored. If Super Hacker can be abused in some way because of it, then it should be reverted.

## Original
shot_20220730_165202_4

shot_20220730_165206_5

## Patched
shot_20220730_165004_1

shot_20220730_165014_2


Jundiyy:
👍, yup, could become quite strong if you have many.
Most likely help defend easier.

commy2:
Was this tested? Seems like only 5 could still put the Hacker in range, because terrain and that weird rounding the game does when moving diagonally on the map.

xezon:
Yes this was tested. I attached images in first post to illustrate the fundamental change. Note this only works against idle tanks. This will not work against tanks on Attack Move, Guard Mode or Retaliation.

I doubt this change has any tangible consequence in regular battle scenarios. The only impact I see for this is for sneaky play where you try to disable idle units parked somewhere or freshly build from a Tank Factory. For example you could try to sneak some Super Hackers and Lotus in enemy base, try to disable some tanks with hackers, and then capture a building with Lotus. I doubt this is easy to pull off however.

As for your concern, I did not test on angled terrain and I don't expect it to be a problem. As for the diagonal movement bug, this should be of no concern here, because the attack ranges would scale identically for all units facing opposing direction.

Jundiyy:
Think it's okay to merge. There are only a few players that use vehicle hack for Hackers, perhaps more will with this change.

ReLaX82:
I like the idea to have this functionslity even more implemented. Its a Late game thing anyway and increasing it is okay.

PR #791, opened 2022-07-30
Fix: Sat Hack 1 Fails To Reveal Newly Built Command Centers

### LINKED ISSUE #748: Internet Center with Satellite Hack I does not spy on newly constructed Command Centers
Internet Center with Satellite Hack I does not spy on newly constructed Command Centers.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/748

1.04:

- Satellite Hack 1 will only reveal Command Centers that existed when the upgrade was finished or the latest Internet Center was built. It will not reveal Command Centers that were built afterwards, until it is destroyed or sold and then rebuild.

This patch: The Internet Center will continously reveal new Command Centers as one may expect.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -2927,8 +2927,8 @@ Object SatelliteHackOneDummy
   End
 
   Behavior = LifetimeUpdate ModuleTag_02
-    MinLifetime = 250   ; min lifetime in msec
-    MaxLifetime = 250   ; max lifetime in msec
+    MinLifetime = 1000   ; min lifetime in msec
+    MaxLifetime = 1000   ; max lifetime in msec
   End
 
   Behavior         = SpyVisionUpdate ModuleTag_15
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -2880,7 +2880,6 @@ Object DummyWeaponProjectile
 End
 
 ; Patch104p @bugfix commy2 24/10/2021 Logic that grants player Cash Bounty.
-
 ;------------------------------------------------------------------------------
 Object CashBountyDummy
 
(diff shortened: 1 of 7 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -2912,3 +2911,30 @@ Object CashBountyDummy
...
+    MaxHealth = 1.0
+    InitialHealth = 1.0
+  End
+
+  Behavior = LifetimeUpdate ModuleTag_02
+    MinLifetime = 250   ; min lifetime in msec
+    MaxLifetime = 250   ; max lifetime in msec
+  End
+
+  Behavior         = SpyVisionUpdate ModuleTag_15
```
xezon, line 2934: Indentation

Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -2912,3 +2911,30 @@ Object CashBountyDummy
     Bounty                  = 20%
   End
 End
+
+; Patch104p @bugfix commy2 30/07/2022 Logic for Sathack 1 to reveal Command Centers.
+;------------------------------------------------------------------------------
+Object SatelliteHackOneDummy
+
+  ; ***DESIGN parameters ***
+  EditorSorting   = SYSTEM
```
xezon, line 2920: Indentation

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8703,9 +8703,15 @@ Weapon Demo_ScudStormDeathWeapon
...
 ;------------------------------------------------------------------------------
 Weapon CashBountyDummyWeapon
   DelayBetweenShots = 250                ; time between shots, msec
   FireOCL = OCL_CashBountyDummy
 End
+
+; Patch104p @bugfix commy2 24/10/2021 Spawns dummy object with sathack 1 logic over and over.
+;------------------------------------------------------------------------------
+Weapon SatelliteHackOneDummyWeapon
+  DelayBetweenShots = 250                ; time between shots, msec
```
xezon: We could set Delay to 1000 ms to spawn it less regularly and save on resources.

Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
commy2: It's copy pasted from the 1.04 SpyVisionUpdate of the Internet Center for convenience.

xezon: For new code we can fix the indentation so it reads nice. If not, also ok.

commy2:
Now scans every 1 second for a new CC.

xezon (approved):
Looks good.

PR #793, opened 2022-07-30
Fix: Worker Dies First Time Trying To Rebuild Scud Storm

### LINKED ISSUE #757: Anthrax poison kills first Worker in GLA SCUD Storm Rebuild
Anthrax poison kills first Worker in GLA SCUD Storm Rebuild. This looks silly. It would be more logical to have the Worker spawn after the Poison disappeared.

shot_20220725_191421_4

shot_20220725_191426_7

### LINKED ISSUE #758: Demo GLA Scud Storm rebuilds around 20 seconds quicker than all other SCUD Storms
Demo GLA Scud Storm rebuilds around 20 seconds quicker than all other SCUD Storms.

Potentially introduced with
* #70

Related to
* #757

## Toxin SCUD Is Killed

shot_20220725_191259_2

## Toxin SCUD Starts Rebuild

shot_20220725_191318_3

## Demo SCUD Is Killed

shot_20220725_191616_8

## Demo SCUD Starts Rebuild

shot_20220725_191627_9

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/757
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/758

1.04: First Worker dies in poison cloud

Patch: Worker waits until poison cloud is gone

This has one drawback. If the new first Worker is killed, it will take 40 seconds instead of 20 seconds for another one to come out. Is this fine?


Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Slth_GLABuildings.ini
```diff
@@ -8154,7 +8154,7 @@ Object GC_Slth_GLAScudStorm
     SubdualDamageHealAmount = 100
   End
   Behavior        = RebuildHoleExposeDie ModuleTag_08
-    HoleName      = GC_Slth_GLAHoleScudStorm
+    HoleName      = GLAHoleScudStorm ; Patch104p @bugfix commy2 30/07/2022 Fix Worker dies from poison on first try.
```
xezon, line 8157: Why is this applied to `GC_Slth_GLAScudStorm` only?

Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
```diff
@@ -13296,9 +13302,68 @@ ObjectReskin GLAHoleScudStorm GLAHole
...
+  Behavior                    = RebuildHoleBehavior ModuleTag_04
+    WorkerObjectName          = GLAInfantryWorker
+    WorkerRespawnDelay        = 40000 ;in milliseconds
+    HoleHealthRegen%PerSecond = 0.5%    ;regen this % of HoleMaxHealth per second
+  End
+
+  Behavior             = CreateObjectDie ModuleTag_13
+    CreationList  = OCL_LargeStructureDebris
+  End
+  Behavior             = FXListDie ModuleTag_14
```
xezon, line 13354: Can fix identations.

Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Slth_GLABuildings.ini
commy2: The other ones already use GLAHoleScudStorm (despite there existing other variants which are completely unused in 1.04 and this patch).

Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
commy2: But then the code is different from the GLAHole, and it becomes harder to grep identical snippets of code :(

Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Slth_GLABuildings.ini
xezon: Ok so all Scudstorms will use this new `GLAHoleScudStorm`, yes?

commy2: Yes, all seven of them^^ :pain:

xezon:
> This has one drawback. If the new first Worker is killed, it will take 40 seconds instead of 20 seconds for another one to come out. Is this fine?

Damn. That stinks. I suspect it is fine in most practical cases.

Jundiyy:
I don't think this is a good idea, in the normal game, the Worker comes out and starts the scaffold, like 2% or something and then dies.
By having the Worker come out later, it means the Scud has less health.

xezon:
If that is true, then can give the hole more health. In my test it was 0% I think.

Jundiyy:
40 seconds is quite long, the poison clears after 30+ seconds, is it 33s? So make it straight after that? But that also means a longer wait time for the next Worker?

commy2:
> But that also means a longer wait time for the next Worker?

Yes.

I think the drawback of this fix outweighs the benefits. 

As for https://github.com/TheSuperHackers/GeneralsGamePatch/issues/758, to compensate I made the explosion of the Demo Scud Storm damage the hole in my patch, but left that out for this project. It's not a fix, but a trade off. Unsure though if that is too "gimmicky" for this patch. It makes sense visually though.

Jundiyy:
I would like to leave this as is, Worker shouldn't take so long. I am sure hole has more than 0% before Worker dies.

xezon:
The fix we certainly need is postponing Demo Scudstorm Rebuild somehow. Because right now it rebuilds 20 seconds faster than other GLA SCUD Storms and this is not original behavior.

Jundiyy:
Demo Scud shouldn't be delayed. Perhaps the Workers (could be cloned and) be given more immunity to Toxins so it doesn't die, although that means other units would also have a harder time killing them. So maybe not a good idea. 

Is there perhaps a way to force the Workers to move out of the scaffolds? With just a bit extra immunity to toxins, so it can survive as it moves out?

xezon:
In original, Demo SCUD spawns toxin that kills worker. In Game Patch Demo SCUD does not spawn toxin that kills worker.

commy2:
It's possible to make the Scud Storm death cloud disappear after 15 seconds already instead of 30.

It is also possible to make the Worker that spawns from the Scud only to be immune to poison.

All fixes in the wrong direction though.

Jundiyy:
Wait, I think I'm confused with everything here lol. I will reply to this topic once I am home.

Jundiyy:
Okay, now I get it, so Demo Scud had its toxins removed, We can get back to Demo Scud somewhere else.

I still don't like delaying the rebuild for the rest, as far as I know, the Scud scaffold gets 2% hp after being destroyed, which I think is more important than some EVA sound being annoying. Having the scaffold up can mean the hole is saved from being destroyed from the next attack. Scud Storm being killed and Worker dying isn't happening every 10 seconds. Low priority imo. And shouldn't be a reason to nerf the Scud.

ExiLe-ZH:
Is this correct summary?

All Scuds in 1.04:
Toxin disappears after 30 seconds.
Worker (re)spawns after 20 seconds, so first worker dies.
This means all Scuds will be rebuild after 40 seconds by the 2nd worker.

Patch:
Demo Scud Toxins are removed (good change), so it will be rebuild by the first worker, 20 seconds after hole mode.

So we have only 1 proposal so far? 
Make workers (re)spawn after 40 seconds?
Anyways, I think this is alright, can always manually order another worker to start rebuilding right?
I wouldn't recommended making (re)spawn any faster for any faction, Scudstorms need no buff.

ZekeDlyoung:
Perhaps we could just make the scud storm toxin field be cleared before 20 secs? It's not like units will be able to pass by it anyway with the hole in the way

commy2:
@ Exile
Summary of the issue is correct, however, making the Worker respawn after 40 seconds would mean that if the first one is killed anyway, the next one would also be delayed by 20 more seconds (40 total). It would also be weird that just for this one building and faction, the Worker would be slower to arrive.

It is correct that a Worker could be ordered to manually rebuild the hole, which means in 1.04 any Scud Storm could be rebuild in ~151 seconds.

Total effective from hole to finished rebuild time in current main branch and without such tricks is 160 seconds for non-Demo and 140 seconds for Demo. The difference is not so big and SWs are rarely destroyed just to the hole and then left alone, so I propose just calling this a loss. The scenario is just too rare to worry and make strange exceptions/work-arounds about it.

Edit: fixed numbers

xezon:
I would prefer all SCUD Storms to repair consistently and without glitches.

I agree with commy2 that this is overall rare occurence. We do not need to overthink this. It will be non consequential for the vast majority of matches. And who is to say that the original behaviour is really optimal?

I suggest comparing SCUD Storm build and rebuild times with the other Superweapons to get an overview.

ZekeDlyoung:
Hmmm, is it possible to fire a clearing weapon before/when the worker comes out? That way the toxin field will be cleared when the hole is left and the worker comes out, but if the Scud Storm is destroyed completely the toxin field will stay there for its full lifetime.

The GLA might be crazy and uncaring, but I imagine even they wouldn't want to waste Workers if they can help it

xezon:
Is this correct?

Time in seconds.

| Object                | Build Time | Repair Time | Hole Repair Time |
|-----------------------|------------|-------------|------------------|
| Original USA Particle | 60         | 60 max      |                  |
| Original China Nuke   | 60         | 60 max      |                  |
| Original GLA SCUD     | 120        | 120 max     | 160              |
| Patched Demo GLA SCUD | 120        | 120 max     | 140              |

ZekeDlyoung:
Or perhaps we can change it so that the toxin puddle is created after the hole is destroyed and not just when the Scud Storm is

ExiLe-ZH:
@commy2 
> making the Worker respawn after 40 seconds would mean that if the first one is killed anyway, the next one would also be delayed by 20 more seconds (40 total). It would also be weird that just for this one building and faction, the Worker would be slower to arrive.

Thought the 40 second proposal would be for all factions because it would fix the dying worker aswell,
 then only downside will be the situation where the worker gets killed right? Hence my argument that a player can manually order faster.

I just read there's a second downside? Where the hole has a lower HP before the worker spawns? What are the exact values?

xezon:
If my time values in above table are correct, then I do not think we need to worry about having all SCUD Storm repair 20 seconds quicker from Hole. It still takes 2 minutes and 20 seconds. This is long time.

commy2:
Table is correct +/- a few seconds due to Worker movement and death randomness.

xezon:
Maybe we can agree on 30 seconds spawn time for SCUD Storm worker + Toxin Puddle disappear before 30 seconds?

ExiLe-ZH:
> If my time values in above table are correct, then I do not think we need to worry about having all SCUD Storm repair 20 seconds quicker from Hole. It still takes 2 minutes and 20 seconds. This is long time.

How to achieve this? Faster decaying Toxins?

What about this? 
- Toxins 25seconds
- workers 30seconds.

Opposed to 1.04 all Scuds start rebuilding 10 seconds sooner. Killing the worker now gives enemy an extra 10 seconds to kill the hole while it's weak (basically buying back those 10 seconds).

xezon:
Sounds fair to me.

commy2:
Why would the Workers respawn slower on just this building? Why would the toxin cloud disappear faster just here? This seems very contrived for a non issue.

ExiLe-ZH:
> Why would the Workers respawn slower on just this building? Why would the toxin cloud disappear faster just here? This seems very contrived for a non issue.

Building Scudstorm is big project, workers need mental prep. 

It's honestly really about balance in this case, halving the timeframe for hole mode can make quite a difference. 20 seconds instead of the effectively 40 seconds in 1.04 isn't a whole lot of time, at all. The total rebuild time isn't really the most relevant factor in actual games.

xezon:
> Why would the Workers respawn slower on just this building? Why would the toxin cloud disappear faster just here? This seems very contrived for a non issue.

I am confused. The issue is that first worker dies in toxin puddle on Hole rebuild and Demo SCUD Storm has 20 seconds better rebuild time. Why would the current state be preferable over the above suggestion?

Jundiyy:
Just randomly came to me just now, but it would naturally make sense if Demo Scud was built 20s faster. You can simply see there are no toxins, why would there be a delay. Whilst it does help Demo over others, it also makes sense if we don't see any toxins.

xezon:
But why should only Demo General have this benefit over other GLA factions? It already has the best Superweapon damage...

Jundiyy:
I totally understand that, all I'm saying is that it makes sense when you look at it. If there are no toxins, then they are safe to rebuild.

What about, simple solution. Toxins clear after 20s and Worker builds at 20s?

xezon:
I am fine with that. I am also fine with rebuild after 30 seconds.

ExiLe-ZH:
No from me to reduce all Scuds to 20 seconds, like I said, it's an important timeframe for enemies to try and finish the hole. The rebuild time after that doesn't matter too much.

It's a huge buff from effectively 40 seconds, but I'm willing to settle on 30 seconds, otherwise I think it should be kept at 40 seconds (make this the respawn time).

Jundiyy:
Well I think 30s is fair as well.
In original ZH, Scud repairs to 2%, so it gets some extra hp.
With a 30s delay, it means, for them 10 seconds, it won't have that hp.

This is still 10s faster than ZH but now with a small cost. So yea, 30s seems fair.

xezon:
I can't believe we agreed on something in one session.

Jundiyy:
Lol haha,  A team!

PR #803, opened 2022-07-31
Fix wrong bonus logic in DPS Calculator

xezon:
Global and weapon bonuses were not added together. Fixed.


commy2:
They are averaged, not added :)

xezon:
1 + (bonus% - 1) + (bonus% - 1) ...

That is the math I used.

PR #804, opened 2022-07-31
Change USA Colonel Burton gun audio to classic sound

### LINKED ISSUE #446: USA Colonel Burton uses Sentry Drone fire sound instead of Burton fire sound
USA Colonel Burton uses Sentry Drone shooting sound instead of Burton shooting sound.

xezon:
* Fixes #446

This change adds Colonel Burton classic gun audio.

It currently does not sound right. Audio is not synced with the gun. Despite, the Burton gun audio in Zero Hour sounds good as is.


xezon:
Closing, as there is no need to do anything about it.

PR #809, opened 2022-08-02
Fix: Missing particle effects

### LINKED ISSUE #424: GLA Terrorist suicide particle effects looks more spectacular in Generals
**Generals**
sshot005

**Generals Zero Hour**
sshot004

### LINKED ISSUE #543: Particles may not render at all if their Alpha min max values are zero
Particles may not render at all if their Alpha min max values are zero.

Reported by Enlima29:
> Particle System with zero values is a bit broken. Zero values for initial alpha or color wont scale/change properly, if it is supposed to change on later iteration. Values must be 0.01 at least. This causes some particles to not render correctly while still counting towards the total particle count. Stalker from Operation Firestorm Mod (NLS Discord Server) may have more information on this.

Example
Bad:
```
  Alpha1 = 0.00 0.00 1
  Alpha2 = 1.00 1.00 5
```

Good:
```
  Alpha1 = 0.01 0.01 1
  Alpha2 = 1.00 1.00 5
```

Note:
Alpha1 = [min value] [max value] [frame time]

xezon:
Closes #543
Fixes #424

This change recovers 13 broken particles.

## Fixed with tweaking color values

airCarrierExplosion2
airCarrierHotPillarArms
airCarrierJet01Explosion
airCarrierJetExplosion1
airCarrierJetExplosion2
airCarrierJetExplosion3
ArmExplosionSmall01
BarrelExplosion
BuggyNewExplosionArms
FireBaseHowitzerPillarArms
HotPillarArms
MammothTankExplosionArms
SpectreHotPillarArms

## Original Zero Hour

https://user-images.githubusercontent.com/4720891/182839993-bd081f25-3129-4a0b-8b09-11006d3b2f9e.mp4

## Patched

https://user-images.githubusercontent.com/4720891/182840036-1a85e3d3-2a65-48b4-90c3-02cd067d698c.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
@@ -10336,7 +10336,7 @@ ParticleSystem ArmExplosionSmall01
   Alpha6 = 0.00 0.00 0
   Alpha7 = 0.00 0.00 0
   Alpha8 = 0.00 0.00 0
-  Color1 = R:3 G:0 B:0 0
+  Color1 = R:6 G:6 B:6 0 ; Patch104p @bugfix xezon 02/08/2022 From R:3 G:0 B:0 0 fixes broken effect in game
   Color2 = R:0 G:0 B:0 0
   Color3 = R:0 G:0 B:0 0
   Color4 = R:0 G:0 B:0 0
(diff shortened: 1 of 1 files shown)


commy2:
Second commit doesn't load. How many changes did you make?

I support changing alpha 0 to alpha 0.01 if that really fixes these effects. Not a fan of mass replacing here. It will be pain, but we should actually check what each one does.

First commit: So lets say `R:3 G:0 B:0 0` doesn't work for whatever reason. Why is the solution to use `R:6 G:6 B:6 0`? Why not `R:3 G:0.01 B:0.01 0.01`?

xezon:
The second commit is 247 lines. Let me break it up a bit to try make it show.

xezon:
I broke it into 5 smaller commits and they should load now.

xezon:
I will split this change into 2.

xezon:
Split completed.

xezon:
I tested with `Color1 = R:5 G:5 B:5 0` as opposed to `Color1 = R:6 G:6 B:6 0` and the fix did not work.

This fix looks good to me.

commy2:
So these 13 have been confirmed to not work in 1.04 and are fixed with this?

I wonder what these have in common that the others lack.

xezon:
Stalker:
> These particlesystems all have 2 things in common:
> They have AttachedSystems
> Their Color1 = R:3 G:0 B:0 0

PR #812, opened 2022-08-03
Change: Increase China Nuke Cannon and China Inferno Cannon Armor against Infantry Missiles by 100%

### LINKED ISSUE #811: China Nuke Cannon has weak armor
China Nuke Cannon has weak armor. It appears small things like a Rocket Buggy or a lousy RPG Trooper can take it out in no time. Coupled with the deployment times and the small movement speed, this makes the units very vulnerable. Maybe this should be looked into.

## Stats

WIP

xezon:
Fixes #811

This change increase China Nuke Cannon and China Inferno Cannon Armor against Infantry Missiles by 100%. All other Armor types remain unchanged.

## Rationale

China Artillery Tanks are much weaker against Infantry Missiles compared to the competition.

| Object                                | Armor Type         | Health | Missile hits to kill |
|---------------------------------------|--------------------|--------|----------------------|
| Original GLA Rocket Buggy             | TruckArmor         | 120    | 6                    |
| Original GLA Scud Launcher            | TruckArmor         | 180    | 9                    |
| Original USA Tomahawk                 | TruckArmor         | 180    | 9                    |
| Original China Inferno Cannon         | TankArmor          | 120    | 3                    |
| Original China Nuke Cannon            | TankArmor          | 240    | 6                    |
| Patched China Inferno Cannon (this)   | ArtilleryTankArmor | 120    | 6                    |
| Patched China Nuke Cannon (this)      | ArtilleryTankArmor | 240    | 12                   |

Rocket Men make a good amount of forces of all factions. They are cheap and can come in masses. China's artillery is unnecessarily weak against their missiles, and an Armor buff against them should help to increase their survive-ability.

The Inferno Cannon now takes as many rockets as a GLA Rocket Buggy does. And the Nuke Cannon takes the most hits of all Artillery Tanks. However, the Nuke Cannon is very slow and requires deployment time, so Rocket Soldiers typically will have an easier time chasing it down.

## Original

https://user-images.githubusercontent.com/4720891/182667885-88828e0a-8870-49fe-8efe-61413a795642.mp4

## Patched

https://user-images.githubusercontent.com/4720891/182667939-3b44e72b-1c7f-4f0b-a0ee-9430b06fc36f.mp4


Jundiyy:
Whilst I do like the change, worrying if double armour is too much.

ExiLe-ZH:
> Whilst I do like the change, worrying if double armour is too much.

- I think it's fair when comparing it to Buggies, which is the direct counterpart. Buggy armor already feels weak, but they atleast the ability to retreat fast.

- Inferno's rely on each other and taking 1 out in a group of 3 instantly removes their whole strength.
Taking Inferno's out usually renders the whole China army useless, it provides the firepower to be able to push and clear, the other units are just there to support.

- Now Inferno's often die to random Rockets, GLA can just hammer a bus straight into an army and chances are big it takes out the Inferno's, or a good chunk of them, this is painful. With this change, GLA needs better tactics and coordination to take them out. 

For these reasons it does sound fair to me for Inferno's to atleast have the same armor as Buggies.

Jundiyy:
Yea, hopefully it should be okay, it's only an infantry missile armour change, which isn't the only thing that destroys them.

MTKing4:
Nuke cannon feels too strong now, we haven't discussed why is it a good idea to buff it?

ExiLe-ZH:
> Nuke cannon feels too strong now, we haven't discussed why is it a good idea to buff it?

+1

I personally think it's too much, 6 is maybe too little for such slow clunky unit, 9 would probably be fair so it's even with tommies and scuds, which both are cheaper (and easier to use). That last bit makes it sound 12 is justified, but still feels a bit radical, let's first try 9.

MTKing4:
I think 9 is good as well, it makes it even with the other heavy artilleries

xezon:
Nuke Cannon does cost significantly more than other Artillery units though, and a Generals Point, and is moving slow and has deployment times. 12 RPG hits look ok to me.

EnigmaZift:
Just so you guys know, the Tank armor comes with trade-offs... Even though the Inferno and Nuke Cannon take less rocket shots than the Buggy - they take double the amount of damage from stuff like Gatts and Quads than light vehicles.

Merging the Tank Armor with the 50% infantry missile bonus can make it so that they have no good counters besides Tanks (lol), Jets and Explosives... 

Just something to keep in mind

Jundiyy:
I still feel 12 is too much for Nuke, but it might be safe. Would prefer 9 tbh.

The reason it might be safe is because usually a Bus would go rampaging through and destroy it. So now it just takes a 2s longer.

The times it would matter, would be some single infantry units trying to kill it.

Wouldn't make a difference for tanks or other units.

Infernos, all good.

xezon:
I will revisit this because I think we need to have a closer look at the other Armor values.

commy2:
> I still feel 12 is too much for Nuke, but it might be safe. Would prefer 9 tbh.

This. Nuke Cannon does not need armor/health buffs.

ReLaX82:
Yes, doubling is too much. Maybe afactor of 1.5 to 1.6 is enough.

PR #820, opened 2022-08-05
Fix: Toxin Tractor Model Scrap Stages

### LINKED ISSUE #814: GLA Toxin Tractor does not change model correctly when scrapped
GLA Toxin Tractor does not change model correctly when scrapped.

Enlima29
> It is supposed to be like this. Bigger side barrels on first salvage. Is easily fixable through ini coding.

gla_toxin_tractor_upgrades

commy2:
Enlima29 pointed this one out.

- fix: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/814

1.04:

- Toxin Tractor only change their gun when scrapped up. Toxin Tractor also looks different than CCG Toxin Tractor even before scrapping up.

This patch:

- Toxin Tractor gets bigger side barrels when scrapping up once.
- Toxin Tractor gets back barrel when scrapped up twice.

shot_20220805_142649_2
shot_20220805_142709_3

- USA05 Toxin Tractor is visually unchanged, but I removed the sections that are covered by bigger model parts anyway to be in line with a regular twice scrapped Toxin Truck.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -17109,6 +17109,7 @@ Object Chem_GLAVehicleToxinTruck
   ;UpgradeCameo4 = NONE
   ;UpgradeCameo5 = NONE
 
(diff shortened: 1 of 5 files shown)


xezon (approved):
Looks good to me.

PR #821, opened 2022-08-05
Change: Reduce AFG_ComancheArmor HEALING by 10% before Upgrade

xezon:
This change makes `HEALING` of `AFG_ComancheArmor` consistent with `HEALING` of `AFG_CountermeasuresComancheArmor`.

The Airforce Comanche will repair a bit slower on the Airfield before Upgrade accordingly.

## Stats

| Object                                    | Full repair time |
|-------------------------------------------|------------------|
| Original AirF Comanche                    | 22.0             |
| Original AirF Comanche w. Countermeasures | 24.4             |
| Patched AirF Comanche                     | 24.4             |
| Patched AirF Comanche w. Countermeasures  | 24.4             |

## Additional Stats

helihealing2

## Rationale

This change makes healing times on Airforce Comanche helicopter consistent. It no longer get a penalty in healing time with Countermeasures upgrade, because the healing time is now the same before the upgrade. Healing time is not bumped to 100% on Countermeasures, as that would be a buff for the Airforce Comanche, and it is already powerful enough as is.


xezon:
Commit #822 before this one.

Jundiyy:
Don't think it's a problem, but...I don't see why it needs changing? I know it says consistent, but I don't see it as a good reason. One armour before upgrade, second armour after. Same for healing.

xezon:
Leaving it as is because it has always been broken is not a good reason either. If this change is detrimental in any capacity, then please write so and explain how, otherwise we can assume this tweak for consistency is an improvement.

ExiLe-ZH:
If the logic is to equalize pre- and post upgrade, shouldn't pre-upgrade actually be around 12 seconds? Or post-upgrade 48.8 seconds?

ExiLe-ZH:
> Leaving it as is because it has always been broken is not a good reason either. If this change is detrimental in any capacity, then please write so and explain how, otherwise we can assume this tweak for consistency is an improvement.

Also, is there proof whether it's broken or that it's intentionally balanced like so?

xezon:
No, that would be a 100% buff pre upgrade and 100% nerf post upgrade. That does not sound right.

It is not intentional because the Countermeasures Upgrade Tooltip does not explain this Healing change.

ExiLe-ZH:
> No, that would be a 100% buff pre upgrade and 100% nerf post upgrade. That does not sound right.

It is a big difference, however if you compare to original helix repair, which also has a lot of armor, then 48.8 seconds doesn't sound off.

> It is not intentional because the Countermeasures Upgrade Tooltip does not explain this Healing change.

But there's a change in both cases, in original case the upgrade makes it heal almost twice as fast (relatively, in health points per second), with 24.4 seconds it would be exactly twice as fast.

xezon:
It is not clear to me what values you are looking at, but the difference is 2.4 seconds.

Jundiyy:
As I already said in my first post, it's not a problem to change. But I just don't see a good reason to do the change. Hence why I responded. 

> Leaving it as is because it has always been broken is not a good reason either. If this change is detrimental in any capacity, then please write so and explain how, otherwise we can assume this tweak for consistency is an improvement.

I didn't say, 'leave it as it because it has always been like so' , and I have not said this for a long time, as you very well know.

The way we left Chinooks the same, it has default 50%, yet we left it as it. Simply add HEALING code. Then what difference is there here? Why can't we simply add the HEALING?

ExiLe-ZH:
> It is not clear to me what values you are looking at, but the difference is 2.4 seconds.

If you consider full health to be 100% pre- and post upgrade, then yes. However in reality post-upgrade makes armor about 200%. Then repair is twice as fast if time is the same.

Issue #816, opened 2022-08-04
Generals Promotion buttons do not work reliably


xezon:
Generals Promotion buttons do not work reliably. This is pretty much 100% to reproduce on higher frame rates, such as 240 FPS, but also happens occasionally at lower ones, such as 30.

This issue does not happen in original game.

Production buttons on the Control Bar do not have this issue.

![shot_20220804_171309_2](https://user-images.githubusercontent.com/4720891/182883121-3f88e6fd-9e65-49fa-9380-3e61276cca24.jpg)

Jundiyy:
It happens with lag as well. Sometimes you click a promotion but don't get it. Have to click 2 times.

xezon:
This issue does not happen in original game.

commy2:
Is it easy to reproduce? If so, I guess it is time to bisect the commit history.

Jundiyy:
Maybe this is a different issue then, if you play pvp and there is lag, normal 1.04, sometimes a player may not get his promotion on the first click. Especially if he clicks multiple different ones.

xezon:
> Is it easy to reproduce? If so, I guess it is time to bisect the commit history.

Yes. Load cheap Skirmish map with any General, set unlimited frame rate, click on Promotion button. It will not click ever.

ThePredatorBG:
Unlike most buttons which get activated on mouse down, GP buttons get activated on mouse up (when mouse button is pressed, then released). Most likely, if those two mouse events are not captured within a certain period, the click doesn't register.

Try adding `ON_MOUSE_DOWN` status to the buttons. The file is `GeneralsExpPoints.wnd` I believe. I did not test this.
For example, first button:
```
    NAME = "GeneralsExpPoints.wnd:ButtonRank1Number0";
    STATUS = ENABLED+IMAGE;
```
to
```
    NAME = "GeneralsExpPoints.wnd:ButtonRank1Number0";
    STATUS = ENABLED+IMAGE+ON_MOUSE_DOWN;
```

xezon:
This works.

ExiLe-ZH:
very pro+++++

Jundiyy:
The only downside to this fix is that, if you make a mistake in the GP menu, you can't cancel it.
In the original game, if you mistakenly choose a GP, you can keep the mouse button pressed, then move it away and let go, it will cancel the GP selection.

[...]

PR #825, opened 2022-08-05
Fix inconsistent button styles in Control Bar

xezon:
Related to
* #816

This change streamlines Control Bar button behavior. The following buttons now behave like all the other similar buttons by triggering on mouse button release.

ControlBar.wnd:ButtonIdleWorker
ControlBar.wnd:ButtonPlaceBeacon

menubuttonfix


Jundiyy:
The only downside to this fix is that, if you make a mistake in the GP menu, you can't cancel it.
In the original game, if you mistakenly choose a GP, you can keep the mouse button pressed, then move it away and let go, it will cancel the GP selection.

Which is similar to other things, for example, you click to train a Dozer, you can still cancel it, or you click to upgrade mines, you can still cancel.

xezon:
Comment by wrong change.

PR #827, opened 2022-08-05
Fix: Grant Faction Inherent Upgrades By Dummy Instead Of By Building, Version 2

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/539

Same as https://github.com/TheSuperHackers/GeneralsGamePatch/pull/539, but without taking another precious slot in StartingUnits.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
@@ -58,7 +58,7 @@ PlayerTemplate FactionAmerica
   SpecialPowerShortcutButtonCount = 10
   DisplayName       = INI:FactionAmerica
   StartingBuilding  = AmericaCommandCenter
-  StartingUnit0     = AmericaVehicleDozer
+  StartingUnit0     = AmericaStartingUnitsUpgrades
   ScoreScreenImage  = America_ScoreScreen
   LoadScreenImage   = SAFactionLogoPage_US
   LoadScreenMusic   = Load_USA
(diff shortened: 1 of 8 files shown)


Patch104pZH/GameFilesEdited/Data/INI/PlayerTemplate.ini
```diff
@@ -58,7 +58,7 @@ PlayerTemplate FactionAmerica
   SpecialPowerShortcutButtonCount = 10
   DisplayName       = INI:FactionAmerica
   StartingBuilding  = AmericaCommandCenter
-  StartingUnit0     = AmericaVehicleDozer
+  StartingUnit0     = AmericaStartingUnitsUpgrades
```
xezon: Maybe call this `AmericaStartingThings` as a more generic wording?

xezon:
I think this is very nice that we can bundle everything into one item. Well done.

Now for the core rationale of the change, I have not yet understood why it is a problem that a player gets the Faction perks when stealing their Dozer and building their Structures. Could you elaborate?

commy2:
I think it is a very obscure mechanic. It is hard to understand why building a Infa War Factory would grant Nationalism. It's not just buildings either. Building a Stealthed Rebel grants Camouflage. Additionally, there are subtile bugs with this. Capturing an Infa War Factory does not grant Nationalism. You have to actually complete a scaffold for it to count. When the Stealthed Rebel comes from a Combat Bike or Rebel Ambush from a Stealth General Command Center, it does not grant Camouflage either. It has to be one trained from a Barracks.

It seems more intuitive to me that a faction would start with upgrades, and what happens with capturing their buildings / Dozers is entirely divorced from that.

xezon:
I just tested with Stealth General, built a Combat Bike, evacuated, and the Rebel was stealthed.

commy2:
The Stealthed Rebel grants Camouflage, but does itself not need Camouflage to be stealthed. 

Repro steps:

- Do not build a Stealthed Rebel from Barracks.
- Build a Comabt Bike
- Capture a Tox Barracks
- Build Toxin Rebel
- Build another Combat Bike (optional)
- Observe how Toxin Rebel is not stealthed
- Build a Stealthed Rebel from Barracks
- Observe how Toxin Rebel is now stelthed.

xezon:
Ok. Will this change affect veterancy level on unit spawn?

commy2:
No.

xezon (approved):
Looks good to me.

PR #830, opened 2022-08-06
Make Dummy Upgrade Global

commy2:
The current uses of the dummy upgrade don't require it to be `Type = Object`. If such a case were found, it could always utilize a upgrade like OverlordGatling or FakeBuildingsWorker.


Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
```diff
@@ -811,9 +811,9 @@ End
 ; Added a new dummy object-class upgrade for use in places where a local upgrade is required.
 ; 23/08/2021
 
+; @bugfix commy2 06/08/2022 Upgrade is now global.
```
xezon: Remove this comment and update the one above.

xezon (approved):

PR #831, opened 2022-08-06
Clean Up Dummy Upgrade Code

commy2:
- requires and forks https://github.com/TheSuperHackers/GeneralsGamePatch/pull/830

- Remove the superfluous `ExemptStatus`. `ExemptStatus` flag does nothing here. `GrantUpgradeCreate` is only triggered once the scaffold is completed (or the unit leaves a factory or the at the start of the map for preplaced objects).
- Use `Upgrade_DummyUpgrade` instead of `Upgrade_ChinaOverlordGattlingCannon` on the vUSA SDZ. This appears to be a minor mistake.

Despite this being a global upgrade, I left it so the buildings grant it. This helps for maps where the player does not chose their faction.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -5105,7 +5105,7 @@ Object AmericaPowerPlant
     ReallyDamagedParticleSystem3 = Bone:None RandomBone:No PSys:StructureTransitionSmallShockwave
   End
 
-  ; @bugfix - hanfield
+  ; Patch104p @bugfix - hanfield
   ; Added two extra modules for Microwave Tank bug fix.
   ;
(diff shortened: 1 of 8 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
```diff
@@ -20415,12 +20413,12 @@ Object AmericaSupplyDropZone
   ; 22/08/2021
 
   Behavior = GrantUpgradeCreate ModuleTag_MicrowaveTankFix01
-    UpgradeToGrant = Upgrade_ChinaOverlordGattlingCannon
+    UpgradeToGrant = Upgrade_DummyUpgrade
     ExemptStatus = UNDER_CONSTRUCTION
```
xezon, line 20417: What about this one?

xezon (approved):

PR #833, opened 2022-08-06
Change: Add China Propaganda Center to Prerequisite of Internet Center

xezon:
This change adds the Propaganda Center or Warfactory prerequisite to the Internet Center. Reason being, that Hackers also require the Propaganda Center. This makes the requirement more intuitive.

## Original

Internet Center requires Warfactory.

## Patched

Internet Center requires Warfactory OR Propaganda Center.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -35045,7 +35045,7 @@ Object ChinaInternetCenter
   Side              = China
   EditorSorting     = STRUCTURE
   Prerequisites
-    Object            = ChinaWarFactory
+    Object            = ChinaWarFactory ChinaPropagandaCenter ; Patch104p @tweak from ChinaWarFactory to be consistent with Prerequisites of ChinaInfantryHacker
   End
   BuildCost         = 2500
   BuildTime         = 30.0           ; in seconds
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
@@ -13753,7 +13753,7 @@ Object Infa_ChinaInternetCenter
   ExperienceValue     = 200 200 200 200  ; Experience point value at each level
 
   Prerequisites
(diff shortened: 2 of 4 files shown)


xezon:
Could this change break any Campaign mission?

commy2:
You could use the IC to access Sathack I without the intention of building any Hackers. Not that that is useful much, but you could.

Jundiyy:
I dislike this change, whilst I understand the reasoning, it's not a needed a change and does ever so slightly nerf China. Of course, this is mainly in FFA as IC is hardly built in 1v1s. 2v2s/team games? Not sure how often IC is built for vision.
Think this should not be implemented.

Stubbjax:
The primary rationale for this change is that it makes far more logical and intuitive sense that the Internet Centre's availability is connected to Hacker availability, as the two are effectively counterparts that belong together. It is incredibly rare and often disadvantageous to construct an Internet Centre _before_ a Propaganda Centre, so there is little merit to the minuscule loss of flexibility as a result of this change.

A lot of confusion arises when a player has a Propaganda Centre but has lost their War Factories, and can no longer construct an Internet Centre to protect their existing Hackers. A solution that would hopefully satisfy the purists while also solving the counterintuitive building prerequisite would be to add the Propaganda Centre as an additional (optional) prerequisite.
```
Object ChinaInternetCenter
  Prerequisites
    Object = ChinaWarFactory ChinaPropagandaCenter
  End
End
```

xezon:
500 IQ suggestion @Stubbjax

xezon:
Updated.

Jundiyy:
Yes, I think that's a better change.

ExiLe-ZH:
What about losing your Propaganda Center? Then you can't (re)build an IC to protect your Hackers?

I also think the Satellite Upgrades should be more accessible with this patch, this change doesn't help then.

Or am I understanding it all wrong?

xezon:
In layman terms:

## Original

Internet Center requires Warfactory.

## Patched

Internet Center requires Warfactory OR Propaganda Center.

ExiLe-ZH:
> In layman terms:
> ## Original
> 
> Internet Center requires Warfactory.
> ## Patched
> 
> Internet Center requires Warfactory OR Propaganda Center.

Ahhhh the OR is VERY PROFESSIONAL!

Issue #745, opened 2022-07-23
China Helix and USA Chinook repair at half rate


xezon:
China Helix and USA Chinook and USA Combat Chinook repair at half rate on an Airfield.

```
Armor ChinookArmor
  Armor = DEFAULT           50%  ;this sets the level for all nonspecified damage types
```

Jundiyy:
I think add the 50% code to the Armour and leave as is. Giving them 100% repair rate might be too powerful.

ExiLe-ZH:
Lix feels painfully slow tbf, some sort of buff wouldn't hurt. Would help vs GLA that can instantly heal all it's units. 

Nooks are rarely repaired by players, it's not worth to build an airfield for this purpose, neither with a buff. However when you already have an airfield it might be an option to repair it if it's faster. Currently it's just better to make a new Nook.

How do they compare to Comanche healing exactly though?

commy2:
Related: [GLA Structures repair at 75% with Fortified Structure upgrade](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/228)
[Fix: Make Reduced Repair Rate On Fortified Structures Explicit](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/246)

Also: This also happens to AFG Comanches, but only after the Countermeasures upgrade. They heal at 90% rate. Only AFG, not the other USA sub-factions.

Jundiyy:
As mentioned on Discord, 100% would be too much and make it quite powerful. But having 100% might mean we don't need any other changes to China army since it would make Helix stronger, faster to repair and reuse. 

Perhaps a small increase would be okay.

Chinooks don't need to be changes imo, perhaps make a copy of the armour and give Helix the edited one.

xezon:
I doubt a faster Helix repair will buff China so much that it is on par with USA and GLA. Airforce will just send one King Raptor that penetrates 20 Rocket men and kill it right in the air when the Helix repair is done.

Jundiyy:
That is true, but in games where China spams Helices, this would become even better. But, thinking over it, perhaps maybe that buff is needed. Maybe it wouldn't overdo it.

commy2:
Note that repair ports in War Factories and Airfields work differently in general.

For War Factories, any unit is repaired by 20% of their total health per second.
Airfields on the other hand heal an absolute amount. They heal 10 health points per second.

Health points vary widely between units. A MIG has 160 health, a Helix has 300 health.
[...]

PR #834, opened 2022-08-06
Change: Add Airfield or Warfactory Prerequisites to China Speaker Tower

### LINKED ISSUE #818: China Propaganda Tower is rarely used
China Propaganda Tower is rarely used.

## Stats
BuildCost 500
BuildTime 10
Requires 1 Power (1/10 of Power Plant)
Requires Propaganda Center

## Why is it not used so much?
It is used much less than Tunnel Networks, Bunkers and Firebases. Rarely it is dropped at a defensive line near a forward supply stash as extra defensive boost. For Infantry General it is next to useless, because all Outposts emit Propaganda already. The Propaganda Center requirement means it is not eligible for early game. And 500 cost is not particularly cheap to spam with. It also does not take much to destroy it.

## What would making it more accessible mean?
Propaganda Tower would mainly boost China defense if it was more accessible. It can be placed next to Warfactory to defend against powerful Humvee or Combat Chinook rushes. It can be placed next to Airfield to repair Migs and Helixes faster. It can be placed next to Bunkers to make Rocket Soldiers shoot more. It can be placed at strategic positions in Home Base to boost defensive units.

Related:
* #640
* #745

### Proposal 1
Reduce Propaganda Tower cost.

### Proposal 2
Change Propaganda Center requirement to Warfactory or Airfield requirement. Other Propaganda things can also be build before Center, for example Infantry Outpost and Helix Propaganda Upgrade.

### Proposal 3
Make Speaker Tower repair structures after Subliminal upgrade.

xezon:
Closes #818
* #818

Related:
* #745

This change adds the Airfield or Warfactory prerequisites to China Speaker Tower.

## Original

China Speaker Tower requires Propaganda Center.

## Patched

China Speaker Tower requires Propaganda Center OR Airfield OR Warfactory.

# Rationale

This will make the China Speaker Tower accessible in early game. It is mainly useful to establish a stronghold within the China home base to defend against enemy forces. It can also be used to repair Aircraft quicker near the Airfield. And it can be used for propaganda pushes, if China can protect a Dozer to build on the front lines.

A Speaker Tower gives twice the bonus that a propaganda unit gives (except Assault Troop Crawler).

## Reference Stats

| Stat   | Value |
|--------|-------|
| BuildCost | 500 |
| BuildTime | 10 |
| MaxHealth | 300 |
| EnergyProduction | -1 |
| Radius | 150 |
| HealPercentEachSecond | 2% |
| UpgradedHealPercentEachSecond | 4% |

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -8804,7 +8804,7 @@ Object Boss_SpeakerTower
     DamageFX          = StructureDamageFXNoShake
   End
   Prerequisites
-    Object            = Boss_Airfield
+    Object            = Boss_Airfield Boss_WarFactory ; Patch104p @balance from Boss_Airfield to make it much more accessible throughout the game
   End
   CommandSet          = ChinaSpeakerTowerCommandSet
   ExperienceValue     = 50 50 50 50  ; Experience point value at each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -15313,7 +15313,7 @@ Object ChinaSpeakerTower
     DamageFX          = StructureDamageFXNoShake
   End
   Prerequisites
-    Object            = ChinaPropagandaCenter
+    Object            = ChinaPropagandaCenter ChinaAirfield ChinaWarFactory ; Patch104p @balance from ChinaPropagandaCenter to make it much more accessible throughout the game
   End
   CommandSet          = ChinaSpeakerTowerCommandSet
   ExperienceValue     = 50 50 50 50  ; Experience point value at each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
@@ -7418,7 +7418,7 @@ Object Infa_ChinaSpeakerTower
     DamageFX          = StructureDamageFXNoShake
   End
   Prerequisites
-    Object            = Infa_ChinaPropagandaCenter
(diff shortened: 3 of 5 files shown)


Jundiyy:
Interesting change. But why Power Plant? Wouldn't it be better to lock behind WF?
Some mods have it locked behind Barracks, which I do like (even though it requires Power - opposite of Firebase).

I think this would help Chinas to some level, giving a bit higher chance at defending a no eco, although it does mean the benefit is only in one place and costs nearly a Gat price...so maybe not. Barracks rushes could be considered, although I think a Bunker would be better. Gives protection, range and damage. 

At the same time, this can still be achieved when using the Lix Strat, costs more of course but is needed when doing Lix.

Would need some good testing, I know a lot of players don't do it regularly, but sometimes we see the Speaker Tower for rushing. Bring the Dozer behind the army and build a Speaker Tower half way or more so that units can fall back a bit if needed.

A negative point would be, that it is quite an obvious change from original game. Which is also quite important.

commy2:
I do not think a change like this is necessary. This will just throw people off.

xezon:
In Original game, Speaker Tower cannot be used for early rush, as it requires Propaganda Center.

Originally I tried to put it behind Warfactory OR Airfield, but figured that does not work. The setting does not allow for a choice of 2. So I thought about Barracks, but chose Power Plant. Barracks would also work.

Jundiyy:
Wasn't clear, my mistake, by rush I meant pushing forward. Not obvious, I know.

And you can set it behind them buildings all at once.
In the prerequisites, you place both object names on one line.
ChinaWarFactory ChinaAirfield

This makes it WF or AF. Check Strat Centre to compare. 

Or you can even use 2 lines, so 2 buildings are a must.
China Barracks
ChinaWarFactory ChinaAirfield

This would mean, Barracks + WF or AF.

xezon:
Oh nice. I will update accordingly.

xezon:
I now set it to Propaganda or Airfield or Warfactory. This makes it much more accessible. The tooltip only shows "Airfield or Warfactory". Looks like the UI code is not fit to display 3.

Jundiyy:
👍
And yes, for some reason it won't show everything.

ExiLe-ZH:
With the fact that GLA and USA can heal their units much more efficiently early game, it might be a fair change (it's also without Subliminal Messaging).

So I'd like to test this.

PR #835, opened 2022-08-06
Change: Reduce cost of China Speaker Tower by 20 %

xezon:
Reference: #818

Reduces cost of China Speaker Tower by 20 %


xezon:
I do not think this is necessary currently.

xezon:
We have no data that this is necessary at this time. Closing.

PR #836, opened 2022-08-06
Change: Make USA Supply Drop Zone Cargo Plane unattackable

### LINKED ISSUE #79: USA Supply Drop Zone Cargo Planes are attackable

commy2:
* Closes #713
* Fixes #79
* Related #1390
* Related #1391

USA cargo planes have a significant detrimental impact on gameplay and performance. With this change cargo planes from Drop Zones are no longer ~~attackable~~ auto attacked (#2279). Consequences:

* Gets rid of attack notification spams on Radar and Voice
* Improves performance in spam/AOD games where many planes no longer draw fire
* Mig, Raptor, etc will no longer auto attack Cargo Planes on Guard Mode
* Cargo Planes will fly the same route, so enemy player knows USA has a Drop Zone where it comes from
* USA Faction loses its free cannon fodder unit
* ~~Cargo Planes can NOT be attacked manually to prevent the money drop~~
* ~~The opposition is no longer awarded 40 XP per cargo plane when not able to destroy them~~

## Revision

This change has been improved with

* #2279

# Rationale

USA is the only faction that can build a structure, the Supply Drop Zone, that triggers a scripted event that spawns a plane flying across the map every 2 minutes. Owning 12 such Drop Zones would spawn a plane every 10 seconds on average. These planes drop ~~6~~ 3 crates with parachutes, trigger audio events and additional particle effects. All this has cost. On top of that, these cargo planes will draw enemy fire, which in turn triggers more logic, more effects to render on screen and bring the performance down further. In long matches, or matches with multiple USA players, these planes contribute to poor performance. On top of that, cargo planes, like General Promotion planes, will actively influence game state. This can be advantageous for USA, and disadvantageous for the opposition. Enemy base defenses, ground units and air units will automatically engage cargo planes, unless their current AI state prevents them. This can cause scenarios where cargo planes trigger friendly fire. It will also act as a free cannon fodder unit for USA that will tank enemy fire, unless the opposition can actively avoid shooting on it, by not using Guard Mode and Attack Move and Idle State. USA is a quite strong faction late game. Arguably it does not need a free cannon fodder unit like this. On top of all that, the cargo planes will render Radar Alarm notifications useless for the owning player once there are enough flying to draw fire regularly. It will trigger alarm, and if player looks at the alarm event, he will see a cargo plane getting shot at, which typically is of no major concern, other than revealing where the enemy forces or defenses are.

~~There is no important reason to have cargo planes auto-attackable. However, making cargo planes force-attackable would allow to shoot them down before they drop the money crates. This however is currently not possible to implement properly. It would have to be revisited with Thyme. With current limitations, making cargo planes not attackable at all is a reasonable action. Secondary income of all 3 factions, USA, China, GLA, is now entirely passive and does not interfere with game state beyond taking space on the map.~~


xezon:
I do not think SDZ and Reinforcement Pad planes should be treated in the very same change. It is not yet clear if Auto Attack on Reinforcement Pad plane is undesirable.

commy2:
I noticed that my cursor changes to the red attack cursor when hovering over the plane. However, left clicking just gives a movement order to my unit.

I can only manually engage the plane by ctrl+clicking the plane. Am I doing something wrong? Is this the same as in the other branch?

xezon:
Yes. It is same on my initial branch. Maybe it can be fixed by messing with the planes KindOf flags, such as removing CAN_ATTACK ???

commy2:
I doubt it. CAN_ATTACK makes it able to fire weapons, not for weapons to be fired on it (unsure why the plane has it).

I find the need to hold ctrl very confusing, especially because the cursor changes to attack, but left clicking does move instead. If the interface is broken like this, then I would rather have the plane not attackable at all (which should be feasible).

xezon:
Ask hanfield, he will know tricks :P

commy2:
This looks fine to me now. Until the Thyme Gods descend and give us a new KindOf that does what we want.

Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
```diff
@@ -2700,11 +2700,6 @@ Object AmericaJetCargoPlane_SupplyDropZone
     Mass = 500.0
   End
 
-  Behavior = StatusBitsUpgrade ModuleTag_NoAutoAttack
-    TriggeredBy = Upgrade_DummyUpgrade
-    StatusToSet = NO_ATTACK_FROM_AI
-  End
```
xezon: Maybe we leave this commented here to act as documentation for the alternative of force fireable planes?

xezon (approved):

xezon:
scuba:
> Why does this need changing?
> I already read it. It's a unit of a particular player. Why should it not be part of the gameplay?There is no reason that they should not be able to be targeted.

Sexiee:
> Whats the point in changing such things? There is literally no need to do it nor did people ever cry about it. leave it as it is.

User reactions to this particular Pull Request. It appears Supply Drop Zone planes and their negative impact on gameplay are not well understood and not sufficiently described in this Pull Request. We need to explain it better.

Excal:
> Takes more APM to control units, removing it makes it easier and less APM. Disagree strongly.

This was brought forward as an argument to keep auto targeting on Supply Drop Zone planes. It would take more APM (actions per minute) to control units around cargo planes. It is unclear why this player thinks it is a good thing to invest APM into this particular scenario, when there are unlimited ways to invest APM into other avenues, such as attacking enemy from 2,3,4,5 different angles and microing multiple armies, which is seldom done in practice, because players are already busy enough with the few things they are doing.

On top of that, Supply Drop Zone planes drawing fire all over the map will inevitable lower performance. And if one players drops below 30 FPS, it will reduce game speed for everyone, which automatically lowers the APM requirement to control things, as there will be more time to do more things with lower game speed.

Stubbjax:
> It's a unit of a particular player. Why should it not be part of the gameplay?

Because the sole purpose and design of the cargo plane is _flavour_. It serves as a realistic and visually distinct way for USA to earn its secondary income. The fact that they have even the slightest potential to change the outcomes of games is seriously questionable.

> There is no reason that they should not be able to be targeted.
> There is literally no need to do it nor did people ever cry about it.

Completely false. Here are several reasons:

1. They disrupt other vital mechanics such as aircraft guard mode.
2. They can end up getting your own units / buildings killed due to collateral damage.
3. They distract AA units from shooting at meaningful targets.
4. They create endless Under Attack or Unit Lost alerts to the owning player.
5. They create additional performance costs with all the shooting / destruction / effects.
6. They distract players from the important battles with all the shooting / destruction / effects.

I'd love to see some well-reasoned arguments as to why USA's secondary economy should come with free periodic cannon fodder that the player can opportunistically utilise at no cost. The only two I can think of is the introduced inconsistency with other cargo planes (which I really don't like, but the alternative is worse), and the relatively substantial difference with 1.04.

> Takes more APM to control units, removing it makes it easier and less APM. Disagree strongly.

These are weak arguments for such strong disagreement. Why not force players to manually target every unit to attack? Perhaps resource gatherers should be manually ordered to collect and drop off every delivery? Should units automatically shoot enemy buildings in addition to enemy units and defences? All of these changes would increase APM and certainly make the game harder, but the game would not be as fun. It's worth considering why. Various mechanics are automated for a reason.

ExiLe-ZH:
@Stubbjax 

I agree with every single point you made. 

However to me, your very first point is exactly the reason to start over from the core-design and improve it further by making it more believable. Like the whole background story of this game and the faction identities are based on the real world (conflict). That's what makes this game cool and gives it character.

In an all-out war scenario it just makes zero sense that planes don't get shot when entering hostile territory. Ofcourse they will get shot down in real life! That's why it's the way it is in 1.04! The performance impact and other gameplay issues simply have to be oversights from the devs, same how it's a design flaw (probably laziness) how planes straight up suicide over enemy bases.

Yes, the current approach does improve performance and solves gameplay issues, but the implementation isn't entirely satisfying yet. Why stop here? Why not combine and implement both things? Then you fixed both the design and gameplay/performance issues. Now that we have the chance to improve the design from the core up, there's zero reason not to implement smart routing for transportation planes.

#GivePlanesBrains.

xezon:
> Why stop here?

Because it means more work for us.

ExiLe-ZH:
I'll figure it out, don't want to put more weight on you guys.

xezon:
I appended a rationale in first post.

ImTimK:
I wanted to leave a final (hopefully unbiased) case summary here:

1) Why the Dropzone design for USA? 
Because it's a realistic, thematic and cool visual representation of USA as a faction, same how Black Markets thematically represents GLA as a guerrilla style of faction e.g.

2) Intended gameplay dynamics, the advantages:
-SDZ gains more money than Black Markets/Hackers.

3) Intended gameplay dynamics, the disadvantages:
-Planes are interceptable before the money drop-off.
-Planes give away intel of the positions and amount of DropZones (mostly likely intended).

4) Gameplay side effects:
USA players learned that Cargo Planes offer more advantages after the money drop-off, for example distracting Units or Base Defenses so that USA can combine attacks or genpowers more successfully. Also does it make Air Guard mode less useful because aircraft auto-attack the plane.
It's very unlikely though that these advantages were intended gameplay dynamics, does USA need these advantages in the first place? 
Some players prefer how these dynamics add more skill and tactical opportunities, while others feel it only adds random frustrating moments. Would be good to have data of how many people like/don't like these side effects.

4) Performance:
It's not exactly clear how much performance is saved with a fix, it would be good to have quantifiable data.

5) Final decision:
It's not a case of bugs or exploits, so I think it's fair to base the final decision on how much performance is saved and how many people would (not) prefer a redesign.

kABUSE1:
> Completely false. Here are several reasons:
> 
>     1. They disrupt other vital mechanics such as aircraft guard mode.
> 
>     2. They can end up getting your own units / buildings killed due to collateral damage.
> 
>     3. They distract AA units from shooting at meaningful targets.
> 
>     4. They create endless Under Attack or Unit Lost alerts to the owning player.
> 
>     5. They create additional performance costs with all the shooting / destruction / effects.
> 
>     6. They distract players from the important battles with all the shooting / destruction / effects.

I strongly disagree with all of these points.

1 and 2: Fixing this by changing the alliance status of dropzone cargo planes is the worst imaginable way this could be approached. Having planes idling on guard in your base is considered a "skill issue" and can and will be punished by placing scaffolds or spawning rebel ambush. Also don't forget that all other large planes would still be auto-attacked, making this an inconsistent fix.
There is this ongoing issue that would perfectly fix all of this: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1400

3: That's the point why it is desired to keep it. It's considered part of USA that you can use cargo planes as cannon fodder.
See https://youtu.be/1d3R72I6p9E?t=784

4: That can most definitely be fixed by working on these sound triggers.

5: Makes no sense. Attackable drop zone planes -> planes die sooner -> less planes per frame. I absolutely legitimately think that this wouldn't make any difference, unless you compare it to entirely removing the cargo planes.

6: Remove "Under attack" and "Unit lost" warning will be a good enough way of fixing this. If desired, we could also reduce the noise level of dying cargo drop zone planes.

Another reason not to go forward with this change is that it would be extremely silly to have drop-zone cargo planes not auto-attacked but ranger drop planes are auto-attacked, even though it's the same exact plane model. So you'd also have to change that into the other usa cargo plane model, which is already used for leaflet / fuel air bomb, changing it would have the effect that ranger drop could be perceived as a fuel air bomb / leaflet.

This is a huge change and should not be commited lightly.

As others have already pointed out, if you were to go forward with this change, there is absolutely no point in even having the planes, hence they should be removed entirely. This would fix the lag, fix the problem that crates can register double, if drop-zones are build close together, fixes the problem that your teammate can have large units/buildings next to your dropzone to steal your cash and all other mentioned points.

If you don't want to remove the planes entirely, there is hope that you understand my point of why I also don't want you to change their auto-targetability. It's just not right.

ExiLe-ZH:
@kABUSE1 

1. Well #1400 is also considered a skill issue, the best players don't guard mode. Saying this from neutral perspective, I'm not for or against it. Making timer longer before auto-land has my preference though.

5. It definitely takes computing power to render hundreds of bullets and rockets going at that plane, it also seems the final explosion can drop the fps a little.

For me personally, the best solution would be to make them shootable, but make them go high altitude and unshootable after the cargo drop.

kABUSE1:
Why would you make them shootable until they dropped their cargo? What difference would that make?


>     2. It definitely takes computing power to render hundreds of bullets and rockets going at that plane, it also seems the final explosion can drop the fps a little.

That can't be denied but it's an exaggeration. At the point where a non-neglectable amount of planes keep streaming in, you most definitely have a lot of stuff that takes the cargo planes out without hundreds of rockets being permanently in the air, such as quads, avengers and gattlings.

Something else that could be done would be to reduce particle number for dead planes to further reduce the lag they could cause, again this could be applied to all sort of cargo planes, as they all have that issue.

ExiLe-ZH:
> That can't be denied but it's an exaggeration. At the point where a non-neglectable amount of planes keep streaming in, you most definitely have a lot of stuff that takes the cargo planes out without hundreds of rockets being permanently in the air, such as quads, avengers and gattlings.
> 
> Something else that could be done would be to reduce particle number for dead planes to further reduce the lag they could cause, again this could be applied to all sort of cargo planes, as they all have that issue.

Well I've paid some attention to this on Domi/Excal streams (both have high-end hardware) and I've seen the fps go down in multiple occasions by 30-40 frames (from 144) when only one plane gets shot.

If something could be done without major visual downgrades that would be great, same applies to genpower planes indeed.

Btw, with high altitude I still mean still visible (just out of firing range), so that the flight paths can still be used as intel to track down SDZ positions.

kABUSE1:
> Btw, with high altitude I still mean still visible (just out of firing range), so that the flight paths can still be used as intel to track down SDZ positions.

Height doesn't influence range vs. planes. No matter how high you spawn a plane in, your units will still auto-attack it. Range vs. planes seems to be calculated only on a 2D level. You'd have to fake it by changing player alliance / making targetable as height changes. Or work on hardcoded stuff...

It's interesting tho how even rocket "fuel" range is only done on 2D level, as you can see the MD's rockets travel impossibly far, it's funny.

https://user-images.githubusercontent.com/47530719/232922827-b06228f9-c091-4110-9913-c4ad41ac59ae.mp4

ExiLe-ZH:
> > Btw, with high altitude I still mean still visible (just out of firing range), so that the flight paths can still be used as intel to track down SDZ positions.
> 
> Height doesn't influence range vs. planes. No matter how high you spawn a plane in, your units will still auto-attack it. Range vs. planes seems to be calculated only on a 2D level. You'd have to fake it by changing player alliance / making targetable as height changes. Or work on hardcoded stuff...
> 2023-04-19.00-58-30.mp4

Ahh ok, thank you for checking that. I wanted to educate myself on this, but hadn't got to it yet. It's unfortunate such fix isn't that easy though.

MTKing4:
having high-altitude planes is bad, they will look too big on the screen

still think this is a good change but too lazy to argue about it, Stubjax made a compelling argument

i will say this though, it's not "Skill issue" if it means you have to babysit your planes as opposed to doing more important tasks i.e increasing the overall quality of the games when you have more successful airstrikes, it will be epic and you'll see

commy2:
What hasn't been mentioned yet is the fact that shooting down an SDZ plane gives 40 experience points. 5000 are needed from Rank 1 to Rank 5.

xezon:
Good observation. I have added this information to opening post.

I think 5000 points are needed from Rank 1 to 5.

xezon:
I have updated opening post to reflect improvement with #2279.

PR #839, opened 2022-08-06
Change: Enable Speaker Tower On Propaganda Center

commy2:
There is a speaker tower on the Prop Center model. This change makes it actually work just like a regular Speaker Tower.

Strat Center detects stealthed units or has a gun, while Palace has 5 fire ports. So this seems fair to me.

shot_20220806_133551_3

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -6701,6 +6701,25 @@ FXList FX_PropagandaTowerSubliminalPulse
   End
 End
 
+; Patch104p @balance commy2 06/08/2022 Propaganda effect with offsets for Propaganda Center
+; ----------------------------------------------
+FXList FX_PropagandaCenterPropagandaPulse
+  ParticleSystem
+    Name = SonicRange
+    Offset = X:-18.6 Y:-48.7 Z:45.0
+    OrientToObject = Yes
+  End
+End
+
+; ----------------------------------------------
+FXList FX_PropagandaCenterSubliminalPulse
+  ParticleSystem
+    Name = SonicRangeUpgraded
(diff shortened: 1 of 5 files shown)


ExiLe-ZH:
Love it.

Or should the tower be installed after purchasing Subliminal Messaging?

Jundiyy:
Since it's not 1.04+ anymore, I'm open for this, it doesn't make the biggest of differences to the game, but still helps a little.
But...now the next question, should it shut down with low power like in mods? I think no, that would be too big of a change.

commy2:
> should it shut down with low power

No.

xezon:
I wonder if this change would incentivize players to strategically place this. Placing it between 2 Airfields for example would repair all planes quicker, but also make it a better target for enemy Generals Powers.

ExiLe-ZH:
> Love it.
> 
> Or should the tower be installed after purchasing Subliminal Messaging?

Rephrased this, it was a question.

commy2:
Subliminal already upgrades normal propaganda to subliminal propaganda, including on this proposed speaker tower, so I don't think so.

xezon:
It's kinda fair to give China Prop this. GLA Palace has 5 Bunker slots and USA Strategy Center has an optional Cannon, but can only have one. China has no perks for it.

ExiLe-ZH:
> Subliminal already upgrades normal propaganda to subliminal propaganda, including on this proposed speaker tower, so I don't think so.

So small pp tower to big pp tower upgrade, oki

> I wonder if this change would incentivize players to strategically place this. Placing it between 2 Airfields for example would repair all planes quicker, but also make it a better target for enemy Generals Powers.

Proxy prop centers, big rush tactics.

MTKing4:
Good Change.

xezon (approved):
Looks good to me.

PR #840, opened 2022-08-06
Fix: Nuke General Mig Tooltip

commy2:
The tooltip of the Nuke Mig says that it "creates a fire-storm in large groups".
It does not create a fire storm.


ExiLe-ZH:
Implement Nuclear Storm pls

commy2:
No.

ExiLe-ZH:
> No.

Why not? 🙁

xezon (approved):

commy2:
> Why not? 🙁

The first mod I ever made was turning the missiles of the Scud Storm into the Chinese nuke missiles. A classmate wanted that back in like 2004.
It was fun.

PR #842, opened 2022-08-06
Fix: Air Force And Super Weapon General Spectre Tooltip

commy2:
The AFG and SWG 3 stage Spectre tooltip claims the recharge time is 4 minutes. But the actual recharge time is only 3 minutes. 

SWG uses the same buttons / special power as AFG. The recharge time actually is 4 minutes for vUSA and Laser.


ExiLe-ZH:
Hmmm should it be 4 for air/sw aswell?

xezon (approved):

commy2:
> Hmmm should it be 4 for air/sw aswell?

Perhaps. This pull request only fixes the tooltips, making the text state the facts.

PR #847, opened 2022-08-06
Fix: Patch104p Disclaimer In Main Menu

commy2:
before:
shot_20220806_162938_1

after:

shot_20220806_170706_2


xezon:
While at it, you can also fix the position.

Can use this as reference, it has proper position set

https://github.com/TheSuperHackers/GeneralsModBuilder/commit/fb07dbbd4dc4cdb35e5625b9d8d27ca488601edb

xezon (approved):

PR #848, opened 2022-08-06
Fix: Air Force General Misspelling

commy2:
Literally unplayable.


Jundiyy:
Lol haha

Anyway, I think both are correct?

commy2:
Not in Arstotzka.

xezon (approved):

PR #864, opened 2022-08-11
Change: GLA Bomb Truck no longer reveals itself in close proximity of its attack target

### LINKED ISSUE #855: GLA Bomb Truck reveals itself before reaching its attack target
If you disguise Bomb truck as a unit then attack an enemy unit, where there's no stealth detection there, the bomb truck will reveal itself before reaching the target

Trick to avoid that is just driving there manually and press detonate when near the target

xezon:
Fixes #855

This change sets the RevealDistanceFromTarget of Bomb Truck from 100 to 0.

Bomb Truck will no longer reveal itself in close proximity of its attack target.

Originally the same behavior can be achieved by moving Bomb Truck close to the target unit and self detonate it there. Originally Bomb Truck will also always apply full damage on destruction.

## Original

https://user-images.githubusercontent.com/4720891/184116679-5eb5b62f-a9e6-4c25-8f76-b0b09e93b6f4.mp4

## Patched

https://user-images.githubusercontent.com/4720891/184116697-189ba9d1-8111-4d64-9ff9-6afc4a4114dc.mp4


Jundiyy:
Think it's good 👍. Although I'm somehow scared of a new bug appearing if we keep it at 0. No idea why but just have this strange feeling.

xezon:
There are other units that have it on 0 already. For example USA Stealth Fighter.

PR #868, opened 2022-08-11
Add Superweapon test map

xezon:


ExiLe-ZH:
Oh ty, that's very useful.

PR #873, opened 2022-08-12
Change: Allow Supply Centers To Be Placed As Close To Supplies As Other Buildings

### LINKED ISSUE #550: Reduce geometry a bit for the Supply Piles
It's always been quite big but what makes it worse is when you try to build at an angle or the Supply Pile is angled itself.

No angle
Supply01
Angled
Supply02
Distance
Supply 2

This also means other buildings can't be built close by as well.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/550

# Before:

- Supply Centers / Supply Stashes have to be placed a certain distance away from Supplies and Supply Piles. All other buildings can be placed very close to the Supplies.

shot_20220812_205219_2
shot_20220812_205222_3
shot_20220812_205232_5
shot_20220812_205228_4

# After:

shot_20220812_205400_2

- All buildings can be placed freely.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/GameData.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/GameData.ini
@@ -249,6 +249,8 @@ GameData
 
   BuildSpeed = 1.0
   MinDistFromEdgeOfMapForBuild = 30.0  ; buildings may not be constructed this close to a map edge
+
+  ; Patch104p @bugfix 12/08/2022 Allow Supply Centers to be placed right next to Supplies and Supply Piles.
   SupplyBuildBorder = 0.0  ; min dist you can put a supply center from a supply source
 
   ;Terrain height at structure footprint must be within this much to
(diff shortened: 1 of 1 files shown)


xezon:
I love it :-)

ExiLe-ZH:
I love that this gets attention but wait, shouldn't it be other way around? So that there's always enough space for trucks and workers? That would fix the whole bad mining problem no? :)

Jundiyy:
Nah, I prefer it like this. 'More room for weapons.'

Patch104pZH/GameFilesEdited/Data/INI/GameData.ini
```diff
@@ -249,7 +249,10 @@ GameData
 
   BuildSpeed = 1.0
   MinDistFromEdgeOfMapForBuild = 30.0  ; buildings may not be constructed this close to a map edge
-  SupplyBuildBorder = 20.0  ; min dist you can put a supply center from a supply source
+
+  ; Patch104p @bugfix 12/08/2022 Allow Supply Centers to be placed right next to Supplies and Supply Piles.
```
xezon: It is not a `@bugfix`, but a `@tweak` I would say.

commy2:
```suggestion
  ; Patch104p @tweak 12/08/2022 Allow Supply Centers to be placed right next to Supplies and Supply Piles.
```

xezon: Looks good

xezon (approved):

PR #876, opened 2022-08-12
Change: Allow Jarmen To Snipe Pilots Of Landed Comanches

### LINKED ISSUE #180: Grounded USA Comanche cannot be sniped by GLA Jarmen
Grounded USA Comanche cannot be sniped by GLA Jarmen. China Helix (and USA Chinook?) can.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/180
- close #877

UNTESTED

Because Chinook and Helix can be sniped when landed, this change would make Comanche able to be sniped when landed as well.

UNTESTED


Jundiyy:
Think it's okay to do this change.

ReLaX82:
Totally fine.

PR #878, opened 2022-08-12
Fix: Make Supply Warehouses And Piles Unattackable

### LINKED ISSUE #35: Supply stashes have damage effects but don't take any damage
Supply stashes should not have damage effects.

commy2:
- I _think_ this fixes https://github.com/TheSuperHackers/GeneralsGamePatch/issues/35

These are indestructible, it follows that they should not be force-attackable.


xezon:
This change makes is impossible to force fire on Supply Stashes. It is somewhat weird, because everything else in the world can be force fired.

xezon:
I vote for close on this one as per reason above.

PR #879, opened 2022-08-12
Fix: Prevent Hijacker From Stealing Combat Bike

### LINKED ISSUE #218: Player cannot build new GLA Jarmen Kell if Jarmen on Combat Cycle was hijacked by Hijacker
Player cannot build new GLA Jarmen Kell if Jarmen on Combat Cycle was hijacked by Hijacker. This lasts until the Jarmen is evacuated from the Combat Cycle. Reported by Goldfish.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/218

This pull request makes Combat Bikes immune to being hijacked.

# Rationale

Logically it makes no sense to steal a Combat Bike with rider on top when you think about it. And it leads to funny bugs, worst of which is trapping an enemy Jarmen Kell.


xezon:
In this change you do touch GC units. In some other change you advised me to not change GC units. Any guidelines on when and when not to touch?

As for the essence of this change, this looks good to me. Combat Bikes are practically never stolen in regular competitive matches.

commy2:
> In some other change you advised me to not change GC units.

Huh, where? Always fix them, as they appear in the campaign.

xezon:
Ups. I can not recall which one. But it was a recent one.

commy2:
All I remember is that I said not to bother with CINE units.

xezon:
Ah. That could be yes. Perhaps I am confused.

xezon (approved):

PR #888, opened 2022-08-13
Change: Implement Previously Unused Nuke Helix Texture

commy2:
There is an unused texture for the Nuke Helix. Since the Nuke Helix can indeed drop nuclear bombs, I think using this would be neat.

shot_20220813_180742_1


xezon:
Does this reveal any secret Upgrade information to the opponent?

MTKing4:
> Does this reveal any secret Upgrade information to the opponent?

All Upgrades have visuals to tell what the helix has, Napalm/Nuclear Bomb Gives them those little Side wings

I like this change, also does Nuke Mig and Nuke Ovie have similar models as well? If not i think having it would be cool, a small nuclear decal would look sick on both.

commy2:
> Does this reveal any secret Upgrade information to the opponent?

No, but it reveals the faction. In theory the opponent could just click on the Helix and see that "Napalm Bomb" is replaced with "Nuke Bomb" and know anyway, but this does make it more obvious. I don't think that is a problem though.

> does Nuke Mig and Nuke Ovie have similar models as well?

Nuke Mig already uses a different texture with the red star replaced by a radiation sign. Overlord does not have an alternative texture. Nuke Cannon has one where the red star is also replaced with a radiation sign, however, the Nuke Cannon is identical unit wise, so idk.

xezon:
My thoughts on this:
* On the side wings it looks okay
* On the the back side it does not look good, mainly because it is on the inside as well as on seams on the outside

shot_20220813_194754_5

shot_20220813_194806_6

commy2:
I agree it is far from optimal, but the issues you mentioned exist for the normal Helix too. 

shot_20220813_203511_1

MTKing4:
> Nuke Mig already uses a different texture with the red star replaced by a radiation sign.

Yes it has, i just checked.

nukeovi

+ We Can Replace This Star with A Radiation Sign, would be cool
+ I just noticed that it uses that same radar dish that a quad cannon has, and it also floats in the air, if we allow it to spin here, why not on quad as well?

xezon (approved):

Jundiyy:
I think it's quite cool to have faction visual changes. Maybe on earlier units it would make a diff in FFA games, but other units I don't think it matters.

Issue #806, opened 2022-08-01
GLA Toxin General is very strong


xezon:
GLA Toxin General is very strong. There are 2 main reasons for this.

1. Toxin RPG uses Stinger missiles that can track targets
2. Toxin Gamma Upgrade grants hefty damage bonuses

## Related reports and changes

* #1094
* #699
* #872
* #882
* #889
* #890
* #1334
* #2002

## Gamma Upgrade Stats

Anthrax Gamma damage bonuses over respective Anthrax Beta damages.

| Thing                          | Original Gamma Bonus | Patched Gamma Bonus |
|------------------------------  |----------------------|---------------------|
| Toxin Terrorist                | -29%                 | 0%                  |
| Toxin Rebel                    | +25%                 | +25%                |
| Toxin RPG Trooper              | +30%                 | +30%                |
| Toxin Tunnel                   | +25%                 | +25%                |
| Toxin Truck Nozzle (Scrap 1,2) | +64% ,+63% ,+42%     | +50%, +50%, +50%    |
| Toxin Truck Spray (Scrap 1,2)  | +00%, +33%, +28%     | +20%, +33%, +28%    |
| Toxin SCUD Storm Primary       | +10%                 | 0%                  |
| Toxin SCUD Storm Secondary     | +33%                 | +33%                |
| Toxin Stinger Site vs ground   | +100%                | +50%                |
| Toxin Stinger Site vs air      | +66%                 | +50%                |

## RPG Trooper Stats

WIP

### Proposal 1

Remove RPG Trooper missile tracking. No other GLA Faction has it.

### Proposal 2

Reduce RPG Trooper missile tracking range.

### Proposal 3

Reduce Gamma Upgrade Toxin Stinger damage bonus. (Done)

* #882

xezon:
Rage:
> Gamma and this tracking ability is definitely what makes tox so strong lategame.

> I think the tracking is mostly a problem for migs and helixes as toxin becomes so insanely hard to play against for china late game if the toxin general builds some nests / busses / puts some rgs in a building. And the ground army doesn't really stand a chance anyway except for perhaps the inferno spam (but 1 flanking buss and gg for that too)

> Perhaps a slight tweak might already do a lot: make toxin RPG troopers with gamma deal slightly less damage to aircraft but keep the tracking ability and keep the same damage to ground units?

Rising:
> I would remove it. Makes no sense to me for tox RPG’s to be better than other gla’s.

[...]

PR #889, opened 2022-08-13
Change: Redistribute Toxin Tractor Gamma damage bonus to stay within 50% max

xezon:
Reference:
* #806

This change redistributes GLA Toxin Tractor Gamma bonuses to be consistent with the intended bonus from the tool tip.

```
CONTROLBAR:ToolTipGLAUpgradeAnthraxGamma
US: "+50% damage bonus to all toxin units"
```

Note: There is a follow up change to this one:
* #900 

## Stats

fixed_toxin_truck

# Rationale

The Toxin Tractor damage values are just a bit wild. The first 2 Gamma ToxinTruckGun's give 64% and 63% relative bonus over Anthrax Beta, while the 3rd gives 42%. The first Gamma ToxinTruckSprayer gives 0% bonus over Anthrax Beta, but the others 33% and 28% relative bonus over Anthrax Beta. This change redistributes the damages to make more sense.
| Weapon                                        | Bonus over Anthrax Beta | Bonus over Toxin   | Change over Toxin  |
|-----------------------------------------------|-------------------------|--------------------|---------------------|
| Original Chem_ToxinTruckGunGamma              | +64%                    | +105%              |                     |
| Original Chem_ToxinTruckGunGammaPlusOne       | +63%                    | +145%              |                     |
| Original Chem_ToxinTruckGunGammaPlusTwo       | +42%                    | +185%              |                     |
| Original Chem_ToxinTruckSprayerGamma          | +0%                     | +25%               |                     |
| Original Chem_ToxinTruckSprayerGammaPlusOne   | +33%                    | +75%               |                     |
| Original Chem_ToxinTruckSprayerGammaPlusTwo   | +28%                    | +125%              |                     |
| Patched Chem_ToxinTruckGunGamma (this)        | +40%                    | +75%               | -30%                |
| Patched Chem_ToxinTruckGunGammaPlusOne (this) | +50%                    | +125%              | -20%                |
| Patched Chem_ToxinTruckGunGammaPlusTwo (this) | +50%                    | +200%              | +15%                |
| Patched Chem_ToxinTruckSprayerGamma  (this)   | +20%                    | +50%               | +25%                |

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -5663,7 +5663,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon Chem_ToxinTruckGunGamma
-  PrimaryDamage               = 20.5
+  PrimaryDamage               = 17.5 ; Patch104p @tweak from 20.5 to give 40% bonus over Anthrax Beta (-30% total bonus over default)
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
@@ -5688,7 +5688,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon Chem_ToxinTruckGunGammaPlusOne
-  PrimaryDamage               = 24.5
+  PrimaryDamage               = 22.5 ; Patch104p @tweak from 24.5 to give 50% bonus over Anthrax Beta (-20% total bonus over default)
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
@@ -5713,7 +5713,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon Chem_ToxinTruckGunGammaPlusTwo
-  PrimaryDamage               = 28.5
+  PrimaryDamage               = 30.0 ; Patch104p @tweak from 28.5 to give 50% bonus over Anthrax Beta (+15% total bonus over default)
   PrimaryDamageRadius         = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 100.0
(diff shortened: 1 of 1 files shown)


Jundiyy:
Haven't you weakened the Tractor quite a bit? I don't think this is a good change. The Tractor is used with original weapon more than it is with scrap up.

xezon:
It has a tiny bit less bonus overall. Although the Nozzle gun is weaker before Scrap, the Sprayer gun is stronger before Scrap, so there is benefit still.

Jundiyy:
Yes I can see that, perhaps I'm reading it wrong, but it was 64% before and 40% now? That's quite a big jump without a proper reasoning I'd say. And the gun is used quite a lot, I'd say more important than the sprayer.

xezon:
You can make new suggestions on how to redistribute if you take issue with it. Right now the damage progression is very organized and logical and certainly better than originally.

Jundiyy:
Whilst it is organised and logical, it doesn't make it the right one as you've taken away a lot of damage from the unit for no real reason, I don't see as a reason to nerf it. Make them all at 50% at the minimum. I don't see why it should go down to 40%.

How about, all damages at 60%? This nerfs the first 2 very slightly, but buffs the last, which is hardly seen, so would not make much of a difference. 

Although I don't agree with this next idea myself, (at the moment anyway,) perhaps the original damage staying the same and then increasing the others to match it.

xezon:
We can make the Chem_ToxinTruckGunGamma 18.75 damage to be a 50% bonus over Anthrax Beta. Anything higher will exceed the 50% Gamma bonus.

Jundiyy:
Yup, I'd prefer a flat 50%.

ExiLe-ZH:
> We can make the Chem_ToxinTruckGunGamma 18.75 damage to be a 50% bonus over Anthrax Beta. Anything higher will exceed the 50% Gamma bonus.

this sounds more logical.

ExiLe-ZH:
The top table is probably not so easy to read for outsiders to be fair:
- better add top headers that splits clearly splits it in groups of 3 (non scrapped / scrapped / double scrapped).
- rename current headers (to standard toxin / beta upgrade / gamma upgrade) instead of the code names.

I'm trying to understand the rationale of the change though, the maingun is nerfed just to align with the tooltip?
Are 64% and 63% really "too much"? Is there any other reason? 

Guess changes are unnoticeable in practice though, so it's not that I care too much about it. But I feel in the general the fixation on value alignment is a bit too much?

xezon:
Take some time and study the presented data.

ExiLe-ZH:
> Take some time and study the presented data.

It's easy to get lost in how the stuff is ordered and readability is bad.

PR #893, opened 2022-08-13
Fix: Granger Fails To Call In Carpet Bomber In Generals' Challenge

commy2:
At the first wave he's attacking the players base, he is supposed to call in a Carpet Bomber, but he lacks the genpower unlock for it. This worked in Zero Hour 1.00, because no unlock was required to call in the Carpet Bomber.

This patch unlocks the Carpet Bomber for him.

US Military spending at its best.


xezon (dismissed):

Patch104pZH/GameFilesEdited/Maps/GC_AirGeneral/map.ini
```diff
@@ -0,0 +1,46 @@
...
+
+Object AirfieldHanger01
+ RemoveModule ModuleTag_03
+END
+
+Object CivilianAirport
+ RemoveModule ModuleTag_05
+END
+
+Object BaikonurGroundControlCenter
```
xezon, line 25: What are these 3 objects here for?

commy2: That's the map.ini of this map. Additions are at the end of the file.

I believe the BGCC is a prop in that mission.

xezon:
Oh. May I suggest to first add the unedited map.ini and then add the edit on the next commit? This way the diff can be preserved in main branch.

commy2:
We could add all the 1.04 map.ini files from Maps folder / MapZH.big.

Jundiyy:
I'm happy to bring back the Carpet. I remember it vanishing after the update.

xezon:
Merge #895 first.

PR #895, opened 2022-08-13
Add original relevant game map files

commy2:
From maps.big and MapsZH.big respectively.


xezon:
I will look into this.

PR #914, opened 2022-08-15
Fix: Demo General Units Unable To Suicide When Close Or Inside Buildings

### LINKED ISSUE #338: Demo Rebel Ambush may be unable to suicide when spawned close to a building
Video evidence 1: https://youtu.be/fgsCqZR4uGk?t=10132
Video evidence 2: https://youtu.be/A05dAVGn4Uc?t=11940

Potentially related to
* #26

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/338

The weapon is only ever used by the Demogen Suicide button, so range can be safely changed to max. This makes the units explode instantly instead of attempt to leave the building, potentially getting stuck or not suiciding immediatly when ordered.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -4871,7 +4871,7 @@ End
 ; WITHOUT BEING PERMITTED TO TARGET ENEMIES WITH IT!!!!
 Weapon TerroristSuicideNotARealWeapon
   LeechRangeWeapon = Yes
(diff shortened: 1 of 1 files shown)


Jundiyy:
Good fix, this does happen in pvp.

xezon:
Does this affect force fire attack range?

commy2:
No, tested with Technical, Tunnel and Stinger Site.
Units can force-fire anywhere, because they can move. However, force-fire does not work with TERTIARY weapon slot.

xezon (approved):

PR #918, opened 2022-08-16
Fix: Fuel Air Bomb Animation Bug

### LINKED ISSUE #610: Weird behavior fuel airbomb
Weird dynamics fuel airbomb. Similar to the alpha aurora bomb it detonates very erratically. Sometimes it has a significant delay and weird visuals when detonating.

Think this mostly occurs when it is detonating in between buildings somehow? Unsure

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/610

When Fuel Air Bomb gas cloud is hit by crossfire, it may die thus aborting the explosion animation. The damage still occurs normally anyway, so this fix is only visual.

Crossfire was simulated using:

```ini
  Behavior = FireWeaponUpdate ModuleTag_DEBUG
    Weapon = CrossfireSimulatorWeapon
  End
```
on building and
```ini
Weapon CrossfireSimulatorWeapon
  PrimaryDamage = 8.0
  PrimaryDamageRadius = 100.0
  DamageDealtAtSelfPosition = Yes
  AttackRange = 100.0
  DamageType = EXPLOSION
  DeathType = EXPLODED
  WeaponSpeed = 600                     ;  dist/sec
  RadiusDamageAffects = ENEMIES
  DelayBetweenShots = 250                ; time between shots, msec
End
```


xezon:
Would it be possible to make before and after video of this? Or is that to complicated to make? I do not know what this bug means.

commy2:
I will make a video on Saturday when I'm on my PC.

commy2:
The fix is not complete, as there is a different issue with the Fuel Air Bomb.

commy2:
Can't reproduce initial bug on PC, only laptop. More research is required.

PR #929, opened 2022-08-19
Fix: Rocket Buggy Missile Deleted Before Reaching Reinforcement Pad At Max Distance

### LINKED ISSUE #142: GLA Rocket Buggy Can't Damage Reinforcement Pad From Maximum Distance

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGameCode/issues/142
- close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/931

1850 (+1 frame) wasn't enough to make all rockets hit, only half of them did.


xezon:
What would happen if the fuel time was 9999 ?

Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
```diff
@@ -3019,7 +3019,7 @@ Object RocketBuggyMissile
   End
   Behavior = MissileAIUpdate ModuleTag_07
     TryToFollowTarget               = Yes
-    FuelLifetime                    = 1800
+    FuelLifetime                    = 1900 ; Patch104p @bugfix commy2 19/08/2022 Give just enough range to hit Reinforcement Pads at max range. Was 1800.
```
xezon, line 3022: What would happen if the fuel time was 9999 ?

commy2: The missile would follow moving targets for roughly 10 seconds before reaching them or running out of fuel and being deleted, making it effectively impossible to dodge them by making your units retreat.

xezon: Ok. So the 100 ms added here are a tiny buff in principle, as it allows Buggies to hit evading units for 5.5% longer time, correct?

commy2: Pretty much.

MTKing4:
This is a buff for buggies isn't it?

I thought that we needed to adjust reinforcement pad itself not the buggy, something about its bounding box being too big or something? Because buggy doesn't have this issue with any other building, surely the problem is with the building not the unit.

Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
xezon: Ok. In that case the other approach likely is preferable.

commy2:
It's the only building that is flat in its center, aside from the Supply Drop Zone, but that one is not as wide.
Range of the Buggy has to be buffed, or the missile has to explode when running out of fuel. No other work around, aside from remodelling the Reinforcement Pad.

MTKing4:
Well can't you make it so that missiles just collide with the geometry of the building? Meaning they do it on the edge of the model itself, if modelling is the only other solution then I prefer that over doing changes on the buggy

commy2:
> Well can't you make it so

Idk. I won't make that anytime soon.

PR #931, opened 2022-08-19
Fix: Rocket Buggy Missile Deleted When Running Out Of Fuel

### LINKED ISSUE #141: GLA Rocket Buggy missiles vanish ungracefully in long distance
GLA Rocket Buggy missiles vanish ungracefully in long distance. Perhaps this can be made better. It seems one issue here is that the explosions to the ground are barely audible, maybe because the vehicle hit sound takes priority.

https://user-images.githubusercontent.com/4720891/184429594-90f0327d-353f-4c84-b463-9d71f9ba1ea3.mp4

### LINKED ISSUE #142: GLA Rocket Buggy Can't Damage Reinforcement Pad From Maximum Distance

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGameCode/issues/142
- also fix https://github.com/TheSuperHackers/GeneralsGameCode/issues/141
- close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/929

`MissileAIUpdate` module has an optional `DetonateOnNoFuel` token. When set to yes, the missile will no longer be deleted after `FuelLifetime`, but instead killed with `DETONATED` death type. This triggers the damage and effects from `BuggyRocketWeapon`, which otherwise would not happen.

Benefits: 
- no extended range like the other fix
- fixes the particle issue, which would persist with the other pull request

Drawbacks:
- stray Buggy Missiles that would previously be outrunned now hit unsuspecting units that happen to be where the missile runs out of fuel


xezon:
On first look this change makes sense to me.

commy2:
Did you have in mind making the missile explode when low on fuel? I would've expected it to just fall on the ground.

MTKing4:
That is a bit better change than the other one, but it will still do some damage when exploding to the escaping vehicle when its an inch away from the missile dying

xezon:
Explosion on ground hit is what I would expect, as that is what happens with other missiles, such as from Rocket Men and Planes. But straight disappear without doing anything does not sound right.

xezon:
Is there perhaps a setting that makes missile drop to ground after X time and then detonate on ground? This is what happens with Missile Defender projectile I think.

commy2:
> This is what happens with Missile Defender projectile I think.

Do they explode or do they just disappear like the Buggy Missiles do?

xezon:
If my memory does not fail me, then these missiles follow target for a while, then stop follow, drop to ground and explode on ground.

ExiLe-ZH:
> Is there perhaps a setting that makes missile drop to ground after X time and then detonate on ground? This is what happens with Missile Defender projectile I think.

This would look the best for the buggy, I believe works like this for all rocketmen, aswell as stingers. 

The splash damage should be removed though, because the max range gives fleeing units a chance to survive, if survival becomes harder after change, then that be a big unnecessary buff.

commy2:
It appears to me that the rocket already falls to the ground, but since it is so low above ground, it is deleted a few frames after.
And I don't think rocketeer missiles explode when they hit the ground, but are deleted as well.
I honestly prefer the other fix. If making the missile fly 3 more frames is too much of a buff (LOL), then someone savvy in model editing has to add an invisible box into the center of the Reinforcement Pad / fix whatever else is wrong with it.

xezon:
In principle missiles can detonate on ground hit, but I do not understand how to tweak this gracefully. There is KillSelfDelay = xxx setting, but it seems to work odd.

https://user-images.githubusercontent.com/4720891/185737174-51e08907-05d9-40f8-9658-d63de35e7ff1.mp4

commy2:
Hmm, I was looking at the Tank Hunter and RPG. Maybe Missile Defender works differently.

PR #932, opened 2022-08-20
Recover some missing language files

xezon:
This change recovers a few missing language files. The UITRST_SKNP.w3d file is for GLA Terrorist.


xezon:
I have not yet tested what the practical consequence of the missing UITRST_SKNP.w3d file is.

commy2:
It's the model for the Gamma poisoned Terrorist. In 1.04, the Gamma poisoned Terrorist is broken, but I don't know if that is realted to this model.

PR #933, opened 2022-08-20
Improve exbinarystream32 texture

xezon:
Improves exbinarystream32 texture. New numbers, a bit larger and less strong fade.

shot_20220820_091033_1

shot_20220820_091244_2


xezon:
Made 2 more tweaks. A bit more fade to reduce segment clipping and removal of Mip Maps on the texture, because the blur really does not look good when viewing from far away. Now it is always crispy.

shot_20220820_093518_4

commy2:
Wait, we can convert PSD to TGA/DDS by building?

xezon:
Yes. Following conversions are currently supported:

PSD to TGA, BMP, DDS
TGA to BMP, DDS

PR #934, opened 2022-08-20
Add MappedImages Folders To ModBundleItems.json

commy2:


xezon:
We don't need to merge this. It will be superseeded today.

PR #935, opened 2022-08-20
Fix: Nuke Battlemaster Uses Button And Portrait From Normal Battlemaster 

commy2:
Art is from NProject.

# before / 1.04

shot_20220820_094917_1

shot_20220820_094921_2

- same icon as normal Battlemaster, art does not resemble unit
- button and portrait are different, portrait has radiation sign

# after / patch

shot_20220820_094208_1

shot_20220820_094213_3

- art resembles unit
- radiation sign on button and portrait


xezon:
The title needs change.

PR #937, opened 2022-08-20
Fix: Bones Visible On Double Scrapped Quad Cannon

commy2:
# before
shot_20220820_110222_1

# after
shot_20220820_110411_4

These cubes are where muzzle effect will show and are not supposed to be drawn.


commy2:
Muzzle flash still shows and cartridge particle are still spawned correctly after this.

xezon (approved):
Very nice.

PR #938, opened 2022-08-20
Fix: Infantry Muzzle Flash Stuck During Transition Animations

commy2:
# before
https://user-images.githubusercontent.com/6576312/185739197-da89e98d-232c-462b-bd7b-0f4dd1767143.mp4

# after
https://user-images.githubusercontent.com/6576312/185739198-b6818e40-50ed-4b7e-9f69-841697db3161.mp4

- [x] Red Guard / Minigunner
- [x] Ranger
- [x] Angry Mob

- Note: muzzle flash, including glow are still drawn during the actual fire animation
- Also note that this bug happens sporadically and is apparently more likely on night or snow maps

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -3369,7 +3369,7 @@ Object AirF_AmericaInfantryRanger
       WeaponFireFXBone  = SECONDARY Muzzle
       WeaponLaunchBone  = SECONDARY Muzzle
       TransitionKey     = TRANS_Stand
-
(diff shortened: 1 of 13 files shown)


commy2:
Looks done, unless someone finds another unit affected by this.

xezon:
Perhaps Burton and Jarmen?

commy2:
Tested both Jarmen and Burton. Burton has no transitions firing->standing and vice versa, so not a problem with him.
Jarmen apparenlty doesn't even have muzzle flash.
However, I found that the same bug happens with the pistol and AK47 Angry Mobs.

ExiLe-ZH:
Jarmen using flashhider to not reveal his position too easy 😄

commy2:
Done.

xezon (approved):

PR #939, opened 2022-08-20
Tweak Minigunner Recoil Animation

commy2:
- 1.04: recoil always slow (copy paste from Red Guard), does not fit at all
- patch before this: recoil always fast, fits at full firing speed
- this pull request: recoil rate increases with firing speed

https://user-images.githubusercontent.com/6576312/185745806-e8e3f9d0-8e76-4171-ac7d-67907b451333.mp4


xezon (approved):
So many INI lines for such innocent change :D

commy2:
They're necessary, otherwise the animation system picks the wrong matches. E.g. capturing the flag with spun up weapon would display firing animation instead of capturing animation if it weren't for some of those AliasCondition types.

xezon:
Enlima29
> Why use USING_WEAPON_A condition and make them scale based on the rapid firing phase, and had to put lots of additional conditions for the other models? Instead just use FIRING_A condition that it can sync the recoil animation every time the bullet being shot, it's much simpler that way.

commy2
> NProject uses entirely different model and animations for the Red Guard. I ain't touching that. Not saying it wouldn't work with the current assets, but someone else can dig into that. Just replacing USING_WEAPON_X -> FIRING_X obviously doesn't work.

PR #943, opened 2022-08-20
Clarify Refinery Description

commy2:
It does reduce the cost of aircraft. Contrary to the Repair Bay, which also just mentions "vehicles", but does not repair aircraft.


commy2:
I have no idea if anything but EN and DE is correct.

xezon (approved):
If Google translate spits proper sentence, then it is fine.

commy2:
DeepL actually, but yea.

PR #945, opened 2022-08-20
Add: Supply Lines Upgrade To Oil Derrick

commy2:
The upgrade works on them, so even if they are a civilian building, they should feature the icon imo.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
@@ -24888,6 +24888,9 @@ Object TechOilDerrick
   ; *** ART Parameters ***
   SelectPortrait         = SSTechOilDerrick_L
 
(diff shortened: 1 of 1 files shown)


xezon:
Will this show for Neutral or non USA owners?

commy2:
Yes, but those would get this bonus too if they manage to research it from a stolen Strat Center, somehow.

Jundiyy:
Seems okay, although might be confusing at first for some.

xezon:
Hmm yeah I am skeptical because it shows for all.

commy2:
Is it more confusing than Camoflage on the Tox Rebel? Anthrax Beta on the Demo Tractor?

Jundiyy:
Okay, I'm fine with it 😂.
Makes sense.

commy2:
I think this actually clarifies a lot. "Huh Supply Lines? Guess it works on oils too!".

xezon (approved):

PR #947, opened 2022-08-20
Add: Upgrade Voice For MOAB, Sentry Drone Gun, Neutron Shells And Fortified Structure

commy2:
Files are already from the game, but can for this use case only be read from this folder, so they have to be copied there unfortunately.


Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
```diff
@@ -172,6 +172,7 @@ Upgrade Upgrade_AmericaMOAB
   BuildTime          = 60.0
   BuildCost          = 4000
   ButtonImage        = SAMOAB
+  ResearchSound      = EvaVoiceUpgradeMOAB ; @tweak commy2 20/08/2022 Add voice line.
```
xezon: Patch104p keyword missing

commy2: oops, done
good eye

xezon:
All files added. Not tested in game.

commy2:
Tested, werks. 👍

ExiLe-ZH:
> Tested, werks. 👍

Always good, working twerks :D

PR #948, opened 2022-08-21
Fix: Game Performance Dropping Continuously Caused By Patch

### LINKED ISSUE #946: Game performance continuously drops with playtime until it reaches 0 fps
Game performance continuously drops with playtime until it reaches 0 fps. Reproducible with all factions on empty map. Potentially we spawn something every X time that is not cleaned up properly.

commy2:
- should fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/946


xezon (approved):
This fix appears to work.

PR #951, opened 2022-08-21
Fix: USA AI Can Order Fuel Air Bomb And Para Drop From Supply Drop Zone Planes

### LINKED ISSUE #599: USA AI Can Order Fuel Air Bomb And Para Drop From Supply Drop Zone Planes
After playing a lot against AI in this game, on a few rare occasions I would have my units/base get randomly exploded out of nowhere. 

Then to only figure out that an enemy Fuel Air Bomb literally spawned inside my base, this does not seem to happen with any other USA faction.

The AI only needs to unlock the Fuel Air Bomb ability to use it in this way, even after destroying AI's Command Center, the Fuel Air Bomb will keep spawning in your base on its usual timer until the AI is defeated.

https://www.youtube.com/watch?v=-Vygld5QE6A
^ A video showing how this looks like, the video has been edited to speed up the waiting time. ( I knew where the Fuel Air Bomb was going to spawn because I saved the game prior to it spawning. )

USA Laser AI was set on Hard difficulty, I don't know if this would happen on other difficulties.

commy2:
- fix #599

I don't know what these are for. They don't seem to be doing anything. Probably something scrapped?


commy2:
Note commit above, these planes can be selected by arrow keys. The buttons do nothing though. May as well hide them.

Issue #925, opened 2022-08-17
AFG Carpet Bomber may not hit perfectly centered on the target location


xezon:
Carpet Bomber General Powers does not hit target perfectly centered on the target location.

WIP Video

ZekeDlyoung:
I mentioned this before on the discord server, this is because of the "units move faster when moving diagonally" bug, and it affects all GP planes

ExiLe-ZH:
### AFG Carpet bomber test.

**Vertically the first bomb hits the ground too soon and the last bombs will not cover the marker:**

> ![vertical 1](https://user-images.githubusercontent.com/32326277/185180388-625cf067-1e82-4c44-9ea9-f671c43d5a7a.png)
> ![vertical 2](https://user-images.githubusercontent.com/32326277/185180394-6ab80dc2-8c83-4f5e-a4fe-e02315d6d739.png)
> 

**Diagonally the first bomb will hit right on the edge of the marker, but follow up bombs will spread out over a longer area and beyond the marker because it moves faster:**

> ![diagonal 1](https://user-images.githubusercontent.com/32326277/185180075-2613b4f4-8181-43cb-85e6-6f01e15667da.png)
> ![diagonal 2](https://user-images.githubusercontent.com/32326277/185180086-9cab0caf-231d-4832-88c8-90ec3247db3c.png)
> ![diagonal 3](https://user-images.githubusercontent.com/32326277/185180097-2c2f9311-0341-4ac2-a7ea-9f06d54255a5.png)

commy2:
The bomber also makes a sharp turn on diagonal path. The turn also happens for China Bombers and Paradrops, but the B2 locomotor has the turn rate of a Raptor/Mig/Stealth Bomber instead of the one of a large plane.

MTKing4:
Issue title should mention it's AFG Carpet Bomber, to be more descriptive

ZekeDlyoung:
Don't all of AirF Gen's GP planes have a PDL tho?

MTKing4:
We should see if AFG A10/B52/Spectre/C-130 Planes have PDLs too

ZekeDlyoung:
I just realized that I posted this in the wrong thread 🙃

commy2:
> We should see if AFG A10/B52/Spectre/C-130 Planes have PDLs too

None of them have. Only the B2 and only when Carpet Bombing.

xezon:
[...]

PR #953, opened 2022-08-21
Change: Remove PointDefenseLaser from USA Airforce Carpet Bomber

### LINKED ISSUE #926: AFG B2 Bomber Has PDL When Calling Carpet Bombing But Not When Calling MOAB
This is a strange difference between the two.

I don't think either should have a PDL.

xezon:
* Fixes #926

This change removes the PointDefenseLaser from the USA Airforce Carpet Bomber.

# Rationale

All other USA Airforce General Promotion planes have no PointDefenseLaser, including the B52 bomber that delivers the MOAB and looks identical to the B52 Carpet Bomber.

USA Airforce General does not need this PointDefenseLaser on it to be useful either. ~~The USA Airforce Carpet Bomber has the lowest cooldown timer (2:30) of all faction carpet bombers.~~ It flies much faster than China Carpet Bombers. And it is available from Rank 1 unlike China Carpet Bombers at Rank 3.

USA Airforce opponents will now be able to shoot down the Carpet Bomber more easily with Rockets.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -3534,7 +3534,7 @@ CommandButton AirF_Command_PurchaseScienceCarpetBomb
   Command           = PURCHASE_SCIENCE
   Science           = SCIENCE_AirF_CarpetBomb
   ButtonImage       = SAB3Carpet
-  Object            = AirF_AmericaJetB3
+  Object            = AmericaJetB3 ; Patch104p @tweak from AirF_AmericaJetB3 to use the plane without PointDefenseLaser
   ButtonBorderType  = UPGRADE ; Identifier for the User as to what kind of button this is
 End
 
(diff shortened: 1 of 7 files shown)


commy2:
> The USA Airforce Carpet Bomber has the lowest cooldown timer (2:30) of all faction carpet bombers and it is available from Rank 1.

The AFG Carpet Bomber has the highest recharge time at 4:00, with Nuke Carpet being 3:00 and vChina/Inf/Tank Carpet being 2:30.

The tooltip is just wrong in 1.04.

commy2:
Another way to fix this would've been to replace the `Transport` token in ObjectCreationList AirF_SUPERWEAPON_CarpetBomb:
"AirF_AmericaJetB3" with "AmericaJetB3". Then the PDL Bomber version could be kept for custom maps or something.

xezon:
Done. Tested. Worked.

Jundiyy:
All good. And I was about to say what commy said. Would be nice to keep the original for mappers.

RisingZH:
Should fix the issue of afg carpet dropping bombs before reaching the targeted location.

ExiLe-ZH:
> Should fix the issue of afg carpet dropping bombs before reaching the targeted location.

https://github.com/TheSuperHackers/GeneralsGamePatch/issues/925#issuecomment-1218887340

Jundiyy:
> Done. Tested. Worked.

I presume you mean commy's method? If so, I'd prefer this way.

commy2 (approved):

PR #956, opened 2022-08-22
Fix: GLA SCUD missile times out without inflicting damage

Stubbjax:
Fixed a rare case where SCUD missiles would detonate too early under certain conditions and not cause any damage.

### Conditions:
- High terrain
- Salvage upgrade
- Search and Destroy battle plan

### Footage of the problem:

https://user-images.githubusercontent.com/11547761/185892021-90d51436-1c70-42ff-b107-a28fad7d6872.mp4


Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
```diff
@@ -3110,7 +3110,7 @@ Object SCUDMissile
   End
   Behavior = MissileAIUpdate ModuleTag_07
     TryToFollowTarget = No
-    FuelLifetime = 5000
+    FuelLifetime = 6000 ; Patch104p @bugfix Stubbjax 22/08/2022 Fixed premature SCUD missile detonation
```
xezon: We can also set `DetonateOnNoFuel = Yes` just to make sure it always explodes. Might as well set FuelLifetime even higher to 9999 to indicate that it is meant to not timeout. And change comment to "Fix premature missile deletion", because it does not really detonate.

commy2 (approved):

PR #958, opened 2022-08-22
Adds code and assets for nuke cannon range

### LINKED ISSUE #957: Add ground decal to Nuke Cannon when deployed that indicates Min and Max Firing range
Add ground decal to Nuke Cannon when deployed that indicates Min and Max Firing range. Reason being, using Nuke Cannon to force fire target can be a frustrating if Nuke Cannon has to undeploy, move a metter, and redeploy, just because the manual Attack Target was a few meters outside the min or max firing range. Adding ground decal would help assist making a better decision.

alanblack166:
* Fixes #957


xezon:
Thanks hanfield. I will get to it in due time.

Issue #897, opened 2022-08-14
USA Patriot Battery is rarely used


ExiLe-ZH:
- Power output is very low vs vehicles (vs infantry it performs surprisingly well though, when actively repaired).
- Requires power unlike the Firebase (relatively power hungry aswell).
- As or more expensive than the superior competitors like the Laser Patriot.

### Proposal 1

Reduce build cost.

### Proposal 2

Remove line of sight requirement for assisting Patriots.

### Proposal 3

Make Patriot hit moving infantry properly.

### Proposal 4

Add secondary damage in small radius to allow damage moving infantry a bit.

### Proposal 5

Reduce power requirement.

### Proposal 6

Increase range/damage/rof parameters to compete better with EMP/LaserPat.

### Proposal 7

Reduce EXP reward for killing the Vanilla Patriot, since it's easier to take down than EMP/Laser Pats.

ExiLe-ZH:
My proposal would be to make it cheaper.

USA doesn't need better defensive capabilities, therefore I think the performance shouldn't be touched. Now $800 would reflect it's performance better while not changing the meta, the Firebase would still be the better choice.

ReLaX82:
Agree on that FBs are much better and also require no power.

To even it out across all other pats I recommend to make it cheaper.

ZekeDlyoung:
As I mentioned here: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/27#issuecomment-1212734344

The Patriot can become much stronger if it didn't need line of sight thus allowing them to properly support each other more often in more varied formations.

Example:
https://www.moddb.com/mods/generation-x/images/patriot-assisted-targeting-fx#imagebox
https://www.moddb.com/mods/generation-x/images/new-patriots-ingame#imagebox

Another thing I did for my mod is make Patriots able to hit moving infantry, this made it waste less time chasing after running Redguards and spend more time actually attacking the vehicles that could actually kill it.

The Patriot's design implies that it is meant to be used in groups, imo if we are to buff the Patriot it should be with this design in mind

xezon:
3 Proposals added to opening post. I hope I got it right, as this line of sight change is not entirely clear to me. Does it mean it can shoot through big buildings always? Or only when assisted?

commy2:
Can Fire Base shoot over structures? If Fire Base can't, I don't see why Patriot would.

ZekeDlyoung:
[...]

PR #965, opened 2022-08-23
Change: Standard Patriot Missile Systems now track infantry

Stubbjax:
Patriot Missile Systems now track infantry. This is an implementation of proposal 3 for https://github.com/TheSuperHackers/GeneralsGamePatch/issues/897.

### Footage of the change:

https://user-images.githubusercontent.com/11547761/186201457-d59e907c-6a0a-4079-8d5e-68ac3b4bfe32.mp4

### Rationale:

This is an interesting change because it helps distinguish the Patriot more from the Firebase, which similarly tracks vehicles but not infantry. It offers deeper strategic diversity in the form of additional / greater anti-infantry options, which other proposals such as higher damage output or price reductions don't effectively provide. It also gains some similarity / consistency with the Technical's RPG weapon, which tracks infantry and deals the same type of damage. Bonus benefit: It feels nice when the rocket directly hits the infantry.

While the change does take away some micro opportunities - e.g. where the attacker moves an infantry unit around to waste the Patriot's shots or the attacker specifically targets stationary infantry - it should be noted that these micro opportunities rarely manifest in competitive games anyway (and in casual games where they are more common, players are often not skilled enough to do so). These opportunities will remain with the EMP (anti-vehicle) variant and Firebases, where they are already fairly common.

This change also amplifies the dynamic with China's `Red Guard Training` power, as an unranked Red Guard will take 4 missiles to kill, whereas a veteran Red Guard will take 5, thus having a greater influence on Troop Crawler attacks. The difference is much clearer without moving targets to waste shots on.

With this change, the Patriot Missile System gains a few extra use cases, while still remaining a relatively niche option so that the current gameplay / meta is maintained.


MTKing4:
While it is nice to look at, I don't think it's a good change, patriot will be too good

Also makes it inconsistent with EMP if that one doesn't track, not to say that it should because then it would be ultra broken

That being said, it does align with stingers sites now, since those missile do track.

Stubbjax:
> While it is nice to look at, I don't think it's a good change, patriot will be too good
> 
> Also makes it inconsistent with EMP if that one doesn't track, not to say that it should because then it would be ultra broken

How will this change make it too good? While it does add an inconsistency, it can be argued that the EMP variant itself is already an inconsistency with the structure having a different appearance, purpose and effect. It thematically works because it is a vehicle-specialising variant that doesn't track typical heat signatures.

> 
> That being said, it does align with stingers sites now, since those missile do track.

Stinger Sites don't track infantry, but GLA already has tunnels as their anti-infantry base defence.

MTKing4:
> How will this change make it too good?

well not too good, but it'll have a big impact on balance in general, in a way that buffs USAs when not needed to, since now it can't be distracted to draw fire into a infantry unit

> Stinger Sites don't track infantry, but GLA already has tunnels as their anti-infantry base defence.

i must i've misremembered then, so that makes the inconsistency argument even stronger, other rocket defenses don't track, so why should this?

Stubbjax:
> 
> well not too good, but it'll have a big impact on balance in general, in a way that buffs USAs when not needed to, since now it can't be distracted to draw fire into a infantry unit

I don't agree it would affect balance that much, if at all. Players will simply be punished less when they place a Patriot down instead of a Firebase, which is really the primary _aim_ (see what I did there?) of this change.

> 
> i must i've misremembered then, so that makes the inconsistency argument even stronger, other rocket defenses don't track, so why should this?

Because not all factions are the same? This argument makes no sense.

xezon:
> While it is nice to look at, I don't think it's a good change, patriot will be too good.

This claim needs to be supported with data, otherwise it is meaningless. For example create a map with 5 Fire Bases and 5 Patriots and have 100 infantry run at each, and see how they perform side by side.

While I do like the improved functionality in principle, I do take issue with the introduced inconsistency with the EMP Patriot. The EMP Patriot is a specialized version of the Patriot with different ammunition, however apart from that it behaves comparatively with regular Patriot Battery. From Faction Strength perspective, SWG is one of the weakest faction in both Pro and Noob games, below the other USA's, so from that perspective it should be ok to give the same perk to EMP Patriot.

To fine tune the behaviour, it would also be possible to set `ScatterRadiusVsInfantry` somewhere between 0 and 10, so that the splash damage deals a bit less damage to moving infantry units and therefore slightly reduce effectiveness. This would have to be tested to find optimal value.

Note also that this change will make it more difficult to kill Patriot with GLA Terrorists.

I am a bit skeptical about this change. Some lab test results would be good to see how strong this actually compares to other base defenses, such as Tunnel, Stinger, Firebase, Bunker with 5 Rocket Men and to before and after.

Stubbjax:
> 
> While I do like the improved functionality in principle, I do take issue with the introduced inconsistency with the EMP Patriot. The EMP Patriot is a specialized version of the Patriot with different ammunition, however apart from that it behaves comparatively with regular Patriot Battery. From Faction Strength perspective, SWG is one of the weakest faction in both Pro and Noob games, below the other USA's, so from that perspective it should be ok to give the same perk to EMP Patriot.

I also felt the same, but thought it would be acceptable considering EMP Patriots are already incredibly common and don't need the buff, as well as it making thematic sense. But in the grand context of improving SWG, it might be viable to also give the EMP variant the same benefit.

> 
> To fine tune the behaviour, it would also be possible to set `ScatterRadiusVsInfantry` somewhere between 0 and 10, so that the splash damage deals a bit less damage to moving infantry units and therefore slightly reduce effectiveness. This would have to be tested to find optimal value.

I'm fairly sure the scatter radius being greater than zero is what stops the rockets from tracking, so that might not work.

> 
> Note also that this change will make it more difficult to kill Patriot with GLA Terrorists.

It would certainly serve as a counter to rushing terrorists on closer-proximity maps like Defcon 6. But it would be no harder to bomb than a Laser Turret.

> 
> I am a bit skeptical about this change. Some lab test results would be good to see how strong this actually compares to other base defenses, such as Tunnel, Stinger, Firebase, Bunker with 5 Rocket Men and to before and after.

I expect players would still rarely use them. I don't believe the reason players avoid using them is because they don't hit infantry very well - the Firebase simply has much greater benefits. It's still far from optimal vs infantry spam due to the time it takes to shoot each one individually.

Jundiyy:
It's a neat idea but an unneeded buff. USA doesn't need that. 

With this change, it would deny or slow down Tech Terrors, but would also be better vs TC attacks/rush. Might start seeing them more, currently Firebases are used, but you can counter them by moving units closer to them. With Patriots, that wouldn't be the case. 

And as you said, in pro games, or even semi, such as myself, a single unit is used to distract. 

Another thing, it also removes the micro of the Patriot, which is a fun feature vs moving infantry, force firing to hunt them down.

ExiLe-ZH:
Wouldn't be good, can't use Terrorists anymore vs EMP's then, early game GLA can't use anything else.

Would be weird if normal vPat can track inf, but not EMP.

vPat already beats vet1 TC when actively repaired by Dozer. That is when he manually target the shooting guards, not the ones the attacker moves away as distraction.



Anyways, I like the exploration to improve the vPats, 

However I think the main problem is the pure DPS vs your common vehicles, vs Techs and Gats it performs no better than the Firebase, while investing the same or more to meet power requirements.

commy2:
> Wouldn't be good, can't use Terrorists anymore vs EMP's then, early game GLA can't use anything else.

This change does not affect EMP Patriots.

MTKing4:
> I don't agree it would affect balance that much, if at all. Players will simply be punished less when they place a Patriot down instead of a Firebase, which is really the primary aim (see what I did there?) of this change.

It will be a buff for USAs regardless, USAs are already strong enough, so it is a step in the wrong direction. Unless you want to have even more unnecessary changes to mitigate the impact

> Because not all factions are the same? This argument makes no sense.

It is USAs counterpart of the stinger, their main AA defense building, and gatling cannon for chinas, there should be no doubt about that fact, i obviously don't mean exactly the same.

ZekeDlyoung:
I like this change, I suggested it afterall 😛,  and imo this buff is needed. Much like the Nuke Cannon the Patriot has a lot of things going against it that make it undesirable most of the time, You guys already rejected the proposal to make them not require line of sight which hinders its assist ability, imo this is an acceptable compromise.

But if it is really that big of a deal for you guys, one other compromise would be to bring back the scatter radius vs infantry, but give patriots a secondary damage radius that is also 10, so moving infantry will at least get some damage as opposed to full damage.

EDIT: That way it can stay consistent with the EMP Patriot, while still having improved functionality vs infantry

xezon:
This change will make Patriots more self sufficient and be able to defend against everything: Aircraft, Vehicles, Infantry. It takes away some strategic depth and helps USA defend more effortlessly, because base defenses no longer need assistance from units such as Humvee, Ranger, Pathfinder. Critical infantry such as GLA Terrorist can no longer engage from far away. It is not clear to me if this is desireably gameplay change. Perhaps it is better to find different way to make Patriot better and not mismatch with EMP Patriot behaviour.

commy2:
Not keen on changing the behaviour of TOW missiles as I never seen it as a problem or mentioned as a problem before.
The only thing I would change is make the model of the missile consistent between vs ground and vs air mode, but keep the behaviour.
Currently, the model used vs ground is that of the Rocket Buggy and the model used vs air is that of the Raptor.

Stubbjax:
>
> It's a neat idea but an unneeded buff. USA doesn't need that.

> 
> It will be a buff for USAs regardless, USAs are already strong enough, so it is a step in the wrong direction. Unless you want to have even more unnecessary changes to mitigate the impact

It's not completely accurate to call it a buff if the change is more of a punishment reduction and makes something that players never use into something that players rarely use. It's akin to reducing the price / research speed of the Booby Trap upgrade for GLA, or reducing Comanche / Stealth Drone prices for SWG. While they are technically buffs, they don't practically make the faction any stronger, and are still disadvantageous strategies to employ.

>
> With this change, it would deny or slow down Tech Terrors, but would also be better vs TC attacks/rush. Might start seeing them more, currently Firebases are used, but you can counter them by moving units closer to them. With Patriots, that wouldn't be the case.

It certainly would affect those situations - but only if they are constructed, which I believe would remain rare, as the Firebase still remains the better option. The idea is to see standard Patriots used slightly more - but not so much more that they become the primary choice of base defence. Instead of e.g. 99% Firebases vs 1% Patriots, it could become something like 95% Firebases vs 5% Patriots, which is still extremely skewed in terms of diversity but maintaining the current gameplay / meta takes priority. GLA already has appropriate responses to Laser Turrets, so it would not be much different in the few additional cases where standard Patriots are used.

>
> Wouldn't be good, can't use Terrorists anymore vs EMP's then, early game GLA can't use anything else.

Yeah, this was also my concern with giving it to the EMP variant. I believe it would be better to leave it alone. I think it still makes sense that EMP missiles miss infantry seeing as they are specifically intended for vehicles, but I agree that the introduced inconsistency is less than ideal.

> 
> vPat already beats vet1 TC when actively repaired by Dozer. That is when he manually target the shooting guards, not the ones the attacker moves away as distraction.

I tested this thoroughly before submitting the change. A veteran Troop Crawler will actually win against a standard Patriot (even with infantry tracking) that is under active repair, while an unranked TC will require at least another unit such as a Gattling Tank to assist in order to win. This assumes the first two volleys go towards destroying the Troop Crawler.

> 
> Anyways, I like the exploration to improve the vPats,
> 
> However I think the main problem is the pure DPS vs your common vehicles, vs Techs and Gats it performs no better than the Firebase, while investing the same or more to meet power requirements.

That is true. Its damage output is not much better, which is why diversifying its use case seems like a viable solution and much a more interesting one than a simple cost reduction or damage boost.

>
> This change will make Patriots more self sufficient and be able to defend against everything: Aircraft, Vehicles, Infantry. It takes away some strategic depth and helps USA defend more effortlessly, because base defenses no longer need assistance from units such as Humvee, Ranger, Pathfinder. Critical infantry such as GLA Terrorist can no longer engage from far away. It is not clear to me if this is desireably gameplay change. Perhaps it is better to find different way to make Patriot better and not mismatch with EMP Patriot behaviour.

It is true that there are a few implications, though I'd still point out this is not much different to the current situation with Laser Turrets, which actually do see common usage. While it does take away some strategic depth, it also adds more strategic depth at the same time. Any lost strategic depth is largely negligible because it rarely comes into play. It really comes down to how the addition and subtraction balance out.

Issue #862, opened 2022-08-10
Sentry Drone is rarely used


ExiLe-ZH:
Potential good uses for the unit:
- Extending map vision for several purposes.
- Supporting armies by extending Shroud Reveal and Stealth Detection ranges (to avoid DemoTraps, Tunnels etc.).
- Intercepting (stealth) Workers and Hero's.*
- Protecting oils from being captured.*
- Scouting and killing Hackers.*
- Harassing Worker/Truck collection.*

*MG needed.

Negatives:
- Drone is expensive and fairly useless without the Machine Gun upgrade.
- Shroud Clearing and Stealth Detection range aren't great.
- Machine Upgrade expensive and extremely slow building.
- DPS is fairly weak.
- Deployment time makes it slow reacting, some people may experience this as buggy as there is no visual/audio clues for this behavior.
- Tire-tracks are visible for enemies.
- Damaged smoke reveals it permanently, without ways to field repair other than using Emergency Repair or capturing a Repair Bay.

### Stats
**Price & Build Time:**
| |Price|Build Time|
|-|-----|----------|
|Drone|800/800/850/1000 (usa/las/sw/af)|13.3s|
|MG Upgrade|1000|30s|

**DPS, AttackRange & Armor:**
| |Sentry Drone|Vee|
|-|------------|---|
|DPS|40|50 (60 with Bombardment*)|
|AttackRange|150|150 (180 with Search&Destroy*)|
|Armor|300 (375 with DroneArmorUpgrade**)|240 (252 with HoldTheLine*)|

_*Battleplan from Strategy Center._
_**Drone Armor upgrade from Strategy Center._

**Vision & Stealth Detection Range:**
| |Sentry Drone|Outpost|
|-|------------|-------|
|Vision range|200|250|
|Shroud clearing range|350|500|
|Stealth detection range|225|300|

### Proposals

- **Proposal 1:**
Remove or decrease deployment time.

- **Proposal 2:**
Add visual and/or audio clues for deployment.

- **Proposal 3:**
Reevaluate pricing and/or build time.

- **Proposal 4:**
Reevaluate pricing and/or build time for the MachineGun upgrade.

- **Proposal 5:**
Give MachineGun by default, or reimplement it as unit upgrade (same like a HellFire Drone or Propaganda Tower).

- **Proposal 6:**
Add auto-repair.

- **Proposal 7:**
Enhance shroud clearing range.

- **Proposal 8:**
[...]

PR #970, opened 2022-08-25
Change: Improved Sentry Drones

Stubbjax:
Improved Sentry Drones so that they don't suck as much. This implementation comprises several proposals from https://github.com/TheSuperHackers/GeneralsGamePatch/issues/862. Further adjustments may be necessary, but this should form a solid, low-risk foundation.

### Summary of changes:

- Removed the invisible 1s pack and unpack delays
- Reduced prices from $1000/$850/$800 to $600
- Reduced build times from 10s to 8s
- Added a minor healing function of 1hp / second (Sentries have 300hp)
- Fixed the lack of stealth / unstealth sounds

### Footage of changes:

https://user-images.githubusercontent.com/11547761/186616126-7ad02caf-25bc-4d73-94ae-d2177e487faf.mp4

The footage should hopefully demonstrate the greatly improved responsiveness of the drones, as well as their minor healing ability.

### Rationale:

These changes should assist in making Sentry Drones more versatile and provide greater strategic diversity for USA factions to employ, without boosting them too much that they become a primary strategy and change the way USA is typically played.

#### Fix the broken weapon deployment behaviour
It goes without saying that Sentry Drones would be better off without the strange and invisible “weapon deployment” delay that occurs between firing and moving.

#### Reduce and standardise prices
Sentry Drones are considerably expensive units for what they do, and a Humvee with a scout drone is almost always a more cost-effective and versatile option. Reducing the price of Sentry Drones from $1000/$850/$800 to $600 would be an effective and low-risk way to boost utility, due to the fact that they serve a very specific purpose that is already readily available elsewhere, and are otherwise never used.

#### Reduce construction time
Another effective means of increasing utility is to decrease the build time from 10 to 8 seconds. Note that Sentry Drones still require the Sentry Drone Gun upgrade to become useful in combat, which prevents a reduced construction time (and several other improvements) from opening up early Sentry Drone-based strategies and potentially affecting traditional gameplay too much.

#### Add auto-repair capability
As Sentry Drones cannot rank up, a good addition would be to give them auto-repair capabilities. This would not only distinguish the unit from other vehicles a bit more, but would allow for a unique tactic where players can command the unit to stop and restealth in order to heal up at the cost of time and the opponent likely remembering the unit’s location.l


xezon:
I am a fan of the Pack & Unpack time removal. It should solve the original frustration problem it poses. However note that this potentially gives a big buff in unit strength, as it will increase damage output, since drones no longer need to wait this time to acquire targets.

The miniscule healing also looks okay on first sight. It would take 300 seconds (05:00) to heal fully. It is a long time.

I am a bit concerned about the sharp price and build time drop. It will make the unit more spammable, especially from 1000 to 600. Which General does it have priced at 1000? Also, could we get a comparison of the amount of damage the Drone can take compared to Humvee and Crusader? I think it would be worthwhile to check that it does not perform better than Humvee or Tank, otherwise the low price could be risk.

Edit: Also stats comparison between Drone Gun vs Humvee Gun vs Humvee Tow would be good to see how strong/weak it is.

ExiLe-ZH:
Nice.

- I think removing delay is a good thing, probably doesn't make it overpowered, though I'd like to see a test where it fights a diversity of (other weaker) units 1v1. Also tests to see how it performs vs infantry, how many it can fight before and after? In both offensive scenarios where the drone attacks first and defensive scenarios where it gets revealed and attacked.

- 600 does sound like a fair price to me, same as the ambulance, which is much more useful overall.
Though if we want to play it safe we could make it 700 instead, which is same as a Vee or an Ambo with Scout Drone.

- 8 seconds BT sounds fair to me aswell, hopefully doesn't make spam with 600 dollar price too crazy. Jundiyy and I made that mistake before, where Logica started to dominate with Sentry Drones in our test tournaments. Think we handled some crazier values though, can't exactly remember.

- The auto repair thing sounds totally fair, it doesn't have Battle Drones available unlike any other unit and the repair time is very safe.

ExiLe-ZH:
> I am a bit concerned about the sharp price and build time drop. It will make the unit more spammable, especially from 1000 to 600. Which General does it have priced at 1000? Also, could we get a comparison of the amount of damage the Drone can take compared to Humvee and Crusader? I think it would be worthwhile to check that it does not perform better than Humvee or Tank, otherwise the low price could be risk.

Originally USA/Laser 800, AFG 850, SWG 1000. 

What we do know is that 800/10s definitely isn't spammable and that 850/1000 is simply too much.

Jundiyy:
I do like these changes, perhaps the only one which might feel odd to some, is the auto repair. Or perhaps players won't notice is it's rarely used. They'd be like 'oh, I didn't know it auto repaired,  should have used it more'

$800/10s is surely a safe one, we've had enough testing with cheaper/less build time, in private 1v1s and through tournaments.
I'd like $600, although might be too good now with the no deploy, otherwise I'd have gone with that. Perhaps $650/$700 is the right price.

It has a really good armour vs rockets, Gats shred through them, Vees and Quads are also very good. But rockets are bad.

MTKing4:
I like these changes, most of them make sense, I'm not sure about the auto repair though

commy2:
> Reduced build times from 10s to 8s

You changed it from 13.3 seconds to 11.3 seconds. :)

Could we try a pack/unpack time of 250 msec (4 frames) first before removing the behaviour entirely?
I don't get the auto repair either.

xezon:
Where are these extra 1.3 build seconds from?

commy2:
USA WF is 3250 msec door opening time in 1.04, and the latest change I observed was you setting it to 3300 (1 extra frame) to fix animation issues.

Stubbjax:
> I am a bit concerned about the sharp price and build time drop. It will make the unit more spammable, especially from 1000 to 600. 

Humvees are still far more versatile and would absolutely remain the unit of choice. The Humvee + MD combo basically counters everything in the game, whereas Sentry Drones will easily succumb to any kind of tank, base defence or aircraft.

> It has a really good armour vs rockets, Gats shred through them, Vees and Quads are also very good. But rockets are bad.

So the armour values between Sentry Drones and Humvees are almost identical except for two (irrelevant) types:

| | SentryDroneArmor | HumveeArmor
-- | -- | --
POISON | 0% | 50%
KILL_PILOT | 0% | 100%

The real difference is that Humvees have 240 health, whereas Sentry Drones have 300 health (not 200 - I should have double-checked this). There could be some potential armour options here too, such as increasing INFANTRY_MISSILE from 50% to 60% or reducing FLAME damage from 50% to 25% (which I can assure you, MTKing, not a single pro would notice).

> Also stats comparison between Drone Gun vs Humvee Gun vs Humvee Tow would be good to see how strong/weak it is.

Concerning weapons, the only difference between the two primary weapons is that the Humvee's machine gun deals 10 damage per 200ms (50 DPS) whereas the Sentry Drone deals 8 damage per 200ms (40 DPS). Otherwise, range, damage type, etc. are all identical. (It is rather unfortunate that the sound effect for the Sentry Drone machine gun makes it seem much more powerful than it actually is.)

It should be noted that Sentry Drones require a $1000 + 30s upgrade to deal any damage in the first place, nor can they run over infantry, which severely limits their early-game utility. They also cannot rank up or utilise support drones as distractions or extra firepower, which restricts their power and survivability.

Humvees, on the other hand, have the TOW missile upgrade for $800 + 30s, which deals 30 damage every 2s and allows them to target aircraft. It is hard to accurately gauge DPS with both weapons due to the different damage types having vastly different effects vs different armour types.

> Or perhaps players won't notice is it's rarely used. They'd be like 'oh, I didn't know it auto repaired, should have used it more'

Yeah, this was my exact thinking as well. I believe it's a nice way of distinguishing it more from the Humvee, and can combo well with its stealth capability.

> I like these changes, most of them make sense, I'm not sure about the auto repair though

You reckon it's too visible of a change?

> Could we try a pack/unpack time of 250 msec (4 frames) first before removing the behaviour entirely? I don't get the auto repair either.

You are welcome to try, but I'd ask why? Do you believe the unit should feel less responsive? I explained the reasoning for auto-repair in the first post.

ExiLe-ZH:
Another good buff would be to match the vision range with the MD, since it's supposed to fill this role. This would make them more useful without the Machine Gun upgrade.

MD's are currently most used as scouts because they're cheap , have good vision, provide defensive cover and can be put in bunkers and garrisons. Obviously MD's aren't stealth, can't detect stealth and are vulnerable in the open, but on the majority of maps you enough garrisons/bunkers on keypoints to not ever having to consider Sentry Drones for scouting purposes. For completely open maps you can just use a ScoutVee, which comes with a better MachineGun for free.

> It should be noted that Sentry Drones require a $1000 + 30s upgrade to deal any damage in the first place, nor can they run over infantry, which severely limits their early-game utility. They also cannot rank up or utilise support drones as distractions or extra firepower, which restricts their power and survivability.
> 
> Humvees, on the other hand, have the TOW missile upgrade for $800 + 30s, which deals 30 damage every 2s and allows them to target aircraft. It is hard to accurately gauge DPS with both weapons due to the different damage types having vastly different effects vs different armour types.

Yes, the pricing doesn't make sense at all. I think swapping prices is the first thing we should do, I really can't think of a counter argument here. The Scorp Rocket is also 1000 and the Scorp totally relies on this, while Vees don't always do. For 1000 it's still a bonus no brainer.

commy2:
> Do you believe the unit should feel less responsive?

Yes. It is a remote controlled drone.

commy2:
Another difference between Vee and SD is, that SD does not benefit from the Strat Center Battle Plan, because it is a drone.

ExiLe-ZH:
> > Do you believe the unit should feel less responsive?
> 
> Yes. It is a remote controlled drone.

It doesn't have a noticeable move delay though if you haven't used the machinegun before? Moving should always react as slow as shooting if this is the logic.

Stubbjax:
> Another good buff would be to match the vision range with the MD, since it's supposed to fill this role. This would make them more useful without the Machine Gun upgrade.

That's a very good point. That would be an increase from 350 to 400.

> Yes, the pricing doesn't make sense at all. I think swapping prices is the first thing we should do, I really can't think of a counter argument here. The Scorp Rocket is also 1000 and the Scorp totally relies on this, while Vees don't always do. For 1000 it's still a bonus no brainer.

While I do agree, this would be a highly visible and likely controversial change. Increasing the research duration could be a safer option?

> Another difference between Vee and SD is, that SD does not benefit from the Strat Center Battle Plan, because it is a drone.

This is also a very important distinction. Though the Drone Armour (+75hp) upgrade does mitigate this flaw somewhat.

ExiLe-ZH:
> While I do agree, this would be a highly visible and likely controversial change. Increasing the research duration could be a safer option?

I believe Tow was either 1000 or 1200 in CCG, but it's another discussion anyway.

Sentry Machine Gun for 800 surely can't be disputed.

Stubbjax:
Oh yeah, the price reduction makes total sense.

ExiLe-ZH:
Just thought of an alternative balance solution.

What if make the MachineGun an upgrade per unit? Same like Battle/Hellfire/Scout Drones?

1. It wouldn't waste valuable production time in the WF. Which is the biggest problem besides the price. Reduction to $800 will not help much if buildtime stays at 30s. Even if we reduce buildtime to 10s you might aswell just make 2 Vees instead of 1 upgraded SD.

2. No need to adjust/cut production time for the unit, 10s (or 13,3) isn't too bad if the upgrade no longer has to be purchased from the WF.

3. We can lower the base price to maybe like $500 and charge $300 for the MachineGun, same like a Battle Drone. 
$500 without MachineGun is definitely not too cheap.
$800 with MachineGun and 13.3s production time doesn't exactly make it a good spam unit.

We could give the MachineGun by default but that's controversial aswell isn't?

commy2:
> Moving should always react as slow as shooting if this is the logic.

Then the logic is that the drone CPU can only handle so much tasks at once and has to reboot when toggling between attack and movement mode.

The unit having delays for taking orders is part of the design of the unit, because it has no pilot that steers it directly. I would not fundamentally change the design except as last resort. I do agree that the delay is currently too much, but even halfing it would make a big difference already.

Similar thing about the prices. It is obviously meant to be a "high-tech" unit. That's the flavour. Making it cheap and mass producable goes against that.

\-\-\-

I think first you have to identify the issues with the Sentry Drone, and then you have to address them. I simply disagree that the price and build time of the Drone are issues, except for SWG, but that is perhaps true for all her units.

As far as I can see, the issue is that the thing is 1) unresponsive and either 2) unarmed or 3) dealing underwhelming damage. And it is definitely too expensive for SWG.

The solutions follow from that. Make the gun deal more damage (why not more than Humvee gun, since no TOW?), the upgrade more accessible (price, research time?) and the unit more responsive (without making it basically not a drone with response lag at all).

And give SWG the price of $800 from vUSA/Laser. SWG kind of has that high tech flavour that would exempt the Sentry Drone from the higher prices for her (see: $1200 Burton).

xezon:
If we keep some delay, it would be good to make this delay somehow readable, for example with a charging sound. In all the years I though the delay is a clumsy bug and not a design choice, mainly because there is nothing that explains it other than imagination.

ExiLe-ZH:
> > Moving should always react as slow as shooting if this is the logic.
> 
> Then the logic is that the drone CPU can only handle so much tasks at once and has to reboot when toggling between attack and movement mode.
> 

I guess so lol.

> The unit having delays for taking orders is part of the design of the unit, because it has no pilot that steers it directly. I would not fundamentally change the design except as last resort. I do agree that the delay is currently too much, but even halfing it would make a big difference already.
> 

Don't get me wrong, I like the whole idea that there's some slight delay in the overall control and it makes sense. It also adds some charm/uniqueness. Just reducing delays is fine aswell imo, if completely removing it is too controversial.

> Similar thing about the prices. It is obviously meant to be a "high-tech" unit. That's the flavour. Making it cheap and mass producable goes against that.
> 

True, but in the current state nobody would want to spam it. The overall investment and buildtime already reflects alot of "high-tech", too much in fact.

> 
> I think first you have to identify the issues with the Sentry Drone, and then you have to address them. I simply disagree that the price and build time of the Drone are issues, except for SWG, but that is perhaps true for all her units.
> 

Cost and BT are a very real issue. USA/Laser can make Vees for less, which are much more versatile and stronger. The MachineGun upgrade is 90% responsible for the whole problem though, $1000 dollars and ESPECIALLY the 30s seconds BuildTime just kills it. If it was included by default, $800 for the SD wouldn't be all that bad actually.

> As far as I can see, the issue is that the thing is 1) unresponsive and either 2) unarmed or 3) dealing underwhelming damage. And it is definitely too expensive for SWG.
> 
> The solutions follow from that. Make the gun deal more damage (why not more than Humvee gun, since no TOW?), the upgrade more accessible (price, research time?) and the unit more responsive (without making it basically not a drone with response lag at all).
> 

Well vs Infantry and Hero's it already performs pretty well right? Which is how it was intended. Don't think there's need to make it better vs Vehicles.

commy2:
> Well vs Infantry and Hero's it already performs pretty well right?

It doesn't perform well against infantry, because the damage is pitiful.

> True, but in the current state nobody would want to spam it.

It should stay that way, which is why slashing price and build time is out of the question imo,

MTKing4:
> ncreasing INFANTRY_MISSILE from 50% to 60% or reducing FLAME damage from 50% to 25% (which I can assure you, MTKing, not a single pro would notice

Maybe, but that's because how it's literally not used at all by pros so they wouldn't know the difference, but bomb trucks on the other hand are used a bit.

> You reckon it's too visible of a change?

No i think it will make it better but more than needed, since it will not require to go to wf for repair and just leave it somewhere emtpy and repair in stealth mode, could prove very useful and frustrating to deal with like killing workers then run away and hide, or trucks, for such a cheap price of 600.

also it doesn't make sense why is it repairing, like why? What's repairing it? You see other units need speaker tower or being an ambulance or junk repair,  but why drones are self repairing?

Also think it won't make it attractive to players still due to the 30s $1000 upgrade (unless of course auto repair proven too good) i say buffing it that way instead is better, make upgrade 15s $500 and remove the self heal

ExiLe-ZH:
> > Well vs Infantry and Hero's it already performs pretty well right?
> 
> It doesn't perform well against infantry, because the damage is pitiful.

Ok we can look into MG damage vs inf then, it's fair that the MG performs just aswell as a Vee vs Hero's for example.

> 
> > True, but in the current state nobody would want to spam it.
> 
> It should stay that way, which is why slashing price and build time is out of the question imo,

Nobody wants it to become a spam unit, all we're trying here is to make them more accessible. 

It's all about the stupid MG upgrade and I'm really starting to hate it. Super cheap/fast is also silly and it will get skipped regardless if you can only get 1 SD for the production time of 2 Vees. 

Maybe it isn't even all that controversial if it comes with the MG by default, IF we make the Drone itself more expensive. Perhaps 850-1000.. yay high tech :D + some further tiny improvements mentioned in this topic).

xezon:
I would advise against changing core Drone gun setup, because it will change how it works and will be a very visible altercation to the original. By tweaking the setup it should be possible to achieve a satisfying result.

Stubbjax:
> Then the logic is that the drone CPU can only handle so much tasks at once and has to reboot when toggling between attack and movement mode.
> 
> The unit having delays for taking orders is part of the design of the unit, because it has no pilot that steers it directly. I would not fundamentally change the design except as last resort. I do agree that the delay is currently too much, but even halfing it would make a big difference already.

Unresponsiveness is terrible design. Regardless of whether it is intended or not, it is not conveyed to the player and should be removed unless it can be properly implemented. If proper pack and unpack animations can be added, then that would be a welcome and even preferred change. But until any such development, disabling it is the best option.

> 
> Similar thing about the prices. It is obviously meant to be a "high-tech" unit. That's the flavour. Making it cheap and mass producable goes against that.

I don't see how this unit is much more "high-tech" than most other units in USA's arsenal, and I don't see how that weighs into the game's pricing paradigm (or a Battle Bus would not be as expensive as a King Raptor). The design indicates that its primary function is surveillance, with an expensive and untimely upgrade also supporting its combat utility being secondary / supplemental.

> 
> I think first you have to identify the issues with the Sentry Drone, and then you have to address them. I simply disagree that the price and build time of the Drone are issues, except for SWG, but that is perhaps true for all her units.

Considering the unit's primary purpose is surveillance, and the fact that USA already has ample surveillance in the form of Spy Drones, Scout Drones, Satellite Scan, etc., it makes sense to make the unit more attractive without compromising the underlying design - as you've highlighted should be important. It is a specialised unit and nowhere near as versatile as other units like the Humvee. The build time reduction also realistically leans into the fact that drones technically require less 'training' due to the lack of pilot and respective casualties.

> 
> As far as I can see, the issue is that the thing is 1) unresponsive and either 2) unarmed or 3) dealing underwhelming damage. And it is definitely too expensive for SWG.

100% agreed. But the goal is not to make them so much more viable that they alter traditional gameplay and strategies, which I believe the proposed changes would maintain.

> The solutions follow from that. Make the gun deal more damage (why not more than Humvee gun, since no TOW?), the upgrade more accessible (price, research time?) and the unit more responsive (without making it basically not a drone with response lag at all).

I'd be happy to explore higher damage output, especially as the associated machine gun sound does not line up here. But I'd also highlight that this would be diverging from the unit's original design / purpose as surveillance. I'd probably go as far as reducing the upgrade from $1000 / 30s to $800 / 20s. If damage was to go from 8 / 200ms (40 DPS) to 12 / 200ms (60 DPS), I'd probably leave the upgrade unchanged and maybe increase the Sentry Drone price back up to $800, and ideally get the unpack / pack animations working.

> It doesn't perform well against infantry, because the damage is pitiful.

It performs roughly the same as the Humvee. A Humvee will kill infantry faster (20% more DPS) and thus take fewer shots, but Sentry Drones have more health (25% more HP), so it roughly balances out. This is, of course, assuming no TOW missile.

> it will not require to go to wf for repair and just leave it somewhere emtpy and repair in stealth mode, could prove very useful and frustrating to deal with like killing workers then run away and hide, or trucks, for such a cheap price of 600.

Would this not be a good thing? This change explicitly targets its niche role as a surveillance unit, as it would no longer need to return to heal. Plus it makes up for its lack of elite-rank repair and Battle Drone repair. Don't forget that it would take a full 5 minutes to repair from 0 to full health.

> also it doesn't make sense why is it repairing, like why? What's repairing it? You see other units need speaker tower or being an ambulance or junk repair, but why drones are self repairing?

Because it's future tech and thematically works. How does Propaganda magically repair everything?

> Also think it won't make it attractive to players still due to the 30s $1000 upgrade (unless of course auto repair proven too good) i say buffing it that way instead is better, make upgrade 15s $500 and remove the self heal

It's not intended to be made attractive! The auto repair really doesn't make that much difference in the grand scheme, it's just a nice distinction from other units, fits in well with the unit's core purpose, and thematically works.

xezon:
There is one major drawback of this slow repair: It will make this animated repair image hover above the drone for very long. This would be reason for me to not implement this as is, because this is an unusual long and obvious repair that can leave players scratching their head as to what takes them so long to repair all the time.

Stubbjax:
> There is one major drawback of this slow repair: It will make this animated repair image hover above the drone for very long. This would be reason for me to not implement this as is, because this is an unusual long and obvious repair that can leave players scratching their head as to what takes them so long to repair all the time.

I did consider this too, and would prefer a faster repair, but that could have greater implications during combat. I could not find a way to only have it repair out of combat or while stealthed. I can remove it if everyone feels it's too controversial.

xezon:
> Unresponsiveness is terrible design. Regardless of whether it is intended or not, it is not conveyed to the player and should be removed unless it can be properly implemented. If proper pack and unpack animations can be added, then that would be a welcome and even preferred change. But until any such development, disabling it is the best option.

I agree with this. Players hate lag for obvious reasons. And this design as is is perceived as lag. I don't know how this could be animated, but as mentioned earlier a charge and discharge sound could help to communicate a few frames of additional engagement delay. Is it possible to trigger sound on engage and disengage once?

commy2:
> Is it possible to trigger sound on engage and disengage once?

There are UnitSpecificSounds/Deploy and UnitSpecificSounds/Undeploy, which are used by the Nuke Cannon, but not the Sentry Drone atm.

xezon:
Very nice. So we could add something to Drone. Would just need to find or make matching sounds.

```
  UnitSpecificSounds
    VoiceCreate         = NukeCannonVoiceCreate
    TurretMoveStart     = NoSound
    TurretMoveLoop      = NukeCannonTurretMoveLoop
    VoiceEnter          = NukeCannonVoiceMove
    Deploy              = NukeCannonDeploy
    Undeploy            = NukeCannonUnDeploy
  End
```

commy2:
dial up modem handshake

ExiLe-ZH:
A little robot sound would be cool, hope that's enough feedback for the player so behaviour is not perceived as gamelag or bug.

xezon:
Ok how we make this sound? xD

ExiLe-ZH:
Lol :D can reuse some sounds?

https://youtu.be/r8OeN72jV0c?t=172

> I would advise against changing core Drone gun setup, because it will change how it works and will be a very visible altercation to the original. By tweaking the setup it should be possible to achieve a satisfying result.

Even if you remove the whole unit from the PvP environment I think noone bats an eye. 

I honestly think there's no way of improving the unit in PVP, without changing the setup. Facts remain: 
- Unit is too expensive without MG.
- MG upgrade even at 10s BT and some cheap price delays normal production too much, can get 2 Vees instead of 1 SD.

Btw, note in the video above how the voiceline says MG should actually be available from the Strat Center, I honestly even believe this will be an improvement. Actually another potentially effective utility could be scouting and killing Hackers, this is only in post-tech scenario anyway.

ExiLe-ZH:
About vision range and stealth detection again, wouldn't it be fair if it's the same as an Outpost? Outpost is much superior at a cheaper price.

xezon:
The Sentry Drone has VisionRange 200. New Outpost will have 150.

ExiLe-ZH:
> The Sentry Drone has VisionRange 200. New Outpost will have 150.

That isn't right? MDs and THs have 400? Outposts have more.

What about stealth detection range though?

Stubbjax:
If the deployment behaviour is retained, there definitely needs to be some form of visual distinction. It cannot be solved with sounds alone.

> I honestly think there's no way of improving the unit in PVP, without changing the setup.

Even with the respective changes? Further enhancements could make it _too_ attractive and alter traditional gameplay.

> The Sentry Drone has VisionRange 200. New Outpost will have 150.

It's important to distinguish vision range from shroud clearing range. The shroud clearing range is the important factor here.

| | Sentry Drone | Outpost
-- | -- | --
Vision range | 200 | 250
Shroud clearing range | 350 | 500
Stealth detection range | 225 | 300

ExiLe-ZH:
> Even with the respective changes? Further enhancements could make it _too_ attractive and alter traditional gameplay.

We have no decent/accepted proposals regarding MG upgrade yet, it's the root of all problems. 

So I don't think we're anywhere close to make the unit accessible in any kind of PvP match (noob, semi or pro). 

*Well maybe in noob only matches, as macro efficiency plays no role here, but I don't think noobs can manage to win games with SD's either. It's usually the pros who are creative (some are robotic af aswell) and have the micro skills to put alternative units to good use. And obviously better players can troll and win vs lesser skilled players in any sort of way.

xezon:
If the Upgrade researchs fairly quick and cheap, when it should be usable, no?

Stubbjax:
> We have no decent/accepted proposals regarding MG upgrade yet, it's the root of all problems. So I don't think we're anywhere close to make the unit accessible in any kind of PvP match (noob, semi or pro).

How about reducing the upgrade from $1000 / 30s to $800 / 20s?

Jundiyy:
I don't think we need to overdo the Sentry Drone. Just fix the bugs and make it a bit more attractive.
It doesn't need to become a standard unit, if it's good enough to be used in some situations or a few more, then that's fine. Not every unit can be used as an all round unit.

USA already has massive stealth detection and vision, does the SD really need to buff that up?

- Price of the upgrade and time should be decreased, otherwise it would probably not be researched. 
- Deploy time reduced to minimum or removed.
- And a slight decrease in price/BT, this makes the unit a bit more attractive. 

Don't think other factors needs to be touched.

xezon:
I suggest to estimate the value of Sentry Gun Upgrade by comparing it with the value provided by the Humvee Tow Upgrade and then price it accordingly.

ExiLe-ZH:
@Stubbjax 
> How about reducing the upgrade from $1000 / 30s to $800 / 20s?

I've already tried to clear this up a few times, but let me try to demonstrate the problem a bit better. Here a straight comparison of cost and BT:
- 1x SD + MG = $1400 @ 31.3s ($600 + $800 @ 11.3s + 20s, with current proposals)
- 2x Vee = $1400 @ 26.6s

It's an improvement over 1.04 (but still):
- 1x SD + MG = $1800 @ 43.3s
- 3x Vee = $2100 @ 39.9s

@xezon 
> I suggest to estimate the value of Sentry Gun Upgrade by comparing it with the value provided by the Humvee Tow Upgrade and then price it accordingly.

I feel the comparison is a little weird, Vee has a similar but better performing MG for free.

@Jundiyy 
> I don't think we need to overdo the Sentry Drone. Just fix the bugs and make it a bit more attractive. It doesn't need to become a standard unit, if it's good enough to be used in some situations or a few more, then that's fine. Not every unit can be used as an all round unit.
> 
> USA already has massive stealth detection and vision, does the SD really need to buff that up?
> 
>     * Price of the upgrade and time should be decreased, otherwise it would probably not be researched.
> 
>     * Deploy time reduced to minimum or removed.
> 
>     * And a slight decrease in price/BT, this makes the unit a bit more attractive.
> 
> 
> Don't think other factors needs to be touched.

I don't disagree with your main point.

It shouldn't become meta, or spammable. It just needs better accessibility without the huge compromises, so it can be used for specific applications.

xezon:
Both TOW Missile and Sentry Gun are exclusive weapons added to Vehicle. The TOW Missile weapon is likely factually superior to Sentry Gun, e.g. by being able to target Aircraft. I do believe a comparison of both weapons is possible and necessary to determine how good one is compared to the other. This comparison can also help determining if the Sentry Gun itself could be tweaked for its damage capabilities.

Stubbjax:
> I've already tried to clear this up a few times, but let me try to demonstrate the problem a bit better. Here a straight comparison of cost and BT:
> 
>     * 1x SD + MG = $1400 @ 31.3s ($600 + $800 @ 11.3s + 20s).
> 
>     * 2x Vee = $1400 @ 26.6s

Of course, at the lower end, Humvees have better value, however with the reduced prices and costs, this eventually evens out and then surpasses Humvees. Considering the Sentry Drone Gun being purchased first, it would take roughly 9 units to equalize value. Reducing the machine gun upgrade research time to 10 seconds will roughly halve the number of units to match the time investment.

Quantity | Sentry Time | Humvee Time | Sentry Cost | Humvee Cost
| - | - | - | - | -
1 | 31.3 | 13.3 | 1400 | 700
2 | 42.6 | 26.6 | 2000 | 1400
3 | 53.9 | 39.9 | 2600 | 2100
4 | 65.2 | 53.2 | 3200 | 2800
5 | 76.5 | 66.5 | 3800 | 3500
6 | 87.8 | 79.8 | 4400 | 4200
7 | 99.1 | 93.1 | 5000 | 4900
8 | 110.4 | 106.4	| 5600 | 5600
9 | 121.7 | 119.7	| 6200 | 6300

image

> Both TOW Missile and Sentry Gun are exclusive weapons added to Vehicle. The TOW Missile weapon is likely factually superior to Sentry Gun, e.g. by being able to target Aircraft. I do believe a comparison of both weapons is possible and necessary to determine how good one is compared to the other. This comparison can also help determining if the Sentry Gun itself could be tweaked for its damage capabilities.

TOW missiles deal EXPLOSION damage, which is not resisted by a lot of units. Both machine guns are of COMANCHE_VULCAN type, which are resisted by a lot of units (though typically 50%). The DPS vs a tank is thus vastly different when combining both weapons vs a light vehicle or infantry.

Jundiyy:
I don't think SDs will be spammed enough. 
Rockvees will be needed. Rockvees are good vs everything whereas SDs are not. Maybe early game players might use them but as they move around the map, destroying buildings or heavy units, they will have to move over to Rockvees.

ExiLe-ZH:
@Stubbjax 
Factually you are correct, however:
1. In real matches there is (will be) no reason to waste more time/money on SD production to get more value out of it.
2. It's not our intent either to make mass production more valuable.

All we want is making the unit more accessible in lower quantities right? Higher quantity value would mean meta (or atleast an alternative meta strat).

I'm sure we're all aligned in what we want to achieve here right?

We just don't agree with the methods yet, that's ok, we're having a healthy discussion here.

Stubbjax:
> I don't think SDs will be spammed enough. Rockvees will be needed. Rockvees are good vs everything whereas SDs are not. Maybe early game players might use them but as they move around the map, destroying buildings or heavy units, they will have to move over to Rockvees.

I don't believe so either. Are you suggesting they are still too expensive / slow to produce? The general consensus seems to be that the proposed changes won't make Sentry Drones good enough to have any meaningful impact on traditional gameplay, while still reducing the disadvantage that typically comes with their use.

> All we want is making the unit more accessible in lower quantities right? Higher quantity value would mean meta (or atleast an alternative meta strat).

That is certainly true, though any cost or build time reduction is going to apply to both low and high quantities. It's important to consider higher quantities, because that's what makes the upgrade viable. It doesn't make as much sense to have an upgrade if only one or two Sentries are ever built.

> I'm sure we're all aligned in what we want to achieve here right?
> 
> We just don't agree with the methods yet, that's ok, we're having a healthy discussion here.

Absolutely; I'm really not fussed however the problem is approached, I just want the end result to be an interesting, relatively subtle and high quality approach. Would you agree or disagree that the current proposal fits your exact criteria of making the unit more accessible without altering traditional gameplay?

ExiLe-ZH:
> Are you suggesting they are still too expensive / slow to produce? 

> Would you agree or disagree that the current proposal fits your exact criteria of making the unit more accessible without altering traditional gameplay?

Yea I don't agree with the proposals yet, I think the graphs show the exact problem, that low quantity still is too inaccessible/slow/invaluable vs the alternative. Ideally the model should show an inverted line.

My earlier proposal does do this (per unit upgrade), but I do understand it's controversial.

Jundiyy:
> I don't believe so either. Are you suggesting they are still too expensive / slow to produce? The general consensus seems to be that the proposed changes won't make Sentry Drones good enough to have any meaningful impact on traditional gameplay, while still reducing the disadvantage that typically comes with their use.

Price at $700/8s (unit BT) or around that much is what I'm hinting towards.
I'm not asking for them to be spammed, I think with the changes I mentioned in my earlier post, they would be right, not left out and not spammed.
Just saying they don't need more than that.

ExiLe-ZH:
> Both TOW Missile and Sentry Gun are exclusive weapons added to Vehicle. The TOW Missile weapon is likely factually superior to Sentry Gun, e.g. by being able to target Aircraft. I do believe a comparison of both weapons is possible and necessary to determine how good one is compared to the other. This comparison can also help determining if the Sentry Gun itself could be tweaked for its damage capabilities.

One major reason why I think the comparison is not really valid is the fact that the SD totally relies on the upgrade, $800 for just vision is not worth, unit can not even crush inf either.

Whereas TOW is optional depending on the matchup, vs GLA players like Excal even skip it, because it renders the WF useless for a while. 

When China goes air (Gatlix/Migs) it's a very crucial upgrade, though in some situations it's also good to prioritize Avengers. 

In USA mirrors it's a super valuable upgrade, it helps alot with harassing Nook collection. Vs Commie Spam in teamgames it's also crucial, Avengers don't offer enough mobility to provide enough cover on the bigger teamgame maps.

Overall TOW gives a good extra punch in firepower, but it's still a bonus in most cases.

MTKing4:
> Unresponsiveness is terrible design

100% Agree

> Would this not be a good thing? This change explicitly targets its niche role as a surveillance unit, as it would no longer need to return to heal. Plus it makes up for its lack of elite-rank repair and Battle Drone repair. Don't forget that it would take a full 5 minutes to repair from 0 to full health.

It is a good thing, maybe too good, but it's definitely not a role of a surveillance unit, more like a sneaky harrasment unit, if this change goes through, you'll definitely going to be seeing me confusing my enemies to death with them :D

The rest of the point do make sense though tbf, it does take ages to repair.

> Because it's future tech and thematically works. How does Propaganda magically repair everything?

Well at least it's been defined in the game that propaganda heals, where this one isn't, and surely unfitting in this universe, we don't have nanotech/nanobots in this game/timeline nor in USA itself

> It's not intended to be made attractive! The auto repair really doesn't make that much difference in the grand scheme, it's just a nice distinction from other units, fits in well with the unit's core purpose, and thematically works.

No I'm saying people will still not use it, that upgrade just sucks as it is, too long and too expensive when you can just get a humvee to do the job, sentry drone has to have a specific purpose or role to justify such an expensive and long upgrade. And currently it isn't, nor is planned to, so reducing the upgrade BT and cost is the best solution imo.

Issue #337, opened 2021-09-13
China Nuke Cannon needs to have better deployment handling


xezon:
Discuss if Nuke Cannon needs to have better deployment handling. Nuke Cannon can deploy in Guard Mode, but will not work well because it may try to chase units poorly instead of staying deployed. Investigate if this can be improved.

xezon:
Maybe a simple fix here would be to adjust Guard Mode range to Firing range.

commy2:
This fix https://github.com/xezon/GeneralsGamePatch/pull/269 already is a huge help.

ZekeDlyoung:
> Maybe a simple fix here would be to adjust Guard Mode range to Firing range.

If you mean make the guard range the same as its weapon range, that is a bad idea. There is a bug in ZH where artillery can auto acquire units inside fog of war, if the Nuke Cannon's guard range is extended, people can guard into the fog and attack targets even if they can't see them

xezon:
Oh I thought Guard range is larger than Firing range, and that is why they undeploy, to move closer to the target and shoot. Right now Nuke Cannons will not wait for enemy units to come close, but instead try to move to enemy units for a kill, which is bad given the time it requires to deploy.

commy2:
1.04

Attack range: 150 - 350
Guard radius: 180
Shroud clearing range: 350

So guard range already is much smaller than attack range.

Stubbjax:
Here is a demonstration of the unnecessary undeploy / redeploy behaviour after a target is killed.

https://user-images.githubusercontent.com/11547761/133054117-e407aafc-0ff4-433f-a103-0a2dc4563630.mp4

Also shows some funky retaliation behaviour on the Laser Crusader.

xezon:
Ok judging by that video it looks like a clear bug.

xezon:
Zeke:
> You either:
> 1. Allow it to auto attack, which let's it guard at full range without weird redeploy bug, at the cost of it wasting shots and blowing up your own troops due to reinforcement GPs
> 2. Make it need to manually change modes using switch weapon/bike logic, which makes it cumbersome to use, and bike logic makes it instantly die from neutron and pilot kill
> 3. Keep it the way it is and just don't use guard mode with it, just make it deploy with force fire and just manually attack the stuff you want

ZekeDlyoung:
There's another method I forgot, Option 1 but with hold fire button, basically a weapon switch to a dummy weapon to stop it from firing.

[...]

PR #975, opened 2022-08-25
Fix: Listening Outpost Guard Mode Range Too Large

### LINKED ISSUE #964: Guard mode area for Outposts is too large
Guard mode is often very unpractical to use with Outposts, because it makes your unit wander off very far to engage targets. 

Instead of matching the vision range, it should be closer to it's actual attack range?

It could also be a pro, when defending vs Dragons flanks on wider more open maps for example, unlike a BM which can only cover a tiny area.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/964

Note, this does not affect the detection range, because the Outpost's StealthDetectorUpdate module overwrites it via DetectionRange.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -2620,7 +2620,7 @@ Object ChinaVehicleListeningOutpost
   End
   BuildCost       = 800
   BuildTime       = 15.0          ;in seconds
-  VisionRange     = 250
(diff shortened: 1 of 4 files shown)


xezon:
Any specific reason 150 was picked or just lowered?

I am asking because Vision Range was not yet fully understood and investigation was not yet completed.

See
* #337

We can revisit this when we have gathered more knowledge.

commy2:
150 was chosen, because it is the same value as Tank Hunter, Battlemaster, Gatling tank, Overlord and ECM, all units that would be mixed together with the Listening Outpost.

xezon:
Ok that makes sense.

xezon (approved):

PR #982, opened 2022-08-26
Fix: Angry Mob Deals No Damage With AK47

commy2:
Fixes bug introduced in https://github.com/TheSuperHackers/GeneralsGamePatch/pull/279

I gave the dummy projectile to GLAAngryMobAK47Weapon instead of GLAAngryMobAK47**NoDamageWeapon**.

LMAO


xezon (approved):
This was good for balance.

PR #985, opened 2022-08-27
Fix: Broken Angry Mob Animations

commy2:
- Fix: Broken pistol reloading animation
- Fix: Broken molotov throw animation
- Fix: Mob member with molotov is holding rock
- Fix: Can't attack ground with AK47
- Fix: Dust trail when moving disappears after Arm The Mob upgrade
- Fix: Angry Mob members "walks in place" instead of playing death animation when killed by certain weapon types after Arm The Mob

# before

https://user-images.githubusercontent.com/6576312/187032918-ee9e5c22-c42f-4055-bfe8-6e0cc1ddc58f.mp4

# after

https://user-images.githubusercontent.com/6576312/187032930-940810b4-5398-4c7d-8dd8-1ceac877eecf.mp4


xezon:
It looks great. Mob looks extra angry now. Exceptional throws. I think someone more qualified needs to review INI. I do not have experience with these settings.

PR #1003, opened 2022-08-28
Change: Implement Previously Unused Nuke Helix Texture

commy2:
There is an unused texture for the Nuke Helix. Since the Nuke Helix can indeed drop nuclear bombs, I think using this would be neat.

shot_20220813_180742_1


xezon:
* old PR: #888

xezon (approved):

PR #1005, opened 2022-08-28

commy2:
ZH 1.04

patch:

It keeps the toxin trail.


xezon:
* old PR: #394

commy2:
Please don't force push again. I squashed two commits into one, because the first one was missing something.

xezon:
Ok sorry.

PR #1006, opened 2022-08-28

### LINKED ISSUE #300: Inf China Minigunner in Helix has extended firing range when attacking air unit together with a grounded Minigunner

commy2:
- fix https://github.com/xezon/GeneralsGamePatch/issues/300

ZH 1.04

After patch:

Drawback:


xezon:
* old PR: #342

ExiLe-ZH:
Wouldn't change, Helix vs Helix battles are fun now.

Also Inflix is gonna be OP vs chinooks.

PR #1018, opened 2022-08-29
Fix: Delete Immortal Aurora Wrecks Circling In Air

commy2:
Please ignore the Nuke and the Migs. I needed a savegame where this reliably happens after load for debugging.

https://user-images.githubusercontent.com/6576312/187041891-093f65c0-268b-4088-9852-e5059d531728.mp4

They still bug out and circle, but the spectactle will be cleaned up after 10 seconds maximum. This is the best I can do with INI scripting.

previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/994


xezon (approved):
Looks good to me.

PR #1020, opened 2022-08-29
Fix: SWG Particle Cannon Uses Portrait Of Normal Particle Cannon

commy2:
# before
shot_20220820_104753_2
shot_20220820_104744_1

- SWG has unique button, but the same portrait as other USA
- button has blue and red team color mixed in different places

# after
shot_20220820_104337_1

- portrai and button match
- team color changed to blue

previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/998


xezon (commented):
It shows "Git LFS file not shown" here in the TGA's

xezon (approved):

PR #1022, opened 2022-08-29
Tweak: Replace Generic Tank On USA War Factory Button With Crusader

commy2:
to match the art on the portrait

previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/996


xezon:
It shows "Git LFS file not shown" here in the TGA's

xezon (approved):

PR #1026, opened 2022-08-29
Add: Upgrade Voice For MOAB, Sentry Drone Gun, Neutron Shells And Fortified Structure

commy2:
Files are already from the game, but can for this use case only be read from this folder, so they have to be copied there unfortunately.

previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/999


xezon (commented):
It shows "Git LFS file not shown" here in the WAVs

xezon (approved):

PR #1012, opened 2022-08-28
Add: Cheering Animation For Rebel

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1011

PR #1030, opened 2022-08-29
Add: Missing Cheering Animations

commy2:
previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1012
previous: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1013

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -921,6 +921,7 @@ CommandSet AmericaBarracksCommandSet
   2 = Command_ConstructAmericaInfantryMissileDefender
   3 = Command_ConstructAmericaInfantryColonelBurton
   4 = Command_ConstructAmericaInfantryPathfinder
+  5 = Command_ConstructAmericaInfantryPilot
   6 = Command_ConstructAmericaInfantryBiohazardTech
   7 = Command_UpgradeAmericaRangerFlashBangGrenade
   8 = Command_UpgradeAmericaRangerCaptureBuilding
@@ -1327,6 +1328,7 @@ CommandSet GLABarracksCommandSet
   6  = Command_ConstructGLAInfantryJarmenKell
   7  = Command_ConstructGLAInfantrySaboteur
   8  = Command_UpgradeGLAInfantryRebelBoobyTrapAttack
+  9  = Command_UpgradeGLAArmTheMob
 ; camo net
   11  = Command_UpgradeGLARebelCaptureBuilding
   13 = Command_SetRallyPoint
@@ -2070,6 +2072,7 @@ CommandSet AirF_AmericaBarracksCommandSet
   2 = AirF_Command_ConstructAmericaInfantryMissileDefender
   3 = AirF_Command_ConstructAmericaInfantryColonelBurton
   4 = AirF_Command_ConstructAmericaInfantryPathfinder
+  5 = Command_ConstructAmericaInfantryPilot
   7 = Command_UpgradeAmericaRangerFlashBangGrenade
   8 = Command_UpgradeAmericaRangerCaptureBuilding
  13 = Command_SetRallyPoint
@@ -2333,6 +2336,7 @@ CommandSet Demo_GLABarracksCommandSet
   4  = Demo_Command_ConstructGLAInfantryAngryMob
   6  = Demo_Command_ConstructGLAInfantryJarmenKell
 ; 8  = Command_UpgradeGLAInfantryRebelBoobyTrapAttack
+  9  = Command_UpgradeGLAArmTheMob
   11 = Command_UpgradeGLARebelCaptureBuilding
   13 = Command_SetRallyPoint
   14 = Command_Sell
@@ -3008,6 +3012,7 @@ CommandSet Slth_GLABarracksCommandSet
   5  = Slth_Command_ConstructGLAInfantryHijacker
   6  = Slth_Command_ConstructGLAInfantryJarmenKell
   7  = Slth_Command_ConstructGLAInfantrySaboteur
+  9  = Command_UpgradeGLAArmTheMob
   11 = Command_UpgradeGLARebelCaptureBuilding
   12 = Command_UpgradeGLACamoNetting
   13 = Command_SetRallyPoint
@@ -3289,6 +3294,7 @@ CommandSet Chem_GLABarracksCommandSet
   3  = Chem_Command_ConstructGLAInfantryTerrorist
   4  = Chem_Command_ConstructGLAInfantryAngryMob
   6  = Chem_Command_ConstructGLAInfantryJarmenKell
+  9  = Command_UpgradeGLAArmTheMob
   11 = Command_UpgradeGLARebelCaptureBuilding
   13 = Command_SetRallyPoint
   14 = Command_Sell
@@ -3941,6 +3947,7 @@ CommandSet SupW_AmericaBarracksCommandSet
   2 = SupW_Command_ConstructAmericaInfantryMissileDefender
   3 = SupW_Command_ConstructAmericaInfantryColonelBurton
   4 = SupW_Command_ConstructAmericaInfantryPathfinder
+  5 = Command_ConstructAmericaInfantryPilot
   7 = Command_UpgradeAmericaRangerFlashBangGrenade
   8 = Command_UpgradeAmericaRangerCaptureBuilding
  13 = Command_SetRallyPoint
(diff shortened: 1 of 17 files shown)


commy2:
WIP. I want to do all cheers in one PR. Non vGLA Rebels are still using the wrong animation. Do not merge yet.

xezon:
x = has animation
(GOOD) = good transition
(OKAY) = okay transition
(bad) = bad transition
## Original

| Object                   | Regular Cheer  | Injured Cheer  |
|--------------------------|----------------|----------------|
| USA Ranger               | x              | x              |
| USA Missile Defender     |                |                |
| USA Pathfinder           |                |                |
| USA Burton               | x              | x              |
| China Red Guard          | x              |                |
| China Tank Hunter        |                |                |
| China Lotus              | x              | x              |
| China Hacker             | x              |                |
| GLA Rebel                |                |                |
| GLA Tunnel Defender      | x              | x              |
| GLA Terror               | x              |                |
| GLA Hijacker             | x              |                |
| GLA Saboteur             | x              | x              |
| GLA Jarmen               | x              | x              |
| GLA Mob                  | x              | x              |

## Patched

| Object                   | Regular Cheer  | Injured Cheer  |
|------------------------  |----------------|----------------|
| USA Ranger               | x (bad)        | x (GOOD)       |
| USA Missile Defender     | x (bad)        | x (bad)        |
| USA Pathfinder           | x (GOOD)       | x (GOOD)       |
| USA Burton               | x (bad)        | x (GOOD)       |
| USA Pilot                | x (GOOD)       | x (GOOD)       |
| China Red Guard          | x (GOOD)       | x (GOOD)       |
| China Minigunner         | x (GOOD)       | x (GOOD)       |
| China Tank Hunter        | x (GOOD)       | x (GOOD)       |
| China Lotus              | x (bad)        | x (GOOD)       |
| China Hacker             | x (bad)        | x (bad)        |
| GLA Worker               | x (GOOD)       | x (GOOD)       |
| GLA Rebel                | x (GOOD)       | x (GOOD)       |
| GLA Tunnel Defender      | x (GOOD)       | x (GOOD)       |
| GLA Terror               | x (bad)        | x (bad)        |
| GLA Hijacker             | x (OKAY)       | x (OKAY)       |
| GLA Saboteur             | x (GOOD)       | x (GOOD)       |
| GLA Jarmen               | x (bad)        | x (GOOD)       |
| GLA Mob                  | x (bad)        | x (bad)        |
| GLA Mob AK47             | x (bad)        | x (bad)        |

commy2:
>  USA Missile Defender Damaged - No animation

Interesting. Then I _think_ this is bugged in NProject too. I figured it would not work with the setup they had, but didn't bother testing. No problem to fix though.

I don't think these have graceful transitions.

ThePredatorBG:
> The Missile Defender Animation works, but there is no graceful transition in and out of it. Same for Ranger.
> 
> USA Ranger - Works, but bad transition USA Ranger Damaged - Works, but bad transition USA Missile Defender (this) - Works, but bad transition USA Missile Defender Damaged - No animation USA Pathfinder - No animation USA Pathfinder Damaged - No Animation USA Burton - Works, looks acceptable USA Burton Damaged - Works, looks acceptable

USA Pathfinder has animation - "AIPFDR_CHA". Also, normal animations can be used for damaged states if damaged animations don't exist.

commy2:
Should work for all USA units and all Rebels now.

xezon:
Tested most infantry units and updated table above. Pilot and Minigunner missing. "Transition Ok" means a very generous judgement whether or not the clipping looks alright. Ideally the animations would properly interpolate. All of them do not. Unclear to me if that can be fixed in INI.

### Videos of current state with this branch

https://user-images.githubusercontent.com/4720891/187472708-bf775bbe-8d8b-46ec-ad7a-4dee30441cf3.mp4

https://user-images.githubusercontent.com/4720891/187472720-6a62c566-45fe-4482-bad9-d477a771fc57.mp4

https://user-images.githubusercontent.com/4720891/187472726-81b3f46b-7179-4df2-914f-b23aed554ef6.mp4

commy2:
Fixed the Stand -> StandInjured transition playing for Ranger and Burton after cheering when hurt. Also removed the freeze of MD during the cheer.

commy2:
Should be done now.

xezon:
I will make cheering test map and test this.

commy2:
You can add the CIA Agent, Biohazard Tech, and USA and China Officer. I check, it should work for them correctly already.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -3056,7 +3058,7 @@ Object AirF_AmericaInfantryColonelBurton
   ShroudClearingRange = 500
   Prerequisites
     Object = AirF_AmericaBarracks
-    Object = AirF_AmericaStrategyCenter
+    ;Object = AirF_AmericaStrategyCenter
```
xezon, line 3061: This looks like it should not be here.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -4371,6 +4378,7 @@ CommandSet Lazr_AmericaBarracksCommandSet
   2 = Lazr_Command_ConstructAmericaInfantryMissileDefender
   3 = Lazr_Command_ConstructAmericaInfantryColonelBurton
   4 = Lazr_Command_ConstructAmericaInfantryPathfinder
+  5 = Command_ConstructAmericaInfantryPilot
```
xezon, line 4929: This looks like it should not be here.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
xezon, line 3078: This looks like it should not be here.
Many times in this change.

commy2:
Could you check again? Should be complete.

Don't forget the Worker.

xezon:
Last time I checked what I did not like about the cheering is that the animations have no proper transitions. They just jump from one state to another. Not clear to me if this can be improved.

There is a case to make that if the animation jumps cannot be solved, then perhaps there should be no cheering animation.

commy2:
> Last time I checked what I did not like about the cheering is that the animations have no proper transitions. They just jump from one state to another. Not clear to me if this can be improved.

I wouldn't know how. It's like when they start moving or die. It's all instantly. I don't think there even are any transitions for these. Some naturally end up at the standing position though.

> There is a case to make that if the animation jumps cannot be solved, then perhaps there should be no cheering animation.

The sound plays anyway. No animation is weird. You could have a group of units without any animations.

xezon:
Hmm for other animations it appears to be less of a visual problem. Perhaps because the animation jump is not as obvious?

MTKing4:
Instant Transition is Fine

commy2:
It's better than the current state anyway.

xezon:
I rebased on latest main.

I noticed something. The cheering animations stop when the cheering event stops. This way the animation does not gracefully transition to idle but clips back. If the cheering is continued until the animation end is reached, then the cheering looks good.

Can we perhaps force the cheering animation to finish, as long as unit is in cheering or idle state? This way we could get much better results.

See attached video for reference. The first cheer is ended early and animation looks bad. But second cheer is ended late and animation properly finishes until it reaches idle stance.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/d9da2977-0f6e-44f8-9116-5089d1b79c07

commy2:
> The first cheer is ended early and animation looks bad. But second cheer is ended late and animation properly finishes until it reaches idle stance.

Is the cheer duration random?

xezon:
The cheer state has fixed duration. It's timeout duration is reset when pressing CTRL + C anew. The ideal fix here would be to allow for the animation to finish even across cheer state.

commy2:
No idea how to do that.

xezon:
Perhaps requires code support.

commy2:
Implemented some ideas, USA only for now. Please check.

There're two types of cheering animations

1) those that loop
2) those that begin and end in a standing pose

It's possible to make smooth transitions for (2), but not (1). Some units have 2 cheering animations - healthy and injured, some units have only 1 cheering animation, especially those that don't have an injured walk/standing animation to begin with.

MD only has 1 animation of type (1)
Pathfinder only has 1 and of type (2)
Ranger and Burton have two animations, where the healthy one is type (1), the injured one is type (2) 
etc.

I added the smooth transitions were possible.

Another idea I have is adding randomization to these animations, so large homogenous groups of units don't look so synchronized when cheering. Idk if that is overkill though.

Will need to go through civs too eventually, especially the recently added armed civs.

xezon:
I tested the 5 usa infantry.

Pilot and Pathfinder look very good.

commy2:
China batch.

commy2:
GLA Batch. It's getting quite big. I propose merge this if all good and move on to civilians separately. They likely need more tweaks anyway.

xezon:
Good effort.

I have updated table in second post.

14 cheer transitions look bad.
2 cheer transitions look ok.
22 cheer transitions look good.

I think this is as good as can be without touching w3d animations.

xezon:
Could get rid of 4 bad transitions by removing cheer animations from Angry Mob. They are cheering all the time anyway. It looks as if their idle animations are identical or similar to the cheering anyway.

xezon (approved):
Looks good to me. Debug needs to be removed. Documentation needs to be added. Perhaps remove obsolete cheering from Angry Mobs.

commy2:
> They are cheering all the time anyway.

Same animation files, but idle "cheering" is played slower than cheering cheering.

xezon:
Now you submitted this with all the extra debug. And without documentation.

PR #1038, opened 2022-08-29
Move unedited W3D files into GameFilesOriginalZH folder

xezon:


commy2:
> BIN -25.6 KB
[Patch104pZH/GameFilesEdited/Art/W3D/ABSDILink_L.W3D](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1038/files#diff-db5c17c44cb4f15009b70820820fea170221fd9ccd64f579c1d984ded2f13b0d)

This should stay.

xezon:
I was wondering about it. It is unchanged to original. What does it do?

commy2:
Ah, I am wrong.
ABSDILink_L.W3D = original animated particle cannon explosion beams 
ABSWGLink_L.W3D = hex edited using purple texture for SWG
ABLGCLink_L.W3D = hex edited using red texture for Laser Cannon

ABSDILink_L.W3D was added, because you requested I add the original reference. 
The file is unchanged from original. Your call where it goes or if we delete it.

xezon:
It already is in the other folder, that is why it does not show as moved, but deleted. I will create new Repo (Private) with all reference files and clean this Repo. Maybe I should do that straight away and not even bother moving files.

commy2 (approved):

PR #1042, opened 2022-08-30
Change: Improved Sentry Drones

### LINKED ISSUE #862: Sentry Drone is rarely used

xezon:
* old PR: #970
* Fixes: #862

Improved Sentry Drones so that they don't suck as much. This implementation comprises several proposals from https://github.com/TheSuperHackers/GeneralsGamePatch/issues/862. Further adjustments may be necessary, but this should form a solid, low-risk foundation.

### Summary of changes:

- Removed the invisible 1s pack and unpack delays
- Reduced prices from $1000/$850/$800 to $600
- Reduced build times from 10s to 8s
- Added a minor healing function of 1hp / second (Sentries have 300hp)
- Fixed the lack of stealth / unstealth sounds
- Reduced machine gun upgrade cost and research time from $1000 / 30s to $500 / 15s

### Footage of changes:

https://user-images.githubusercontent.com/11547761/186616126-7ad02caf-25bc-4d73-94ae-d2177e487faf.mp4

The footage should hopefully demonstrate the greatly improved responsiveness of the drones, as well as their minor healing ability.

### Rationale:

These changes should assist in making Sentry Drones more versatile and provide greater strategic diversity for USA factions to employ, without boosting them too much that they become a primary strategy and change the way USA is typically played.

#### Fix the broken weapon deployment behaviour

It goes without saying that Sentry Drones would be better off without the strange and invisible “weapon deployment” delay that occurs between firing and moving.

#### Reduce and standardise prices

Sentry Drones are considerably expensive units for what they do, and a Humvee with a scout drone is almost always a more cost-effective and versatile option. Reducing the price of Sentry Drones from $1000/$850/$800 to $600 would be an effective and low-risk way to boost utility, due to the fact that they serve a very specific purpose that is already readily available elsewhere, and are otherwise never used.

#### Reduce construction time

Another effective means of increasing utility is to decrease the build time from 10 to 8 seconds. Note that Sentry Drones still require the Sentry Drone Gun upgrade to become useful in combat, which prevents a reduced construction time (and several other improvements) from opening up early Sentry Drone-based strategies and potentially affecting traditional gameplay too much.

#### Add auto-repair capability

As Sentry Drones cannot rank up, a good addition would be to give them auto-repair capabilities. This would not only distinguish the unit from other vehicles a bit more, but would allow for a unique tactic where players can command the unit to stop and restealth in order to heal up at the cost of time and the opponent likely remembering the unit’s location.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -6955,11 +6955,11 @@ Object AirF_AmericaVehicleSentryDrone
     DetectionRate            = 900   ; how often to rescan for stealthed things in my sight (msec)
     DetectionRange           = 225   ;Dustin, enable this for independant balancing!
   End
-
-;  Behavior = GrantUpgradeCreate ModuleTag_13
-;    UpgradeToGrant = Upgrade_AmericaSentryDroneGun
-;    ExemptStatus = UNDER_CONSTRUCTION
-;  End
+  Behavior = AutoHealBehavior ModuleTag_13
+    StartsActive  = Yes
+    HealingAmount = 1
+    HealingDelay  = 1000
+  End
 
   Geometry = BOX
   GeometryMajorRadius = 9.0
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -6819,7 +6819,7 @@ Object AirF_AmericaVehicleSentryDrone
     Armor           = SentryDroneArmor
     DamageFX        = TruckDamageFX
   End
-  BuildCost       = 850
+  BuildCost       = 600
   BuildTime       = 10.0          ;in seconds
   VisionRange     = 200
   ShroudClearingRange = 350
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -6895,10 +6895,10 @@ Object AirF_AmericaVehicleSentryDrone
       MaxIdleScanAngle      = 360     ; in degrees off the natural turret angle
     End
     AutoAcquireEnemiesWhenIdle = Yes ;Stealthed ; Means "Yes when idle, even if I am stealthed"
-    PackTime = 1000
-    UnpackTime = 1000
+    PackTime = 0
+    UnpackTime = 0
     TurretsFunctionOnlyWhenDeployed = Yes
-    TurretsMustCenterBeforePacking = Yes
+    TurretsMustCenterBeforePacking = No
   End
 
   Locomotor = SET_NORMAL SentryLocomotor
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -6958,8 +6958,9 @@ Object AirF_AmericaVehicleSentryDrone
   End
   Behavior = AutoHealBehavior ModuleTag_13
     StartsActive  = Yes
-    HealingAmount = 1
+    HealingAmount = 2
     HealingDelay  = 1000
+    StartHealingDelay = 3000
   End
 
   Geometry = BOX
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -6836,6 +6836,7 @@ Object AirF_AmericaVehicleSentryDrone
 
   ; Patch104p @bugfix commy2 04/09/2021 Added missing move start sound effects from vanilla USA Sentry Drone.
   ; Patch104p @bugfix commy2 09/09/2021 Added missing sound when unit leaves factory.
+  ; Patch104p @balance Stubbjax 25/08/2022 Removed deployment times, reduced price, reduced build time, fixed stealth sounds, added auto-healing ability.
 
   ; *** AUDIO Parameters ***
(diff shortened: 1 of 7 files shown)


Jundiyy:
- Removed the invisible 1s pack and unpack delays
- Reduced prices from $1000/$850/$800 to $600
- Reduced build times from 10s to 8s
- Added a minor healing function of 1hp / second (Sentries have 300hp)
- Fixed the lack of stealth / unstealth sounds

I think all are good to go, other than the healing (I'm sure I said this in the other topic), not sure how everyone will take that.
I don't think it's a big problem.

commy2:
This doesn't address the Sentry Drone Gun upgrade at all.
What about the track marks thing?

Jundiyy:
Gun upgrade is in a different PR?
If not, then yes, needs to be discussed here.

Could be as cheap as $500 if the weapon is not edited. 10/15s upgrade time.

Track marks? If you mean that you can see the tracks, then I don't mind both ways.

Stubbjax:
I've committed additional changes to the machine gun upgrade so that it is now $500 / 15s - a straight 50% reduction from $1000 / 30s. I don't anticipate this having a significant impact on the meta or traditional gameplay, though a well-executed 'sentry drone rush' may be able catch an unsuspecting opponent off-guard once every blue moon (and be a bit of fun).

The track marks thing doesn't really phase me too much because they are only spawned if the unit is visible to the opponent at the time, and not in any shrouded areas. So the primary behaviour of trying to subtly position these units outside the opponent's sight range, akin to Outposts or Pathfinders, remains roughly the same.

Regarding the auto-heal feature, there still seems to be some indecision. I've tried to summarise the **for** and **against** arguments below.

**For:**
- Distinguishes the unit from other vehicles and gives it a bit more uniqueness
- Ties in nicely with the unit's purpose as a lone, self-sufficient scout
- They cannot rank up to acquire auto-repair capabilities at elite rank
- They cannot purchase a Battle Drone for repairs
- They are often far from home / hiding and returning to a War Factory to repair is not viable
- It is a relatively low-visibility change due to the unit's rarity and shouldn't phase purists too much
- Introduces additional ~~'hit-n-hide'~~ 'hit-n-heal' tactics that combo well with the stealth ability
- Thematically fits the faction and drone's 'high-tech' design

**Against:**
- The heal speed is so low that it is unnecessary, and any higher value would be too strong
- The slow repair speed means the repair icon stays on the unit for a long time which could look unsightly
- It adds complexity
- It introduces more potential for controversy
- It is an extra behaviour that players have to learn and become accustomed to
- It doesn't realistically make sense for drones to repair themselves
- It makes the unit too strong

I'm happy to remove the auto-heal behaviour if there are too many reservations. I thought it would be a nice little bonus addition to give Sentry Drones a bit more appeal, but it is certainly not a necessary change.

ExiLe-ZH:
I think auto-repair is quite neat aswell. I mentioned some good reasons in the other topic (pretty identical to yours actually):

https://github.com/TheSuperHackers/GeneralsGamePatch/issues/862#issuecomment-1229164560

xezon:
This PR has been superseded by #1162. It can be reopened if there is a strong feel towards buffing Sentry Drone, as #1162 has just made an attempt to make it not useless.

My immediate thoughts:
* 600 cost is cheap
* 8 seconds build time may be ok
* Stealth sounds are debatable. Outpost also does not have this sound
* Auto Repair is interesting feature, but seeing the Repair Icon for 5 minutes seems not optimal

xezon:
Regarding self healing Sentry Drone, there is this KindOf that could hide repair icon: `NO_HEAL_ICON`. But it would add confusion because then player would never see Drone repairing at all.

PR #1047, opened 2022-08-30
Add: Heal Effect to Tech Radio Station

xezon:
* old PR: #623

Jundiyy thinks it could have a bit more health.

He knows it wasn't given any modules in the end but he thinks it's good to have something as a starting ground, so it can at least be placed and used by the simple mappers, if anyone wants to give it some extra abilities and have map.ini knowledge, they can do what they want.

He thought having a Speaker Tower styled Tech building seemed quite cool. Has a larger radius than the Speaker Tower, allowing players to use the Tech as a key fighting area.

(Of course it's not going to be built by the USA Dozer but just to show how it works.)

https://youtu.be/eNZvtTivvqQ


MTKing4:
I think i can settle with this just having a very large Vision range just like Observation Towers in Red Alert 3

PR #1048, opened 2022-08-30
Fix firing elevation limits

xezon:
* old PR: #111

All weapons which previously had TargetPitch parameters no longer have this. This allows them to acquire targets regardless of minor elevation differences.


ExiLe-ZH:
Does this also fix Dragons missing/not clearing elevated bunkers?

commy2:
No, seems only to be for shell based weapons. Tanks, Fire Base, Strat Center ... and Angry Mob Molotov for some reason.

xezon:
We tested this already. With this change tanks can shoot up and down any steep terrain. Realistically we can only close this, but I will leave it open for now for reference, because this is ongoing investigation.

xezon:
Actually, we can close already, it is linked in the other Pull Request, so can be viewed from there.

PR #1049, opened 2022-08-30
Scud Storm launch can no longer be cancelled by accident

xezon:
* old PR: #234

Scud Storm, if given a target or ordered to force fire while launching, would try to swap to a non-existent primary weapon and thereby stop the launch. This commit fixes that.


commy2:
Unfortunately, this commit re-introduces the Scudbug.

xezon:
I suspect we can close it then?

commy2:
Yeah, unfortunately, this is not the path to fix.

PR #728, opened 2022-07-22
Balance: Invert and Redistribute Quad Cannon Guns DPS

xezon:
Reference issue:
* #56

This change redistributes the Quad veterancy damage values so that VETERAN has the least bonus damage and HERO has the most. The sum of all bonus is identical to the sum of all original bonus. A Quad Cannon with veterancy and without scrap will now deal about the same damage as a China Gatling Tank with veterancy. Before it dealt around 33% damage more than its Gatling Tank counterpart. Quad Cannon will unleash full potential with scrap upgrades only, where it will deal up to around 33% more damage than Gatling Tank.

## Original
before

## Patched
after

## Original
https://user-images.githubusercontent.com/4720891/182035421-266e6e12-fce2-44ca-a4f5-2abe67ba91e7.mp4

## Patched
https://user-images.githubusercontent.com/4720891/182035426-c30d780b-54f5-4b0e-86d9-7c86c4916b3b.mp4

PR #1055, opened 2022-08-30
Fix: Invert and Redistribute Quad Cannon Guns DPS

### LINKED ISSUE #56: GLA Quad Cannon doubles DPS when vet 1 and becomes weaker when scrapped and vetted

xezon:
**Merge with Rebase**

* old PR: #728
* Fixes #56
* Closes #1052
* Closes #1053
* Closes #1054

This change redistributes ground gun damage values of the Quad Cannon to achieve a proper damage progression from one salvage upgrade to another. Originally the Quad Cannon would become weaker after picking up salvage crates while being promoted to VETERAN, ELITE or HERO.

The redistributed sum of all patched veteran bonus is identical to the sum of all original veteran bonus. A Quad Cannon with veterancy and without salvage upgrade will now deal about the same damage as a China Gatling Tank with veterancy. Before it dealt around 33% damage more than its Gattling Tank counterpart. The Quad Cannon will unleash its full potential with salvage upgrades only, where it will deal up to around 33% more damage than the Gattling Tank.

Additionally the damage bonus for salvage upgrade 1 and salvage upgrade 2 without veterancy have been redistributed to achieve a progression. They now give +40% and +80% bonus damage instead of +60% and +60% bonus damage.

Additionally the damage bonus for Easy AI, Hard AI, Subliminal and Faerie (Avenger Target) have been adjusted to achieve a similar level of progression from one salvage upgrade to another. The redistribution sum does vary a bit due implementation constraints, but is overall fairly redistributed.

The air guns of the Quad Cannon are entirely unchanged, because they do not have this issue.

## Original

before

## Patched (Version 3)

quadcanndmg

## Original

https://user-images.githubusercontent.com/4720891/182035421-266e6e12-fce2-44ca-a4f5-2abe67ba91e7.mp4

## Patched (Version 3)

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/449d2da1-1ad5-4a7c-a715-12d97b8dc21b

# Rationale

The scrap crate is supposed to make GLA weapons stronger and not weaker. It is counter intuitive that not picking up scrap crates would be the better choice for the Quad Cannon. Therefore fixing the damage progression of the Quad Cannon is logical.

Overall the patched Quad Cannon performs comparable to the original Quad Cannon, but with its damage bonuses inverted. The impact of this change is also affected by the removal of the salvage crate scaffold exploit with #1676. This means Quad Cannons will have more opportunities to pick up salvage crates in practice because the enemy player can no longer remove them with the exploit.


xezon:
@commy2 wants to look into this as well.

xezon:
I tested Unscrapped Quad vs Gatt. Here are relevant results. Table shows winner with rough percentage of health left, judged by health bar.

## Original Unscrapped Quad vs Gatt, No Pre Spin

|  VS       | Vet0 Quad | Vet1 Quad | Vet2 Quad | Vet3 Quad |
|-----------|-----------|-----------|-----------|-----------|
| Vet0 Gatt | Gatt 20%  |           |           |           |
| Vet1 Gatt |           | Quad 30%  |           |           |
| Vet2 Gatt |           | Quat 15%  |           |           |
| Vet3 Gatt |           | Gatt 2%   | Quad 5%   |           |

## Original Unscrapped Quad vs Gatt, WITH PRE SPIN

|  VS       | Vet0 Quad | Vet1 Quad | Vet2 Quad | Vet3 Quad |
|-----------|-----------|-----------|-----------|-----------|
| Vet0 Gatt | Gatt 30%  |           |           |           |
| Vet1 Gatt |           | Quad 20%  |           |           |
| Vet2 Gatt |           | Quad 5%   | Quad 20%  |           |
| Vet3 Gatt |           | Gatt 5%   | Quad 2%   |           |

## Patched Unscrapped Quad vs Gatt, No Pre Spin

|  VS       | Vet0 Quad | Vet1 Quad | Vet2 Quad | Vet3 Quad |
|-----------|-----------|-----------|-----------|-----------|
| Vet0 Gatt | Gatt 20%  |           |           |           |
| Vet1 Gatt |           | Quad 5%   |           |           |
| Vet2 Gatt |           | Gatt 5%   | Quad 5%   |           |
| Vet3 Gatt |           |           | Gatt 15%  | Quad 1%   |

## Patched Unscrapped Quad vs Gatt, WITH PRE SPIN

|  VS       | Vet0 Quad | Vet1 Quad | Vet2 Quad | Vet3 Quad |
|-----------|-----------|-----------|-----------|-----------|
| Vet0 Gatt | Gatt 30%  |           |           |           |
| Vet1 Gatt |           | Gatt 2%   |           |           |
| Vet2 Gatt |           |           | Gatt 1%   |           |
| Vet3 Gatt |           |           |           | Gatt 1%   |


# Conclusion

China Gattling has better fighting chance with this change, but Quad still wins in direct 1v1 on same veteran level if the Gattling is not spun up, albeit just barely surviving. If Gattling is spun up, it will just barely win and survive with the health bar pretty much empty. Quad would require real nerf for Gattling to be more comfortable. Or the Gattling a buff. Either in weapon or armor.

xezon:
Jundiyy:
> It seems reasonable, especially since vet 0 has been kept the same, which I think is very important. Since most of the units will be vet 0. Changes seem good but perhaps they have been tweaked a bit too much?

commy2:
The current PR doesn't address the bug happening with Quads when combined with Propaganda (happens every other moon in team games) and Avengers (never happens) or when fighting Hard AI. It also has no boost from v0s1 to v0s2. 

I also want to do more testing with FPS != 30 (skrimish).

It should also be considered that AA units have really strange veterancy requirements compared to other units (100xp for vet 1, but then only 50 more xp for vet 2) as mentioned here https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1175.

xezon:
> The current PR doesn't address the bug happening with Quads when combined with Propaganda (happens every other moon in team games) and Avengers (never happens)

I think that is fine. They give no bonuses as per Original setup. No need to change here. If it needs to be changed, then that can be done in a separate change as well, as that would be a definite Buff.

> or when fighting Hard AI.

I can look into that.

> It also has no boost from v0s1 to v0s2.

This was deliberate to not change more than necessary. Essentially it is not as big of a problem when a Scrap does not give more damage, as long as it does not give less.

But yes we could redistibute the +60% +60% to +40% +80% if I am not mistaken.

MTKing4:
> Essentially it is not as big of a problem when a Scrap does not give more damage, as long as it does not give less

It doesn't give less on vet 0, it doesn't do anything, making it useless and a waste of a scrap that could've been on a different unit, this is one of the main issues that we wanted to fix the quad scrap for and we should fix it

xezon:
## Version 1

after

## Version 2

How about this?

invert_redist_2

ExiLe-ZH:
I think I like this model, because it doesn't snowball so crazy fast. This brings it closer to the Gatling Tank model (vet1 +32%, vet2 +80%, vet3 +95%). Now scraps have to be collected to get the most firepower, instead of instant firepower gained through killing units and especially buildings, something Gats can't do. 

The only concern is versus Airforce, because scraps will be non-existent when Air flies.

If we want it to be closer to the original then this might be more adequate (UpgradeOne are simply in between values):

-- | QuadCannonGun | QuadCannonGunUpgradeOne | QuadCannonGunUpgradeTwo
-- | -- | -- | --
NONE | 0% | 50% | 100%
VETERAN | 76% | 98% | 120%
ELITE | 92% | 116% | 140%
HERO | 108% | 134% | 160%

xezon:
> The only concern is versus Airforce, because scraps will be non-existent when Air flies.

Not a concern. The original Air guns are separate and do not suffer from bug (see columns 2, 4, 6). Now I did apply some tweaks to it in my last proposal, but I can revert it to original, so performance vs Aircraft can be completely unchanged.

> If we want it to be closer to the original then this might be more adequate (UpgradeOne are simply in between values):

That would be a definite buff for Quad Cannon because in sum the Damage Output would be significantly higher. Plus it means that Quad will always win vs Gattling at Vet1, Vet2, Vet3 without Scrap.

ExiLe-ZH:
> That would be a definite buff for Quad Cannon because in sum the Damage Output would be significantly higher. Plus it means that Quad will always win vs Gattling at Vet1, Vet2, Vet3 without Scrap.

Sum is not really relevant to determine whether it's a buff or nerf, it's about game/battle dynamics. 

What previously happened in mass (Quad vs Gatling) engagements is that Quads turning Vet1 instantly change fights with their 120% dps increase, making things very snowbally. This doesn't happen anymore. Same for base races where no scraps are obtained, it's very snowbally in 1.04.

With the new model a fight has to be completely won first and scrap has to be collected, they will only be stronger in the next battle. You don't want to collect scrap whilst in battle becausing moving and shooting reduces the damage output.

So I'm pretty sure all models we came up with so far are actually nerfs.

xezon:
That is relying on assumptions. There are also scenarios where some Rocket Soldiers kill tanks and Quads drive over Scrap before killing their first targets.

This change intends to Redistribute the damages, not buff or nerf them. Likewise, the aim of this change is to not have a Buff or Nerf label asociated.

And on top of that we also removed Scrap Bug, so GLA actually has better chances to pickup scrap now in those top level matches.

ExiLe-ZH:
> That is relying on assumptions. There are also scenarios where some Rocket Soldiers kill tanks and Quads drive over Scrap before killing their first targets.

I'm not assuming anything, obviously you have different situations, but same like your example they are mostly smaller scale and less impacting. The dynamics I described are the most common/gamechanging and that's the experience of playing/watching MANY games. Ask any expert, they will say the same.

Actually Quads gaining advantage through these smaller scale situations (and be nerfed at a bigger scale) fits GLA as a faction. It promotes hit and run style tactics instead of mass engage. 

> This change intends to Redistribute the damages, not buff or nerf them. Likewise, the aim of this change is to not have a Buff or Nerf label asociated.

It's inevitably going to change game dynamics.

ExiLe-ZH:
Furthermore, in 1.04 people try to avoid scraps, so basically 120% - 140% - 160% is what applies the most.

xezon:
Anyway, I am against buffing the sum of damages at this point. It should only be done if GLA turns out to be weak.

ExiLe-ZH:
Well like I mentioned, I predict any redistributed model is going to be a nerf because of this:

> Furthermore, in 1.04 people try to avoid scraps, so basically 120% - 140% - 160% is what applies the most.

xezon:
I would not classify it as Nerf per se, for the same reason I would not classify Scud Bug fix as a Nerf. It is supposed to be a bug fix first, where the design clearly intended to make things stronger on Scrap up and not weaker. I expect that the majority of players are oblivious to the broken damage distribution, because it is difficult to measure in natural battle situations.

But yes players who actively avoided picking scrap with Quads will eventually have weaker units now. But that is the way it should be.

ExiLe-ZH:
> I would not classify it as Nerf per se, for the same reason I would not classify Scud Bug fix as a Nerf. It is supposed to be a bug fix first, where the design clearly intended to make things stronger on Scrap up and not weaker. I expect that the majority of players are oblivious to the broken damage distribution, because it is difficult to measure in natural battle situations.
> 

It still affects the general battle dynamics for everyone, which I tried to explain. It's just a simple fact that veterancy bonus is something that happens instantly during battles and has more direct impact than Scraps, Scraps are only picked up after the battle is over. 

It's also a fact that non-scrapped vet0 Quads are the most common type on the battlefield. So when an engagement happens and Quads become veteran, they now only do +60% damage, instead of the 1.04's +120% and it changes the whole course of that engagement, that's what I meant with the snowball effect. It's a pretty huge nerf isn't it?

It's all easily tested by simulating it, which is something we definitely need to do. It's not useful enough just looking at sum averages. I hope @MTKing4 could help me with this soon, because we were working on a testmap but it needs some scripting still.

Anyways, I hope this explanation is more clear.

xezon:
Potential Quad damage buff should be a separate change. It is mistake to intermingle it with the redistribution. I do not have data on how much buff is required, so I am unable to drive that discussion. However, redistribution is fair approach and this change does it. I will make a second pass on redistribution later.

ExiLe-ZH:
My suggestion absolutely does not intent to buff Quads, actually to do less of a nerf, because I predict that redistributing the values (inverting the model) is effectively a nerf, thus not a fair approach. 

I could be wrong, but let's see what happens in a simulation.

xezon:
Yeah but how much do you intent to upscale? 5%, 10%, 20%, 50%? It is not clear how much would be ok. Simple redistribution is fairest method as it does not discriminate or promote the overall unit abilities, aka on average the Quad Cannon will perform similar over all its modification states. That is why I intent to aim for redistribution first. If you want to buff it afterwards and can explain why it makes sense, then ok.

ExiLe-ZH:
> Yeah but how much do you intent to upscale? 5%, 10%, 20%, 50%? It is not clear how much would be ok. Simple redistribution is fairest method as it does not discriminate or promote the overall unit abilities, aka on average the Quad Cannon will perform similar over all its modification states. That is why I intent to aim for redistribution first. If you want to buff it afterwards and can explain why it makes sense, then ok.

But it depends on how common each state is on the battlefield isn't it?

MTKing4:
> I think I like this model, because it doesn't snowball so crazy fast. This brings it closer to the Gatling Tank model (vet1 +32%, vet2 +80%, vet3 +95%). Now scraps have to be collected to get the most firepower, instead of instant firepower gained through killing units and especially buildings, something Gats can't do.
> 
> The only concern is versus Airforce, because scraps will be non-existent when Air flies.
> 
> If we want it to be closer to the original then this might be more adequate (UpgradeOne are simply in between values):
> 
> --	QuadCannonGun	QuadCannonGunUpgradeOne	QuadCannonGunUpgradeTwo
> NONE	0%	50%	100%
> VETERAN	76%	98%	120%
> ELITE	92%	116%	140%
> HERO	108%	134%	160%

i like this model the best so far, it makes exact sense, there has to be some sort of ladder of strength for all the combinations, i still don't understand this whole redistribution thing, aren't you doing the same thing Xezon? buffs and nerfs to each level are inevitable here

> I hope @MTKing4 could help me with this soon

i've been meaning to get back to testing but i've been to busy in life these couple of weeks, i might be available next week i'll let you know

xezon:
## Patched (Version 3)

> The current PR doesn't address the bug happening with Quads when combined with Propaganda (happens every other moon in team games) and Avengers (never happens)

Fixed.

> or when fighting Hard AI.

Fixed.

> It also has no boost from v0s1 to v0s2.

Fixed.

This is my final result. Changes are implemented and pushed. Change description is updated. Change log documentation left to be added.

All changes:

1. Inverts and redistributes Veteran damages
2. Redistributes Salvage Gun damages
3. Fixes Easy AI and Hard AI damages
4. Fixes Subliminal and Faerie damages

quadcanndmg

commy2:
👍 for fixing the issues with other ROF bonuses. Can't speak for the regular vet/scrap numbers.

MTKing4:
This last suggestion is also fine and can work

One thing to note, v0s2 can beat v1s0, v1s1 and v2s0
And this can be seen on other combinations as well where a scrapped quad can beat a higher vet and lower scrapped quad

I don't really have an issue with that though.

xezon:
> One thing to note, v0s2 can beat v1s0, v1s1 and v2s0

I think that is fine too.

My suggestions for next steps:

1. Finalize documentation, update video and merge this, because this is a good baseline
2. Think about buffing Veteran damages if required

ExiLe-ZH:
Well a pro of this 'nerf' is that Gatling buffs like horde-bonus and other ideas aren't needed anymore, though a con of that is that we can't help making gats better vs USA's.

As for the next steps and updating video, it would be good to include test simulations vs Gatlings, like 10vs10 because it gives much better overview of overall strength and snowball effects than 1v1 test.

xezon:
Can you make test map or test result? I do not want to spend more time at this than I already did.

xezon:
Any more input on this? I would like to wrap this up.

commy2:
This definitely needs field testing and potential tweaks later down the line. We have a tag for that or are all changes monitored like this?

ExiLe-ZH:
MTking and me have been trying to find some free time, but has been little difficult, hopefully we can make those testmaps and videos soon.

xezon:
> We have a tag for that or are all changes monitored like this?

Anything labeled controversial is under heightened scrutiny.

xezon:
Any more input on this? I would like to wrap this up.

MTKing4:
We still have to test this, I'm available this weekend

xezon:
Any more comments this?

ExiLe-ZH:
Mtking been working on test map, we can test in the next few days hopefully

commy2:
> Any more comments this?

Probably should look into the anti air versions too. They don't have any damage regressions in 1.04 (unlike the anti ground versions), but they still likely scale strangely.

xezon:
Hmm that is another can of worms. I would prefer to not touch air weapons for this change, because it would widen the focus of the discussion.

xezon:
Any further comments on this?

xezon:
Exile
> Quad change seems safe so far. Think ok in main for further testing.

xezon:
MTKing
> * Vet 0 Gattling Tanks Always win vs Vet 0 Scrap 0 Quad Cannons
> * Other Scrap 0 matchups are situational, but Quad Cannons win 90% of times
> * Quad Cannons destroy Gattling Tanks in every other combination
> * Results are pretty much identical in Patch and Original 1.04 with the exception that Vet 3 Gattling Tanks can win vs scrapped Quad Cannons, but it is also situational

xezon:
Anything else to add?

commy2:
Should update video in OP.

xezon:
Yes. I will do this last.

xezon:
Documentation added.

xezon:
commy2
> Looks good to me, aside from not handling the AA versions at all.

MTKing
> It is good enough.

xezon:
Video updated. Rationale added. Looks good.

PR #1056, opened 2022-08-30
Remove some reference files

xezon:
This change removes obsolete reference files.

* Closes #1038


commy2:
> [Remove Generals reference files](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1056/commits/65403a622a32d968e495a7f345301b3e0f7be70c)

These should stay. They are occasionally useful.
Instead, "C:\dev\GeneralsGamePatch\Patch104pZH\GameFilesOriginalCCG\Art" can go entirely.

Apart from that, lgtm.

commy2:
Instead, "C:\dev\GeneralsGamePatch\Patch104pZH\GameFilesOriginalCCG\Art" can go entirely.

commy2 (changes requested):
Keep CCG .ini files.

xezon:
I would have liked to put the original Generals ini files into a new repository. What is the reason you would like to keep them here instead of somewhere else? Note: They are entirely unedited.

Regarding Patch104pZH\GameFilesOriginalCCG\Art, the files in there are used for build! So these files are good to stay.

commy2:
> Regarding Patch104pZH\GameFilesOriginalCCG\Art, the files in there are used for build! So these files are good to stay.

I see. Interesting.

> I would have liked to put the original Generals ini files into a new repository. What is the reason you would like to keep them here instead of somewhere else? Note: They are entirely unedited.

Why keep the unedited ZH files then?

xezon:
> Why keep the unedited ZH files then?

The only reason I see to keep these is because we can keep their formatting in sync with our edited ones - for what its worth. Plus they have potentially been referenced in past Discord and Issue conversations, so removing them would break the URLs.

commy2:
> Plus they have potentially been referenced in past Discord and Issue conversation

Yea, the same is true for the CCG files. I'm indifferent on moving them or not, but I think CCG files and ZH INI files should be in one repository.

xezon:
We can leave Generals INI for now if you want. I do not mind either way.

commy2:
Please do. They are very helpful and do not take much space, and it is good to keep them close to the ZH ref files.

commy2 (approved):

PR #1057, opened 2022-08-31
Add Patch104p keyword to Listening Outpost change comments

xezon:
Just updating comments.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
@@ -4792,10 +4792,9 @@ Object Nuke_ChinaVehicleListeningOutpost
     IRParticleSysBone         = IRFX
   End
 
-  ; @bugfix - hanfield
+  ; Patch104p @bugfix hanfield 22/08/2021
   ; Replaced the AssaultTransportAIUpdate module with TransportAIUpdate.
   ; This should make the Listening Outpost behave better in combat.
-  ; 22/08/2021
 
   ;Behavior = AssaultTransportAIUpdate ModuleTag_NewAI
   ;  MembersGetHealedAtLifeRatio = 0.5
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
```diff
@@ -4792,10 +4792,9 @@ Object Nuke_ChinaVehicleListeningOutpost
...
   End
 
-  ; @bugfix - hanfield
+  ; Patch104p @bugfix hanfield 22/08/2021
   ; Replaced the AssaultTransportAIUpdate module with TransportAIUpdate.
   ; This should make the Listening Outpost behave better in combat.
-  ; 22/08/2021
 
   ;Behavior = AssaultTransportAIUpdate ModuleTag_NewAI
   ;  MembersGetHealedAtLifeRatio = 0.5
```
xezon, line 4800: I am curious about this setting. The new `TransportAIUpdate` that we added does not have the `MembersGetHealedAtLifeRatio = 0.5` setting. Does this mean the Outposts no longer heal passengers?

xezon: Seems `TransportAIUpdate` does not support `MembersGetHealedAtLifeRatio` setting.

MTKing4: Attack outposts don't heal passengers afaik, only when dismounted

xezon: Actually they do, same as Troopcrawler, when the infantry is boarded. I just tested.

commy2 (approved):

commy2 (approved):
Could also delete the commented out module. Nothing interesting to see.

PR #1059, opened 2022-08-31
Change: Remove assault behavior from China Troopcrawler

xezon:
This change removes the assault behavior from China Troopcrawler, because it is buggy and exploitable to a certain extend.

The Troopcrawler now can be evacuated manually. The passengers cannot shoot out of the vehicle. Giving attack orders is still possible, so it can drive to attack targets with other units on attack command. The attack audio has been replaced with move audio, because it essentially no longer attacks.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaMiscUnit.ini
@@ -246,7 +246,8 @@ Object ChinaVehicleTroopCrawlerEmpty
   VoiceSelect = TroopCrawlerVoiceSelect
   VoiceMove = TroopCrawlerVoiceMove
   VoiceGuard = TroopCrawlerVoiceMove
-  VoiceAttack = TroopCrawlerVoiceAttack
+  ;VoiceAttack = TroopCrawlerVoiceAttack ; Patch104p @tweak disable because attack no longer does much other than driving
+  VoiceAttack = TroopCrawlerVoiceMove
   SoundMoveStart = TroopCrawlerMoveStart
   SoundMoveStartDamaged = TroopCrawlerMoveStart
   SoundEnter = HumveeEnter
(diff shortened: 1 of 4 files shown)


commy2:
This first https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1060, probably.

Probably should remove weapon. The unit doesn't do anything if ordered to attack after all. Would be similar to Radar Van or Ambulance then. 
Maybe this can be salvaged in such a way that AI can still use it.
Sad to see this go. Maybe it would be better to acknowledge the issues and postpone a proper fix until Thyme?

xezon:
I tested removing the weapon. It has significant disadvantage:

Having Group of units + Troopcrawler can make whole group not be able to attack as long as Troopcrawler is selected. It is similar to USA Ambulance situation, just much worse, because Ambulance at least allows attack targeting in group always, but will stop moving.

Yes #1060 shall be submitted first.

Agree it is not best to remove this feature. But I have also seen other bug with it:
a) Sometimes not all men would disembark on attack order
b) Sometimes not all men would embark on stop order

It appears to have multiple issues here. It would need proper Thyme fixes indeed.

commy2:
> Having Group of units + Troopcrawler can make whole group not be able to attack as long as Troopcrawler is selected.

It only disables force-fire on the ground if Troop Crawler is the highest priority unit selected (i.e. was built first) meaning its portrait is shown. Enemies can still be targeted.

This is the same behaviour as a Supply Truck or Radar Van grouped with the units.

I find the non functional fake weapon more jarring. You click on the enemy building next to you with attack cursor, but nothing happens.

xezon:
So attack on enemy with group of units with and without weapons is always possible?

commy2:
> So attack on enemy with group of units with and without weapons is always possible?

You need at least one unit in the selection that can attack the target. That is also enough to display the Attack Move button.

There are also bugs associated wih this. If you attack a target, the voice line of the highest priority unit will play. So if you select a Chinook + Humvee and click on an enemy, and Chinook portrait is shown, the "Combat Drop" sound will play, even though the Chinook ignores the order.

xezon:
Tracked.

Stubbjax:
> There are also bugs associated wih this. If you attack a target, the voice line of the highest priority unit will play.

I'm pretty sure priority = lifetime.

This reminds me of another issue: Attack-moving does not work while a demo worker has selection priority (different command set?). Can be quite frustrating when popping a tunnel with a worker in the mix and being unable to attack-move. Is this tracked anywhere?

commy2:
For Demo Worker, Advanced Demo Trap shadows Attack Move. This _should_ be indirectly fixed in main branch patch already, because of the streamlining of the Suicide Button. Needs testing.
Should definitely be added to the patch notes though.

commy2:
Tested Demo Worker in multi select scenario, and the Suicide button no longer obscures Attack Move in patch.

xezon:
Close? Or merge?

MTKing4:
Close, leave the issue to thyme

Issue #640, opened 2022-01-19
Reduce effectiveness of USA All-In Humvee rush


xezon:
Explore changes for reducing effectiveness of USA All-In Humvee rush.

How it works:
As USA, build one Supply, one Warfactory and one Barracks. Build 2 Humvees with Rocketmen and an Ambulance. Drive into the enemy China base and kill everything. Works everytime as long as the USA player controls his Humvees appropriately.

Current approach in combating this, is decreasing costs and build times of its cannon fodder units, so it can throw more of them against the Humvee rush and eventually overwhelm them.

Are there alternative solutions to give China a better chance?

### Proposal 1
Missile Defenders in passenger seat have slower firing rate by default. Passenger Training Upgrade in Barracks will unlock regular firing rate. This delays maximum Humvee damage output post upgrade research. Affects both Combat Chinook and Humvee.

### Proposal 2
Reduce Ambulance healing rate.

### Proposal 3
Ambulance does not heal vehicles by default. Ambulance either needs upgrade to heal vehicles or needs Strategy Center to unlock Ambulance that can heal vehicles. Note: There are already 2 different Ambulance models in the game. One of them has a big yellow repair icon instead of a red cross on its roof. Red cross = repair infantry. Repair icon = repair infantry + vehicles.

### Proposal 4
Slightly increase ground attack range of China Gattling Tank and/or decrease China Gattling spin up time. Helps China only. Affects China VS GLA and China vs China too.

### Proposal 5
Add propaganda tower to China Warfactory to heal and buff units in range. This way China can defend better against All-In Rush near its production facility.

### Proposal 6
Make USA Ambulance vehicle repair a manual action, similar to cleaning up toxins. This way repair is more micro intensive and naturally slowed down.

### Proposal 7
Make USA Ambulance auto-repair only while being stationary. This slows down healing naturally.

### Proposal 8
Reduce auto-repair radius of USA Ambulance. This does not prevent effectiveness of small Humvee rush, but effectiveness in big humvee armies.

RisingZH:
Firstly I think addressing no-eco’s should definitely be looked at. Balance aside, it just makes for really boring 3 minute games and is one of if not the worst part of ZH. If USA player micros well they win, if not they lose, the China player might as well be a brutal ai.

The problem with trying to solve it is humvees/Ambo and gats/outposts are such a big part of the game that if you change them at all whilst trying to fix no eco you will change a lot of other matchups/strategies.

[...]

PR #1067, opened 2022-09-01
Emergency Repair Proposal 5, Repair Vehicles In Area Over Time

xezon:
* old PR: TheSuperHackers/GeneralsGamePatch#744

Implementation of proposal 5 for TheSuperHackers/GeneralsGamePatch#707.

After this pull request, the repair effect lingers for 10 seconds (with visible decal). The vehicles are repaired over time instead of in one burst. Upgrading the ability increases how long the effect lasts (to 20 and 30 seconds respectively). Upgrading the ability does not increase the healing speed. The total healing amount only differs due to the time the ability lasts. The radius is increased by +50% (which equals +125% area).

Vehicles can move in and out of the repair area. They are only repaired while inside the area.

https://user-images.githubusercontent.com/6576312/180599717-03c0d0d9-29dd-43f8-bc17-97725ab3f7c3.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -419,6 +419,7 @@ Object GPSScrambler_InvisibleMarker
 
 End
 
+; Patch104p @balance commy2 23/07/2022 Make repair effect linger and heal over time.
 ;------------------------------------------------------------------------------
 Object RepairVehiclesInArea_InvisibleMarker_Level1
   ; ***DESIGN parameters ***
@@ -428,13 +429,13 @@ Object RepairVehiclesInArea_InvisibleMarker_Level1
   Draw = W3DModelDraw ModuleTag_NOTREALLYADRAW
 
     DefaultConditionState
-      Model = None
+      Model = EXAMineGroup
       ParticleSysBone = None GPSRotisserie
       ParticleSysBone = None RepairCloud
     End
   End
 
-  KindOf = NO_COLLIDE IMMOBILE UNATTACKABLE
+  KindOf = NO_COLLIDE IMMOBILE UNATTACKABLE FS_FAKE
 
   ; *** ENGINEERING Parameters ***
   Body = ImmortalBody ModuleTag_01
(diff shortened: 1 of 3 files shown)


commy2:
This is the only change that would make me ever consider chosing ER.

xezon:
We need to test this on Overlords.

commy2:
Made it also repair buildings. Translation missing from Arabic.

xezon:
The level 3 Emergency takes 30 seconds to repair an Overlord half.
The Subliminal Messaging (upgraded) takes 25 seconds to repair an Overlord half.

xezon:
The repair ability for buildings is nice. Makes it more useful, especially for China. But it does not repair much, just this much from 1 health:

shot_20230801_182441_3

Perhaps repair amount should be higher?

xezon:
The problem of increasing healing amount with this particular implementation could be that units could become quite difficult to kill while inside the affect radius. Right now one Gattling Tank is still enough to damage a Humvee with Repair + Ambulance, but increased Repair amount could change this. On the other hand it is nice that it becomes no wasted ability when there are just a few damaged units at the time of use but there is the expectation of more damage to heal over time.

xezon:
I found that repairing of structures does not stack. So if ability repairs structure first, then Dozer does not repair, and vice versa. And Dozer repairs structure much faster, perhaps 3 times as fast, so this repair ability is really not great on structures as is.

shot_20230801_184007_4

commy2:
> The problem of increasing healing amount with this particular implementation could be that units could become quite difficult to kill while inside the affect radius. Right now one Gattling Tank is still enough to damage a Humvee with Repair + Ambulance, but increased Repair amount could change this. On the other hand it is nice that it becomes no wasted ability when there are just a few damaged units at the time of use but there is the expectation of more damage to heal over time.

It's possible to have one module repair vehicles, another repair buildings, where the one that repairs buildings repairs a bigger amount.

xezon:
Repair also does not stack on vehicles. Tested with Ambulance and Subliminal Messaging and repair rate of 10.

This means the Repair ability is useless on a blob of units that have an Ambulance or Prop Tower in the mix.

commy2:
Seems to be an engine thing then. In 1.04, ER is completely swallowed by Propaganda Towers: the entire burst is.

xezon:
> Seems to be an engine thing then. In 1.04, ER is completely swallowed by Propaganda Towers: the entire burst is.

That is even more useless then :D

I think this needs a bug report.

commy2:
Ok, doubled repair rate on buildings. Feel free to tune.

commy2:
> I think this needs a bug report.

https://github.com/TheSuperHackers/GeneralsGameCode/issues/240

xezon:
With fixed stacking I think the effects would be quite nice. However, I do expect problem with Early_Command_EmergencyRepair.

Airforce, Stealth, Nuke and Tank have Rank 1 Emergency Repair. Especially with Airforce this could be problematic, as this could exacerbate problems with strong Humvee Rush (#640).

commy2:
Yeah, could get rid of early version, or distribute it only to faction that need it (SWG?). This would need lots of playtesting though.

xezon:
On the other hand, in normal circumstances the Command Center will be sold, so perhaps will not make a practical difference in early game.

PR #1069, opened 2022-09-01
Fix for ground aligned flares

### LINKED ISSUE #969: Fire storm textures show artifacts
Fire storm textures show artifacts.

shot_20220824_170026_5

shot_20220824_170211_9

xezon:
* Fixes #969

Fixes artifacts in all 7 flare textures.

Optimizes the following ground aligned flares:

* InfernoCannonFlare
* InfernoCannonFlareUpgraded
* MineFlash
* ScudLauncherTransitionLenzflare
* BattleMasterTransitionLenzflare
* BattleMasterTransitionLenzflareSmall
* OverlordTransitionLenzflare
* EMPPatriotLensFlare2

## Original

https://user-images.githubusercontent.com/4720891/186900257-a1873b71-16e4-4f83-9ef0-b9767c9e76a2.mp4

## Patched

https://user-images.githubusercontent.com/4720891/186900281-e87caad2-92ff-489c-a302-0ecb38eba23e.mp4


commy2:
I do understand this fix for Inferno Cannon, because multiple shells may hit the same location. But why is any change necessary to damage transition of vehicles, Overlords, Scud Launchers and battle drones? They don't stack on top of each other.

xezon:
I looked through all flare particles that are ground aligned and these are the ones that appear to be able to stack on each other, either by spawning multiple instances themselves without any positional or rotational randomness or by being able to spawn on top of each other by multiple spawners (e.g. EMP Patriot Missile).

I have not tested them all. Would you need visual confirmation for all of them?

commy2:
Yes.

xezon:
### BattleMasterTransitionLenzflare Original

original

### BattleMasterTransitionLenzflare Patched

patched

xezon:
I looked at video recordings of both original and patched battlemaster flares closely and can conclude that it is a very minor quality improvement. From the screenshots it is not possible to judge properly, neither will be from compressed video.

As for the more blurry flare textures, I have reviewed the edit and reverted the blurring of the flares, as the pristine sharpness of the original texture can be preserved and the DDS export actually yields the same results without any further quality loss. Therefore the only fixes that remain on textures are the rough outline fixes by masking them better.

xezon:
EMP hits appear to look ok originally, because they do not hit perfectly at the same time and therefore there is little overlap. Only the non-ground-aligned effect shows obvious clipping. The patched result looks comparable to original.

### Original EMPPatriotLensFlare2

shot_20220902_142022_2

### Patched EMPPatriotLensFlare2

after

xezon:
My test is concluded.

MTKing4:
To me it looks the patched EMP looks worse than the original, or at least not different.

Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
```diff
@@ -12085,9 +12085,9 @@ ParticleSystem MineFlash
   Gravity = 0.00
   Lifetime = 100.00 100.00
   SystemLifetime = 2
-  Size = 1.00 1.00
+  Size = 1.00 2.00 ; Patch104 @bugfix xezon 26/08/2022 from 1.00 1.00 to not overlap stacked pixelated textures
```
commy2: This does increase their size on average, is that correct?

xezon: Essentially the Mine Flash is growing from [1-2] to [37-43].
We can change this to 0.75 1.25 to follow more the averaging norm of the other changes.

xezon: Fixed.

commy2: Oh, THAT is what SizeRate means. What a terrible name.
Yeah, 0.75-1.25 would make more sense I think.

PR #1071, opened 2022-09-01
Fix: Recover all missing language files: fixes Gamma Terror death, Battlemaster audio lines

xezon:
* old PR: #932

This change recovers all missing language files. This fixes

* Missing China Battlemaster voice in French, German
* Missing GLA Terrorist toxin death visual in Chinese, French, Korean


xezon:
I tested the Battlemaster Audio fix in French. That worked.

As for the Terror W3D, it appears to make no difference. Terror unit just disappears on Gamma spray. No idea what is wrong.

xezon:
I had the missing w3d files copied to the wrong folders previously. Mistake was corrected. Now Gamma toxin death works in Chinese, French, Korean as expected. Originally Terrorist will simply disappear.

shot_20221106_142250_2

PR #1076, opened 2022-09-02
Change: Use Different Explosion Effect For Passive Demogen Destruction Weapon

### LINKED ISSUE #850: No Audio Visual Difference Between Suicide And Regular Death With Demolitions Upgrade.
Background Demogen mechanics:
After Demolitions upgrade, if a Demogen unit is killed, it explodes with a small 50 damage that only hurts enemies and neutrals.
Alternatively, the player can suicide which does 500 instead and also only hurts enemies and neutrals.

The issue from a UI perspective is that there is almost no audio-visual difference between the two explosions. The same particle effects are used and the same explosion sound is played. One difference is that infantry will be flung into the air when suicided, but just collapse as normal when killed.

Since the damage is 50 vs 500, I think there should be better feedback given to the players what happened. I think the explosion should be smaller when killed.

Thoughts?

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/850

Much smaller explosion without the red donut cloud, because the weapon does small damage (wouldn't even two shot a Red Guard!). Most importantly, it no longer plays the Terrorist Suicide sound effect.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -8791,8 +8791,32 @@ FXList WeaponFX_DemoSuicideDynamitePackDetonationPlusFire
   ParticleSystem
     Name = HotPillarArms
   End
+  ParticleSystem
+    Name = DemoTrapLenzFlare
+  End
   ViewShake
     Type = NORMAL
+  End
+  TerrainScorch
+    Type = RANDOM
+    Radius = 15
+  End
+End
+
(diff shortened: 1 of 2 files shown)


xezon:
Nice. Could we have a before and after video for this for safe keeping?

commy2:
# original

https://user-images.githubusercontent.com/6576312/188205462-e4720190-bd09-41e8-bd45-8008be12f28f.mp4

# patched

https://user-images.githubusercontent.com/6576312/188205808-73301c5f-0bfb-4588-a8d6-d916bb6bd962.mp4

xezon (approved):
The new effect also looks quite spectacular. Could also add it to the CrushedDynamitePack and CrushedBikeBomb.

PR #1078, opened 2022-09-02
Fix: Visible Permanent Muzzle Flash When Placing First Fire Base

### LINKED ISSUE #1077: First Fire Base Placement Ghost Displays Permanent Muzzle Flash
https://user-images.githubusercontent.com/6576312/188199245-5db027e3-c3e5-4b4d-ae16-179d915d3f7c.mp4

After the first Fire Base has been placed (even if just as scaffold), the permanent muzzle flash disappears.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1077

This is a hack. The muzzle flash is still shown when firing correctly despite HideSubObject.


xezon:
You write it is a hack. Does this mean there could be unknown side effects?

commy2:
Anything could have side effects in this engine^^

Can't think of anything. The muzzle flash is still correctly unhidden when firing. Nothing else comes to mind that involves the muzzle flash besides actually firing the gun.

xezon (approved):
If it works, it works.

PR #1080, opened 2022-09-02
Fix: Destroyed Listening Outpost Radar Dish Spawns In Air, Has Very Low Weight

commy2:
# Original
https://user-images.githubusercontent.com/6576312/188211721-5f60c456-575a-4a4a-9df9-59a0e4ae918b.mp4

# Patched
https://user-images.githubusercontent.com/6576312/188211725-b15f6518-14fa-423f-a127-118aa15486bf.mp4


xezon (approved):
Never noticed myself ha. Is this 100% ?

commy2:
>Is this 100% ?

Yes. There is some randomness to debris, but it will always spawn above the unit and always fly into SPACE

PR #1085, opened 2022-09-03
Fix: Granger Fails To Call In Carpet Bomber In Generals' Challenge

commy2:
He doesn't have the science that is required since ZH 1.02 patch and thus fails to call it.


xezon:
Is this branch a duplicate of this branch:
https://github.com/TheSuperHackers/GeneralsGamePatch/tree/fix-afg-challenge-carpet-bomber

If yes, then the other should be deleted.

xezon (approved):

PR #1090, opened 2022-09-03
Fix: China Command Center Has Second Clipping Building At Night When Damaged

commy2:
# before
shot_20220903_150126_1

# after 

shot_20220903_152110_1

Note: 'after' is essentially how the building looks already when damaged on day, snow and snow+night maps.

It may be preferable to remove the rogue objects from the model instead.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -353,6 +353,7 @@ Object Boss_CommandCenter
       ParticleSysBone    = Smoke01 SmolderingSmoke
       ParticleSysBone    = Smoke02 SmolderingSmoke
       ParticleSysBone    = Smoke03 SmolderingSmoke
+      HideSubObject      = Object03 Mesh03 ; Patch104p @bugfix commy2 03/09/2022 Fix two buildings in different damage states clipping inside each other.
     End
     ConditionState    = REALLYDAMAGED RUBBLE NIGHT
       Model           = NBConYard_EN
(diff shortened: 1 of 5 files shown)


xezon:
The only art changed for nbconyard in NProject are these 4 textures

Art\Textures\nbconyard2_d.dds
Art\Textures\nbconyard2_ds.dds
Art\Textures\nbconyard2_e.dds
Art\Textures\nbconyard2_es.dds

commy2:
Texture changes won't fix this. It needs model changes in `NBConYard?_DN.W3D`, where ? is a sub-faction specific character.

xezon:
Yes. There is no conyard model change. Just textures.

commy2:
I checked the NProject INI files. The fix is not applied there either. I think this is not fixed in NProject.

A similar PR is this: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/133

Instead of INI work-arounds, the model was changed. From my understanding, no HideSubObject work-around could have been applied there, because the issue was with a texture, not with rogue model parts.

I think this PR is most reasonable way to fix this issue.

xezon (approved):

xezon:
Enlima29
> there are 3 objects that need to be hidden in the damaged night condition: Object03, Object04 and Mesh03. For the snow damaged condition, sadly these sub-objects were welded into the main building so actual model editing need to be done (unless you want to salvage the parts from snow night damaged model and hide the night bits to look like the snow damaged model)

commy2:
While the clipping happens on the SNOW+DAMAGED CC's and OBJECT03 is merged into the rest of the model and thus cannot be hidden by INI, at least the team colour is working correctly:

shot_20220904_110745_2


This and the lack of a model makes me think this fix is good enough, unless somebody else wants to have a go at it.

xezon:
Fix looks good enough.

PR #1091, opened 2022-09-03

### LINKED ISSUE #300: Inf China Minigunner in Helix has extended firing range when attacking air unit together with a grounded Minigunner

xezon:
* old PR: #342

- Fixes #300

ZH 1.04

After patch:

Drawback:

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -24,7 +24,7 @@ https://github.com/commy2/zerohour/issues/206 [MAYBE]                 Combat Chi
 https://github.com/commy2/zerohour/issues/205 [DONE]                  Listening Outpost Damaged Smoke Effect Sticks After Repair
 https://github.com/commy2/zerohour/issues/204 [WONTFIX]               Fire Base Cannot Order Passengers To Attack Airborne Targets
 https://github.com/commy2/zerohour/issues/203 [MAYBE][NPROJECT]       Overlord With Bunker Cannot Order Passengers To Attack Airborne Targets
-https://github.com/commy2/zerohour/issues/202 [MAYBE][NPROJECT]       Helix With Bunker Cannot Order Passengers To Attack Airborne Targets
+https://github.com/commy2/zerohour/issues/202 [DONE][NPROJECT]        Helix With Bunker Cannot Order Passengers To Attack Airborne Targets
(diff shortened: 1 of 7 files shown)


xezon:
xezon:
> It was raised as highly controversial by Jundiyy in regards to China Helix if I remember correctly. In Original Game, Helix cannot target Air units, but passengers will attack when getting close to Air target. Changing this would make Helix fights play out differently he says. For reference, with Combat Chinook it is possible to target Air units. To me this is not balance question, but game controls bug. As of right now, the way to circumvent this control issue, is to group Helix with Ground unit. Then Helix can attack Air units as expected.

Rising:
> Yeah seems clear that a unit that can shoot air should be able to target air, and it is illogical that it can not. And I agree that this is a clear case of fix the bug and then worry about balance (with that said I do not see how this could cause any balance issues anyway).

xezon:
> Minigunner will need consideration if Helix attacking is fixed. Because Minigunner has incredible firing range in Bunkers. This was already experimented with and apparently is is possible to reduce firing range inside Helix. In Original Game, Minigunner will start firing on close distance only, but will then keep firing when flying away.

Exile:
> It could change Helix vs Helix micro. I have seen in a video how rockets track forever after manually targeting (not sure if there were other changes involved), but if that is the case then lix battles are more point and click rather than the current situation where players have to move smart to target and dodge rockets. Also the Minigunner Helix is a big one indeed, if target range is too large it would not be fun for USA. USA would have barely any chance of getting away with Chinooks.

xezon:
## Original

GARRISON gives 33% range bonus
S&D (Search and Destroy) gives 20% range bonus

| Object                | Anti Air Attack Range  |
|-----------------------|------------------------|
| Grounded TankHunter   | 175                    |
| Grounded Minigunner   | 350                    |
| GARRISONed TankHunter | 232 (+57)              |
| GARRISONed Minigunner | 465 (+115)             |
| Gattling Tank         | 350                    |
| Gattling Cannon       | 400                    |
| Quad Cannon           | 350                    |
| Humvee TOW            | 320                    |
| Humvee TOW + S & D    | 384 (+64)              |

kABUSE1:
Please keep in mind that mini gunner helixes have insane range when ordered to attack air by having a ground minigunner in the selection with them. The range is so large that a king raptor cannot even shoot them down anymore.

Having to manually swap the mini gunners by getting close to enemy air units makes a huge difference, especially because they switch back whenever you get close to ground any enemy ground unit again.

PR #1094, opened 2022-09-03

xezon:
* old PR: #394

ZH 1.04

patch:

### DIFF
--- a/Patch104pZH/Design/Tasks/commy2_tasks.txt
+++ b/Patch104pZH/Design/Tasks/commy2_tasks.txt
@@ -77,7 +77,7 @@ https://github.com/commy2/zerohour/issues/145 [DONE][NPROJEC!]        Toxin Gene
 https://github.com/commy2/zerohour/issues/144 [IMPROVEMENT]           Overlord Does Not Fit Into Chinook But Only Takes 3 Slots Inside Helix
 https://github.com/commy2/zerohour/issues/143 [DONE]                  GPS Scrambled Rangers And Red Guards Remain Stealthed While Capturing Buildings
 https://github.com/commy2/zerohour/issues/142 [IMPROVEMENT][NPROJECT] Fire Base Shoots At Awkward Angle
-https://github.com/commy2/zerohour/issues/141 [PENDING][NPROJECT]     Toxin General RPG-Trooper Shoots Stinger Missiles
+https://github.com/commy2/zerohour/issues/141 [DONE][NPROJECT]        Toxin General RPG-Trooper Shoots Stinger Missiles
 https://github.com/commy2/zerohour/issues/140 [DONE][NPROJEC!]        Gattling Helix Sometimes Fires At Airborne Units
 https://github.com/commy2/zerohour/issues/139 [DONE][NPROJEC!]        Dummy Weapons Cause Notifications And Hit Effects
 https://github.com/commy2/zerohour/issues/138 [IMPROVEMENT][NPROJECT] Nuke Battlemaster Uses Normal Battlemaster Button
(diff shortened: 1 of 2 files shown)


xezon:
# RPG vs Toxin RPG Weapon

https://user-images.githubusercontent.com/4720891/184421124-0bb4e984-efda-4a53-8b02-50f3dd626aa9.mp4

https://user-images.githubusercontent.com/4720891/184421141-a8689573-d49b-4fe5-baf9-5d9e28f04e1d.mp4

# RPG and Toxin RPG vs Plane

https://user-images.githubusercontent.com/4720891/184421273-c55df056-0d36-48b2-9983-595e09854229.mp4

https://user-images.githubusercontent.com/4720891/184421289-1c848293-2aca-4617-951f-a6e52c436129.mp4

# RPG and Toxin RPG vs Helix

https://user-images.githubusercontent.com/4720891/184421816-bbe4d80c-22ee-43e0-b659-f7881d65d944.mp4

https://user-images.githubusercontent.com/4720891/184421840-6c31c812-6b36-4571-8414-bc71268e2e77.mp4

# Conclusion

Both RPG Troopers have same firing interval, but the Toxin Stinger RPG will always hit quicker. Against planes, the Toxin RPG will always hit twice, but RPG only ever once on a straight fly over. Against Helix, both RPG will deal about the same damage. In the video, the Toxin RPG shoots 1 more time, but this should be just coincidence depending on Helix route.

The advantage is 2 fold:
1. Toxin RPG Tracking follows planes far away
2. Toxin RPG flies faster and can kill targets quicker

xezon:
So what shall we do with this then? Close?

MTKing4:
I think there was some agreement to reduce tracking interval?
I guess that means they will still use stinger missiles so you could close this

xezon:
Jundiyy:
> I suppose ground battle wouldn't make a massive difference, other than vs Buggies and Technicals. Can we agree to like half tracking? lol. I would still like the speed of the rocket to stay the same. Plus damage. Tracking/life time reduced by half?

ExiLe-ZH:
Would be fine with me to keep the speed boost.

xezon:
Difference in numbers:
```
Object TunnelDefenderMissile
    TryToFollowTarget = Yes
    FuelLifetime = 1250
    InitialVelocity = 150
    IgnitionDelay = 0
    DistanceToTravelBeforeTurning = 20
  End
End

Object Chem_StingerMissile
    TryToFollowTarget = Yes
    FuelLifetime = 5000
    InitialVelocity = 200
    IgnitionDelay = 0
    DistanceToTravelBeforeTurning = 5
    DistanceToTargetForLock = 100  ; If it gets within 100 of the target, it kills the target.
  End
End
```

xezon:
```
Object MissileDefenderMissile
    TryToFollowTarget = Yes
    FuelLifetime = 3000
    InitialVelocity = 150
    IgnitionDelay = 0
    DistanceToTravelBeforeTurning = 3
  End
End

Object TankHunterMissile
    TryToFollowTarget = Yes
    FuelLifetime = 1250
    InitialVelocity = 150
    IgnitionDelay = 0
    DistanceToTravelBeforeTurning = 20
  End
End
```

MTKing4:
This is Horrible, why should tox RPS have 5x times the fuel time of all other rockets + Extra speed? Also the MD has double why is that?

I think fuel time should be reduced, maybe to 3000 like the MD, and leave the speed the same

xezon:
I suspect Missile Defender Missile has longer life time for Search And Destroy range bonus + Laser Guided range bonus.

Stubbjax:
It seems like the rockets are keeping their purple trail despite it now having no meaning?

(I do like the look of that `; ProjectileDetonationOCL = OCL_PoisonFieldUpgradedSmall` as some form of counterbalance to the nerf, but it would most likely be too different from 1.04.)

ExiLe-ZH:
Purple trail should only show after Gamma upgrade?

Stubbjax:
> Purple trail should only show after Gamma upgrade?

It is slightly blue-ish pre-Gamma in 1.04.

commy2:
> It seems like the rockets are keeping their purple trail despite it now having no meaning?

There is precedence for this. Demo RPG (as well as Marauder and Rocket Buggy and SCUD Launcher) have red explosions, despite being identical in damage and radius.

PR #1097, opened 2022-09-04
Fix: Units Shoot From Center Of Palace In Certain Model States

### LINKED ISSUE #471: GLA Palace with Fortified Structure upgrade is missing many fireports on back and sides
GLA Palace with Fortified Structure upgrade is missing many fireports on back and sides. This issue has been fixed in NProject Mod.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/471

- Even though there are 24 different models that have to support firepoints on the Palace, there are only 3 different setups for firepoints.
- (1) One of them has all 20 firepoints defined correctly
- (2) One of them is missing firepoint 9 and therefore unintentionally disables fireports 9 through 20.
- (3) One of them is missing firepoints 3, 4 and 7 through 12, which disables firepoints 3 through 20.

This means that the fix that requires the least amount of individual changes can be done with a dummy draw module and two custom models that add all the disabled firepoints (setups (2) and (3)).

Dummy model FP9 contains all firepoints from 9 upwards. For this to work correctly, the points are renamed 1 to 12.

Dummy model FPX contains all firepoints of setup (1), except firepoints 1 and 2. The existing firepoints are named 1 to 18 for this to work correctly.

This fix was tested on normal and snow maps with and without Fortified Structures. I double checked and none of the firepoints were used by multiple units.

Note: The firepoint state is cached individually for each ingame object at the moment the GARRISON flag is added to the building. This complicates testing and may make the issue fail to reproduce in 1.04 if done wrong.

An alternative fix would involve changing the 17 models that are missing the firepoints. Because there are only three different firepoint setups, and only two of them are broken, the same effect can be achieved in the way this PR does.

The fix applied in NProject would require more extensive changes to the INI file. GARRISON and USER_1 (Fortified Structures) states are handled with separate objects that add sandbags and barbed wire and certain parts of the model are hidden or shown via Show/HideSubObject in INI. The bug is avoided in NProject by never displaying any of the EALA made models that contain firepoints.

I think this is a simpler fix for a patch and does not require 17 or 6, but only 2 models (that are also smaller in filesize individually).

|State                  | Model  | Available Ports    | Missing Ports
|-----------------------|--------|--------------------|-------------|
|Day                    | G      | 1-20               | none        |
|Night                  | NG     | 1-20               | none        |
|Snow                   | SG     | 1-8                | 9           |
|Snow + Night           | GNS    | 1-8                | 9           |
|Damaged                | DG     | 1-20               | none        |
|Damaged Night          | DGN    | 1, 2               | 3, 4, 7-12  |
|Damaged Snow           | DSG    | 1-8                | 9           |
|Damaged Snow + Night   | DGNS   | 1-8                | 9           |
|Rubble                 | EG     | 1-20               | none        |
|Rubble Night           | ENG    | 1-20               | none        |
|Rubble Snow            | ESG    | 1-20               | none        |
|Rubble Snow + Night    | ENSG   | 1-20               | none        |

Note: Fortified Structures versions all have the "missing 3, 4, 7-12" setup!


xezon:
Very good investigation and interesting fix. How much more complex is the NProject INI change in comparison?

The advantages of NProject fix are
* Reduced model count makes it easier to tweak Palace models for other errors
* Unified setup with NProject enables to benefit from other NProject model fixes
* Less model variants leave smaller memory footprint in game process

commy2:
> Reduced model count makes it easier to tweak Palace models for other errors

This is true. However, if some other issue with the Palace is found and it is determined that the best approach to fix it is to reimplement the NProject Palace or something similar to it, then this PR could be reverted. I don't foresee this happening at this point.

> Less model variants leave smaller memory footprint in game process

There is one less draw module required for this fix than the NProject fix. 
I would say the NProject Palace has a simpler implementation than the EALA version, while the EALA version has less overhead than the NProject version. I don't think this alone is enough reason to favour EALA however, because premature optimization etc. 
I prefer the EALA setup for now simply because it is the current state of 1.04, and acceptable with the additions of this PR.

Here is a diff 1.04 with NProject:

https://gist.github.com/commy2/db4969d79f13994b8c76b09fefc26369/revisions

Sadly, because of the extensive formatting changes in NProject, the diff is not very clean. The "gist" (hue) of it is the following:

- NProject never uses EALA GARRISONED models, because they may or may not miss firepoints. Using them will break it in one way or another without additional fixes like in this PR.
- Instead, firepoints are implemented using a single dummy model (similar to those two used in this PR)
- To make sandbags and barbed wire show up in all necessary combinations of weather, Fortified Structures and garrsion state, 6 models are used that display all additions to the Palace model, but not the building itself (extra draw module).
- These additions are further tweaked as necessary by Show/HideSubObject

One could think that just applying the firepoint dummy model from NProject would solve this as well. However, this only works in conjunction with the new sandbag and barbed wire setup, because there still would be the firepoints from the EALA models that contain duplicate firepoints and would result in two units essentially sharing one position.

While this "sandbag and barbed wire setup" of NProject is much better than whatever EALA cooked up, the changes as demonstrated are rather extensive compared to this PR and may or may not alter the model states in undocumented ways.

MTKing4:
## Day
Screenshot (360)


## Night
Screenshot (359)

## Snow
Screenshot (361)

## Snow Night
Screenshot (362)

Everything Seems to be working Correct
Back and side Fireports work as well, and they look Sick actually! Good Job on that

commy2:
To clarify, was this a multiplayer test?

MTKing4:
No, i asked you earlier why does it need to be tested on multiplayer you haven't responded so i just did single player.

xezon:
Perhaps there is worry of mismatch?

MTKing4:
We can test multiplayer next then

MTKing4:
Multiplayer Test, Everything Working as it should.

## Day
Day



## Night
Night


## Snow
Snow


## Snow Night
Night Snow

xezon:
On snow, did you test the damaged model? That is where the issue is according to the table above.

MTKing4:
Did not, only Snow day? And only damaged without reallydamaged? Also needs multiplayer test?

ExiLe-ZH:
wow this is pretty cool.

does it affect the effective range though?

commy2:
> does it affect the effective range though?

It doesn't affect range, however, projectiles will spawn closer to the enemy thus leaving less time to avoid them / zap them / deflect them etc.

I don't think it is a problem though. It already works as shown above in 1.04 on all maps, except snow maps as long as the Palace was garrisoned before it was damaged or Fortified Structures was upgraded. Only damage states and Fortified Structures break (some) hard points.

> Did not, only Snow day? And only damaged without reallydamaged? Also needs multiplayer test?

Snow Day and Snow Night tests already prove that the concept of the fix works in multiplayer. You can test the other configurations as well, but I already did that anyway in SP. When testing, make sure to damage / upgrade the building before it was garrisoned for the first time. The fire spots are cached on per object basis, which may obscure the issue.

1. build Palace
2. damage to appropriate state
3. upgrade Fortified Structures (optional, only if that is the state you want to check)
4. then put units inside and fire on targets

MTKing4:
Ok I'll test that next

xezon (approved):
Looks good.

PR #1099, opened 2022-09-04
Fix: Rocket Buggy Missile Deleted Before Reaching Reinforcement Pad At Max Distance

xezon:
* old PR TheSuperHackers/GeneralsGamePatch#929
* Fixes TheSuperHackers/GeneralsGameCode#142
* Closes TheSuperHackers/GeneralsGamePatch#1100
* Closes https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1101

1850 (+1 frame) wasn't enough to make all rockets hit, only half of them did.

PR #1100, opened 2022-09-04
Fix: Rocket Buggy Missile Deleted When Running Out Of Fuel

### LINKED ISSUE #141: GLA Rocket Buggy missiles vanish ungracefully in long distance

### LINKED ISSUE #142: GLA Rocket Buggy Can't Damage Reinforcement Pad From Maximum Distance

### LINKED ISSUE #874: GLA Rocket Buggy missiles vanish ungracefully in long distance
GLA Rocket Buggy missiles vanish ungracefully in long distance. Perhaps this can be made better. It seems one issue here is that the explosions to the ground are barely audible, maybe because the vehicle hit sound takes priority.

https://user-images.githubusercontent.com/4720891/184429594-90f0327d-353f-4c84-b463-9d71f9ba1ea3.mp4

### LINKED ISSUE #894: GLA Rocket Buggy Can't Shoot Reinforcement Pad From Maximum Distance

+ Moving A bit Closer will get it to shoot
+ Force Firing on the Edges will naturally make buggy move Closer therefore Damage the Target
+ Problem is on All maps, Enemy, Player, or Netrual Pad, Force firing and Regular Targetting all give the same Result

xezon:
* old PR TheSuperHackers/GeneralsGamePatch#931
* Fixes TheSuperHackers/GeneralsGameCode#142
* Fixes TheSuperHackers/GeneralsGameCode#141
* closes TheSuperHackers/GeneralsGamePatch#1099
* Closes https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1101

`MissileAIUpdate` module has an optional `DetonateOnNoFuel` token. When set to yes, the missile will no longer be deleted after `FuelLifetime`, but instead killed with `DETONATED` death type. This triggers the damage and effects from `BuggyRocketWeapon`, which otherwise would not happen.

Benefits: 
- no extended range like the other fix
- fixes the particle issue, which would persist with the other pull request

Drawbacks:
- stray Buggy Missiles that would previously be outrunned now hit unsuspecting units that happen to be where the missile runs out of fuel

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -3019,8 +3019,12 @@ Object RocketBuggyMissile
   Behavior = PhysicsBehavior ModuleTag_06
     Mass = 1
   End
+
+  ; Patch104p @bugfix commy2 19/08/2022 Make Buggy rocket explode when running out of fuel.
+  ; This will enable the rocket to hit Reinforcement Pads and display particle effects instead of vanishing.
   Behavior = MissileAIUpdate ModuleTag_07
     TryToFollowTarget               = Yes
+    DetonateOnNoFuel                = Yes
     FuelLifetime                    = 1800
     InitialVelocity                 = 150                ; in dist/sec
     IgnitionDelay                   = 0


xezon:
Splash damage consideration:

https://user-images.githubusercontent.com/4720891/199596876-8b08a8a2-4d42-439e-a957-e34fef9c0504.mp4

commy2:
This would mean only secondary damage is dealt vs Reinforcement Pads at max range if everything works as I think it does. Which makes this only a partial fix.

xezon:
> This would mean only secondary damage is dealt vs Reinforcement Pads at max range if everything works as I think it does. Which makes this only a partial fix.

I tested. Damage from shooting close and damage from shooting at max range with this change deals identical damage. This is a working fix.

xezon:
I tested outrunning Rocket Buggy and I am pleased by how it plays out. It looks plausible and it only applies a little splash damage (25%) if it does hit. So yes on paper Rocket Buggy is better with this change, but it is a marginal buff only and for the majority of battle situations it is expected to be inconsequential.

I do like this change much better than the alternatives. It fixes 2 issues.

https://user-images.githubusercontent.com/4720891/210646011-a6c57110-fd36-4315-b4d3-8aed9f5266d1.mp4

MTKing4:
It does look fine seeing it now, yes it's a little buff, but it always looked odd before when missiles reach running enemy units but do no damage, i guess one can nerf it some other way later on if proven too good

xezon:
So do we want to go with this change?

ExiLe-ZH:
Can't we make out of fuel explosions harmless? 

Cause this will change battles quite a bit, the splash damage makes it impossible for the enemy to flee with Vees or his own Buggies without causaulties, if the chasing player targets a unit that isn't the last one.

xezon:
> Can't we make out of fuel explosions harmless?

That is the status quo. Rockets vanish.

> Cause this will change battles quite a bit, the splash damage makes it impossible for the enemy to flee with Vees or his own Buggies without causaulties, if the chasing player targets a unit that isn't the last one.

This is correct. There is a penalty to targeting units further away however: The Rocket Buggy has to be standing closer to the enemy forces, and eventually will not fire early enough, whereas targeting the closest enemy units has the best chance for quicker kill due to closer initial proximity to begin firing with. It is a trade off. This change certainly is a net positive for Rocket Buggy, but it is a highly situational buff.

For most common practical purposes, the performance will be comparable with original.

https://user-images.githubusercontent.com/4720891/210810957-2ca608a5-6e9c-4554-b766-5a8f82fe9a46.mp4

MTKing4:
>  generals.2023-01-05.15-59-04-13.mp4

Here we can see the direction impact of of missiles explosion damage, it killed units on its path, so targeting further units will be the new preferable approach for players when Targetting enemies giving higher chances of hitting something.

I think we have two options here
We either

1. Merge this change temporarily Until we have a better one with thyme
2. Close PR entirely and wait for a thyme fix

xezon:
> new preferable approach for players when Targetting enemies giving higher chances of hitting something.

The downside to attacking units further away is that attacking Rocket Buggy eventually takes longer to move into position and initiate shooting. So in no way it is guaranteed that hitting further away targets is the recommended choice. It is situational.

MTKing4:
I beg to differ, players already micro them back when buggies move too close as buggies tend to do that a lot, it's an essential skill with buggies, without it you won't be a good buggy player

I think we should merge and see where this goes.

xezon:
There is nothing you should have to disagree with. If a Target A is closer than Target B and the time to move into firing position is meant to be kept to a minimum, then Target B is not a more attractive target than Target A, because Target B would require more distance and time to move into firing position.

ExiLe-ZH:
Buggies are pretty fast so with buggy vs buggy wars the lead distance is quite big, it could indeed be risky to target the first one in order to hit the last one.

Maybe with it's different with Vees since they're slower? What is the lead distance with them?

MTKing4:
> There is nothing you should have to disagree with. If a Target A is closer than Target B and the time to move into firing position is meant to be kept to a minimum, then Target B is not a more attractive target than Target A, because Target B would require more distance and time to move into firing position.

that is fine most of the time because buggies have large range as it is, so if you're going to higher chances of damage by shooting further units without taking damage yourself you will take it every single time, the only issue with that is if the enemy player notices it and moves that single unit back so that you get closer to his front units and die, but that can also be avoided if the player is paying attention

> Maybe with it's different with Vees since they're slower? What is the lead distance with them?

i do think it works best with vees, yes

xezon:
If Rocket Buggies attempt to target Humvees at the end of the pack then they run risk getting shot at by Humvees. Humvee Group can use X to scatter and then the Rocket Buggies could die as the closest Humvees come closer.

ExiLe-ZH:
> If Rocket Buggies attempt to target Humvees at the end of the pack then they run risk getting shot at by Humvees. Humvee Group can use X to scatter and then the Rocket Buggies could die as the closest Humvees come closer.

My point is, if the lead distance isn't as big with vees, maybe the buggies can target the second last vee so they can keep a greater distance?

xezon:
If Rocket Buggy targets a unit that stands behind another then naturally the attack distance is not maxed out. The closest unit then has a better chance of charging the Rocket Buggy, because it is not the attack target of said Rocket Buggy.

ExiLe-ZH:
Not if it dies ^^

But yes, I suppose it could be microed to both dodge splash damage and potentially pick off buggies.

xezon:
## Targeting back Humvee

https://user-images.githubusercontent.com/4720891/211174340-0db9700c-7d0c-4d8c-ae5b-474924e6d1ca.mp4

## Targeting front Humvee

https://user-images.githubusercontent.com/4720891/211174355-d4edfa21-6484-4c88-8363-c94379c97128.mp4

ExiLe-ZH:
Distance to vees (when targetting the first one) seems alot safer than vs other buggies. Too much of a buff? Not sure, got to be tested in actual games, hard to tell with just these tests.

MTKing4:
> ## Targeting back Humvee
>  generals.2023-01-08.00-44-21-12.mp4 
> ## Targeting front Humvee
>  generals.2023-01-08.00-32-32-60.mp4

That's what i was talking about, bingo! 
Yes, clearly this will be taken advantage of in games, it's a considerable buff for sure, you couldn't do that before, getting extra damage when you should be missing

but it might be a little amplified with your example here, and players can adapt to react properly, so it doesn't automatically mean it's OP now

Stubbjax:
I really don't like the way the missed rockets are exploding like that. It makes it seem like you're hitting something when you're not, and I can't recall any other rockets behaving this way. I don't know of a better solution, but this one definitely feels weird and not very 1.04-like.

xezon:
USA Missile Defender rockets fall to the ground and explode there as far as I am aware. There is precedent to this behavior.

To me the disappearing rockets of Rocket Buggy, Comanche look odd.

Judging by the last test videos, it indeed looks like a tangible buff. Rocket Buggy eventually will be able to unleash more damage when rockets explode more often. This change can be considered risky, given the monumental strength of Rocket Buggy in Original Game. Perhaps this change is only ok if China and USA are more potent to deal with it too.

ExiLe-ZH:
I know I would much prefer it if it was just a particle effect, so that game balance doesn't change in any way. Maybe upkicking dirt would look better, so there's a distinct visual difference between hitting and missing?

There's not a single balancing reason to implement this. The sole reason to fix it is because disappearing rockets looks odd right? Which I do agree with btw.

xezon:
* Fixes TheSuperHackers/GeneralsGameCode#142
* Fixes TheSuperHackers/GeneralsGameCode#141

ExiLe-ZH:
> * Fixes TheSuperHackers/GeneralsGameCode#142

Well ok that's a small issue, but has such small implication on gameplay compared to the splash damage introduced by the fix.

xezon:
Close this?

ExiLe-ZH:
Well that's a little fast.

> * Fixes [GLA Rocket Buggy Can't Damage Reinforcement Pad From Maximum Distance #894](https://github.com/TheSuperHackers/GeneralsGameCode/issues/142)

Can't this be fixed by making the pad's hitbox larger?

> * Fixes [GLA Rocket Buggy missiles vanish ungracefully in long distance #874](https://github.com/TheSuperHackers/GeneralsGameCode/issues/141)

And visually fix this by adding harmless particles?

xezon:
I do not know how.

ExiLe-ZH:
Maybe Thyme, useful to keep open until we know/can do more?

xezon:
I would like to reduce Pull count. So close is preferred. There are still 2 open Issue reports.

xezon:
We do not want to buff Rocket Buggy.

PR #1101, opened 2022-09-04
Fix: Rocket Buggy Missile Times Out Before Reaching Reinforcement Pad

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1099
- close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1100

This removes the NO_COLLIDE flag from Reinforcement Pad (and RepairBay), which makes Rocket Buggy Missiles collide with the borders of the building and prevents them from timing out.

HOWEVER, this means that the parachuting vehicle will avoid the building before landing. Example of this bug from NProject:

https://user-images.githubusercontent.com/6576312/188304421-864536f8-b8d8-400b-872e-a77a1ef7db5f.mp4


xezon:
This fix breaks parachute movement.

PR #1102, opened 2022-09-04
Change: Make Lotus Hack Abort When Out Of Range

xezon:
ref https://github.com/xezon/GeneralsGamePatch/issues/415#issuecomment-931352064

Demo:

https://user-images.githubusercontent.com/6576312/135530330-8b70110b-f7c0-45dd-973a-f4fc2f582e0c.mp4


xezon:
I suggest we test Lotus Vehicle Hack vs fast vehicle evasion - most notably GLA Technical. Then see how far away the Hack can actually succeed and if that makes sense.

MTKing4:
I think this is a bad idea, nerfing the lotus who can't even target vehicles, i think we should leave it as it is

ExiLe-ZH:
I agree with MTking, I don't see a reason to do this either.

PR #1103, opened 2022-09-04
Add: Fire Visuals For Aflame Units

xezon:
* old PR #526
* ref #482

- [x] USA vehicles
- [x] China vehicles
- [x] GLA vehicles
- [ ] Civilian and misc vehicles
- [x] ~~buildings?~~ Buildings already handle AFLAME status via FireS/M/L hard-coded bones.


xezon:
group of aflame quads performance test

https://user-images.githubusercontent.com/83613147/136849408-17a321b2-4c63-4bf8-b2bc-f693efec330e.mp4

xezon:
I tested it on Quad. My thoughts:

* Idea is good
* Fire effects last too long
* Small fire balls do not look good when driving
* Multiple particles can be costly

I suggest to

* Reduce particles count
* Reduce particles life time
* Use larger particles to blend better when driving

Perhaps use a bigger fire and smoke at the center of the vehicle, or one at the front + back.

commy2:
> Fire effects last too long

The effect is removed immediately when the vehicle stops losing health.

xezon:
Ah okay, when it is linked to actual damage duration, then that duration makes sense. Is quite long.

commy2:
Done for the base factions.

We need a map with all the civ vehicles (Crush test has no player controllable flamer).

Buildings already burn when damaged and AFLAME: engine creates larger fires and less smoke at FireS/M/Lxx memory points while burning.

Base defenses lack these, so I recommend to at least add the flame to Tunnel/Stinger/Patiot/Fire Base/Bunker/Gattling/Prop Tower and Demo Trap.

xezon:
Would it perhaps be possible to put the flames at custom positions on the chassis so that the fire does not spawn from below the vehicle? The flame underneath vehicles does not look optimal in my opinion.

shot_20230812_183708_3

commy2:
z-Positions of the fire now editable in INI. Should look better now. Can be fine tuned for each unit.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -5292,6 +5292,17 @@ Object AirF_AmericaVehicleHumvee
...
   End
 
+  Draw = W3DModelDraw ModuleTag_Aflame
+    MinLODRequired = MEDIUM
+    DefaultConditionState
+      Model = NONE
+    End
+    ConditionState = AFLAME
+      Model = FireSource
+      ParticleSysBone = Source05 AflameVehicleMedium
```
commy2: SourceXX where XX is z position in feet. Range [00-20] where 00 is the floor.

xezon: Nice. Perhaps call the bones X00Y00Z00-X00Y00Z20 or Center00-Center20? This would communicate their position more clearly.

We need to check that carrying these extra 20 bones per each object is ok.

commy2: The bones only exist when AFLAME. No model otherwise.

commy2: Renamed to CenterZxx. It's supposed to mean the center x,y at z=...

xezon:
The fires are too intense for my liking. I have made a pass on them.

I have used InfernoCannonFire and DragonTransitionFire as references.

## Before

shot_20230824_165825_3

## After

shot_20230824_181444_1

xezon:
Flamed aircraft appears to produce no fire

shot_20230824_170011_4

commy2:
I haven't added it to aircraft or civilian cars yet. It may also need to be added to base defenses.

PR #749, opened 2022-07-23
Anthrax Bomb Proposal 7, Allow Flame Weapons To Clear Poison Clouds

commy2:
Some weird idea to try regarding https://github.com/TheSuperHackers/GeneralsGamePatch/issues/55.

This makes FLAME weapons (Dragon Tank, Inferno Cannon, Mig) clear up poison fields. It does not make it clear up radiation fields.

Issue #55, opened 2021-08-27
Anthrax Bomb and Toxin Tractor Contamination can be cleared by enemy Poison and Radiation Clouds


commy2:
When the enemy drops an Anthrax bomb, force fire the ground near with a Toxin Shell Scorpion or Nuke Battlemaster to delete it.

### Proposal 1
Stacked poison no longer disappears.

### Proposal 2
Stacked poison only disappears for ally attacks.

### Proposal 3
Anthrax bomb spawns multiple puddles and requires more shots at different locations to be cleared.

### Proposal 4
If exploit is removed entirely, then reduce Anthrax Bomb radius to mitigate impact on gameplay.

### Proposal 5
If exploit is removed entirely, then reduce Anthrax Bomb poison duration to mitigate impact on gameplay.

### Proposal 6
If exploit is removed entirely, then reduce Anthrax Bomb poison damage to mitigate impact on gameplay.

### Proposal 7
Allow China to remove toxins with fire if we keep this mechanic to keep it consistent across factions.

### Proposal 8
Anthrax Bomb puddle requires more damage to die. (Done)
* #2023

commy2:
Why is this even a thing?

This is done to prevent poison clouds from stacking, but this also makes it so the player can just delete enemy large poison fields by creating small ones in place. 

What is my suggestion?

Ideally poison fields could only clear up allied poison fields. That way stacking is prevented for the most part without being able to overwrite enemy poison attacks altogether. Implementation can be seen here: https://github.com/commy2/zerohour/commit/c5ba420cd7a7d0b0a2db1cf157ccc0510cabedf7

xezon:
I highly prefer a clean fix for this, then adjust balance if necessary. The current way the anthrax can be removed with 1 shot is not satisfying behavior and does not make the gameplay better.

xezon:
Added proposals to opening post. I still think clean fix is best and then mitigate the inherent buff of Anthrax Bomb by tweaking it. Meaning Proposal 1+4 or 1+5 for example.

Proposal 2 I think is confusing.

Proposal 3 is better than status quo, but not really a good solution. Radiation also does not disappear on force fire. So why should Toxins?

commy2:
[...]

PR #1113, opened 2022-09-04
Add: Decontamination Ability To Toxin Tractor

xezon:
* old PR #972

Idea from [MTKing4](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/55#issuecomment-1193170691).

This adds a decontamination ability to the Toxin Tractor.

Combined with https://github.com/TheSuperHackers/GeneralsGamePatch/pull/749 it would mean that all 12 factions have a cleanup unit to counter the Anthrax Bomb.

PR only works with vGLA Tractor.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -1920,6 +1920,21 @@ CommandButton Command_GLAToxinTractorContaminateGround
   UnitSpecificSound       = ToxinTractorVoiceModeContam
 End
 
+CommandButton Command_GLAToxinTractorCleanupArea
+  Command                 = FIRE_WEAPON
+  SpecialPower            = SpecialAbilityToxinTractorCleanupArea
+  WeaponSlot              = TERTIARY
+  Options                 = NEED_TARGET_POS OK_FOR_MULTI_SELECT ; CONTEXTMODE_COMMAND <- uncomment to remove attacking voice when area is selected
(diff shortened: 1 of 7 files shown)


MTKing4:
Looks Correct, i Approve 

Toxin/Stealth have it Grayed out while demo doesn't have the ability currently

image

commy2:
Yeah, this is only added to vGLA, because it is for testing.

It will not be possible to add this to the Demo Toxin Tractor without other changes. The Demo Toxin Tractor already uses 3 weapons: spray, sprinkler and suicide. Suicide is slot 3 and 3 is the maximum.

One could remove the sprinkler from Demo I suppose, or just not add suicide ability to the Toxin Tractor. Both would be strange though.

xezon:
I am skeptical of this, because it adds new functionality to Tractor which is counter intuitive and as for Demo Tractor would introduce inconsistency as highlighted previously.

MTKing4:
If nothing works, we could just leave it the way it is now, i.e you clean with sprinkler for all glas

commy2:
As for Demogen, could consider removing the Suicide ability from Tractor. Other units are missing it too, like Terrorist, Bomb Truck and Angry Mob.

commy2:
If https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2303 is merged (even if the delay is set to 0), then it's possible to have Demo Tractor Cleanup + Srinkler + Suicide all at once. What would have to be sacrified though is the Anthrax Beta goo. I think that would be fine, because Anthrax Beta is inaccessible for Demo anyway.

MTKing4:
That sounds perfect actually

PR #1114, opened 2022-09-04
Change: Increase Super Hacker Vehicle Hack range by +3%

xezon:
* old PR #790

Increase Super Hacker Vehicle Hack range by +3%.

3% does not sound like much, but this is the difference between life and death vs idle units. With this change Super Hacker can try engage idle tanks. When idle, meaning no guard or retaliation, Super Hacker can successfully disable Tank without getting shot at.

Most Tanks Attack Range: 150
Original Super Hacker Attack Range: 150
Patched Super Hacker Attack Range: 155

This change needs to be closely monitored. If Super Hacker can be abused in some way because of it, then it should be reverted.

## Original
shot_20220730_165202_4

shot_20220730_165206_5

## Patched
shot_20220730_165004_1

shot_20220730_165014_2


xezon:
Jundiyy
> Think it's okay to merge. There are only a few players that use vehicle hack for Hackers, perhaps more will with this change.

ReLaX82
> I like the idea to have this functionality even more implemented. Its a Late game thing anyway and increasing it is okay.

PR #1122, opened 2022-09-04
Fix: Nuke Cannon Shell Detonation Effect Has Offset

commy2:
Spawns effect via dummy, which should not suffer impressions.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -4984,6 +4984,35 @@ Object NukeCannonShell
 
 End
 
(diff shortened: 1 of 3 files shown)


xezon (dismissed):
Looks good to me.

xezon:
This change does not work correctly when targeting unit instead of terrain. Same for both weapons.

https://user-images.githubusercontent.com/4720891/188618957-5a2a2c14-e4a3-4967-973c-929dbf83595f.mp4

commy2:
Good catch. Dumb game. Should work now. Note to self: Don't test nuke explosion effects on Nuke Battlemasters.

xezon (approved):
Looks good now.

xezon:
Good to merge?

PR #1123, opened 2022-09-04
Fix: Red Guard and Minigunner Recoil Animations

commy2:
Heavily inspired by the code from NProject.

### TODO

- [ ] Fix firing transition animation

## Patched

https://user-images.githubusercontent.com/6576312/188334843-d7db437b-45fa-49ba-8f65-7d3fce12d4d7.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
@@ -12,15 +12,12 @@ Object ChinaInfantryRedguard
   ;UpgradeCameo4 = NONE
   ;UpgradeCameo5 = NONE
 
+  ; Patch104p @bugfix commy2 04/09/2022 inspired by NProject
+  ; Fix recoil animations and various transitions
+  ; Fix dust trail missing when moving while injured
   Draw = W3DModelDraw ModuleTag_01
 
     OkToChangeModelColor = Yes
-
-    ; this says "we don't use these condition states at all, so completely
-    ; ignore them for purposes of matchmaking"... this is useful to help
-    ; reduce the number of AliasConditionState clauses you must add in
-    ; order to avoid ambiguity in some cases.
-    IgnoreConditionStates = PREATTACK_A FIRING_A BETWEEN_FIRING_SHOTS_A RELOADING_A
 
     ; ------- Standing-Around Animations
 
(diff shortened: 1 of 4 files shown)


ExiLe-ZH:
That syncs very nicely

Transition animation from idle to shooting is kinda instant though

commy2:
Yeah, that seems to be the drawback of that setup.
It is basically Burton animation logic instead of Ranger logic.
The lack of a proper stand->firing transition is a major regression.

xezon:
Can this transition be re-added?

commy2:
I doubt it. There are essentially two different types of firing animation implementations. One for e.g. the Ranger (USING_WEAPON) and one for e.g. Burton (FIRING BETWEEN_FIRING_SHOTS). They each have perks and shortcomings.

It seems that Ranger type allows for Stand->Firing transitions, while Burton type does not. Burton type on the other hand displays an animation for every individual shot and not the same animation on loop while firing (regardless of clip sizes etc.).

Switching Red Guard to Burton type is what NProject does. It appears not to be a direct improvement. I hope do not come across as biased, but if I look at my Minigunner change over 1.04 ZH, it is a direct improvement. This on the other hand is a change in behaviour with aspects one may or may not like. The lack of a firing transition is very noticeable.

Of course, maybe Enlima et al know how to get a transition animation to show up correctly anyway. Hopefully without having to change unit statistics.

So far, all my changes on this units firing behaviour are of course purely astetic.

commy2:
I hope so, because the lack of transition is very noticeable in game. 

There are essentially two different types of firing animation implementations. One for e.g. the Ranger (USING_WEAPON) and one for e.g. Burton (FIRING BETWEEN_FIRING_SHOTS). They each have perks and shortcomings.

It seems that Ranger type allows for Stand->Firing transitions, while Burton type does not. Burton type on the other hand displays an animation for every individual shot and not the same animation on loop while firing (regardless of clip sizes etc.).

xezon:
This needs work.

commy2:
I give up on this.

xezon:
Red Guard end boss :D

What is the problem? Nothing works?

xezon:
The original transition animation is bugged too. The first shot is sideways, because the animation is so slow - or the shot too early.

I see recoil timing on Original Red Guard is also a problem. After a few shots it goes out of sync.

commy2:
> What is the problem? Nothing works?

I think I exhausted all the possible states and combinations between ATTACKING, PREATTACK, USING_WEAPON, FIRING, BETWEEN_FIRING_SHOTS and RELOAD as well as all transition modifiers and nothing gives a workable result.

> I see recoil timing on Original Red Guard is also a problem. After a few shots it goes out of sync.

Yes. This PR fixes it. Tradeoff is the Stand -> Firing transition being removed.

xezon:
The original infantry start firing transition animations have this problem:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/f3a76971-9c8f-42d4-afbe-911ed6ee0ee2

commy2:
Yes. That is because units have no wind up for firing, except from turning into the right direction. If they had, this would be EZ, since one could use PREATTACK state.

I see two ways:

- keep vanilla: transition, but desynched firing
- go NPM: no transition, but synced firing

If there is a trade-off like this, I'm inclined to keep vanilla: don't introduce new bugs, unless the benefit decisviely outweigh the drawbacks. I don't see that here.

xezon:
There is another problem:

When sending Ranger to attack while walk, then there is no transition animation.

When sending Ranger to attack while standing, then there is transition animation.

To me it looks better without the transition, because the initial backwards muzzle flash is jank.

MTKing4:
Go NPM, Constant Fire is more visible because it's longer so it takes priority over the short period-ed transition animation which will be less visible, so I'd rather fix the recoil animation here.

commy2:
Ok.

PR #1126, opened 2022-09-05
Add object experience document

xezon:


xezon:
Issues:

* Sheet does not allow to set patched values for units that start off with 0 XP, because 0 * Anything remains 0
* Sheet is missing actual values currently applied in Main branch, incl Hacker, Lotus and Tanks
* Sheet is missing distinction for Super Hacker, Super Lotus, Demo Jarmen
* Sheet value columns have varying widths. It would be better to look at with consistent widths

I suggest to flip the inputs. Allow set patched raw values, and calculate the percentage off of it, and not the other way around.

xezon:
Exile posted new version in Discord.

ExiLe-ZH (approved):

PR #1128, opened 2022-09-05
Change: Add sound effect for building placement

xezon:
This change adds a gentle sound effect on building placement.

https://user-images.githubusercontent.com/4720891/188427621-6da86e0b-c45a-4ad8-8f94-712aa32eecff.mp4


commy2:
Please no. This is super irritating.

xezon:
Could you elaborate why? Everything else has sound feedback: selecting Building in UI, selecting building in World, selling Building, cancelling Building. Just placing Building has no sound, and it appears odd, as one could expect audible confirmation on placing a building. Other Strategy games have that as well. For example check video:

https://www.youtube.com/watch?v=4C93tY8Bhkg

commy2:
And other RTS games do not have audible scaffold placement:

https://youtu.be/dneEsiAFy_Q?t=41

The audio feedback already exists. The Worker or Dozer confirms the order or rejects the order in case of bad terrain ("cannot build there"). 

This sound effect on the other hand adds nothing but sound clutter. It may even distract you from hearing more important sounds going on, e.g. what the enemy is up to. But most of all, it is an unnecessary change that makes me think I am playing a mod rather than a patch of direct improvements.

xezon:
The priority of the sound could be tweaked if it would hide other sounds. Medium sounds about right however. Volume can be reduced, that way it would be more gentle.

commy2:
It's not the priority. It is that it exists at all. It is distracting even if nothing else happened, because it sounds like something happened, when it did not. The preferable volume of this effect is 0.

xezon:
I think it is nice. Adds to audio feedback to base build. Perhaps could have lower volume, but seems to fit with current volume.

https://user-images.githubusercontent.com/4720891/188466715-3caebdda-d652-4a1b-b702-2866a8a90c42.mp4

xezon:
I will make it optional content. Then purist does not need to hear it.

ExiLe-ZH:
I agree it's annoying aswell, it's like there's some sort of collision happening that needs my attention. Maybe because I'm not used to it.

But is there there any feedback needed in the first place? Like you already get feedback when it's not possible to place.

xezon:
Yes the building feedback logically makes sense, because there is no feedback for cancelling a building before placing it either. This means if player accidentally pressed the wrong mouse button, it sounds the same, unless the Worker unit says something (which it may not if other Worker unit sounds are queued as illustrated in video above).

xezon:
As for explosion sound, as far as I am aware there is no explosion in the game that sounds like this placement sound.

ExiLe-ZH:
> Yes the building feedback logically makes sense, because there is no feedback for cancelling a building before placing it either. This means if player accidentally pressed the wrong mouse button, it sounds the same, unless the Worker unit says something (which it may not if other Worker unit sounds are queued as illustrated in video above).

Cancelling how? Deselecting worker?

xezon:
You press one mouse button to place the building, and the opposite mouse button to abort the placement. Both actions have no sound in original game, so there is no audible distinction.

ExiLe-ZH:
Right clicking simply deselects worker, there already is visual feedback for what you have selected.

Also, worker already gives voiceline when placing down scaffold.

xezon:
The worker voice line is not guaranteed, because if Worker already talks then he will not talk another line.

ExiLe-ZH:
A real QOL improvement would be a way to deselect scaffold without deselecting worker, so you can give a straight move order without having to reselect your worker.

xezon:
Feel free to implement that, but this change is solely about build sound.

ExiLe-ZH:
Not sure why such remark is necessary. 

Just brainstorming for useful features to have, I'm no techy.

Guess you can already do ingame when you number your worker, or idle button when it's your only idle worker. Though pressing S to cancel scaffold selection would be quite nice.

Btw, Guard or Attack Move order cursor can already be 'cancelled' by right clicking, without deselecting the unit.

ExiLe-ZH:
As for sound, maybe some sort of UI sound would work better, not a big boomy sounds that activates my subwoofer.

xezon:
You can add sound suggestion here. I think sound is ok, but certainly it is possible that there is much better sound.

ExiLe-ZH:
Same sound as choosing/selecting a building from worker UI?

xezon:
I do not think that is optimal, because that sound would indicate that user clicked on UI instead of placing building on Terrain.

MTKing4:
Hmmm, listening to it in game it does sound a bit better than when listening to it raw.

MTKing4:
Also, I'm not sure about this, but i think the voiceline feedback from worker can be interrupted by other voicelines, where this placement sound doesn't get interrupted by voicelines, right? 

So basically this is the only true audible feedback, not the voicelines, assuming I'm correct on what i said above.

This can be helpful in GLAs especially, when i sometimes place a fake barracks/arms dealer, the worker was already talking, i don't get audible feedback, i come back a minute later, i have no arms dealer (a noob i know)

ExiLe-ZH:
> The worker voice line is not guaranteed, because if Worker already talks then he will not talk another line.

Btw, can happen with any order for any unit. 

But besides voicelines, you also have visual indicators for all of these orders except for GuardMode or direct AttackOrder.

Move and AttackMove do have visual indicators (little cursor animation the ground) and when placing down a Scaffold it appears right away, you can not really miss it unless you scroll really fast in a laggy match.

There probably are actions for which better feedback is needed, but I can't really think of any examples right now.

ExiLe-ZH:
> This can be helpful in GLAs especially, when i sometimes place a fake barracks/arms dealer, the worker was already talking, i don't get audible feedback, i come back a minute later, i have no arms dealer (a noob i know)

The laggy switch between fake and real buildings is definitely a problem, because the scaffold seems to appear at first but then just disappears? So you scroll away and come back to discover nothing was build, this sucks.

To get around the problem, one has to develop a habit of "pre-switching", which is learned and discovered pretty easy (if you have a brain :P), however sometimes you still may want to switch and build fast.  Or you simply forgot, happens aswell.

xezon:
I was unable to extract build sound from Red Alert 3, so this is the best sound I can get for now. The sound is now part of Optional Audio, not Core.

MTKing4:
> To get around the problem, one has to develop a habit of "pre-switching", which is learned and discovered pretty easy (if you have a brain :P), however sometimes you still may want to switch and build fast. Or you simply forgot, happens aswell.

I started to estimate the delay in the end yeah, but it's annoying, and still happens sometimes, Especially when being rusty and just came back :D

ExiLe-ZH:
What I meant is not even having to estimate the delay, like you can already press F straight after placing down the Supply Scaffold (if you plan to play your BO with fake AD).

MTKing4:
i guess that works for AD yeah true, but still the barracks is annoying too.

commy2:
> because the scaffold seems to appear at first but then just disappears?

I don't remember this happening. If a scaffold appears, then the fake building should be placed without disappearing later.
I do admit though that I don't use fake buildings very much in multiplayer (even though they have their advantages).

xezon:
3 people do not like it. 1 likes it. Closing.

PR #1129, opened 2022-09-05
Change: Use better sound for TechRepairbay and TechRepairPad

xezon:
This change uses better fitting sound for TechRepairbay and TechRepairPad. The sound already exists in game data.


xezon:
I tested sound again. I don't like it.

xezon:
Rejected

PR #1131, opened 2022-09-05
Change: Add new sound for Demo Trap select

xezon:
This change adds better fitting sound for Demo Trap select.

## Original

Plays generator sound on Demo Trap select.

https://user-images.githubusercontent.com/4720891/188462853-54203e09-b436-440e-a640-c3cd9d98baa5.mp4

## Patched Core Sound

Plays machinery on Demo Trap select.

https://user-images.githubusercontent.com/4720891/188566739-e5b3d67e-1eb3-4e52-8544-366db6459ef6.mp4

## ~~Patched Optional Sound~~

~~Plays beeps and machinery on Demo Trap select.~~

https://user-images.githubusercontent.com/4720891/188566756-a73ac5f3-e3f8-4089-9c15-01cff1fe891f.mp4


commy2:
Hmm, while the original sound is strange, this new one sounds very similar to a demo trap about to blow up. This will require getting used to and may be confusing.

xezon:
It should sound different, because the pitch of the beeps is different and it has longer gaps between the beeps.

ExiLe-ZH:
Good improvement, original sounds out of place.

xezon:
The beeping sound is now optional content. I added another sound into core files without the beeping.

MTKing4:
I like it, commy has a point though, no preference for me either way.

xezon:
I am removing the Controversial label because the sound without the beeping is now a clean substitute for the original generator sound which should come without annoyance.

MTKing4:
A demonstration of the new edited sound?

xezon:
New videos added.

xezon:
Any further opinions on this? From my side is is ok, although I am not sure how the beeping select will experience in actual match. Would have to be tested, but can be changed again if deemed annoying.

xezon:
I removed the optional sound with beeping for now. This way this change is less likely to cause upset.

Issue #1016, opened 2022-08-29
Comanches vet up extremely quick (and give off little XP)


ExiLe-ZH:
Compared to other meta units, Commies vet up extremely quick because of the strong onboard weapons which are capable of leveling buildings very quickly (buildings give of alot of XP). Whereas Vees or Lixes often rely on their passengers to make the kills, which makes a passenger vet up , but not the vehicle itself.

If you further compare Commies to other killing machines like **Overlords** and **Alphas** we can see the requirement is very low, while the Commies in groups are much more efficient usually.

Same story for the ExperienceValue, vetted Comanches are incredibly hard to catch, unless the Air player commits to a bad engagement or doesn't pay attention to get away in time.

It seems EA wasn't really aware of how capable Comanches are.

```
Comanche:
  ExperienceValue     = 50 50 100 200   ; Experience point value at each level
  ExperienceRequired  = 0 100 200 400   ; Experience points needed to gain each level

Helix:
  ExperienceValue     = 50 100 150 200 ;Experience point value at each level
  ExperienceRequired  = 0 100 200 400  ;Experience points needed to gain each level

Tank Hunter/MD:
  ExperienceValue = 20 20 40 60    ;Experience point value at each level
  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level

Vee:
  ExperienceValue = 50 50 100 150   ;Experience point value at each level
  ExperienceRequired = 0 100 150 300  ;Experience points needed to gain each level

Crusader/Paladin:
  ExperienceValue        = 100 100 200 400 ;Experience point value at each level
  ExperienceRequired     = 0 200 300 600 ;Experience points needed to gain each level

Crusader/Paladin Patched:
  ExperienceValue        = 100 100 200 300 ;Experience point value at each level
  ExperienceRequired     = 0 200 300 600 ;Experience points needed to gain each level

Overlord:
  ExperienceValue = 200 200 400 600   ;Experience point value at each level
  ExperienceRequired = 0 400 600 1200  ;Experience points needed to gain each level

Alpha Aurora:
  ExperienceValue = 200 200 400 400   ;Experience point value at each level
  ExperienceRequired = 0 200 400 800   ;Experience points needed to gain each level
```

### Proposal:
Since buffing is prioritized with the gamepatch approach, +25% is atleast a balance improvement over the original.

```
ExperienceValue     = 60 60 125 250 ;Experience point value at each level
[...]

PR #1135, opened 2022-09-06
Change: Reduce AFG_ComancheArmor HEALING by 10% before Upgrade

xezon:
* old PR #821

This change makes `HEALING` of `AFG_ComancheArmor` consistent with `HEALING` of `AFG_CountermeasuresComancheArmor`.

The Airforce Comanche will repair a bit slower on the Airfield before Upgrade accordingly.

## Stats

| Object                                    | Full repair time |
|-------------------------------------------|------------------|
| Original AirF Comanche                    | 22.0             |
| Original AirF Comanche w. Countermeasures | 24.4             |
| Patched AirF Comanche                     | 24.4             |
| Patched AirF Comanche w. Countermeasures  | 24.4             |

## Additional Stats

helihealing2

## Rationale

This change makes healing times on Airforce Comanche helicopter consistent. It no longer gets a penalty in healing time with Countermeasures upgrade, because the healing time is now the same before the upgrade. Healing time is not bumped to 100% on Countermeasures, as that would be a buff for the Airforce Comanche, and it is already powerful enough as is.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -1080,7 +1080,7 @@ Armor AFG_ComancheArmor
   Armor = EXPLOSION         100%    ;gives patriot some more punch
   Armor = INFANTRY_MISSILE  100%    ;gives missile guys some more punch
   Armor = LASER               0%    ;lasers are anti-personel and anti-projectile only (for point defense laser)
-  Armor = HEALING           100%
+  Armor = HEALING            90%    ; Patch104p @tweak from 100% to make healing time consistent with AFG_CountermeasuresComancheArmor
   Armor = HAZARD_CLEANUP      0%    ;Not harmed by cleaning weapons
   Armor = KILL_PILOT          0%    ;Jarmen Kell uses against vehicles only.
   Armor = SURRENDER           0%    ;Capture type weapons are effective only against infantry.


xezon:
So what shall we do with this? Jundiyy is against changing it.

> Don't think it's a problem, but...I don't see why it needs changing? I know it says consistent, but I don't see it as a good reason. One armour before upgrade, second armour after. Same for healing.

commy2:
What _should_ we do? Entirely rework Countermeasures bonuses, so it is a 25% flat upgrade, and not a 37.5% upgrade, which would perfectly counteract AP Bullets.

Short term? Nothing. as no consensus can be reached.

ExiLe-ZH:
Oh no, definitely don't nerf Counter Measures, would be the same as nerfing Vee HP.

Without CM Commies are useless. Or you'd have to increase HP pre-upgrade a little to compensate, but not a fan of doing this.

As for balancing Commies, we should look into XP values first.

xezon:
This change is not about Countermeasures. It is about HEALING before upgrade.

Jundiyy:
In that message, the first part means, don't think it's problem to make the change, meaning it seems safe.
And the second part is saying, I don't see a reason to actually make the change. And I explained why.
If people think it's better to make this change then sure.

ExiLe-ZH:
In reality CM makes your Commies repair faster, almost twice the HP per second. 

The way it is can be considered fair, since pre-CM Commies are pretty much useless vs Quads or Gats. We don't want to buff pre-CM either.

xezon:
This is not true. Stealth Comanche without Countermeasures are still very good. I know because I rarely built Countermeasures, because I never knew it gives more Armor. I was always under the impression it just diverts some RPG missiles. I think the tool tip does not explain this correctly.

ExiLe-ZH:
You must have picked and microed your fights very well then. I've seen the mistake so many times where ppl don't get CM (in time) and see them go down like flies when they are forced to defend vs Quad pushes.

Expert vs expert has no chance whatsoever of winning vs GLA without CM.

Stubbjax:
> Oh no, definitely don't nerf Counter Measures, would be the same as nerfing Vee HP.
> 
> Without CM Commies are useless. Or you'd have to increase HP pre-upgrade a little to compensate, but not a fan of doing this.

I strongly disagree. Air Force Comanche Countermeasure armour is ridiculously strong and could absolutely do with a nerf. Comanches aren't meant to be flying tanks. The entire point of the stealth upgrade is to enhance their survivability and combo with their role as highly mobile glass cannons. The bonus armour is an unnecessary double-up in survivability which, coupled with their relatively low price, leads to a huge imbalance in numbers as well as between Comanches of other USAs. It also leads to their complete domination in 3v3s, where opponents are almost always forced to embark on a wild goose chase across the map with limited firepower and stealth detection capabilities. I'd be questioning why your Comanches are taking so much damage / dying if you really believe they are useless without it.

> As for balancing Commies, we should look into XP values first.

Changing xp values will not solve the imbalance. The problem is that Air Force Comanches are too difficult to kill in the first place due to the tough armour and limited attack windows. If an Air Force loses enough Comanches for any heightened xp yield to make a meaningful difference, then the damage is already done.

> I've seen the mistake so many times where ppl don't get CM (in time) and see them go down like flies when they are forced to defend vs Quad pushes.

As they should! That is a clear error in judgement / execution on the user's part. It does not make sense to try and preserve the usage of helicopters against anti-air or incentivise one-man armies. Comanches are not supposed to be anti-everything. Strength / weakness relationships are key factors in a balanced, well-designed game. The bonus countermeasure armour goes against this.

Anyway, I believe a reasonable approach to the overall Comanche issues would be to match `ComancheArmor` with `AFG_ComancheArmor`, which would make standard Comanches a bit more viable and bring about some consistency. These are the **current** differences in 1.04:

| Type | ComancheArmor | AFG_ComancheArmor |
|-|-|-|
| SMALL_ARMS | 120% | 100% |
| GATTLING | 120% | 100% |
| EXPLOSION | 130% | 100% |
| INFANTRY_MISSILE | 120% | 100% |

Reducing the `AFG_CountermeasuresComancheArmor` bonus from 37.5% to 25% is also a great idea, as @commy2 suggested. It may even be best to simply match `AFG_CountermeasuresComancheArmor` with `CountermeasuresComancheArmor`, and leave the stealth ability as the sole distinction between Comanches. These are the **current** differences in 1.04:

| Type | CountermeasuresComancheArmor | AFG_CountermeasuresComancheArmor |
|-|-|-|
| DEFAULT | 100% | 90% |
| SMALL_ARMS | 75% | 62% |
| GATTLING | 75% | 62% |
| EXPLOSION | 110% | 90% |
| INFANTRY_MISSILE | 100% | 90% |

These changes would help to improve the utility of standard Comanches, and take a bit of the edge off of Air Force Comanches by amplifying the impact of user-error that often comes as a result of direct anti-air engagement, as well as improving the counterplay abilities of opponents. And best of all, achieved by reducing inconsistencies instead of adding them! The bizarre 90% `HEALING` modifier would no longer be a factor either.

To summarise, I'm suggesting an equivalent approach to the following:
```
OLD:
Standard Comanche:  ComancheArmor     → CountermeasuresComancheArmor
Air Force Comanche: AFG_ComancheArmor → AFG_CountermeasuresComancheArmor
```
```
NEW:
Standard Comanche:  AFG_ComancheArmor → CountermeasuresComancheArmor
Air Force Comanche: AFG_ComancheArmor → CountermeasuresComancheArmor
```
Also note that the `EXPLOSION` modifier would need to be reduced from 110% to 100% or 90% for `CountermeasuresComancheArmor`.

And one final note; it could also be good to increase the `StealthDelay` from 1500ms to 2000ms to add weight to attack commands, provide opponents with more counterplay opportunities, further amplify the impact of mistakes and increase the overall skill ceiling.

xezon:
Nice analysis. Do you want to open a new Pull with implementation for it?

ExiLe-ZH:
Lets open new topic for that. 

However, I'm gonna say this right away: While I completely agree upgraded AFG Commies are OP, I'm not a fan of nerfing them. It would be same as nerfing Vee armor, range or whatever. Maybe the Stealth thing could be acceptable though.

Stubbjax:
> However, I'm gonna say this right away: While I completely agree upgraded AFG Commies are OP, I'm not a fan of nerfing them. It would be same as nerfing Vee armor, range or whatever. Maybe the Stealth thing could be acceptable though.

What particular reservations do you have? Are you suggesting Air Force Comanches should remain OP? Or do you believe the balance issues would be better addressed another way?

I think it is somewhat different to your comparison, as Humvees are USA's bread and butter, whereas Air Force Comanches are more specialised / situational units.

It's also important to note that the overwhelming majority of damage that stealth Comanches receive is limited to their window of attack / aggression, and is usually a direct consequence of the user's actions, rather than the opponent's. Essentially, the respective change would strictly target the user's skill requirement while leaving the opponent's skill requirement largely unaffected; which would not be the case if applicable nerfs were applied to Humvees.

Anyway, I've pushed the respective changes to the branch `comanche-survivability-balance`, so feel free to check it out and see how it feels before I open a pull request. I tested the changes against three GLA Stealth Hard Armies on Defcon 6 with no War Factory (quite the challenge!), and the changes feel very reasonable.

xezon:
Realistically no one is going to playtest branches. There has been very little precedent.

ExiLe-ZH:
My reservations are that Commies are the bread and butter for AFG. When he commits to fly it's his only unit on the playing field (and maybe a Rap or 2 for support and sometimes a SF in later stages).

Yes, in teamgames it's often mega OP when teammates combine with strong groundforces that requires a different counter units than Gats/Quads. Commies can just go wherever they want and wreck havoc on those bigger teammaps and it's very hard for the opponent to cover it all.

It's a different story in 1v1 though, going Commies hardly means auto-win. It works best vs Demo since he has more expensive Quads and no stealth tunnel upgrade. Vs GLA and especially Stealth it's tricky and going Vees is generally the better choice.

I consider CM-Commie as the standard Commie, there's no point in ever using them without CM and therefore I'm hesitant to nerf the upgrade.

Your branch needs to be tested 1v1 and I'm always happy to test, tested the Helix branch with MTKing recently aswell.

Stubbjax:
> My reservations are that Commies are the bread and butter for AFG. When he commits to fly it's his only unit on the playing field (and maybe a Rap or 2 for support and sometimes a SF in later stages).

Hmm, I'd probably give that designation to King Raptors before Comanches, and still likely Humvees overall. And don't forget the odd Combat Chinook (which are also overpowered in their own right, but that's a whole other topic). There is actually some nice strategic diversity with Air Force. But this is also why it's a less impactful change, particularly relative to Humvees, because Air Force has options, and it's ultimately a player's decision whether they commit to Comanches or not.

> It's a different story in 1v1 though, going Commies hardly means auto-win. It works best vs Demo since he has more expensive Quads and no stealth tunnel upgrade. Vs GLA and especially Stealth it's tricky and going Vees is generally the better choice.

That is a certainly a valid concern, but as you point out, Humvees are the meta where it's the best / safest option. If going Comanches is already a risky strategy vs those armies, then such a change would simply increase the risk and require more skill to pull off, and amplify the existing dynamic between those armies that players are already familiar with. It's ultimately Air Force's game to lose, rather than the opponent's game to win.

> I consider CM-Commie as the standard Commie, there's no point in ever using them without CM and therefore I'm hesitant to nerf the upgrade.
> 
> Your branch needs to be tested 1v1 and I'm always happy to test, tested the Helix branch with MTKing recently aswell.

I think you'd be surprised just how many (skilled) players don't realise the true impact of Countermeasures, especially as it's a hidden mechanic and counterintuitive. I suspect it'd be pretty unlikely that anybody would consciously notice a ~13% armour difference. Despite this, Air Force Comanches are still incredibly capable units even without Countermeasures as their stealth ability serves as their primary means of damage mitigation, and aligns with their proficiency as harassers / counterpunchers. They are certainly trickier to use in 1v1 due to the strict (anti-air) focus of the opponent and smaller map size, but the key distinction of the respective change would be their heightened skill requirement rather than lowered performance.

Nevertheless, there aren't many (if any) alternatives to curbing their domination in team-games without affecting 1v1s in some way. Inconsistencies are ideally avoided, so the only other option seems to be to adjust the stealth upgrade's cost or research time to increase the initial window of vulnerability. Restealthing could be denied while under fire, though that'd be a very noticeable and potentially confusing change. Do note that other changes may also have some influence, such as #1055 and #1094. Plus a `HEALING` modifier of 100% instead of 90%, as trivial as it may be.

Also happy to test!

RisingZH:
Can rocket pod guard mode be fixed? That’s certainly a bug and would be a better way to nerf commies I think.

Stubbjax:
While indeed effective, that is a very situational and seldom used tactic and is not the primary source of their power. Reducing the velocity of their rockets could be an interesting idea, as it would allow opponents more time to get some damage in before their entire army is wiped. But the ultimate outcome of heightened vulnerability is the same.

ExiLe-ZH:
> Can rocket pod guard mode be fixed? That’s certainly a bug and would be a better way to nerf commies I think.

That's very good thinking, Rocketpods are meant to attack an area, not to be mixed in as a primary weapon for normal engagements, it's overly efficient and takes away all micro. It's pretty much the same bug where Rangers auto switch their Rifles and Flashbangs in a perfectly efficient manner.

I'd say this is a good and fair nerf.

xezon:
Can that be considered as bug though? I never thought of it as that.

Stubbjax:
It's more of an exploit than a bug.

ExiLe-ZH:
> Can that be considered as bug though? I never thought of it as that.

I didn't either actually, but Risings comment made me think about it and I came to the conclusion that it is, or rather an exploit indeed.

RisingZH:
Imagine if you could guard mode laser lock.

ExiLe-ZH:
> Hmm, I'd probably give that designation to King Raptors before Comanches, and still likely Humvees overall. And don't forget the odd Combat Chinook (which are also overpowered in their own right, but that's a whole other topic). There is actually some nice strategic diversity with Air Force. But this is also why it's a less impactful change, particularly relative to Humvees, because Air Force has options, and it's ultimately a player's decision whether they commit to Comanches or not.

Raps are also AFGs bread and butter, it's just whatever he commits to, he either goes all Raps, all Commies or all Vees. Full Rap commitment isn't good in 1v1 except Airmirrors ofcourse, not on higher (equal) skill level atleast. 

> I think you'd be surprised just how many (skilled) players don't realise the true impact of Countermeasures, especially as it's a hidden mechanic and counterintuitive. I suspect it'd be pretty unlikely that anybody would consciously notice a ~13% armour difference. Despite this, Air Force Comanches are still incredibly capable units even without Countermeasures as their stealth ability serves as their primary means of damage mitigation, and aligns with their proficiency as harassers / counterpunchers.
> 

I don't agree with this one, everyone of higher level knows how important CM is, it is considered a HUGE mistake not to get it early.

Sure, you can still roam around without CM on bigger maps in teamgames and surprise attack, but you can't take direct engagements.

> Nevertheless, there aren't many (if any) alternatives to curbing their domination in team-games without affecting 1v1s in some way. Inconsistencies are ideally avoided, so the only other option seems to be to adjust the stealth upgrade's cost or research time to increase the initial window of vulnerability. Restealthing could be denied while under fire, though that'd be a very noticeable and potentially confusing change. Do note that other changes may also have some influence, such as #1055 and #1094. Plus a `HEALING` modifier of 100% instead of 90%, as trivial as it may be.

Maybe the restealth thing is fair, you could state that it was an over-commitment from AFG in the first place if he has to escape during engagement.

How does this work with other stealth units though? Do they restealth while getting shot?

Actually don't even think it makes a huge difference, they're fast and when they are out of range... they're out of range. Maybe the last fleeing Commie (or 2) can be taken out. Could be wrong about this, needs testing.

Stubbjax:
> I don't agree with this one, everyone of higher level knows how important CM is, it is considered a HUGE mistake not to get it early.

Again, I think you would be surprised. It really depends how you define higher level / everyone. I've partaken in enough 3v3 GG EXPERTS +++++ games to know that it's definitely not universal knowledge. I'm sure I could dig up a bunch of high-level replays where Air Force Comanches dominate even without the upgrade.

> Maybe the restealth thing is fair, you could state that it was an over-commitment from AFG in the first place if he has to escape during engagement.

Well that's the thing, it has more or less of the same impact as the armour change and ties in incredibly well with their design / purpose. Half a dozen heroic Comanches are still incredibly hard to take down, and do a disproportionate amount of damage, even without countermeasures.

> How does this work with other stealth units though? Do they restealth while getting shot?
> 
> Actually don't even think it makes a huge difference, they're fast and when they are out of range... they're out of range. Maybe the last fleeing Commie (or 2) can be taken out. Could be wrong about this, needs testing.

There is a condition that prevents stealthing while under attack, though I'm fairly sure the only unit it exists on is stealth rebels and nothing else (though maybe it should be removed from them for consistency). It applies to buildings though.

ExiLe-ZH:
> Again, I think you would be surprised. It really depends how you define higher level / everyone. I've partaken in enough 3v3 GG EXPERTS +++++ games to know that it's definitely not universal knowledge. I'm sure I could dig up a bunch of high-level replays where Air Force Comanches dominate even without the upgrade.

It's still their mistake to be fair, you don't get away with it when playing vs opponents like Hummi, Ganja etc. (standard higher level non expert players).

It would be more user friendly for everyone though to improve the upgrade description.

> Well that's the thing, it has more or less of the same impact as the armour change and ties in incredibly well with their design / purpose. Half a dozen heroic Comanches are still incredibly hard to take down, and do a disproportionate amount of damage, even without countermeasures.

That's why I prefer to reduce the snowball effect by slowing down the rate at which Commies vet up (and increase reward for killing them), rather than touching base values.

Stubbjax:
> It's still their mistake to be fair, you don't get away with it when playing vs opponents like Hummi, Ganja etc. (standard higher level non expert players).

Just like it's a mistake to open up your Comanches to attack and suffer losses! Countermeasures still provides a 25% armour in addition to the stealth upgrade - the change doesn't make them useless by any means of the imagination.

> That's why I prefer to reduce the snowball effect by slowing down the rate at which Commies vet up (and increase reward for killing them), rather than touching base values.

But would you do the same for standard Comanches? Because the added inconsistency would not be ideal. If the 13% armour bonus absolutely must remain, then I'd suggest it should apply to standard Comanches as well.

xezon:
Standard Comanche does have 37.5 % armor bonus too, no?

Reducing Armor buff from 37.5 to 25 seems reasonable, however it will make Regular Comanche weaker, so it would need buff elsewhere, for example increase base armor to match +25 Armor buff with current values.

Stubbjax:
> Standard Comanche does have 37.5 % armor bonus too, no?

Nope, only Air Force countermeasures give the 37.5% bonus, whereas all other USAs provide 25%.

ExiLe-ZH:
> Just like it's a mistake to open up your Comanches to attack and suffer losses! 

Easier said than done, try to defend early Quad RPG push in 1v1 without CM. I assure this doesn't fare well in majority of cases, it's just silly to skip CM. 

Again, we can try the changes in testgames. I'm just stating my preferences on how to take on this case.

> But would you do the same for standard Comanches? Because the added inconsistency would not be ideal. If the 13% armour bonus absolutely must remain, then I'd suggest it should apply to standard Comanches as well.

This argument also came to light here https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1016

xezon:
> Nope, only Air Force countermeasures give the 37.5% bonus, whereas all other USAs provide 25%.

I checked. This is not correct statement.
```
Armor ComancheArmor
  Armor = SMALL_ARMS        120%  ;gives Quad and gattling a little more punch.
  Armor = GATTLING          120%  ;resistant to gattling tank

Armor CountermeasuresComancheArmor
  Armor = SMALL_ARMS        75%  ; currently set to 37.5% increase (120 x 0.75 -- 25% reduction off ComancheArmor)
  Armor = GATTLING          75%  ; currently set to 37.5% increase (120 x 0.75 -- 25% reduction off ComancheArmor)

```

120 * 0.625 = 75 equals 37.5% buff.
Same for planes.

commy2:
>  Same for planes.

Except Spectre. 
\*leaves*

Stubbjax:
> > Just like it's a mistake to open up your Comanches to attack and suffer losses!
> 
> Easier said than done, try to defend early Quad RPG push in 1v1 without CM. I assure this doesn't fare well in majority of cases, it's just silly to skip CM.

True, though it's very situational and depends how and when you decide to engage. I'd still argue a quad / RPG push should be effective against the Comanches though, and the change would primarily affect the quad damage rather than the RPG damage.

> Again, we can try the changes in testgames. I'm just stating my preferences on how to take on this case.
> 
> > But would you do the same for standard Comanches? Because the added inconsistency would not be ideal. If the 13% armour bonus absolutely must remain, then I'd suggest it should apply to standard Comanches as well.
> 
> This argument also came to light here #1016

Oh yeah, considering that the Helix has the same requirement, it definitely could do with some alteration.

> > Nope, only Air Force countermeasures give the 37.5% bonus, whereas all other USAs provide 25%.
> 
> I checked. This is not correct statement.
> 120 * 0.625 = 75 equals 37.5% buff. Same for planes.

Sorry, I meant in comparison to the standardised values of 100% (AFG_ComancheArmor). The bonus is technically the same for both, but the overall outcome is that Air Force Comanches take 62% damage while standard Comanches take 75%.

Jundiyy:
Such a long discussion and it's gone off topic as far as I can see.

I'm sure this topic is simply for the 90% and 100% healing. 

Regarding all other talks, I don't think there is any harm for having different stats for Air Force Comanches, they are different, they have a different model.

ExiLe-ZH:
> Such a long discussion and it's gone off topic as far as I can see.
> 
> I'm sure this topic is simply for the 90% and 100% healing.
> 
> Regarding all other talks, I don't think there is any harm for having different stats for Air Force Comanches, they are different, they have a different model.

I agree, same for vanilla/king raps (should cost the same), though yet another topic again.

MTKing4:
> Can rocket pod guard mode be fixed? That’s certainly a bug and would be a better way to nerf commies I think.

I don't think this is a good idea to change, it's rather a nice pro trick that makes the best of said ability, sort of like planned attacks like in RA2, also removing that means also removing it from Dragon tank flamewalls, which proves to be extremely beneficial to use over normal firewalls, as the unit can now be forgotten about and it will still do the damage needed, which i also like it to stay that way.

ExiLe-ZH:
Doesn't need fixing for Dragons, it's not nearly as powerful, it's handy to clear up undefended buildings, but not when fighting enemies, because guard mode makes Dragons retaliate, where they try to flame wall units close up and keep switching targets, it's not very reliable.

With Commies it's a whole different story, each individual Commie alternates between Rocket Pods and their standard weapons perfectly efficient and it's all perfectly aimed at the enemy, it's no longer an area of effect weapon. With timing and placement taken away, I can't consider this trick very pro, it's very fun and satisfying to use though admittedly.

MTKing4:
i think for consistency's sake, if feature is removed here it should be removed there as well, which i totally don't want removed btw, never seen anyone considering Commies OP because they have this ability so i'ts fine to stay imo.

xezon:
Perfect China Supply Placement was argued to be a separator between Noobs and Pros and deemed positive. The same logic should apply to the Rocket Pod Guard Mode combo, because it does separate Pros from Noobs. Noobs will not know or use it effectively.

ExiLe-ZH:
Except there are no observable factors that hints towards the existence of efficient Rocket Pod usage. As a beginner you would have to find it out by accident or have someone tell you.

Jundiyy:
I'd like to keep it, it's a neat feature, ZH doesn't have many of them available.
And I don't think you can remove it anyway without changing the type of button/weapon?

PR #1144, opened 2022-09-06
Fix: z clipping muzzle wave particles

xezon:
This change fixes z clipping muzzle wave effects.

shot_20220906_201535_2


commy2:
Example of clipped particle on flat terrain?

xezon:
I do not have video. Basically the ground texture shown above is Z clipped into terrain.

Issue #690, opened 2022-07-11
USA Pathfinder is very strong


xezon:
USA Pathfinder is very strong. It does require a Level 3 Promotion, but still it is an absolute game changer. It kills all Infantry and Hero's instantaneously. It can even shoot when loaded in Humvee and Combat Chinook with laser precision.

Very strong vs:
* China Infantry
* USA
* GLA

It costs just 600 money and builds in 10 seconds. Perhaps it can be a bit less powerful while still remaining very useful.

### Proposal 1
Increase unit build cost.

### Proposal 2
Increase unit build time.

### Proposal 3
Decrease rate of fire.

### Proposal 4
Decrease damage.

### Proposal 5
Decrease rate of fire when loaded in vehicle only.

### Proposal 6
Decrease damage when loaded in vehicle only.

### Proposal 7
Reveals itself while shooting on foot (same as Jarmen Kell).

### Proposal 8
Does not reveal enemy stealth units.

### Proposal 9
Has reduced enemy stealth detection range.

### ~~Proposal 10~~
Has to aim for 1 second before shooting. This decreases kills per second.
Implementation: https://github.com/commy2/zerohour/commit/48b06f0d096ba37a7fd48ce210a534524c37929f
Would make unit feel unresponsive.

### Proposal 11
Make Pathfinder no longer use a hitscan weapon and use a dummy projectile instead. Dummy projectile doesn't track targets, but the weapon has a secondary damage radius with lower damage. Therefore, if Pathfinder hits stationary target, it gets hit with full damage and dies and if Pathfinder's target is moving, it'll only be hit by secondary damage, and require multiple hits to die.

### Proposal 12
Make Hero infantry more resistant to Pathfinders. This could also give Jarmen Kell an edge in 1v1 situations.

### ~~Proposal 13~~
Require Strategy Center to build Pathfinder.

ExiLe-ZH:
think you meant production time with either proposal 1 or 2

ExiLe-ZH:
Additional proposals:
- Reveals when shooting (like Jarmen Kell).
- Does not reveal stealth (a Lotus for example, Jarmen Kell doesn't either).

commy2:
### Proposal 3b

Instead of straight decreasing the ROF, have it have to aim for 1 second before shooting. This decreases kills / second.

Implementation: https://github.com/commy2/zerohour/commit/48b06f0d096ba37a7fd48ce210a534524c37929f

[...]

PR #1152, opened 2022-09-07
Change: Increase XP reward for China Minigunner by 100%

### LINKED ISSUE #1150: Mini Gunner grants very little XP
| Regular | Veteran | Elite | Hero
-- | -- | -- | -- | --
Ranger | 20 | 20 | 40 | 60
Red Guard | 5 | 5 | 10 | 20
Mini Gunner | 5 | 5 | 10 | 20
Rebel | 15 | 15 | 30 | 40
Tox Rebel | 25 | 30 | 35 | 45

- Grunt Ranger grants twice the XP over an Elite MG.
- Red Guards also grant double since they come in pairs (when produced from Rax).
- Hero MG grants 20% less than a regular Tox Rebel.
- Veteran AssaultTroopCrawler grants 40xp (8 x 5xp), this is 20% less than a Vee/Quad/Gat.
- Elite ATC grants 80xp (8 x 10xp), 20xp less than a regular Tank in 1.04 (120xp less than an Elite BM).

xezon:
* Fixes #1150

This change increases the XP reward for China Minigunner by 100%.

| Object                               | Regular | Veteran | Elite | Hero |
|--------------------------------------|---------|---------|-------|------|
| Original USA Ranger                  | 20      | 20      | 40    | 60   |
| Original GLA Rebel                   | 15      | 15      | 30    | 40   |
| Original GLA Toxin Rebel             | 25      | 30      | 35    | 45   |
| Original China Tank Hunter           | 20      | 20      | 40    | 60   |
| Original China Red Guard             | 5       | 5       | 10    | 20   |
| Original Inf China Tank Hunter       | 20      | 20      | 40    | 60   |
| Original Inf China Minigunner        | 5       | 5       | 10    | 20   |
| Patched Inf China Minigunner (this)  | 10      | 10      | 20    | 40   |

## Rationale

The XP reward of the Minigunner is very low in comparison to what it can do. It can attack ground units with +25% range compared to Red Guard. It can also attack Air units with outrageous range of 350 or more. It deals 400% more damage than Red Guard when Minigun is spun.

Identical XP reward with Red Guard is not justified. The Minigunner is a much better unit and more difficult to deal with. An even higher XP reward may be justified, however, Infantry soldiers come in high numbers, especially with China Infantry General, so it should not be unfairly set.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
@@ -15205,7 +15205,7 @@ Object Infa_ChinaInfantryMiniGunner
   BuildCost     = 350
   BuildTime     = 10.0          ;in seconds
 
-  ExperienceValue = 5 5 10 20   ;Experience point value at each level
+  ExperienceValue = 10 10 20 40   ; Patch104p @balance from 5 5 10 20 to reward more than Red Guard
   ExperienceRequired = 0 20 40 80  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel         = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles


commy2:
This will make USA unlock Pathfinders and A10 even faster. With Elite Training, each Minigunner is already worth 10 XP in 1.04, which is already twice as much as each Red Guard, with or without Veteran Training.
This is also on top of the 5 XP that was already added for Outpost and Troop Crawler kills.

xezon:
Infantry General is pretty good vs USA, so it is not a major concern, but we should link this pull with
* #690

ExiLe-ZH:
Inf performs very well vs USA, because it often wins before USA can change the game with Pathies. Pathies being OP is it's own topic anyway.

Stubbjax (approved):

commy2:
Bad change.

Stubbjax:
> Bad change.

Why?

commy2:
Faster Pathfinders and A10 for xUSA.

xezon:
As of 1.04 all players
```
China Infantry vs USA              winrate  60.80%
China Infantry vs USA Airforce     winrate  58.39%
China Infantry vs USA Laser        winrate  61.80%
China Infantry vs USA Superweapon  winrate  64.92%
```

Right now China Infantry benefits from all the China buffs we add while it dominates all USA factions, so giving more XP reward for Minigunner is not something that should worry, quite contrary, should be good.

Stubbjax:
> Faster Pathfinders and A10 for xUSA.

Which they could all arguably do with considering their poor chances against Infantry General. Elite Training will have to be more considered, and xp will effectively match 1.04 if it is not acquired. I doubt an extra ~50 xp is going to change much in the grand scheme though - it's really more important for individual unit ranks. Keep in mind that other proposed changes, like lowered Supply Truck xp values, will also play a role.

Regardless, as already mentioned, Pathfinders are ideally taken down a peg or two so that they aren't so overpowered, and this change should be considered with that in mind. Don't forget that these xp changes also affect other match-ups, most of which Infantry dominates as well.

Jundiyy:
I also don't think it's a good idea. Same as commy has mentioned. Inf uses Mini Gunners vs USAs, USA does have a hard time vs Inf but only at the start, then it starts evening out and after that USA will have a much easier time.

With this change, USA will still have a hard time early on, but this early on won't last as long.

You have to remember, a lot of the times, Mini Gunner are spammed, possibly just as much as Tank Hunters, never counted. Every ATC is 8 MGs.

xezon:
What about other factions though? Do we need to balance Infantry vs USA with Minigunner XP, or is it possible there are other means to address issues, namely Pathfinder?

Jundiyy:
USAs do find man spam quite hard to deal with, China will also do man spam, although both Red Guards and Tank Hunters are weaker than the Inf versions, but they get to mix in Gats. So, it's mainly just Inf Gen that causes the most problems.

We saw more man spam in 1.04+, since we made the RGs and THs train faster or cheaper for all factions, and this did help to some level, but once the Pathfinders came out, the man spam was not enough.

It's a hard one, nerfing the MG means faster rank for USA and means Pathfinders are out, but nerfing Pathfinders means better chance for Inf late game. They could both go hand in hand (increase xp but somehow nerf Pathfinders), but then we are adding more changes, which I don't know if it's a good thing.

xezon:
Keep in mind this Minigunner XP is still low in comparison to other infantry. USA Ranger gives 4 times as much XP as 1 Minigunner in Original 1.04. In this proposed change USA Ranger still gives 2 times as much XP as 1 Minigunner. And USA also spawns many Rangers from destroyed USA buildings... free XP. This is not fair and favors Infantry General a lot.

Jundiyy:
That is true, but we can't really compare the xp of Rangers and MGs when we see the faction as a whole. Inf Gen relies solely on MGs and THs, whereas all other armies and factions don't. USAs have many other units to lose, whereas Inf has just the 2 mentioned, (until it goes for Migs or Helices). 

I don't think it would be the end of the world to make this change, but from the actual games that are played. Inf does do better early on and will many times take the game from there. So, with this change it would possibly allow USAs to not lose as quickly as before in them situations (remember, USA does also manage to take the game to mid and late game).
Although later on, it does mean Inf will be hurt a lot more than original patch.

Perhaps this change is reasonable, and we just need to do something to fix the late game instead. I don't know.

ExiLe-ZH:
40xp for Veteran and 80xp for Elite ATC is a joke. USA deserves more for cleaning that up.

As Stubbjax said, Inf now has to weigh pros vs cons of choosing Elite Training. Veteran MG's don't completely suck.

Also amount of THs on the field is quite a bit larger than MGs on average, I don't think it will make USA vet up that much faster. 

Plus Pathfinders might get a few nerfs of some sort which balances it out again.

xezon:
Yes other Generals do not rely on Infantry, but they have vehicles and planes that give XP too.

In Original 1.04

8x Vet 2 Minigunner kills give 80 XP
1x Vet 2 Humvee kill gives 100 XP

The issue with Pathfinders is that they are too strong. Surely we can set all Minigunner and Tank Hunter XP rewards to 0 and Pathfinders would never spawn, but that would be wrong approach. Better approach would be to make Pathfinder a fair unit.

I do not feel strong about this Minigunner XP change, and to keep peace we can leave as is in 1.04. It just looks like mistake, because Minigunner is so much better than other Infantry and gives no tangible rewards. China Infantry is one of strongest Factions in Game, so surely it has a lot going for it.

ExiLe-ZH:
1 regular Tank already gave 100xp lol, now 75xp.

Jundiyy:
I am not fully against it as mentioned above, we can go with it if that is what most think is best, it does help USAs early game but backfires more later on. I know it's a lot less xp than other infantry and 10xp is really nothing, but it's still double the amount, which means, apart from rank, USA units will also vet up at that speed.

>Plus Pathfinders might get a few nerfs of some sort which balances it out again.

I don't think this is a reasonable answer, might doesn't fix anything, if we know for sure there is a change up the road and we can see that change, then that could be used.

Regarding Pathfinders, I am both for and against but...we should open a new topic for that.

xezon:
We can put this on hold until Pathfinder PR is setup.

MTKing4:
We should discuss pathfinder solutions now then

ExiLe-ZH:
Honestly don't even know what this has to do with Pathfinders, it's a separate issue. We can make valid comparisons without looking at Pathfinders. 

The following things make very little sense:
- TH's give off 4x more XP.
- ListeningOutpost with 2TH gives off same XP (40) as Veteran ATC.
- AttackOutpost with 3TH gives off 60xp vs ATC (40 or 80xp).
- Veteran ATC gives off same XP as Veteran TC.
- 2 RedGuards from Barracks gives off 2x more XP than 1 MG with same BuildTime.

|Unit|1.04|Patch|
|----|----|------|
ListeningOutpost (2 TH) | 40xp | 40xp |
AttackOutpost (3 TH)| 60xp | 60xp |
Veteran TC (8 RG) | 40xp | 40 xp |
Veteran ATC (8 MG) | 40xp | 80xp |
Elite ATC (8 MG) | 80xp | 160xp |
Full AttackOutpost (5+3 MG) | 115xp | 130xp |
Full AttackOutpost (5+3 EliteMG) | 130xp | 160xp |

*Outpost is typically loaded with 6+2MG or 5+3MG. I chose 5+3 since XP change has more effect here.

If INF is so worried about giving away XP against USA, he can skip MG Elite Training and it's same as before. It's a little nerf, but Veteran MGs don't suck, they still rip through Chinooks, especially with horde. Nerf might be needed since INF is already highly positive in this MU, the Doortime and other general China buffs might swing it further into INFs favor.

So before jumping to the Pathfinder argument, please proof that the above made points (don't) make sense.

ImTimK:
Alternative model could be 10 10 15 30, which also makes things a little more balanced, although makes less sense.

commy2:
Table ^ is wrong, because empty Troop Crawlers and Outposts already give 5 XP in patch.

ExiLe-ZH:
Not yet even sure if I like that change.

commy2:
It must be considered though: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/778

ExiLe-ZH:
Alternative model could be 10 10 15 30, that would make:
- Veteran ATC 80 instead of 40 XP.
- Elite ATC 120 instead of 80 XP.

Doesn't sound too bad either, although makes less sense than the original proposal when comparing to all the other (inferior) infantry units.

PR #1154, opened 2022-09-07
Fix: Use regular size Muzzle Flash effects for China Battlemaster Tank

xezon:
This change replaces `WeaponFX_GenericTankGunNoTracerSmall` with `WeaponFX_GenericTankGunNoTracer` on all Battlemaster tanks.

## Rationale

Original Overlord, Crusader, Paladin, Marauder use regular Muzzle Flash.
Original Battlemaster, Scorpion use small Muzzle Flash.

This makes no sense. Battlemaster Tank deals as much damage as Crusader, Paladin and Marauder, and the gun is also similarly sized, so the particle effect should look similar too.

Scorpion Tank using small Muzzle Flash is justified, because it has a small gun which deals -66% damage.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -305,8 +305,8 @@ Weapon BattleMasterTankGun
   WeaponSpeed             = 400                           ; dist/sec
   WeaponRecoil            = 5                            ; angle to deflect the model when firing
   ProjectileObject        = BattleMasterTankShell
-  FireFX                  = WeaponFX_GenericTankGunNoTracerSmall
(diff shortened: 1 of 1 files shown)


ExiLe-ZH:
Noice

commy2 (approved):
technical aspects ok

commy2:
Note: Tank_BattleMasterTankGun is the Autoloader upgrade. Tankgen otherwise uses vChina weapon before that.

Both `WeaponFX_HeroicGenericTankGunNoTracer` and `xxxSmall` use NORMAL camera shake. Should Autoloader weapon camera shake be reduced considering it is 3 shots per unit at once and thus stacking cam shake that becomes annoying?

xezon:
I checked. Autoloader shake looks ok to me.

PR #1155, opened 2022-09-07
Change: Reduce USA Burton shroud range from 500 to 400

xezon:
This change reduces USA Burton shroud range from 500 to 400. This matches the other Hero units, GLA Jarmen Kell and China Black Lotus. With "Search And Destroy" battle plan, Burton will have a shroud range 480 instead of 600, still much larger than the other Heros.

# Rationale

The USA factions have the best scouting abilities.

Common Infantry:

| Object          | Shroud Range |
|-----------------|--------------|
| USA Ranger      | 400          |
| GLA Rebel       | 300          |
| China Red Guard | 200          |
| USA Rocket Man  | 400          |
| GLA Rocket Man  | 400          |
| China Rocket Man| 400          |

There are only 4 ground units across all factions that have above 400 shroud range

| Object          | Shroud Range | "Search and Destroy" Shroud Range |
|-----------------|--------------|-----------------------------------|
| USA Scout Drone | 500          | 500                               |
| USA Burton      | 500          | 600                               |
| GLA Radar Van   | 500          | (600)                             |
| China Outpost   | 500          | (600)                             |

The USA Burton with a vision range of 500, or most commonly 600 with Search and Destroy, can reveal entire bases on its own, giving USA too much intel on the enemy movement and making it too easy in return to react to the enemy movement.

The reduction of the Burton's vision range takes off his edge as the best infantry scout in the game and will now require USA to work a bit harder to see more of the map, either by using its vast scan abilities or deploying Pathfinders, Sentry Drones or vehicles with Scout Drones.

## Original, 500

shot_20221102_203125_1

## Patched, 400

shot_20221102_202911_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -3053,7 +3053,7 @@ Object AirF_AmericaInfantryColonelBurton
   End
 
   VisionRange         = 150
-  ShroudClearingRange = 500
+  ShroudClearingRange = 400 ; Patch104p @tweak from 500 to match with other Hero units
   Prerequisites
     Object = AirF_AmericaBarracks
     Object = AirF_AmericaStrategyCenter
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -293,7 +293,7 @@ Object AmericaInfantryColonelBurton
   EditorSorting       = INFANTRY
   TransportSlotCount  = 1                 ;how many "slots" we take in a transport (0 == not transportable)
   VisionRange         = 150
-  ShroudClearingRange = 500
+  ShroudClearingRange = 400 ; Patch104p @tweak from 500 to match with other Hero units
   Prerequisites
     Object = AmericaBarracks
     Object = AmericaStrategyCenter
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -16502,7 +16502,7 @@ Object Boss_InfantryColonelBurton
   End
 
   VisionRange         = 150
-  ShroudClearingRange = 500
+  ShroudClearingRange = 400 ; Patch104p @tweak from 500 to match with other Hero units
   Prerequisites
     Object = Boss_Barracks
     Object = Boss_ParticleCannonUplink
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -2609,7 +2609,7 @@ Object Lazr_AmericaInfantryColonelBurton
     DamageFX        = InfantryDamageFX
   End
   VisionRange         = 150
-  ShroudClearingRange = 500
+  ShroudClearingRange = 400 ; Patch104p @tweak from 500 to match with other Hero units
   Prerequisites
     Object = Lazr_AmericaBarracks
     Object = Lazr_AmericaStrategyCenter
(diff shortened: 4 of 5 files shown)


commy2 (approved):
implementation correct

Jundiyy:
Seems okay, although didn't bother me how it was.

OoE-ExCaL:
Has this ever worried you when playing ZH? There seems to be, not just on this particular proposal but on many, a desire to uniform and match the factions. This shroud change lines it up with black lotus and Jarmen Kell. The same logic is applied to opening and closing doors. I simply don't agree with this logic. The factions are different and therefore should have differences. The three heros have completely different heroic abilities - like not even close how they operate and interact. Why should their shroud range match? I don't understand the philosophy behind the change, and why it needs to be changed. 

Think about when you actually use burton within a game. You need a strategy centre. USA as a faction has an advantage of that little bit extra vision. That is unique to the USA. Burton is either used within bases to do some infrastructure damage, or combined with your humvee army to pick off long range units. You need that vision as part of the interaction to take out buggies or whatever else long range.

xezon:
I have appended more information in first post to clarify what the motivation for this change is. If we find that Burton is too weak with reduced shroud range, then it can be reverted.

ExiLe-ZH:
I'm not too sure either if consistency for the sake of consistency is enough motivation for such changes. Yes, 500 is A LOT (400 is still quit a bit) and USA indeed has many more ways to get vision, but like Excal said, that's also an advantage USA has by nature.

Let's think more about actual situations though, for example imagine Burton running into a Gat/Outpost or Quad/Radarvan, Jarmen has the ability to instantly snipe and get away, but Burton would mostly likely lose this fight without veterancy. Having that extra vision might help Burton not getting into those situations in the first place.

xezon:
Black Lotus with 500 would have made some sense. It is a Scout Hero with Stealth Detection. Plus China has no Map scans apart Internet Center. Burton is no Scout Hero. He is Rambo.

commy2:
500 is comically large, especially because in practice it becomes 600 thanks to Search And Destroy being the default choice and Burton being post Strat Center exclusive. 

I think other long vision infantry like rocketeers or non-Burton heroes are a good reference point.

Never thought this change was necessary though, hence my approval only mentioning the implementation being correct.

xezon:
I appended the Search and Destroy note to the post above. As far as I am aware Scout Drones do not benefit from the bonus. This would make Burton the single best visionary on the ground of all factions.

commy2:
600 is Chinook range. Afaik, Chinook doesn't get S&D bonus either, as it flies.

ExiLe-ZH:
So with this change Burton shroud range is basically always 480? Since S&D is mostly preferred?

xezon:
Yes, with Search And Destroy 480 instead of 600. Still 80 more than Lotus and Jarmen.

MTKing4:
I think it's a good change, can revert it back if it affects burton badly, highly unlikely though

xezon:
Excal:
> From CCG to ZH, EA gave USA a spy drone at start of game, something USA didn't have in CCG (just a scan). My opinion is that I feel it was intended to give USA a "vision" advantage over the other factions, which is why I am against unifying all the scout and vison range of the hero units. It was said to me that "this Burton range is too large, and a mistake by EA to give burton this extra range". My question is - how do we know it was a mistake? If you look at the additional spy drone the USA received in ZH, along with other elements such as Intelligence Ability from strategy center, sentry drones etc etc, then to me this infers the opposite, and EA clearly wanted USA to be the "vision faction".

> This is one question I had with regard to a lot of the balance suggestions - why do all the numbers have to be uniform and equal? This is the point of having different factions no?

commy2:
USA did not receive the Spy Drone in Zero Hour.

shot_20230804_180748_1

> This is one question I had with regard to a lot of the balance suggestions - why do all the numbers have to be uniform and equal? This is the point of having different factions no?

Totally agree. Making the vision ranges uniform across factions just to make them uniform is no good. The reason to reduce Burton vision range is that it is excessively large, especially when compounded with S&D. It doesn't matter what values other units have.

MTKing4:
I don't think anyone intended to uniform numbers just for the sake of uniformity, there are balance reasons behind it, it was just a comparison to get insight on how much the range should be.

one of the main reasons why USA excels vs other factions is their vision, and their hero is the easiest to sneak and is probably the most devastating hero in the game, not only he kills buildings but also has a chance of being undetected while doing so due to the lack of announcements, (which black lotus has) his value increases exponentially which each building he kills, i think, when it comes to heros, having equal vision with other heros is fair, this doesn't make him any less unique than before.

And this whole EA argument, if you follow through with that, you will always get a broken game, look at all their other games, literally not one balanced game, RA2 for example gave The Allied (Basically USA) a radar that reveals the whole map all the time, and look at that game's balance, Allied wrecks everyone (at least before Yuri showed up) sorry but their philosophy is stupid, and unbalanceable, and we shouldn't adopt it.

PR #1161, opened 2022-09-08
Replace C comments with INI comments

xezon:


commy2 (approved):
Changes correct, not checked for completeness

PR #1162, opened 2022-09-08
USA Sentry drone improvements 2

### LINKED ISSUE #862: Sentry Drone is rarely used

xezon:
* Fixes #862
* Closes #1042

This change improves the USA Sentry Drone.

* Increases gun damage by 20%, from 40 to 50 DPS
* Decreases gun upgrade cost by 20%, from 1000 to 800
* Decreases gun upgrade time by 66%, from 30 to 10 seconds
* Replaces terrain tracks with less transparent ones
* Removes deployment
* Decreases build cost to match with Humvee cost
* Decreases acceleration time by 70%, from 1000 to 300 milliseconds
* Increases sight by 30%, from 350 to 400 radius

# Rationale

## Build Cost

There is no way the Sentry Drone is more valuable than Humvee. It does have Stealth detection and Stealth capability, but it spawns without gun by default, has no transport slots and cannot buy drones. It is safe to match price with Humvee.

| Object               | Price Regular | Price Laser | Price Air | Price Swg |
|----------------------|---------------|-------------|-----------|-----------|
| Original Humvee      | 700           | 700         | 800       | 850       |
| Original Drone       | 800           | 800         | 850       | 1000      |
| Patched Drone (this) | 700           | 700         | 800       | 850       |

## Locomotor

Sentry Drone And Humvee have very similar locomotor. Both have same top speed and turn speed. There are 2 differences: Sentry Drone drives faster than Humvee when damaged and has lower acceleration than Humvee.

The faster SpeedDamaged looks reasonable. It is a scout unit, so ability to flee the scene quick is nice perk. However, the slower acceleration is not something it needs. It takes 1.00 seconds to accelerate, as opposed to 0.06 seconds for Humvee.

| Object                      | Time until top speed (ms) |
|-----------------------------|---------------------------|
| Original Humvee             | 60                        |
| Original Sentry Drone       | 1000                      |
| Patched Sentry Drone (this) | 300                       |

## Deployment

Sentry Drone has a depoyment time of 1000 millisecond for both packing and unpacking. That means it requires 1 second to acquire a fresh target, and 2 seconds between engaging two targets if it needs to move to new position. This is a huge time and performance penalty. And not just that: It will also make the unit feel clunky and unresponsive, which is comparable to how units feel unresponsive in laggy matches. And no one likes laggy matches. Therefore, a removal of the deployment is reasonable. A small redeployment time may be readded if we can get the assets or sounds for that, or if it is needed to rebalance the unit. Originally the Sentry Drone does not communicate the deployment at all, unlike the China Nuke Cannon.

## Machine Gun Weapon

Machine Gun strength of Humvee and Sentry Drone is comparable. Sentry Drone deals 20% less damage, but the vehicle tanks 30% more damage, because Sentry Drone has more health. This looks fair. However, Sentry Drone is not trainable, meaning it will not benefit from veterancy buffs like Humvee does.

| Gun                             | VETERAN DPS | ELITE DPS | HERO DPS |
|---------------------------------|-------------|-----------|----------|
| Original SentryDroneGun         |             |           |          |
| Original HumveeGun              | +10%        | +80%      | +95%     |
| Original HumveeMissileWeapon    | +10%        | +20%      | +30%     |
| Original HumveeMissileWeaponAir | +10%        | +20%      | +30%     |

Humvee gun gains an astonishing 80% damage boost on Vet 2 with its Machine gun. This will outperform the Sentry Gun more than twice. It is reasonable to raise the Sentry Drone Gun damage, matching with Humvee Gun damage, so that at least in early stages of match the Humvee and Sentry Drone can deal the same damage for the same price.

| Gun                             | Damage per Second |
|---------------------------------|-------------------|
| Original HumveeGun              | 50                |
| Original HumveeGun Vet 1        | 55                |
| Original HumveeGun Vet 2        | 90                |
| Original HumveeGun Vet 3        | 97.5              |
| Original SentryDroneGun         | 40                |
| Patched SentryDroneGun (this)   | 50                |

## Upgrade

Sentry Drone Gun has some things going against it. Drone starts without gun by default, but Humvee has it always. Humvee can get a second gun for 20% less cost at same build time. And that second gun is really good, because it does not just target ground units, but also air units.

| Gun                             | Damage per Second | Type            |
|---------------------------------|-------------------|-----------------|
| Original SentryDroneGun         | 40                | COMANCHE_VULCAN |
| Original HumveeGun              | 50                | COMANCHE_VULCAN |
| Original HumveeMissileWeapon    | 15                | EXPLOSION       |
| Original HumveeMissileWeaponAir | 25                | EXPLOSION       |

Structures and light vehicles have around 50% COMANCHE_VULCAN and 100% EXPLOSION, and tanks have 25% COMANCHE_VULCAN and 100% EXPLOSION. EXPLOSION is a much better damage type than COMANCHE_VULCAN. We can roughly say that Humvee Damage Output vs all things on ground is effectively doubled.

Since SentryDroneGun is not better than HumveeGun, there is no justification to price it at 1000. 800 to match it with TOW Missile Upgrade is fair. As for build time, it looks fair to halve it from 30 seconds to 15 seconds, because it is nowhere near as good as TOW, since it cannot shoot on Aircraft. Another 5 seconds can be cut to compensate for it requiring an upgrade to begin with to do any sorts of combat.

| Upgrade                   | Cost | Time |
|---------------------------|------|------|
| Original TOW Missile      | 800  | 30   |
| Original Sentry Gun       | 1000 | 30   |
| Patched Sentry Gun (this) | 800  | 10   |

## Shroud Clearing Range

| Object                           | Shroud Clearing Range | Area   | Advantage over Sentry Drone |
|----------------------------------|-----------------------|--------|-----------------------------|
| Original USA Sentry Drone        | 350                   | 384845 |                             |
| Original USA Scout Drone         | 500 (+75+10 wander)   | 785398 | +104%                       |
| Original USA Humvee              | 320                   | 321699 | -17%                        |
| Original USA Missile Defender    | 400                   | 502654 | +30%                        |
| Original USA Pathfinder          | 400                   | 502654 | +30%                        |
| Original China Outpost           | 500                   | 785398 | +104%                       |
| Original China Lotus             | 400                   | 502654 | +30%                        |
| Original GLA Radar Van           | 500                   | 785398 | +104%                       |
| Patched USA Sentry Drone         | 400                   | 502654 | +30%                        |

Sentry Drone has slightly better Shroud Clearing Range than Humvee, but it is completely dwarfed by Scout Drone (100 cost). An increase to 400 is reasonable.

### Original Sentry Drone with sight range 350

shot_20220908_150857_1

### Patched Sentry Drone with sight range 400

shot_20220908_150623_1

# Terrain Tracks

Original Sentry Drone leaves Tank Tracks. Tank tracks have opacity of 160/255, while Tire tracks have opacity of around 88/255. China Outpost Tracks are therefore much less visible than Sentry Drone tracks. Sentry Drone gets new Tracks that make it less visible than Tanks and Outpost.

shot_20220908_120339_1


Jundiyy:
Seems to read good.

xezon:
Further tweaks to Sentry Drone are still possible, but these changes should work as a baseline to make it not as useless as originally.

ExiLe-ZH:
| Object          | Shroud Range | "Search and Destroy" Shroud Range |
|-----------------|--------------|-----------------------------------|
| USA Sentry Drone | 400          | 400                               |
| USA Ranger | 400          | 480                               |
| USA Missile Defender | 400          | 480                              |
| USA Scout Drone | 500          | 500                               |
| GLA Radar Van   | 500          | (600)                             |
| China Outpost   | 500          | (600)                             |

I still don't understand why the Sentry Drone's Shroud Range can't compete with other units. It doesn't profit from S&D and without MachineGun it's only utility is scouting and spotting, while still being as expensive as an Humvee. 

For example, a good potential use-case of a non-upgraded Sentry Drone would be to use it to support your main army, it would give the army the ability to look ahead further and spot treats like demo-traps earlier. Right now one $100 Scout Drone can do that better.

Quote from ZH manual: 
“_This stealthed vehicle provides forward-looking eye that doesn’t  need much coordination or supervision. When the drone is upgraded, a  nose-mounted 20mm machine gun is automatically deployed against infantry  and small vehicles_"

xezon:
It is fair point. You can open new Pull Request with new setup if so desired. This change aimed to elevate Sentry Drone to more reasonable values to setup a useful foundation for it. 400 over 350 here was not a big improvement, but it is an improvement.

Stubbjax:
I believe making the Sentry Drone more like a Humvee was / is the wrong direction to go. It would be far more interesting if its utility was diversified to increase viability, rather than trying to imitate the Humvee.

ExiLe-ZH:
> I believe making the Sentry Drone more like a Humvee was / is the wrong direction to go. It would be far more interesting if its utility was diversified to increase viability, rather than trying to imitate the Humvee.

I agree, therefore in my opinion the focus should be on scout or stealth abilities rather than firepower. 

One other idea I had in the main topic was to give it the ability to climb hills for example. That would be a nice pro to diversify it further.

ExiLe-ZH:
I suddenly had an alternative and simple idea to make the Sentry Drone a decent unit.

Basically make it always stealth, also when moving. With this there's no need to buff anything else really, like a vision upgrade wouldn't be needed anymore as it can scout enemy terrain much easier now. 

The Machine Gun at the original price/build time will become worthy, because now it can reach it's target (like Hackers or Workers) much easier.

xezon:
It is an interesting idea. Would make it useful as a sneaky harass and scout unit. Though perhaps it will be really annoying to deal with when there are a lot of these on the map sniping Workers and denying Heros und stealthed units from advancing. It could be bad news for GLA and China.

ExiLe-ZH:
> It is an interesting idea. Would make it useful as a sneaky harass and scout unit. Though perhaps it will be really annoying to deal with when there are a lot of these on the map sniping Workers and denying Heros und stealthed units from advancing. It could be bad news for GLA and China.

Yup, but will it be overpowered? I'm not sure, because it's still an alternative, since it's not cheap and at the expense of a lot WF production time.

Could even increase the Cost/BT if needed.

But yea, I just like the idea of having an actual good alternative to surprise the enemy. Like for example you could produce and sent an unarmed drone to the enemy right at the start of a match, then get the MG upgrade later. You could camp a barracks with it or harrass workers indeed.

Perhaps we can make a test branch with this change?

MTKing4:
Would be nice if we stop revisiting resolved topics and focus more on the unresolved ones :P 

We should leave this for now and only revisit it if our current setup doesn't satisfy, that would be a good practice to have

ExiLe-ZH:
I was thinking about the Vanilla Hijacker case which is still unresolved and then thought of this. I just really like the idea of having a better unit but less accessible. And prefer that over the concept of having a less effective unit, but more accessible.

MTKing4:
But your approach has more impact on balance which is something we were trying to avoid, i also think it's a better design to make it a decent and accessable unit, sentry drone is not meant to be a hero unit to be that strong and inaccessible

xezon:
One huge pain point could be if the agile stealthed Sentry Drones just camp to try kill high value scaffolds, such as Command Center, War Factory or even Superweapons.

ExiLe-ZH:
True those are painfull losses, therefore it would probably be best for drones to be fairly audible when moving, for players who pay attention atleast. 

It wouldn't be the only unit that could do this though, Burton produces faster and is cheaper (in total) and is completely silent. It can even climb mountains to avoid enemy defenses. Also Stealth Commies can do the same aswell but ofcourse are audible at all times.

Anyways, if I have time I'll try to make such branch for testing purposes.

xezon:
Burton, Stealth Rebel, and GPS Scrambled units can do this. But infantry is slow, Burton is singular and GPS is late game. Sentry Drone would be in a league of its own.

MTKing4:
I think having an all time stealthed units is a bad design decision in its entirety, a lot of zh units are abusing this broken attribute, it will only make matters worse to add more units into that category

Check this article for more insight
https://waywardstrategy.com/2023/06/26/fixing-stealth-in-rts/?fbclid=IwAR1gePTXtbJVN4GHK1RNwHi8FELyV523nkcP_45kGsycCpoThZZslZP97Bg

ExiLe-ZH:
> I think having an all time stealthed units is a bad design decision in its entirety, a lot of zh units are abusing this broken attribute, it will only make matters worse to add more units into that category
> 
> 
> 
> Check this article for more insight
> 
> https://waywardstrategy.com/2023/06/26/fixing-stealth-in-rts/?fbclid=IwAR1gePTXtbJVN4GHK1RNwHi8FELyV523nkcP_45kGsycCpoThZZslZP97Bg

It's all subjective anyway, this mega long article still can't objectively conclude it's bad design. It's all about balancing the strengths and weaknesses anyway. I personally like stealth abilities in rts and how it can diversify units.

ExiLe-ZH:
> Burton, Stealth Rebel, and GPS Scrambled units can do this. But infantry is slow, Burton is singular and GPS is late game. Sentry Drone would be in a league of its own.

"A league of it's own" as in better in every way?

xezon:
Better at killing scaffolds I mean. Early in game. For example it could kill tech building scaffold.

xezon:
Btw, I think it is better to not discuss new changes in merged pulls, because conversation here is meant to stop and is also not part of active work log. It is better to create a new issue report and link the the relevant pull(s) to continue from.

ghost:
I'm not a fan of these changes. I remember in a Dominator match, the Sentry drones were absurdly powerful when there are many of them.
I don't think their damage should be buffed. They should be used mainly for scouting.
I gave some of my ideas here: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/2524#issuecomment-2692257831

PR #1163, opened 2022-09-08
Fix: Crusader Wreck Spawned Before Model Disappears

commy2:
From NPM

1.04:

- 50% chance the Crusader flings away the turret on death, if this happens, the wreck is spawned before the original object is deleted, so two models intersect for ~8 frames.

patch:

- wreck model spawns at the same time as when original model is deleted

Also changed CINE, because otherwise that unit would be broken with this change.
The fix was done by removing the OCL that spawns the wreck at MIDPOINT and instead adds the entirety of OCL_GenericTankDeathEffect to OCL_CrusaderTurret.

OCL_GenericTankDeathEffect has a bad name. It is specifically for the Crusader, not a generic tank.


xezon (approved):
Looks good to me.

PR #1164, opened 2022-09-08
Fix: USA Pilots Can Promote GLA Combat Bikes

commy2:
This change prevents USA Pilots from promoting GLA Combat Bikes.

# Rationale

It makes no sense, and it leads to promoted units that should never promote, such as Workers and Saboteurs. These units do not work correctly with promotions, meaning they do not receive any benefits from promotion, not even health or regen, so nothing is lost here.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4280,7 +4280,7 @@ Object AirF_AmericaInfantryPilot
 
   Behavior = VeterancyCrateCollide       ModuleTag_03
     RequiredKindOf = VEHICLE      ; we only give our bonus to VEHICLEs we collide with
-    ForbiddenKindOf = DOZER       ; but not to TRANSPORTs or DOZERs!
+    ForbiddenKindOf = DOZER CLIFF_JUMPER       ; but not to TRANSPORTs or DOZERs! Patch104p @bugfix commy2 08/09/2022 Adds CLIFF_JUMPER to prevent Pilot from promoting Combat Bikes.
     EffectRange = 0               ; 0=="affect only the thing you collide with"
     AddsOwnerVeterancy = Yes      ; we add our own veterancy to the target (rather than just +1 level)
     IsPilot = Yes                 ; set the pilot flag because it's different than the veterancy crate and has extra checking
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1534,7 +1534,7 @@ Object AmericaInfantryPilot
 
   Behavior = VeterancyCrateCollide       ModuleTag_03
     RequiredKindOf = VEHICLE      ; we only give our bonus to VEHICLEs we collide with
-    ForbiddenKindOf = DOZER       ; but not to TRANSPORTs or DOZERs!
+    ForbiddenKindOf = DOZER CLIFF_JUMPER       ; but not to TRANSPORTs or DOZERs! Patch104p @bugfix commy2 08/09/2022 Adds CLIFF_JUMPER to prevent Pilot from promoting Combat Bikes.
     EffectRange = 0               ; 0=="affect only the thing you collide with"
     AddsOwnerVeterancy = Yes      ; we add our own veterancy to the target (rather than just +1 level)
     IsPilot = Yes                 ; set the pilot flag because it's different than the veterancy crate and has extra checking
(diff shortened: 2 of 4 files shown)


Jundiyy:
> or Saboteurs. These units don't work correctly with promotions, meaning they don't receive any benefits from promotion (not even health or regen), so nothing is lost here.

This is not correct.
Firstly, the Stealth Sabo can vet up, it has to go over crates, it doesn't get a better weapon, but it will have more health and auto heal.

Secondly, all units that vet up even though they usually can't, still gain more health. You can check vehicles or infantry.

And lastly, a vet 3 GLA Sabo will auto repair on a Bike (this is because the Bike is repairing, not the Sabo), but once off the bike, he will not heal but still have more health than a regular Sabo.

commy2:
I looked it up and only Stealthgen Saboteur is trainable. vGLA Saboteur is not trainable.
(Demogen Saboteur is also trainable, WB only)

Good point about repairing Bike though.
Still doesn't make sense to put Pilots on Bikes.

Jundiyy:
> I looked it up and only Stealthgen Saboteur is trainable. vGLA Saboteur is not trainable. (Demogen Saboteur is also trainable, WB only)

Yes, that's why I mentioned Stealth :) 'Stealth Sabo'

> 
> Good point about repairing Bike though. Still doesn't make sense to put Pilots on Bikes.

Yup, it does feel odd as the Pilot just becomes the unit, but it's not been a problem.

xezon:
I do not mind either way, though if Pilot can be put into any other vehicle, then it is also ok to put it in Bike. Generally this is not seen as a problem in matches.

How would you like to proceed with this? Also, can we remove `Bug` label? To me it seems `Design` question, not Bug per se.

commy2:
It's obviously a bug, because it can be used to promote Workers and other untrainable units.

xezon:
Ah, because they have set `IsTrainable = No`. The proper fix then would be to have Pilot be able to enter Bike with specific Passengers only. But that likely is Game Code.

commy2:
Yeah. Although even that seems strange to me. Where does the Pilot go? He's obviously not the rider and a Bike has no other crew unlike any other vehicle.

xezon:
He hides in the Bike Cargo Box and gives instructions.

Stubbjax:
> The proper fix then would be to have Pilot be able to enter Bike with specific Passengers only. But that likely is Game Code.

If we're talking code changes, why not just have the pilot actually ride the bike?

xezon:
Does Bike with Worker gain Veterancy when driving over Scrap?

Stubbjax:
> Does Bike with Worker gain Veterancy when driving over Scrap?

Yes.

https://user-images.githubusercontent.com/11547761/190196611-666ab30c-54d0-4817-a025-c272990c5c32.mp4

Jundiyy:
> > Does Bike with Worker gain Veterancy when driving over Scrap?
> 
> Yes.
> 
>  RANK.mp4

Yes, every GLA unit does. Even if they can't otherwise. Vet 3 Demo Bike is very powerful.

Bike will auto repair, but once the unit gets off the bike, it won't auto heal if it usually doesn't gain vet.

xezon:
It is odd that infantry on bike can rank up but on foot it cannot. Best case it would be consistent on Bike and on Foot, or infantry would need to lose veteran after disembarking bike.

Stubbjax (approved):
I think it's reasonable to remove this weird behaviour.

1. It makes no logical sense that the pilot is consumed by the bike.
2. No other USA infantry can ride bikes.
3. It's super rare so it's not like anyone is going to miss it.
4. Promoting infantry with pilots feels like an exploit more than anything.

Jundiyy:
Other than Worker, all infantry vetting up (in any way) wouldn't feel odd imo, on foot or bike. Hence why I mentioned it in the Sabo topic. 

Workers would feel odd since they don't help in combat in any way. And because other such units don't vet up. Such as Dozers and Supply Trucks.
Hijackers, Saboteurs and Terrorists are all combat type units and would feel fine to gain vet. 

I know it's another topic and too controversial, but would have been nice if EA had allowed  Dozers and Trucks vet, to be rewarded for their crushes. A bit like ECMs, can't vet up other than running over Inf.

xezon:
So, if we merge this, then the next logical question is whether or not Bike should be Trainable, because Scrap still gives Vet.

Stubbjax:
> So, if we merge this, then the next logical question is whether or not Bike should be Trainable or not, because Scrap still gives Vet.

Definitely needs to be for cases like Jarmen, which often promotes very quickly when encountering Hackers.

xezon:
Ah right, it needs to be Trainable for the Infantry that can shoot. And Scrap cannot be changed for different types I think. At least not without INI trickery.

commy2:
Probably can make the scrap never level up CLIFF_JUMPER kind of and instead always give money. But that would mean Jarmen and co can no longer vet up by driving over scrap either.

xezon (approved):
Looks good.

Piddox:
Is this issue still relevant if it's already been fixed in Code here: https://github.com/TheSuperHackers/GeneralsGameCode/pull/1970 ?

PR #1166, opened 2022-09-08
Change: Require Stealth Fighter Science To Upgrade Bunker Busters

commy2:
I suppose this could prevent you from upgrade units inherited from a player that surrenders, but that is never going to happen.

Note that AFG spawns with SCIENCE_StealthFighter unlocked automatically in this patch (but not 1.04).

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -1200,9 +1200,12 @@ CommandButton AirF_Command_UpgradeAmericaLaserMissiles
   DescriptLabel           = CONTROLBAR:AirF_TooltipUSAUpgradeLaserMissiles
 End
 
+; @tweak commy2 08/09/2022 Require Stealth Fighter science to upgrade Bunker Busters.
 CommandButton Command_UpgradeAmericaBunkerBusters
   Command       = PLAYER_UPGRADE
   Upgrade       = Upgrade_AmericaBunkerBusters
+  Options       = NEED_SPECIAL_POWER_SCIENCE
+  Science       = SCIENCE_StealthFighter
   TextLabel     = CONTROLBAR:UpgradeAmericaBunkerBusters
   ButtonImage   = SABunkerBust
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is


xezon:
If I understand correctly, player will never be able to buy Bunker Buster Upgrade if unable to purchase Stealth Fighter, but can always buy it when Stealth Fighter is buildable, yes?

> I suppose this could prevent you from upgrade units inherited from a player that surrenders, but that is never going to happen.

Can you clarify this? It is not clear to me.

commy2:
> If I understand correctly, player will never be able to buy Bunker Buster Upgrade if unable to purchase Stealth Fighter, but can always buy it when Stealth Fighter is buildable, yes?

Sort of, except that AFG Stealth Fighters do not require a science in neither 1.04 nor this patch.

> Can you clarify this? It is not clear to me.

Lets say you inherit an Airfield from AFG from a team mate that surrendered. You will be able to train AFG Stealth Fighters, but you will not be able to upgrade them, unless you happen to be USA yourself and either got the science for free (AFG in patch) or picked it by spending a genpoint (vUSA/Laser/SWG).

A science requirement for AFG Stealth Fighters was deliberately not added previously, because in 1.04, anybody can produce AFG Stealth Fighters if they manage to acquire an AFG Airfield.

Options are: 
- Lock AFG Stealth Fighters behind the science. This will not affect AFG, because they spawn with the science in this patch, but it will prevent e.g. GLA from building Stealth Fighters from AFG Airfields.
- Make a unique upgrade research button for AFG Airfields that does not require the Stealth Fighter science.
- ignore issue
- reject this change

The same applies to Nukegen Nuke Cannons. Except the problem is even more complicated there, because Nuke Cannon and Neutron Shells are build/researched from different buildings.

xezon:
> Make a unique upgrade research button for AFG Airfields that does not require the Stealth Fighter science.

This reads like the appropriate course of action if I understand correctly. This way anyone capturing the AFG Airfield can buy Stealth Fighter and Bunker Buster upgrade, just like originally.

commy2:
But what about Neutron Shells and Nuke Cannons? They are researched from different buildings.

xezon:
What is the problem with different buildings? I assume Nuke General also needs a new Button for Neutron Shells without requirement?

commy2:
Yes. The connection between Nukegen War Factory and Nukegen Nuke Silo is just not as close as being literally the same building (AFG Airfield).

MTKing4:
I don't think this is a good change, there's no reason to prevent players from buying the upgrade before aquiring stealth fighters, could be used to save time buying it now and wait to get a gen point to spend on stealth fighters, instead of being forced to upgrade it after.

Also i see no issue if other factions could get BB stealth fighters from Air or Nuke cannons from nuke, i think it's a nice and fun quirk tbh. the neutron thing doesn't even matter, you'd need to capture two buildings or invest in a nuke + upgrade if you got a nuke CC, by that time you kinda deserve to have it.

xezon:
Good point. We can close this and the other one.

xezon:
Close?

Jundiyy:
Don't see a need for this. Allow the players to purchase whatever upgrades they wish.

PR #1167, opened 2022-09-08
Change: Require Nuke Cannon Science To Upgrade Neutron Shells

commy2:
I suppose this could prevent you from upgrade units inherited from a player that surrenders, but that is never going to happen.

Note that Nukegen spawns with SCIENCE_NukeLauncher unlocked automatically in this patch (but not 1.04).


xezon:
> I suppose this could prevent you from upgrade units inherited from a player that surrenders, but that is never going to happen.

Can you clarify this? It is not clear to me.

ExiLe-ZH:
I guess so you can't accidentally purchase neutron shells upgrade if you didn't unlock nuke cannons first.

commy2:
See discussion in https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1166

PR #1170, opened 2022-09-09
Fix: Stealth General Saboteur Can Level Up

commy2:
vGLA Saboteur can not level up. Saboteur also has no weapon, so it should instead just pick up money from salvage crates like Workers, Terrorists and Radar Vans do.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -17082,9 +17082,10 @@ Object Demo_GLAInfantrySaboteur
   BuildCost = 800
   BuildTime = 15.0          ;in seconds
 
+  ; @bugfix commy2 09/09/2022 Fix non vGLA Saboteurs level up by walking over crates.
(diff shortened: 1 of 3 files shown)


Jundiyy:
Whilst it's not the end of the world to make this change, it would be nice to keep, if you can get a more hp + auto regen sabo then why not?

commy2:
It's not a combat unit. That's like asking why Workers don't vet up, even though you'd get health and regeneration from that.

Jundiyy:
Hmmm, that's true.
Although I can still vet up Workers without Pilots :)

xezon (approved):
Makes sense to me. There is no need to keep such inconsistency.

Piddox:
Is this issue still relevant if it's already been fixed in Code here: https://github.com/TheSuperHackers/GeneralsGameCode/pull/1970 ?

Stubbjax:
> Is this issue still relevant if it's already been fixed in Code here: [TheSuperHackers/GeneralsGameCode#1970](https://github.com/TheSuperHackers/GeneralsGameCode/pull/1970) ?

Yes, it can still walk over crates to rank up without requiring a bike to do so.

PR #1171, opened 2022-09-09
Fix: Particle Cannon Loops Deploy Animation When Finished While Disabled

commy2:
Fix is currently only applied to vUSA. Please confirm this solves the issue. 

Repro: 
- Start building Particle Cannon
- Shut it down with EMP or Microwave Tank, but continue building
- After it is finished 100%, let it get re-enabled

The PC should now loop the deploy animation. It does no longer do that after this patch.
Note however, that the super weapon timer will be bugged out regardless of this change.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -10032,7 +10032,7 @@ Object AirF_AmericaParticleCannonUplink
     AliasConditionState  = NIGHT SNOW AWAITING_CONSTRUCTION
     AliasConditionState  = SNOW AWAITING_CONSTRUCTION
 
-    ; @bugfix commy2 09/09/2022 Fix Particle Cannon loops deploy animation when finished while disabled.
+    ; Patch104p @bugfix commy2 09/09/2022 Fix Particle Cannon loops deploy animation when finished while disabled.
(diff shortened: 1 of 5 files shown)


Stubbjax:
The issue can also be reproduced by completing the Particle Cannon whilst out of power.

(Also note that Saboteurs that sabotage a super weapon whilst it is out of power will add the time since its power went out onto the super weapon’s current countdown timer.)

xezon:
Tested. Worked. Animations look also correct when firing afterwards.

commy2:
Nice. Don't ask me why this works. Will implement for the remaining Particle Cannon variants.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -10032,8 +10032,11 @@ Object AirF_AmericaParticleCannonUplink
     AliasConditionState  = NIGHT SNOW AWAITING_CONSTRUCTION
     AliasConditionState  = SNOW AWAITING_CONSTRUCTION
 
+    ; @bugfix commy2 09/09/2022 Fix Particle Cannon loops deploy animation when finished while disabled.
```
xezon: Patch104p keyword missing in comments.

commy2: done

xezon (approved):

PR #1176, opened 2022-09-09
Fix: Technical Uses Wrong Muzzle Flash When Equipped With RPG

commy2:
No entry for `WeaponMuzzleFlash` means it defaults to `WeaponMuzzleFlash = MuzzleFX01` from `DefaultConditionState`, but that is for the machine gun, not the RPG.


MTKing4:
With RPG, you mean Scrap 2?

commy2:
Yes, scrap+2 is called "RPG" in code. MG -> Grenade Launcher -> RPG

MTKing4:
Oh i see, what a weird naming

xezon:
Do we already have a video of original and patched?

Stubbjax:
I believe this is the difference between old (left) and new (right).

image

Stubbjax (approved):

PR #1178, opened 2022-09-10
Change: Auto Reload Battery When Idle

commy2:
1.04

- In case the target is destroyed or moved out of range, but the Patriot fired less than four missiles, the next time a target moves into range it will only shoot less than 4 missiles with its first salvo, because the clip is not reloaded until completely empty.

patch:

- The Patriot will reload the entire clip after being idle for ~2 seconds. (This is also how the Rocket Buggy operates for comparision).

Note: In 1.04, the player has always the option to manually empty the clips by force-firing the ground. I have never seen this used in practice, probably because this issue is rather minor, or because the current mechanics are opaque to the players.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1971,6 +1971,7 @@ Weapon AnthraxGammaBombWeapon
   AutoReloadsClip     = No
 End
 
+; Patch104p @tweak commy2 10/09/2022 Auto reload missile battery when idle.
 ;------------------------------------------------------------------------------
 Weapon PatriotMissileWeapon
   PrimaryDamage               = 30.0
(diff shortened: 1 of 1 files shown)


Stubbjax:
Very nice. Should this change also be applied to the assist weapons?

commy2:
__The assist logic should be removed from the game.__

Jundiyy:
I this this is a good change.
If I'm correct, Auto Loader also does this, I'm not so sure if it is good change for the Battle Master, Auto Loaders are used mainly for Tank vs Stealth and Tank mirrors, in which both match ups are quite evenly matched. Maybe open a new topic and see what others think regarding it?

xezon:
Any particular reason 2100 ms was chosen?

commy2:
`AutoReloadWhenIdle` is always `ClipReloadTime` + 100 (at least 1 frame). See BuggyRocketWeapon or ComancheAntiTankMissileWeapon for examples.
Probably so you can't reload faster by idleing, which would be cheese. Not sure why 1 frame more though. I am blindly trusting what EALA uses.

xezon (approved):
Looks good to me.

PR #1179, opened 2022-09-10
Change: Add Blue Missile Defender Texture From CCG

commy2:
This is something from the NProject changelog.

The blue uniform is from CCG, the green uniform is from ZH. The blue better fits the Ranger look, while the green makes it appear like a China unit.

shot_20220910_114450_1
shot_20220910_114101_1


commy2:
No preference.

Jundiyy:
Doesn't make a difference to me. Looking at both together, I get GLA feeling from the green one.

MTKing4:
Blue looks cool, but i think our eyes have adjusted to green

No preference

ExiLe-ZH:
From a normal viewing distance both appear quite grey'ish and it's hard to tell a difference until you zoom in. Either is fine. Blue fits better thematically I guess.

xezon:
There must have been a reason they made an effort to change blue to green, no? Maybe to visually better distinguish the Missile Defender with the Ranger?

commy2:
Probably a leftover from it being a China unit originally. Can read this in this [code comment](https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini#L1101-L1105)

>A little history -- the MissileDefender now uses the assets of the TankHunter.
>The TH uses assets of the GLA tunnel defender, and the TD uses assets of the NEW
>MD. It was a design/art/code decision because the new missile defender animations
>don't suit that of a mobile/offensive unit with packing and unpacking.

In that case, it would be one of the many regressions that happened from CCG -> ZH.

xezon:
That comment is really confusing. It reads as if they swapped things with all 3 Rocket Men. What makes you believe this texture became a bug?

commy2:
Yeah, they rotated the models around. The texture looks similar to what the Red Guard is wearing, which resembles WW2 Chinese Red Army uniforms.

Gen1_Red_Guard-2969640347

The texture is even named "NI" (tag for chiNese Infantry) "Tank Hunter".

xezon:
Here is image with bunch of infantry.

shot_20220913_172423_3

The primary colors appear to be

USA : Light Blue, Green
GLA : Yellow, White, Black
CHI : Green, Dark Blue

The USA Green looks same as China Green. Colonel Burton also has the green pants and backpack. It is reasonable to do something about it.

Stubbjax:
I didn't expect it to make much of a difference, but after trying the blue look for several dozen games, I actually find it looks and feels quite a bit nicer.

xezon:
Moved to Optional bundle. This way this change is safe to go. Squash on Commit.

xezon (approved):

PR #1181, opened 2022-09-10
Fix: Added missing radio sounds to the Helix voice lines.

Stubbjax:
While copy / pasting units, EA clearly forgot to add the aircraft radio sounds to the Helix voice lines. These sounds are played for every other selectable aircraft in the game.

https://user-images.githubusercontent.com/11547761/189490819-59894992-13e1-4e2e-906d-9a71f793d774.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
@@ -4804,8 +4804,12 @@ End
 
 ; Helix
 
+; Patch104p @tweak Stubbjax 11/09/2022 Added missing radio sounds to the Helix voice lines.
+
 AudioEvent HelixVoiceSelect
   Sounds =  vhelsea vhelseb vhelsec vhelsed vhelsee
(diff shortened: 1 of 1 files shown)


commy2 (approved):

ExiLe-ZH:
I've heard this ingame already

commy2:
What have you heard already?

MTKing4:
not a fan

commy2:
Comanche and Chinook have this as well.

ExiLe-ZH:
Nvm, I didn't hear the added sounds on my phone. Neither did read :D

xezon:
Nice.

MTKing4:
If it exists in Chinook and Comanches, and i think I've heard it before on migs, and maybe all other aircrafts have it as well? Then it probably makes sense

Don't know why it sounds weird on helixes though :/

commy2:
It exists on all air units, including Migs. It also only exists for air units, nothing else. Even Spectre has it.

PR #1189, opened 2022-09-13
Fix Burton texture and add blue pants for USA Infantry

xezon:
Related:
* #1179

With this change plus #1179 all USA infantry have blue uniforms. The new look is optional. It also fixes the texture error on the inner thigh of Colonel Burton, for both classic and new Burton texture.

### Fixed green pants

shot_20220913_232809_1

### Blue pants and backpack

shot_20220913_234032_2


xezon:
The gang in blue.

shot_20220914_090444_3

PR #1202, opened 2022-09-14
Remove DisplayName from vehicle hull objects

xezon:
Not sure what it does, but only 5 hulls in Hulk.ini have it, and only 2 of them are even correctly set. Looks like it can be removed.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -395,7 +395,6 @@ Object ComancheRubbleHull
   End
 
   ; ***DESIGN parameters ***
-  DisplayName      = OBJECT:ComancheHull
   Side = America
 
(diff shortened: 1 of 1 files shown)


commy2:
Why bug? And why remove them? Other props have display names when hovered over, like supply stashes and buildings.

xezon:
But are these Props? They are dead hulls that spawn on vehicle kills. Hovering them does nothing. Only 5 of many have DisplayName and 3 of them are incorrectly set. It makes no sense to me.

Stubbjax:
Why comment these out instead of deleting them?

xezon:
No particular reason. Would you prefer removed?

Stubbjax:
> No particular reason. Would you prefer removed?

Preferably removed because the change doesn't otherwise do anything due to the display name already being ignored for those objects. More importantly though, it makes things (more) visually consistent and cleaner.

xezon:
Lines are now removed instead of commented.

Stubbjax (approved):

PR #1205, opened 2022-09-15
Fix and improve dead vehicle hull sink delays

### LINKED ISSUE #1183: USA Humvee has significantly longer sink time than other vehicles
USA Humvee has significantly longer sink time than other vehicles.

https://user-images.githubusercontent.com/4720891/189884359-3d72e993-ea11-46d0-bb6f-5d5e01728fb5.mp4

xezon:
* Fixes #1183

This change fixes and improves dead vehicle hull sink and destruction delays in 2 steps.

1. Unifies all vehicle sink delays to common 1500 ms, sink speed to 2 m/s and destruction delays to whenever the model is fully sunken into terrain. A few units had very long sink times, such as Humvee, Microwave and Overlord. They are all the same now. Except for Flame Tank, because that one has a special death particle that lasts very long, so dead hull should stay with it. For GLA Battle Bus and Combat Bike the destruction timing does not work correctly originally and patched. It will disappear before sunken into terrain. Helicopters originally had slow sink speed of 1 m/s or fast sink speed of 4 m/s. This has been consolidated to 2 m/s for all hulls. GLA Rocket Buggy, Technical and Combat Bike (on kill) do not sink into terrain, as instead they just disappear on explosion, so their setups are unchanged.

2. Increases all sink and destruction delays by 1500 ms, except for flying Drone hulls and Combat Bike hull. This way dead hulls can be seen properly after explosion particles for a bit of time, otherwise hulls can barely be seen on various vehicle kills. The new sink delay of 3000 ms is also consistent with all Infantry, which is 3000 ms as well.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -38,9 +38,9 @@ Object DeadMicrowaveHulk
   End
 
   Behavior = SlowDeathBehavior ModuleTag_05
-    SinkDelay = 14000
+    SinkDelay = 1500 ; Patch104p @tweak from 14000 to be consistent with other vehicles
     SinkRate = 2     ; in Dist/Sec
-    DestructionDelay = 20000
+    DestructionDelay = 8000 ; Patch104p @performance from 20000 to optimize despawn timing
   End
 
   Behavior = TransitionDamageFX ModuleTag_06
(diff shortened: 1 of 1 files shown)


commy2:
Neat. 14 seconds is whack for 1.04 Humvee. Only worry here is that mass destruction of many units would look very synchronized now.

Delete time should consider sloped terrain.

xezon:
I did not test on sloped terrain. It could make some difference, but that would be original issue too, because most times are set for flat terrain to begin with.

commy2:
Some times are shortened though (e.g. Radar Van), which would exacerbate the issue.

Also, maybe some vehicles like Overlords should show wrecks longer before starting to sink, just because they are big and significant.

xezon:
Yes, all vehicles above 8000 destruction delay have been candidates fort shortening, because most vehicles have about the similar hull dimensions, and there is no need for them to live 4000 or 8000 ms longer than necessary. Longer alive times will incur resource cost for longer periods. To optimize everything for sloped terrain, all vehicles will need to be tested on sloped terrain, if that is so desired. This change does not aim to do that, because regular flat terrain is good enough for most common scenarios.

xezon:
Is this change good?

commy2:
I don't get why thise increases sink delay from 1.5 s to 3 s for so many vehicles. I don't think it should be the same fixed value for all of them.

xezon:
Ok. How should we classify the hulls to determine appropriate delay?

xezon:
> Too many unrelated changes at once.

xezon:
This change has been superseeded by #1613. Closing.

PR #1211, opened 2022-09-15
Change: Add fixed and improved GLA Cargo Plane (NProject)

xezon:
This change adds a fixed and improved GLA Cargo Plane from NProject.

It uses the original textures, apart from the red nonsense on the cockpit. The model shape and texture mapping has changed. The Polygon count is reduced.

The new plane also has variants for Night, whereas the original does not.

This cannot be Optional content because of INI changes.

## Original

original

## Patched, NProject

patched

shot_20220915_194857_2


xezon:
Overall it looks better I think, apart from the center windows that are placed way too close to the wings now.

MTKing4:
It looks very different now from original, also thinner with longer/thinner wings, not sure if i like this one.

Stubbjax:
I prefer the compact form. The fuselage and cockpit on the newer model are definitely an improvement, but I really don't like the conspicuously stretched wings and the cut-off engines. The wings don't align with the fuselage as well as the original model either, as @xezon pointed out. I find the rear stabilisers are a nicer shape and proportion on the original model as well, though I do prefer the narrowing of the front of the tail fin (more aerodynamic).

Below is a nice reference image to further highlight the way the engines should flow into the wing and not be abruptly cut-off. It's hard to unsee once you notice it.

image

xezon:
Looks too different, closing.

PR #1214, opened 2022-09-16
Fix: GLA Scud Launcher head lamps and chassis holes

xezon:
This change fixes head lamps and chassis holes in GLA Scud Launcher. Head lamps can now be properly viewed from all angles. Bottom side of model is no longer see-through. The pivot positions are unchanged. Total polygon count raised from 1066 to 1108. Design of unit is unchanged.

Credits: Jordan.

https://www.moddb.com/mods/cc-generals-crazy-mod/addons/gla-scud-launcher-remake

## Original

shot_20220917_083002_3

## Patched, NProject

shot_20220917_082705_1

shot_20220916_101916_2

shot_20220916_103349_3


commy2 (approved):

xezon:
Original image added. ~~There is one problem I observed:~~

~~The patched damaged Scud Model has damaged texture on the Scud Missile now. Before the Scud Missile looked undamaged. However, when the damaged Scud Launcher shoots then it spawns an undamaged Scud Missile, which does not match the damaged look. Maybe this is something that needs accompanying INI setup.~~

xezon:
Update: I had the images backwards. It is the original Scud Launcher that shows the damaged Scud Texture. The new Scud Launcher looks correct. All good from my side.

Stubbjax:
Is the way the headlight beams diverge between the undamaged and damaged models correct? Looks kinda odd without a corresponding model deformation. It also seems like there should be four (or at least stronger) headlight beams.

xezon:
> Is the way the headlight beams diverge between the undamaged and damaged models correct?

Yes. It is the same in original model. See first image.

Stubbjax:
> > Is the way the headlight beams diverge between the undamaged and damaged models correct?
> 
> Yes. It is the same in original model. See first image.

I can see that. I'm questioning whether it is correct in both.

xezon:
Could be rotated a bit closer yes. Also the lamps are not in perfect position.

PR #1215, opened 2022-09-16
Fix: GLA Radar Van chassis holes

xezon:
This change fixes issues with Radar Van. Visuals are pristine. Pivots look to be in same location as before. Polygon count increased from 324 to 374.

* Fixes chassis holes
* Fixes single sided meshes on back tower
* Increases wheel corners from 6 to 8 sides

Does not fix:
* misplaced radar dish and disconnected column
* headlamps are not correctly viewable from all angles
* radar dish does not match cameo image visually

## Original

shot_20220917_081554_3

## Patched, NProject

shot_20220916_163658_1


xezon:

MTKing4:
Before and after images for comparison

xezon:
Added.

The blinking red dot broke and the smoke emitter on the damaged model is missing.

xezon:
In **NProject** there is a smoke emitter. The Smoke is no longer at the back, but at the front. It does not look good at the front, because it intersects poorly with the front window.

The Radar Tower has been linked to the Scan Upgrade, which makes sense I think. This way the model changes on Upgrade. The blinking dot also works on the Tower.

It looks like INI changes are necessary to complete this work correctly.

shot_20220917_084739_1

shot_20220917_084831_4

xezon:
I copied over the NProject INI changes to make Radar Tower visible on Upgrade only. It works correctly. The Radar Van Model can be used either way, so we can revert the INI changes again if need be.

I rechecked Smoke, and it looks like Smoke Effects spawn at different pivots randomly, so this is not NProject change.

I now reverted my model edits to restore the blinking lights. The models are now as is from NProject. This means the radar dish and column positions are broken again.

To fix them, they have to be moved to following positions:

```
radar
X: 8.572
Y: 0.000
Z: 16.693

radarpart
X: -0.006
Y: -0.085
Z:  0.016

raise top vertices of radarpart by 0.705
```

My 3d software setup prevents me from editing the Radar Van correctly. Upon import, the blinking lights animation breaks.

MTKing4:
Why close this? If we can't fix now surely it can be postponed to when we find a solution

PR #1216, opened 2022-09-16
Fix: GLA Battle Bus chassis holes

xezon:
This change fixes Battle Bus chassis holes ~~and damaged texture~~. The chassis holes on the bottom were minor. A duplicate mesh at the front was deleted. The pivot positions appear to be completely unchanged. The visuals are pristine. ~~The damaged bus texture has a fix for the front window, which originally looked slightly different to the undamaged window. The texture fix has been remade from scratch because the NProject textures use bad DDS compression and the texture was exported with DXT5 instead of DXT1, effectively doubling the texture size for no benefit.~~

The Polygon count increased from 388 to 406.

## Original

shot_20220917_081253_2

## Patched, NProject

shot_20220916_175221_1


xezon:

MTKing4:
Again, looks good, but before and after photos are required

xezon:
Added. Looks like there is a problem on the damaged model. All the particle emitters are missing.

xezon:
Same issue exists in NProject.

xezon:
The lamp lights are properly viewable from different angles. The lamps position could be improved however. They are not placed correctly in front of the headlamps.

shot_20220917_125046_2

xezon:
I separated the texture fix into
* #1231

PR #1217, opened 2022-09-16
Fix: GLA Marauder chassis holes

xezon:
This change fixes chassis holes on GLA Marauder. These holes are small ones at the bottom, and a few small ones on the top rear. The pivot positions appear to be completely unchanged. The visuals are pristine. The Polygon count increased from 484 to 562 (includes all 3 scrap stages).

## Original

shot_20220917_080547_1

## Patched, NProject

shot_20220916_203707_1


xezon:

MTKing4:
It seems to look ok initially, but I'd be better to compare them with original side by side to tell the difference

xezon:
Added Original image. Looks like there is problem with the damaged edit. It looks undamaged model wise. I suspect author edited Marauder, and then just copied the fixed regular model to damaged model.

xezon:
Same issue exists in NProject.

PR #1218, opened 2022-09-17
Fix: GLA Scorpion deformed model, muzzle flash, treads

xezon:
This change fixes some visuals on Scorpion Tank. Visual design is pristine. Some pivots have changed positions, such as for smoke and muzzle flash. The texture fixes have been partly taken from NProject, and partly remade from scratch.

Polygon count changed from 453 to 460.

Fixes:
* ~~Damaged textures no longer look different than undamaged texture~~
* ~~Undamaged textures has less black seams between armor plates~~
* Damaged model now shows deformations
* Muzzle flash can be seen from all angles
* Treads animate with regular speed instead of fast speed
* Tank bottom piece uses different part of texture

Not fixed:
* Minor holes in chassis (bottom)

## Original

shot_20220917_133226_3

## Patched

shot_20220917_163619_2


xezon:
This model has an issue at the barrel. A texture is missing. It references **uvlitetank2.tga** and **uvlitetank2_d.tga** respectively. This looks like an error, because uvlitetank2 is a copy of the original tank within NProject and there is no need to use 2 textures for one model. The original does not either. Likely fix is change texture mapping back to original texture.

xezon:
Texture fixes have been decoupled from this.

commy2:
Deformed chassis can be exported from CCG. There is a separate model there, but it needs to be merged with the scrap turret changes from ZH.

PR #1220, opened 2022-09-18
Fix: Prevent Pilots From Entering Humvees Without Promoting Them

### LINKED ISSUE #126: Investigate pilot + transport inconsistencies
Pilots will sometimes rank up a transport and other times enter it instead. Seems to be tied to movement delta / update - if colliding with the vehicle on a particular frame or when pathing distance is 0.

https://user-images.githubusercontent.com/11547761/131453671-0dbcc4ea-ecb4-4c94-b606-79e733bd32c7.mp4

https://user-images.githubusercontent.com/11547761/131453843-81e2b347-5b7a-40da-b568-61086debf55b.mp4

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/126

This forbids `NO_GARRISON` flag objects from entering Humvee. `NO_GARRISON` is unique to Pilot and makes him unable to enter civilian buildings, Tunnels and (iirc) Bunkers. Cursor still appears and it promotes the Humvee when entered.

This means vet 3 Humvee can no longer transport Pilots.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -5251,6 +5251,7 @@ Object AirF_AmericaVehicleHumvee
 ;    ExitSound           = GarrisonExit
     DamagePercentToUnits = 100% ;10%
     AllowInsideKindOf  = INFANTRY
+    ForbidInsideKindOf = NO_GARRISON ; Patch104p @bugfix commy2 18/09/2022 Reject Pilots.
     ExitDelay = 250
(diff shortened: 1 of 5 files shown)


Stubbjax (approved):
Nice solution. Tested and works.

Pilots will no longer be able to enter heroic Humvees, but I doubt anyone will miss that. The consistent behaviour easily outweighs such a minuscule cost.

commy2:
> Pilots will no longer be able to enter heroic Humvees

Good point. Sad to see that go. 

On a tactical level, you can still carry Pilots with Ambos, and Ambos are usually grouped with Humvees and empty.

xezon (approved):
Looks good to me.

xezon:
With Code, we can probably make a better implementation for it, by still allowing Pilots to enter Humvees with CTRL + Click.

PR #1226, opened 2022-09-20
Fix: Remove wheel from chassis of Overlord Tank

xezon:
This change removes the wheel(s) on the chassis of Overlord Tank. The fix is very minor, thus can be optional content.

## Original

shot_20220920_123822_1

## Patched (Optional)

shot_20220920_124048_1


xezon:
Added before and after image.

commy2:
No need to make this one optional. This just seems like an error and direct improvment without any redesigns.

xezon:
Are you sure? It is such marginal fix, it is not strictly necessary to package in core. The 5 textures will cost 214 kb uncompressed, 128 kb 7z compressed.

Jundiyy:
Yea, don't think anyone will say anything. Just that one wheel thing that visible? And the one on the other side.

xezon:
On the Gattling Turret too.

Jundiyy:
Ah yea, didn't look there. If it costs too much kb then leave it.

commy2:
I am not concerned about size right now. But I am unsure about just how many textures will be replaced, so maybe I have the wrong expectation.

xezon:
There will be more optional texture fixes like this. Basically anything that is not really a bug, but an enhancement, just like this one. The Scorpion texture fix in comparison is much more severe, because it literally looks different in stages. I will change the label of this one.

commy2:
I honestly cannot tell if this is a bug or just dust up. What is this wheel supposed to be? Seems like an unwrapping error.

xezon:
Hard to say what the intention was. It is the same kind of circle that is used for the wheels. It is baked into the texture, so no model error.

commy2:
Copy paste for the wheels and this was a floating layer meant to be hidden?

xezon:
Possible.

xezon:
Is this good to go?

alanblack166:
Not a fix, as that "wheel" is depicted in the original concept art:
13921011_970803903018544_3825368905647755300_n

xezon:
In that case we can probably change the optic a bit to not look identical to the tread wheels. I will take a look.

xezon:
I modified the side circle to look like some sort of vent. This should blend it a bit more gently then the tread wheel.

shot_20220920_192728_1

xezon:
Fixed and Pushed for all variants.

commy2 (approved):

PR #1229, opened 2022-09-20
Fix: utplates snow textures

xezon:
This change increases texture size of utplates snow textures by merging other variants of this texture. It is not AI generated. It is part of the Recovered Art pack which contains 2x upscaled textures from CCG.

### utplates_ds, utplates_es

* Creates 2x upscale to 256x256


xezon:
No idea where this texture is used in game, but does not matter.

xezon:
These textures are used on Scud Storm.

PR #1233, opened 2022-09-20
Fix: abarfrccmd textures

xezon:
This change fixes USA Airfield issues.

Adds alpha channel:
* abarfrccmd_d
* abarfrccmd_ds

2x upscale (from CCG):
* abarfrccmd_d
* abarfrccmd_ds
* abarfrccmd_e

Fixes texture to look damaged:
* abarfrccmd_e

Adds missing texture:
* abarfrccmd_es

abarfrccmd_e, abarfrccmd_es are currently not used by the E W3D models. It just uses the abarfrccmd_d, abarfrccmd_ds variants. We can modify W3D to use new textures, or leave as is and remove abarfrccmd_e, abarfrccmd_es.

### TODO

- [ ] Use E textures in E W3D models


commy2:
If it is just a matter of replacing letter D with E in a texture, it can be done by hex editing the file.

PR #1239, opened 2022-09-21
Fix: Add nbptower_e texture

xezon:
This change adds nbptower_e texture which is missing in original game. nbptower_es (Snow) does exist however. NBPTower_E.W3D references nbptower_d texture.

### TODO

- [x] Use nbptower_e texture in E w3d models


xezon:
This fix looks complete.

commy2 (approved):

xezon:
I think we should apply same fix to China Propaganda Center, which also uses these textures.

PR #1241, opened 2022-09-21
Fix: Add 2x upscale of atcemwall02_es

xezon:
This change adds high res version of atcemwall02_es for Recovered Art pack.

Generals has 2x upscales for all other atcemwall02 variants except this one, so this one is upscaled by hand.


xezon:
Superseeded by #1458

PR #1245, opened 2022-09-21
Fix: atsdislab snow textures

xezon:
This change fixes atsdislab snow texture variants. Originally the snow does not overlap correctly from one stage to another, which will incur visible texture shifts during damage transition of the structure. ~~The fix can be optional content, because this lab appears to be no faction structure.~~

### atsdislab_ds, atsdislab_es

* Fix mispositioned snow
* Fix snow drawing over burns


xezon:
Textures promoted to Core, because they belong to USA Particle Cannon. Both are 256x256.

MTKing4:
Think this is ok to be on main patch

xezon:
Applied a few more tweaks after checking in game. Looks okay now.

PR #1246, opened 2022-09-21
Fix: atstratslab snow textures

xezon:
This change fixes atstratslab (USA Strategy Center) snow texture variants. Originally the snow does not overlap correctly from one stage to another, which will incur visible texture shifts during damage transition of the structure. ~~The fix can be optional content, because this lab appears to be no faction structure.~~

### atstratslab_s

* Fixes deep shadows in snow
* Adds missing snow at wall edges

### atstratslab_ds, atstratslab_es

* Fixes deep shadows in snow
* Adds missing snow at wall edges
* Fixes snow positions


xezon:
Textures promoted to Core, because they belong to USA Strategy Center. They are 256x256.

xezon:
I applied a few more tweaks and fixes. Looks good to me.

PR #1249, opened 2022-09-21
Add avambulancesw_d1 texture

xezon:
This change adds new avambulancesw_d1 texture. avambulancesw, avambulancesw_d already exist. It is the alternative ambulance with the the yellow wrench symbol.

### TODO

- [ ] Add AVAmbulanceSW_E.W3D


commy2:
Not WB, because the model is not used by any object.

xezon:
Hmm. Isn't this Humvee available in World Builder? I am sure I have seen this somewhere.

xezon:
It is this one:

humv

commy2:
I know the model, but what object uses it?

xezon:
Ah I see. Nothing uses it. Well no idea then. Do we want to have a WorldBuilder humvee that uses this model?

commy2:
Could add it. Just copy paste of vUSA Ambulance with this model. It also has audio files btw. I think it's called "Repair Vehicle" (lame).

MTKing4:
We can have that as a separate WB unit yeah, why not

FireRainV:
I feel like this should be an upgrade for ambulances instead of a new unit.
you can still apply the texture if you want, like in a mixed way between the two textures once you get the upgrade.

it will also nerf early rushes of a humvee and an ambu

PR #1252, opened 2022-09-22
Fix: atdiaplate textures

xezon:
This change fixes atdiaplate textures. The texture is used on USA Warfactory repair driveway parts.

### atdiaplate, atdiaplate_d

* Flips texture horizontally to better blend with nearby texture on USA Warfactory

### atdiaplate_e

* Flips texture horizontally
* Removes bullet holes
* Adds more dust

### atdiaplate_s, atdiaplate_ds

* Flips texture horizontally
* Repaints snow (inherently fixes snow position mismatches)

### atdiaplate_es

* Flips texture horizontally
* Repaints snow (inherently fixes snow position mismatches)
* Removes bullet holes
* Adds more dust


xezon:
I made another pass on this. This texture is used on USA war factory. It now blends well with the other war factory fixes.

* #1427

xezon:
The texture flip somehow does not help on the non snow version. No idea why. Will merge anyway. Is ok as is.

xezon:
I noticed there was error in my test setup, with tga files overwriting psd files. After another test, all fixes worked correctly.

PR #1253, opened 2022-09-22
Fix: atconvey01 snow textures

xezon:
This change fixes atconvey01 (USA War Factory) snow texture variants. Originally the snow does not overlap correctly from one stage to another, which will incur visible texture shifts during damage transition of the structure. The textures are small.

### atconvey01_s

* Fixes blue color cast on snow

### atconvey01_ds, atconvey01_es

* Fixes mispositioned snow


xezon:
Appended a few more tweaks. Looks ok.

PR #1255, opened 2022-09-22
Fix: avbattlesh textures

xezon:
This change recreates Battle Ship damaged textures. They were created from the undamaged texture, which is 2x scale. All damages are repainted. Burns are now properly black. Burn highlights are added. Alpha channels were fixed. All bullet holes are removed.

New textures are 512x512 each.

## Original

avbattlesh_d

## Patched

avbattlesh_d_patch


xezon:
The damaged textures looked a bit too clean in comparison to other vehicles so I added a bit more black dust and scratches.

xezon:
Appended improvement to texture layers so it is more straight forward to copy changes from D to D1 without any readjustments.

PR #1260, opened 2022-09-23
Fix: Internet Center select audio

Stubbjax:
Internet Centers of China Nuke, China Tank and China Infantry now play the correct selection audio. Who knows if anyone has ever noticed this.


MTKing4:
I have :)

xezon (approved):
Nice find.

PR #1264, opened 2022-09-23
Fix: atmetal03 snow texture

xezon:
This change fixes atmetal03_s texture.

### atmetal03_s 

* Fixes snow looking different than atmetal03_ds, atmetal03_es


xezon:
Looks good to me.

PR #1093, opened 2022-09-03
Test Branch Helix AI / Helix Gatling Changes

xezon:
* old PR: #573

This change allows Helix with Gattling Cannon to click and attack Air targets.

Issue #991, opened 2022-08-27
Remove initial cooldown for Helix Napalm/Nuke bomb?


ExiLe-ZH:
There are only a very few situations in which the napalm/nuke bomb is useful.

Usually it's used in some last resort when the helix has to evac and can't be brought home anymore, it could still destroy an unguarded (non captured) oil in the enemy's base e.g.

The problem is that it takes very long to upgrade and use the weapon because the cooldown is triggered after upgrading. 

Any other upgrade like proptowers, bunkers, gatling guns, any kind of drone are instantly functional, should this be the case for the napalm/nuke aswell?

xezon:
There is precedent for this. USA Airforce Comanche Rocket Pod upgrade can be used immediately after purchase. Ranger building capture needs to wait for cooldown after purchase.

I am ok with streamlining Helicopter Upgrade behaviour.

ReLaX82:
I like the idea, makes it a little more useful!

commy2:
It should be reviewed how much damage the Nuke Helix Bomb deals. From memory, it is less than half that of a Helix Bomb, even before Black Napalm. If anything, it should do more.

Stubbjax:
Definitely agree with this. It would be nice to see the bombs used more / at all. Innovative manoeuvres like waiting above a tunnel for a pop and then dropping a bomb - which happens maybe once every five years - would become a bit more viable and surely provide some additional strategic depth as well as excitement for observers / viewers.

It might also make sense to **drop** the price somewhat, perhaps from $800 to $500, due to the significantly heightened risk / vulnerability that accompanies their use. It's a tricky design because the bombs are most effective against large groups, but a large group often has some form of anti-air capability that is likely to turn the whole situation into a suicide mission. This is amplified by the clunky behaviour where the Helix has to stop directly above the target location for the ability to function.

It would be interesting if the bombs could be [dropped earlier before the target by inheriting some momentum](https://user-images.githubusercontent.com/11547761/187082776-5a79c481-771b-4737-b417-1c131e2fcabf.png), similarly to how carpet bomb trajectories appear to function. Such a change would make the behaviour feel more fluid and natural, as well as open up potential for more skilled usage by using the Helix's velocity to modify bomb trajectory.

Otherwise, a simple increase to the bomb's velocity so that it falls through the air faster would be an improvement. In its current form, it's far too easy for the opponent to dodge the bomb and completely avoid any harm, which makes the ability much harder to justify.

[...]

Issue #182, opened 2021-09-04
Disabled upgrades and features can be enabled via CommandSet.ini, which acts as a cheat


xezon:
Disabled upgrades and features can be enabled via 
* CommandSet.ini
* CommandMap.ini
* CommandButton.ini

This can give players an advantage. We need to look into that.

Note: Removing legacy features from the game could break Mod maps.

PR #1265, opened 2022-09-23
Change: Improved accessibility and usability of the Helix bomb drop abilities

### LINKED ISSUE #1014: China Helix Bomb is rarely used
China Helix Bomb is rarely used.

### Proposal 1

* #991

### Proposal 2 APPLIED

China Helix with Nuke Bomb deals much less damage than China Helix with Napalm Bomb. Increase Nuke Bomb damage.

Done with
* #1540

### Proposal 3

Reduce cost from 800.

### Proposal 4

Remove/Reduce stand still requirement for bomb drop to increase hit chance.

### Proposal 5

Increase bomb drop velocity to increase hit chance.

### Proposal 6

Make Helix Bomb an Upgrade from China Airfield, just like the USA Rocket Pods is an Upgrade on USA Airfield. This also solves Command Set issue highlighted here:
* #1532

Stubbjax:
Improved accessibility and usability of the Helix bomb drop abilities. Demonstration of the unpack adjustment can be seen [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1014#issuecomment-1231470003). Closes #1014.

- Reduced the research time from 20 seconds to 10 seconds to offset the initial cooldown
- Reduced the price from $800 to $500
- Removed the unpack delay to improve responsiveness
- Added a bomb drop trigger sound effect

The initial cooldown on the ability makes the upgrade a very difficult proposition when adapting to the current state of the battlefield. Waiting for the upgrade _and_ cooldown - an entire 30 seconds in 1.04 - is vital time and is quite simply rarely worth the wait (@ImTimK explained [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/991#issue-1353055385)). Reducing this initial delay from 30 seconds to 20 seconds seems like a reasonable and low-key approach to incentivise reactive / adaptive play while still requiring some initial time investment / prediction.

The price reduction is determined by comparing the unit upgrades of other factions - most notably Comanche Rocket Pods, which is a single $800 investment that then benefits all existing and future Comanches. A good comparison by @ImTimK can be found [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1014#issuecomment-1234050908). This is in stark contrast to the Helix's bomb ability, which not only costs $800, but $800 _per Helix_. In addition, as described [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/991#issuecomment-1229499022), there is a significantly heightened risk / vulnerability that accompanies the ability's use. As the bombs are most effective against large groups, and a large group often has some form of anti-air capability, there is a high risk involved in their usage. This was demonstrated (to an extent) by @MTKing4 [here](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1093#issuecomment-1256202497). The per-unit investment combined with the heightened risk associated with the ability's usage justifies a lower price, and $500 is a reasonable price when also considering the total investment that a Helix typically requires (i.e. unit cost + bunker cost + occupants cost).

The unpack delay improves the ability's responsiveness, and affords the user much greater control and reliability. As explained [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1014#issuecomment-1231470003) and [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1014#issuecomment-1231918630), the original 500ms delay combined with typical network delay greatly reduces a user's ability to accurately and effectively control the ability, and the risk is very rarely worth it. In addtion, it attenuates some of the vulnerability that a Helix has to endure when flying above targets in order to drop the bomb.

The added sound effect serves to provide additional feedback. The visual of the bomb dropping is often hard to see, particularly when there is a lot going on. It allows the user to immediately know when the ability is triggered (though the lack of delay means they no longer need to rely on the visual before giving another order) and, more importantly, it also provides opponents with feedback and a greater chance to react. Plus it just felt odd that it was basically the only action in the game without a corresponding sound effect.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -15618,10 +15618,11 @@ Object Boss_VehicleHelix
     SpecialObjectsPersistWhenOwnerDies = Yes
     SpecialObjectsPersistent = Yes          ;Charges are persistent till lifetime expires or owner detonates them.
     UniqueSpecialObjectTargets = No        ;This would prevent multiple charges placed on the same object.
-    UnpackTime              = 500     ;slight delay to drop bomb
+    UnpackTime              = 0       ;slight delay to drop bomb ; Patch104p @balance Stubbjax 24/09/2022 Remove delay.
     FlipOwnerAfterUnpacking = No
     FleeRangeAfterCompletion = 0.0         ;DOes not run away after finishing ability
     UnpackSound               = NoSound
+    TriggerSound = HelixBombWeapon ; Patch104p @tweak Stubbjax 24/09/2022 Added sound effect.
     LoseStealthOnTrigger      = No
     ApproachRequiresLOS       = No ; we are a helicopter, we can see everything
     NeedToFaceTarget          = No ; can drop the bomb at any angle to target
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaAir.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaAir.ini
@@ -281,6 +281,7 @@ Object ChinaVehicleHelix
     FlipOwnerAfterUnpacking = No
     FleeRangeAfterCompletion = 0.0         ;DOes not run away after finishing ability
     UnpackSound               = NoSound
+    TriggerSound = HelixBombWeapon ; Patch104p @tweak Stubbjax 24/09/2022 Added sound effect.
     LoseStealthOnTrigger      = No
     ApproachRequiresLOS       = No ; we are a helicopter, we can see everything
     NeedToFaceTarget          = No ; can drop the bomb at any angle to target
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaAir.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaAir.ini
@@ -277,7 +277,7 @@ Object ChinaVehicleHelix
     SpecialObjectsPersistWhenOwnerDies = Yes
     SpecialObjectsPersistent = Yes          ;Charges are persistent till lifetime expires or owner detonates them.
     UniqueSpecialObjectTargets = No        ;This would prevent multiple charges placed on the same object.
-    UnpackTime              = 500     ;slight delay to drop bomb
+    UnpackTime              = 0       ;slight delay to drop bomb ; Patch104p @balance Stubbjax 24/09/2022 Remove delay.
     FlipOwnerAfterUnpacking = No
     FleeRangeAfterCompletion = 0.0         ;DOes not run away after finishing ability
     UnpackSound               = NoSound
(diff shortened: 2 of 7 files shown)


commy2:
Please also change for Boss Helix. Don't bother with CINE units though.

xezon:
Can we please get a rationale for the Cost and Time changes as to why the chosen values make most sense?

Stubbjax:
> Can we please get a rationale for the Cost and Time changes as to why the chosen values make most sense?

Were they not rationalised in the respective issue?

xezon:
Gameplay changes need to be most visible, so that visitors can easily look up as to why was something changed and do not leave wondering.

Stubbjax:
> Gameplay changes need to be most visible, so that visitors can easily look up as to why was something changed and do not leave wondering.

Updated!

ExiLe-ZH (approved):

xezon:
Sound is ok. Not a perfect fit, but a known sound in the game.

May I ask what the reason is to decrease the build time instead of removing the initial cooldown? Wouldn't it make more sense to have just 1 timer (20) instead of 2 (10+10)?

In the rationale it says "$500 seems like a reasonable price". Can we update the text to not make it sound like guesswork but instead make the 500 cost a guided choice? The issue with guesses is, it will invite anyone to just suggest new values on a hunch.

ExiLe-ZH:
I'm not sure about the technicality, but removing the initial cooldown apparently opens up exploits. This is a workaround.

$500 is reasonable and safe because upgrading 2 Lixes still makes it more expensive than Rocket Pods ($800 to upgrade all), which is by far the superior upgrade.

Stubbjax:
> May I ask what the reason is to decrease the build time instead of removing the initial cooldown? Wouldn't it make more sense to have just 1 timer (20) instead of 2 (10+10)?
```
Behavior = SpecialAbility ModuleTag_32
  SpecialPowerTemplate = SpecialAbilityHelixNapalmBomb
  UpdateModuleStartsAttack = Yes
  StartsPaused              = Yes; so the UnpauseSpecialPowerUpgrade, below can turn it on
  InitiateSound             = HelixVoiceAttack
End
```
It would make more sense, but the initial cooldown is seemingly tied to the `StartsPaused` field being set to `Yes`. Setting this value to `No` will remove the initial cooldown and allow the ability to be used without requiring the upgrade to be researched. What can be done is to set this field to `No`, but have the upgrade toggle a CommandSet upgrade, meaning the ability is technically already available, but locked in the initial command set and enabled in the upgraded command set. This can be exploited; see #182. @commy2 or @alanblack166 likely know more and can explain it better.

> In the rationale it says "$500 seems like a reasonable price". Can we update the text to not make it sound like guesswork but instead make the 500 cost a guided choice? The issue with guesses is, it will invite anyone to just suggest new values on a hunch.

Anyone is free to suggest new values, and they should be considered if they provide thoughtful and well-reasoned arguments. It cannot be stated with absolute certainty that $500 is the perfect value, but it absolutely can be considered reasonable. I updated it to read a bit better anyway.

xezon:
On that Note, would other Upgrade CommandSet exploits be preventable if we treat them the same way as this Napalm Bomb?

xezon:
commy2:
> Maybe it is enough to set NEEDS_UPGRADE etc. on the ability button.

Stubbjax:
> commy2:
> 
> > Maybe it is enough to set NEEDS_UPGRADE etc. on the ability button.

`NEED_UPGRADE` is required for it to work that way. But someone can just remove / comment the upgrade and then the ability is available without requiring the research.

xezon:
In which file is NEEDS_UPGRADE? Many INI file changes will cause version mismatch.

Stubbjax:
> In which file is NEEDS_UPGRADE? Many INI file changes will cause version mismatch.

You need to append it to the `Options` field for `Command_ChinaHelixDropNapalmBomb`. No plural.

commy2:
> In which file is NEEDS_UPGRADE? Many INI file changes will cause version mismatch.

CommandButton.ini

xezon:
Ok. CommandButton file is not CRC'ed in any way, so NEED_UPGRADE is meaningless.

PR #1267, opened 2022-09-23
Fix: atpanelroof3 snow texture

xezon:
This change fixes atpanelroof3 texture.

### atpanelroof3_ds

* Fixes snow having different offset than atpanelroof3_s


xezon:
The roof texture is used on the USA Strategy Center. Looks good to me.

PR #1268, opened 2022-09-23
Fix: atprisslab snow textures

xezon:
This change fixes atprisslab (USA Prison) snow textures.

### atprisslab_s

* Fixes strong contrasts on snow

### atpanelroof_ds, atpanelroof_es

* Fixes snow to look equal to atprisslab_s
* Fixes snow to not draw over damages, scorches and burns


xezon:
I optimized the fixes a bit. Looks good to me.

PR #1270, opened 2022-09-23
Fix: avpaladin textures

xezon:
This change fixes avpaladin damaged textures. The fixes are very minor and therefore optional.

### avpaladin_d

* Fixes mis-shaped bolts

### avpaladin_d1

* Fixes burn inconsistencies to avpaladin_d


xezon:
Looks okay to me.

PR #1274, opened 2022-09-24
Fix: Set limits and priorities for air unit death sounds

xezon:
This change sets reasonable limits for air unit death sounds, including low fuel and falling. Without limits, too many audio lines can be spammed at once.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
@@ -402,13 +402,14 @@ AudioEvent RaptorVoiceAttack
 End
 
 AudioEvent RaptorVoiceLowFuel
-  Sounds = vrapfub ; vrapfua vrapfuc vrapfud
+  Sounds = vrapfua vrapfub vrapfuc vrapfud ; Patch104p @tweak enable all sound variants
   Control = random
   Attack = gradio1a gradio1b gradio1c gradio1d gradio1e gradio1f
   Decay = gradio3a gradio3b gradio3c gradio3d gradio3e
   Volume  = 120
   MinVolume = 110
   Type = world player global
+  Limit = 1 ; Patch104p @tweak to not spam voice as much
 End
 
 AudioEvent RaptorVoiceUpgradeFuel
@@ -443,6 +444,8 @@ AudioEvent RaptorVoiceFalling
   Control = random
 ;  Attack = gradio1a gradio1b gradio1c gradio1d gradio1e gradio1f gradio1g
 ;  Decay = gradio3a gradio3b gradio3c gradio3d gradio3e gradio3f gradio3g
+  Priority = LOW ; Patch104p @bugfix to be consistent with Voice Die limits
+  Limit = 3 ; Patch104p @bugfix to be consistent with Voice Die limits
   Volume = 100
   MinVolume = 80
   Type = world shrouded player
@@ -512,6 +515,8 @@ End
 AudioEvent ComancheVoiceFalling
   Sounds = vcomfaa vcomfab vcomfac vcomfad vcomfae
   Control = random
+  Priority = LOW ; Patch104p @bugfix to be consistent with Voice Die limits
+  Limit = 3 ; Patch104p @bugfix to be consistent with Voice Die limits
   Volume = 100
   MinVolume = 80
   Type = world shrouded player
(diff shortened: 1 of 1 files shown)


commy2:
? Infantry units also have unlimited death sounds. As for low on fuel, the audio cue is helpful for the player, and the amount of cues gives additional information. It is not clear to me this is an improvement. Verschlimmbesserung in German.

xezon:
I tested 4 planes on low fuel and it sounds awful. I also tested 4 planes dying at the same time and it was too much already.

Unlimited sounds stacking is a bad idea.

Please test it before drawing conclusions.

xezon:
As for amount, maybe low fuel limit can be set to 2 instead of 1. I have not tested it. But 1 sounded fine to me.

An additional problem with RaptorVoiceLowFuel specifically is the its having one sound variant only. So 4 planes would all spam the same line and stack it. With more sound variation this possibly is less of an issue and limit can be higher.

MTKing4:
So what you limited it to only 1 sound? that is a bad idea

if you have to set a limit then that is understandable but make it at least limited to 4

xezon:
I enabled all low fuel sound variations and tested multiple configurations.

There are 2 options that sound ok:
1) No Low Fuel sound variation and Limit set to 2
2) Low Fuel sound variation and Limit set to 1 (this)

Anything else sounds bonkers.

If you have 3 planes spamming the same Low Fuel lines, it sounds awful, because they overlap with slight delays.

If you have 2 or more planes spamming different Low Fuel lines, it sounds awful, because they all say different things and you cannot hear what it says. It just becomes gibberish sound spam.

Please test before commenting again. If you do not test, it will be impossible to make educated statements.

## Original sound spam

https://user-images.githubusercontent.com/4720891/192095159-944b1e3b-1fb0-4aa2-aa45-f2eff8aaa952.mp4

## Sound spam with variation

https://user-images.githubusercontent.com/4720891/192095170-62abde34-b617-41ab-a930-15729d77bc0f.mp4

xezon:
> Infantry units also have unlimited death sounds.

This is not true. All Infantry have Limit 3.

MTKing4:
my comment was simply backed by the fact that we've heard this a hundred times in games and had no issues with it, at least when it's a single line like in the first video, it might sound too much for you but it's fine really

the the second video with the different voice lines is horrible though true

i have no pressing need to keep this, but if you have to go with this, then at least make it 2, 1 doesn't make any sense, how'd you know how many planes are having low fuel? when a singluar and multiple both sound the same? say you've more aircrafts than to fit in airfields, how will you know how many more airfields you need for your planes? having multiple sounds makes it urgent and give an idea that you have many planes that need to land

xezon:
With no sound variation, limit 2 is okay.

commy2:
> This is not true. All Infantry have Limit 3.

Then that should also be the limit for planes. So far PR is good I think now.

Still unsure about the low fuel voice though. What is the cooldown here? Because usually your planes arrive at base at different times and then notify you, instead of all at once.

xezon:
Cooldown? I think as soon as sound playtime is completed, it is removed from audio pool and new one can be pushed. Limit 1 means there can be no 2 sounds overlapping each other.

Btw, when you send a group of units to a location or you select a group of units, then you also don't hear multiple sounds. Just 1. Why would you need X audio lines for low fuel but 1 audio line for many other unit voices are ok?

Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
```diff
@@ -3030,8 +3047,8 @@ End
 AudioEvent TechnicalVoiceDie
   Sounds = vtecdia vtecdib vtecdic vtecdid
   Control = random
-  Priority= high
-  Limit = 2
+  Priority = LOW ; Patch104p @tweak to be consistent with other Voice Die setups
+  Limit = 3 ; Patch104p @tweak to be consistent with other Voice Die setups
```
commy2: This should be its own PR.

commy2:
I will test ingame soonish.

Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
xezon: Fixed

xezon:
Did anyone test this change yet? Does it sound ok?

xezon:
Is this change good to go?

commy2 (approved):

PR #1276, opened 2022-09-24
Fix: avraptorag textures

xezon:
This change fixes avraptorag (King Raptor) textures. Additionally it adds new D1 texture variant.

### avraptorag

* Removes obsolete alpha channel

### avraptorag_d

* Removes obsolete alpha channel
* Removes strong green color tint
* Removes "butterfly" scorch marks

### avraptorag_d1, avraptorag_d1b

* New

### TODO

- [ ] Use D1 texture in D1.w3d and similar


commy2:
shot_20230727_202114_1

Putting the new texture onto the falling Raptor model does not work. Black stripes are gone. There is no model for the destroyed King Raptor. Someone has to make one.

xezon:
That look like the raptor_d1 texture, not the raptorag_d1 texture.

commy2:
Strange.

tex

But wouldn't be the first time I messed up.

commy2:
Yeah, I messed some stuff up.

---
PR is missing the avraptor_d1b texture? (With "B" at the end?)

xezon:
I do not understand question. This pull only touches King Raptor textures.

commy2:
- there are no avraptorAG_d1 and avraptorAG_d1B textures in 1.04
- no model exists for falling (D1B) and crashed (D1) King Raptor, instead they uses the respective Raptor models
- this PR adds avraptorAG_d1, but there still is no avraptorAG_d1B (falling)

xezon:
So we need a avraptorag_d1b texture yes?

commy2:
We need it as much as a avraptorag_d1 without B texture. Without it, there could be no falling King Raptor model.

I'm not talented enough to recreate the rubble and crashed models of Raptor for King Raptor. King Raptor is slightly bigger (like 2 - 5%). What I tried is to copy the D1x models of Raptor and apply your King Raptor texture on them, to see how it looks (probably bad, because mapping is likely different).

xezon:
Does King Raptor model look the same apart from size?

commy2:
To me? It looks the same even in size. I only noticed, because when you put them on an airfield side by side, the KR nose goes a tiny bit further.

xezon:
Ok if these look almost identical then this should be as simple as creating copies of raptor models, renaming them, and applying new textures to them.

The textures have same mapping.

I can add avraptorag_d1b texture later. No problem.

commy2:
Yeah, the model shrinks already in 1.04 and also changes textures. If it only shrank, but kept the markings, it would be a halfway improvement already.

xezon:
If King Raptor model is a bit larger, then we can scale it up by factor 1.x. We can find appropriate scale factor by overlapping Raptor and King Raptor models and then test how much to scale Raptor.

commy2:
Scale attribute in ini is object wide, not model wide. The falling Raptor model is part of the regular King Raptor object. It needs scaling in a model editor to be correct.

xezon:
Yes I was refering to model scale, not ini scale.

xezon:
avraptorag_d1b added.

> there are no avraptorAG_d1 and avraptorAG_d1B textures in 1.04

I noticed I named them incorrectly in the text of the opening post without "ag" in the name. Fixed.

PR #1279, opened 2022-09-24
Fix: avpowtruck textures

xezon:
This change fixes avpowtruck textures. POW Truck is not part of faction, so this can be Optional content, if at all.

### avpowtruck_d (Optional)

* Fixes mis-positioned drivers roof texture
* Fixes sharp edges of burns

### avpowtruck_d1 (Optional)

* Fixes mis-positioned drivers roof texture
* Fixes strong contrast on drivers roof texture
* Fixes sharp edges of burns


xezon:
Looks ok.

PR #1281, opened 2022-09-24
Fix: TechnicalVoiceDie limit and priority

xezon:
Fixes limit and priority of TechnicalVoiceDie to be consistent with all other VoiceDie audio events.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
@@ -3030,8 +3030,8 @@ End
 AudioEvent TechnicalVoiceDie
   Sounds = vtecdia vtecdib vtecdic vtecdid
   Control = random
-  Priority= high
-  Limit = 2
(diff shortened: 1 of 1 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
```diff
@@ -3030,8 +3030,8 @@ End
 AudioEvent TechnicalVoiceDie
   Sounds = vtecdia vtecdib vtecdic vtecdid
   Control = random
-  Priority= high
-  Limit = 2
+  Priority = LOW ; Patch104p @tweak to be consistent with other Voice Die setups
```
Stubbjax, line 3033: I wonder if `LOW` is actually parsed properly. Seems odd that all instances of it are uppercase whereas all other values are lowercase.

Stubbjax (approved):

Patch104pZH/GameFilesEdited/Data/INI/Voice.ini
xezon: We can make them all lower case if in doubt.

PR #1282, opened 2022-09-24
Fix: SaboteurVoiceCreate type

xezon:
Fixes type of SaboteurVoiceCreate to be consistent with all other VoiceCreate audio events, except ~~Angry Mob and~~ Unit Upgrades.


xezon:
Not clear to me what the difference between `global` and `ui` is.

Stubbjax:
`UI` means the sound does not have a position in world space. This change will mean players will hear the voice audio come from the direction of the Barracks from which the unit is produced.

Stubbjax (approved):

PR #1285, opened 2022-09-25
Fix: nvinferno textures

xezon:
This change fixes issues on nvinferno textures. The fixes are so minor, they can be Optional.

### nvinferno
* Removes white seam at the bottom of the texture
* Fixes length of some metal piece to be consistent with nvinferno_d, nvinferno_d1

### nvinferno_d, nvinferno_d1
* Fixes look of one metal plate to be consistent with nvinferno.


xezon:
Looks very good.

PR #1287, opened 2022-09-25
Experimental: Remove Floating Bones On Gattling Tank At Night

### LINKED ISSUE #1286: China Gattling Tank model issues
White boxes on "lamps" and Night map.

shot_20220925_091943_1

commy2:
- would close #1286

Doesn't seem to work.


xezon:
If it does not work, can we close it?

commy2:
It's proof I tried. Seems that HideSubObject is overwritten by engine for HeadlightXY, similar to how you can hide muzzle flahes, but they show up anyway when firing.

The apparent bones in the model should be actual headlights instead of bones.

PR #1289, opened 2022-09-25
Fix: Level 2 and 3 Spectre Gunship damage state animation transitions

Stubbjax:
When transitioning between damage states, the level 2 and 3 Spectre Gunships would reset their animation state to the first frame, and the wings would reanimate outwards, which looks bad. This does not happen for the level 1 Spectre Gunship. This change matches the animation transition values of the level 1 Spectre Gunship to the level 2 and 3 variants.

Below is the original behaviour of level 2 and 3. Note the wings reanimate when transitioning between damage states.

https://user-images.githubusercontent.com/11547761/192134395-6418399d-7fb6-41e1-aa46-8ba861cf7c10.mp4

Below is the original behaviour of level 1 (and now 2 and 3 with this change). Note the wings do not reanimate when transitioning between damage states.

https://user-images.githubusercontent.com/11547761/192134385-7a384659-33af-4d9b-98c9-7d89d0129b06.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -682,7 +682,7 @@ Object AirF_AmericaJetSpectreGunship2
       Model           = AVSGunship
       Animation       = AVSGunship.AVSGunship
       AnimationMode   = ONCE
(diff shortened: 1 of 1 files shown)


xezon (approved):
Nice. Never noticed.

commy2 (approved):

PR #1291, opened 2022-09-25
Fix: Add nvdragon_d1 texture

xezon:
This change adds a nvdragon_d1 texture.

### TODO

- [ ] Recreate base texture, create new UV map for dragon tank
- [ ] Use _D1 texture for _D1.w3d, _D2.w3d, _D3.w3d models


xezon:
Dragon Tank needs new UV mapping.

PR #1295, opened 2022-09-25
Fix: nvnukecn, nvnukecnng textures

xezon:
This change fixes nvnukecn (Nuke Cannon), nvnukecnng (Nuke Cannon for Nuke General) textures. nvnukecnng textures are used by nvnukecnng W3D models which are not yet hooked up in INI. All models look complete, however there might be a reason it is not used in game, so w3d models need inspection and testing. The fixes on the texture are rather minor, thus can be Optional.

### nvnukecn, nvnukecn_d, nvnukecn_d1, nvnukecnng, nvnukecnng_d, nvnukecnng_d1

* Replaces wheel on chassis with vent
* Fixes rotation of star and nuke symbols

### nvnukecnng_d1

* Fixes remnants of former star symbol below nuke symbol

### TODO

- [ ] Hookup nvnukecnng models in INI
- [ ] Inspect nvnukecnng w3d models


MTKing4:
It would be better if you put Unit's Names on all those texture/model changes so that we know what you're talking about. 

Preferably in the title, Right now PL list is just gibberish essentially, we wouldn't know what to look for later on

xezon:
These textures are for Nuke Cannon.

PR #1297, opened 2022-09-26
Fix: nbsign, ntposters snow textures

xezon:
This change creates nbsign snow damage variants. These are chinese posters used on some buildings. Likely these use _d variant now. We can change to _ds variant if applicable. This change may not be worth it if building models have to be touched just for this fix.

### nbsign01_ds, nbsign03_ds, nbsign04_ds, ntposters_ds, ntposters_es

* New

### TODO

- [ ] Use snow damage variants in China buildings


xezon:
Also appended new textures ntposters_ds, ntposters_es to this change as they fall into the same category of fixes as the sign textures do.

xezon:
The sign textures are used on Propaganda Center.

shot_20220926_193944_1

PR #1298, opened 2022-09-26
Fix: ntcement textures

xezon:
This change fixes ntcement textures. There is a slight color mismatch on some rubble between _d and _e variants. This fix is so minor, it can be optional.

### ntcement_e, ntcement_es

* Fixes rubble color mismatch


xezon:
Looks good to me.

PR #1301, opened 2022-09-26
Fix: ntentry snow textures

xezon:
This change fixes ntentry snow textures.

### ntentry_s

* Fixes strong snow shadow on the left side of the door

### ntentry_ds, ntentry_es

* Fixes mispositioned snow


xezon:
Textures look good to me.

PR #1304, opened 2022-09-26
Fix: ntsuppbase snow textures

xezon:
This change fixes ntsuppbase (China Supply) snow textures.

### ntsuppbase_s

* Fixes strong contrast in snow

### ntsuppbase_ds

* Fixes strong contrast in snow
* Fixes snow drawing over rubble, scorch marks

### ntsuppbase_es

* Fixes strong contrast in snow
* Fixes snow drawing over rubble, scorch marks
* Fixes mispositioned snow


xezon:
Applied one more minor tweak to snow fixes. Textures look good to me.

PR #1306, opened 2022-09-27
Fix: pmtorch textures

xezon:
This change fixes pmtorch textures.

### pmtorch_d, pmtorch_n

* Reduces texture size from 256x256 to 64x64 like pmtorch

### pmtorch_ds

* Fixes grey snow


xezon:
Looks good.

PR #1307, opened 2022-09-27
Fix: utapples, utapricots, utoranges textures

xezon:
This change fixes utapples, utapricots, utoranges (GLA) textures.

### utapples_s, utapples_es, utapricots_s, utapricots_es, utoranges_s, utoranges_ds, utoranges_es

* Fixes snow looking different from one damage to another
* Fixes harsh edges on snow

### utapples_ds, utapricots_ds

* New

### TODO

 - [ ] Use utapples_ds, utapricots_ds textures in _D.w3d models


xezon:
I think this needs rework. The damaged fruits do not look good. They are used on the GLA Black Market. The single fruits on the floor have odd outlines. It is no error in the Alpha Channel. It may be shadows.

shot_20221009_140617_1

PR #1311, opened 2022-09-28
Fix: utcloth textures

xezon:
This change fixes utcloth textures.

### utcloth_e

* Reduces bright seam on burns

### utcloth_s

* Fixes snow visual mismatching with utcloth_ds, utcloth_es

### utcloth_ds

* Fixes minor artifacts

### utcloth_es

* Adds alpha channel of utcloth_e and same burn in corner


xezon:
The alpha channel does not render nicely on damaged cloth. It has some visible seams. I tried to improve it a bit by making the edges rough. It helps a bit, but does not get rid of the issue.

### Original

shot_20221009_151152_2

### Patched

shot_20221009_150819_1

xezon:
I think nothing else can be improved here from textures side. Merging.

PR #1315, opened 2022-09-28
Fix: utscudslab snow textures

xezon:
This change fixes utscudslab (GLA Scud Storm) base snow textures. ~~The Snow Scud Storm has an issue where _D and _E models instead of _DS and _ES models are shown.~~

### utscudslab_s

* Fixes strong shadows in snow

### utscudslab_ds, utscudslab_es

* Fixes strong shadows in snow


xezon:
Damaged Snow Storms fixed with
* #1326

xezon:
Looks fine to me.

PR #1317, opened 2022-09-29
Fix: ubdome textures

xezon:
This change fixes and improves ubdome (golden dome on many GLA buildings) textures.

### ubdome, ubdome_d

* Uses 128x128 instead of 64x64
* Fixes seams in dome tip
* Removes strong black seam at sides of dome
* Repaints golden dome to look better

### ubdome_s

* Uses 128x128 instead of 64x64, upscaled with AI

### ubdome_ds

* Uses 128x128 instead of 64x64, upscaled with AI
* Removes black seam at tip
* Removes more snow to look more damaged

### ubdome_e, ubdome_es

* New

### TODO

- [ ] Use new ubdome_e, ubdome_es textures in _E w3d models where applicable

### Image

shot_20221008_195122_1

shot_20220929_172156_2


xezon:
Textures now 2x upscaled and a few more improvements added.

xezon:
I recreated the non-snow dome texture from scratch. Image above updated. From my side this is done now.

xezon:
I reviewed Dome and found that especially the damaged snow variant is not sufficient. It looks too similar to the undamaged one, with too much snow. I removed a bunch of snow from _DS and added new _E and _ES variants. With this we have more artistic freedom to fine tune damage look. But it also means several _E models need to be adjusted accordingly. The Dome is used on various GLA structures, incl. Command Center, Palace, Black Market.

PR #1320, opened 2022-09-29
Fix: utconcroof textures

xezon:
This change fixes utconcroof textures.

### utconcroof_d

* Adds soft black borders from utconcroof
* Replaces damages with utconcroof_ds to look different than utconcroof_e

### utconcroof_e

* Adds soft black borders from utconcroof

### utconcroof_es

* Removes obsolete alpha channel


xezon:
This texture is used on flat conrete roofs of various GLA structures.

xezon:
Textures look good to me.

PR #1325, opened 2022-10-02
Fix: Scud Storm Building Lights Turned Off At Night Until Damaged

commy2:
1.04:
- Lights on Scud Storm are off when the building is unharmed, but get turned on when damaged
- Models for NIGHT state are missing from CCG. (REALLY)DAMAGED+NIGHT exist
- ZH adds NIGHT state models, but never uses them in *ini config

Patch:
- Lights are turned on at night regardless of damage state


xezon (approved):
Looks good to me.

PR #1326, opened 2022-10-02
Fix: Scud Storm Building Snow Disappears When Damaged

commy2:
1.04:
- Snow disappears from Scud Storm when damaged (day or night)

Patch:
- Snow stays on Scud Storm when damaged (day and night)

Notes:

- All 4 combinations of `UBScudStrm_{D,E}A1{S,NS}` are missing from the game. However, with the 4 unused models 
`UBScudStrm_{D,E}{S,NS}` combined with the 4 regular animations `UBScudStrm_{D,E}A1{,N}`, the accurate animated model can be displayed.
- Same as above for A2 (firing) instead of A1 (raising missiles)
- `UBScudStrm_DA3NS` and `UBScudStrm_EA3NS` (reloading missiles at night snow when damaged/reallydamaged) are in ZH, but the models are broken, due to an exporting error by EALA. This patch introduces `UBScudStrm_DA3X` and `UBScudStrm_EA3X` models, which are copies of the former with bytes swapped out to 1) fix the exporting error and 2) avoid name collisions with `UBScudStrm_DA3N` and `UBScudStrm_EA3N` respectively.
- `UBScudStrm_ENS` (as well as the before and after unused `UBScudStrm_EN`) display a floating American flag (without flagpole), which can be hidden using HideSubObject token.

<details> <summary>hex edit script</summary>

```py
with open("UBScudStrm_DA3NS.W3D", "rb") as f:
    orig = f.read()

edited = orig.replace(b"UBSCUDSTRM_DA3N", b"UBSCUDSTRM_DA3X")

with open("UBScudStrm_DA3X.W3D", "wb") as f:
    f.write(edited)
```

</details>


commy2:
related: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1325

xezon:
Maybe we can add UBScudStrm_DA3NS.txt and UBScudStrm_EA3NS.txt (or UBScudStrm_DA3X.txt and UBScudStrm_EA3X.txt) next to the w3d files, refering with a short note to the 3X files. Just as another means of documentation. I already did this for a few textures. This way naming convention difference already becomes obvious when browsing files, and must not be discovered in INI comments. txt files will not end up in build.

commy2:
Like this?

xezon:
Yes looks good.

Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
```diff
@@ -5362,8 +5364,8 @@ Object Boss_ScudStorm
     AliasConditionState = ATTACKING BETWEEN_FIRING_SHOTS_B NIGHT DAMAGED SNOW
 
     TransitionState = Trans_ATTACKING_DAMAGED_NIGHTSNOW Trans_READY_DAMAGED_NIGHTSNOW
-      Model         = UBScudStrm_DA3N
-      Animation     = UBScudStrm_DA3N.UBScudStrm_DA3N
+      Model         = UBScudStrm_DA3X ; Patch104p @bugfix commy2 02/10/2022 Alias for UBScudStrm_DA3NS.
+      Animation     = UBScudStrm_DA3X.UBScudStrm_DA3X
```
xezon, line 5368: Another thing we could try with original files:

`UBScudStrm_DA3NS.UBScudStrm_DA3N`

Maybe that would work too? In that case we would not need to make copies of the existing files.

commy2: I tried:

```ini
      Model         = UBScudStrm_DA3NS
      Animation     = UBScudStrm_DA3NS.UBScudStrm_DA3N
```
It did not work.

Edit:

I also (re)tried:

```ini
      Model         = UBScudStrm_DA3NS
      Animation     = UBScudStrm_DA3N.UBScudStrm_DA3N
```
And that didn't work either.

And finally:

```ini
      Model         = UBScudStrm_DA3NS
      Animation     = UBScudStrm_DA3N.UBScudStrm_DA3NS
```
Didn't work either.

xezon: Dang.

xezon (approved):
Looks good.

PR #1328, opened 2022-10-02
Add upscaled utscudslab textures

xezon:
This change adds 2x AI upscaled versions of the utscudslab textures. Originally they are 256x256, which is too small for a building that large and in need of a sharp ground texture, since not much stands on top of it.

* Textures are based on edits in #1315

### Original

shot_20221002_190938_2

### Patched 

shot_20221002_194038_1

shot_20221002_194306_4


xezon:
I repainted the upscaled burns to add more detail to texture. I have not pushed the PSD files, as they are too big. The TGA should do.

shot_20221008_123037_1

xezon:
I have added a few more minor tweaks to repainted burns. This looks good now.

Issue #1224, opened 2022-09-19
Bunkered Busses can't target buildings


ExiLe-ZH:
They have to be force fired on the edge of buildings, this needs fixing?

xezon:
This bug was fixed for Outposts. We can check if same fix can be applied to Battle Bus.

xezon:
* #4

I tested adding TransportAIUpdate module to Battle Bus but that broke Battle Bus. It will not move at all then.

xezon:
Current module looks like this:

```
  Behavior = AIUpdateInterface ModuleTag_04
    AutoAcquireEnemiesWhenIdle = No ATTACK_BUILDINGS
  End
```

I tried with `ATTACK_BUILDINGS` removed but that made no difference.

```
  Behavior = TransportAIUpdate ModuleTag_04
    AutoAcquireEnemiesWhenIdle = No ATTACK_BUILDINGS
  End
```

Did not work either.

PR #1333, opened 2022-10-05
Fix: Remove duplicate AntiAirborneVehicle entries

xezon:
Not sure if this fixes anything tangible. But something is not right. AntiAirborneVehicle entries are duplicated.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2084,7 +2084,6 @@ Weapon TunnelDefenderRocketWeapon
   ProjectileCollidesWith      = STRUCTURES
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
-  AntiAirborneVehicle = Yes
   WeaponBonus = PLAYER_UPGRADE DAMAGE 125% ; AP rocket upgrade
 End
 
@@ -2113,7 +2112,6 @@ Weapon TunnelDefenderBikerRocketWeapon
   ProjectileCollidesWith      = STRUCTURES
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
-  AntiAirborneVehicle = Yes
   WeaponBonus = PLAYER_UPGRADE DAMAGE 125% ; AP rocket upgrade
   AcceptableAimDelta    = 20.0
 End
@@ -6830,7 +6828,6 @@ Weapon ListeningOutpostUpgradedDummyWeapon
   AcceptableAimDelta    = 180 ; I'm not really shooting, my buddies are.  So no need to turn
   AntiAirborneVehicle   = Yes
   AntiAirborneInfantry  = No
-  AntiAirborneVehicle   = Yes
 End
 
 ;------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8741,7 +8741,7 @@ Weapon BattleBusDummyWeapon
   AutoReloadsClip = Yes
   AntiAirborneVehicle = Yes
   AntiAirborneInfantry = Yes
-  AntiGround = No
+  AntiGround = Yes ; Patch104p @tweak from No
```
xezon: I checked if this fixes #1224 but it did not. All other DummyWeapon have `AntiGround = Yes` so I assume this should have too? I have no idea what this setting does. Maybe it does nothing.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -8693,7 +8693,7 @@ Weapon DemoTunnelDefenderRocketWeapon
   ProjectileCollidesWith      = STRUCTURES
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
-  AntiAirborneVehicle = Yes
+  AntiGround                  = Yes
```
xezon: I cannot decide if simply removing the entry or `AntiGround = Yes` is the proper choice for these.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2: Changing this would alter BB vs ground range to 150, currently 90. BattleBusPassengerDummyWeapon is the one used vs ground.

Not that that is a bad thing.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2: AntiGround defaults to Yes. I'd say, for the sake of simplicity: don't add them, unless set to "No". Also, don't remove them needlessly either.

xezon: So just remove the duplicate AntiAirborneVehicle entries?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok perhaps this should be separate change.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2: Yeah, although I am against sweeping cleanups at this stage. It's 1) waste of time, as it does not affect the game, and 2) bloats diffs vs 1.04, which is more distraction and noise than the duplicate lines themselves cause.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: I tested what would happen with this set to Yes and could not observe any changes in Attack Range. Also Removing Passenger Dummy changed nothing. Not sure how this is supposed to work. Reverted.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: It caused me great headache looking at the duplicate entries as I was not sure what I was looking at. It was indicating that there is possibly a misconfiguration. I do not think these duplicates should remain.

commy2 (approved):

PR #1334, opened 2022-10-05
Fix: Toxin GLA Tunnel Defender double firing against ground and air targets

xezon:
* Reference #806

This change fixes Toxin GLA Tunnel Defender double firing against ground and air targets. It is yet another reason why Toxin General is so good. Note: Tunnel Defenders from other factions do not have this issue because they have just one gun type against ground and air. Toxin General has 2.

### Video of original issue

The first Tunnel Defender is from Demo General. The second is from Toxin General.

https://user-images.githubusercontent.com/4720891/194135459-5b9c8371-a16b-4262-86ef-6cdf91a80a9a.mp4


commy2:
Related: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/294

commy2 (approved):

Jundiyy:
Good fix, it will impact the game, but I think it's a fair fix and players will not notice why Tox has weakened, although they may feel it.
We see this mainly when using Battle Busses, they attack ground units, Migs/Helices come, and the Busses are able to shoot everything at once (ground and air).

ExiLe-ZH:
That's crazy, never knew that, but another reason why Tox rpgs/busses were so OP vs air units.

PR #1338, opened 2022-10-07
Fix: USA Spy Drone can no longer be selected, but attacked

### LINKED ISSUE #115: USA Spy Drone (General Power) is selectable and exploitable
Investigate whether or not USA Spy Drone needs to be selectable at all

xezon:
* Fixes #115

This change attempts to make USA Spy Drone no longer selectable, but attackable.

It can be force fired and killed normally. We need to test if regular attack and Laser Locking also works as usual. It may not.


xezon:
I tested on test map. Hovering Spy Drone will show attack cursor, but it cannot be clicked to attack. Same for Laser Lock.

xezon:
This change does not work properly.

Issue #1332, opened 2022-10-04
GLA Quad Cannon, China Gattling Tank, Gattling Turret, has no recoil and muzzle flash on ground attack


xezon:
GLA Quad Cannon, China Gattling Tank, Gattling Turret, has no recoil and muzzle flash on ground attack

xezon:
Some of the issues were tackled in this change already, however it should probably be redone in isolation.

* #1222

commy2:
Issue does not apply to Helix Gatling Gun or Minigunner.

commy2:
done

ExiLe-ZH:
video examples? :)

commy2:
Gatlings:

https://user-images.githubusercontent.com/6576312/194762649-2570885b-ddc2-46ed-88cf-a6ce1987212b.mp4

commy2:
Quad Cannon:

https://user-images.githubusercontent.com/6576312/194764236-2064b105-d900-46f2-b27f-2cd6951f70c4.mp4

Note that these are already in the game, but only work when aiming at airborne units.

ExiLe-ZH:
Nice work! ❤️

PR #1341, opened 2022-10-08
Fix: Gatling Tank Missing Muzzle Flash When Firing At Grounded Targets

commy2:
- partial fi.x for #1332


xezon (approved):
Looks very good.

commy2:
Adjusted comments, because there is no recoil before or after.

PR #1222, opened 2022-09-18
Fix: China Overlord, Emperor model, texture, Gattling issues

xezon:
This change fixes all jank on Overlord and Emperor models.

Fixes:
* All muzzle flashes are viewable from all sides
* Gattling Cannon muzzle flashes do show now
* Gattling Cannon barrels rotate with proper speeds
* Gattling Cannon stops rotating on cease fire
* Chassis holes on bottom of Overlord closed
* Frankenstein Emperor models replaced with new model
* ~~Removed odd wheels from Overlord chassis textures~~

Does not fix:
* Installing vehicle addon on REALLYDAMAGED vehicle does not spawn damaged addon variant

All W3D Ini changes have been copied as is from NProject. They have been tested and found working. The Emperor Tank has some additional setup merged from Crazy Mod setup due new Emperor model.

New Emperor model uses same textures as original Overlord and Emperor do. The original Emperor is a mash up of Overlord parts poorly put together, with stretched textures, poorly intersecting meshes, treads texture on the top of the chassis, and visible bones on the damaged model variant.

The Emperor model was created by Jordan, Crazy Mod.
https://www.moddb.com/mods/cc-generals-crazy-mod/images/china-emperor-tank-re-make

### TODO

- [ ] Remove all formatting changes
- [ ] Recreate model fixes

# Overlord

## Original

shot_20220918_162112_11

## Patched, NProject

shot_20220918_161359_7

# Emperor

## Original

shot_20220918_161738_1

shot_20220918_161840_2

shot_20220918_161905_3

shot_20220918_161912_4

shot_20220918_163154_3

## Patched, CrazyMod

shot_20220918_083131_1

shot_20220918_161225_3

PR #1343, opened 2022-10-08
Fix: Overlord Gatling Cannon Missing Muzzle Flash When Firing At Grounded Targets

commy2:
- partial fi.x for #1332

This includes the Emperor.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -459,6 +459,8 @@ Object ChinaTankOverlordGattlingCannon
   UpgradeCameo5 = Upgrade_ChinaOverlordGattlingCannon
 
 
+  ; Patch104p @bugfix commy2 08/10/2022 Fix muzzle flash when aiming at grounded targets.
+
   Draw                    = W3DDependencyModelDraw ModuleTag_01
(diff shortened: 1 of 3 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
```diff
@@ -492,15 +508,23 @@ Object ChinaTankOverlordGattlingCannon
       ParticleSysBone    = Muzzle01 GattlingMuzzleSmoke
       ParticleSysBone    = Muzzle02 GattlingMuzzleSmoke
     End
+    ConditionState        = USING_WEAPON_B CONTINUOUS_FIRE_FAST
+      Model               = NVOvrlrd_G
+      Animation           = NVOvrlrd_G.NVOvrlrd_G
+      AnimationMode       = LOOP
+      WeaponFireFXBone    = SECONDARY Muzzle
+      WeaponMuzzleFlash   = SECONDARY MuzzleFX
+      AnimationSpeedFactorRange = 0.3 0.3 ;set this state to animate  vryfst
```
xezon, line 517: After this is merged, next fix is fixing these AnimationSpeedFactorRange values to match other Gattling turrets. Plus there is another issue with Overlord Gattling Cannon where it does not stop spinning. It is fixed in #1222 somehow.

xezon (approved):
Looks very good.

PR #1344, opened 2022-10-08
Fix: Gatling Cannon Defense Missing Muzzle Flash When Firing At Grounded Targets

commy2:
- partial fix for #1332


xezon (approved):

commy2:
Updated to include the most recent fixes: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1345 and https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1347.

commy2:
And https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1348.

PR #1349, opened 2022-10-08
Fix: Overlord Gatling Cannon Keeps Spinning Indefinitely After Firing Once

commy2:
- 1.04

Gatling Cannon on Overlord/Emperor slowly spins forever after firing a single shot.

- patch

Gatling Cannon stops spinning after resting for a few seconds.


commy2:
This is because all references to the animation are missing in default animation mode.

xezon (approved):
Very nice.

PR #1359, opened 2022-10-10
Match Gattling barrel spin animation speeds with fire rates

xezon:
This change matches all Gattling barrel spin animation speeds with their respective fire rates.

This change has multiple commits and should be Merged with Rebase.

The new animation speeds were determined by observing the individual fire rates, barrel spins frame by frame and adjusting the animation speeds accordingly. Surely there is some math to it, but I did not check it.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -13552,7 +13552,7 @@ Object Boss_GattlingCannon
       Model               = NBGattling
       Animation           = NBGattling_A2.NBGattling_A2
       AnimationMode       = LOOP
-      AnimationSpeedFactorRange = 0.5 0.5 ;set this state to animate  s l o w l y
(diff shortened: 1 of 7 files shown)


commy2 (approved):
lgtm
Only worry is, that making effects slower for Gatling tank means, it won't match when vetted.

xezon:
Hmm good point. I did not test that. Perhaps for Gattling Tank we would need to set speed to somewhere between firing rate of Vet 0 and Vet 3. Does Overlord/Helix Gattling Cannon change on Veterancy?

commy2:
> Does Overlord/Helix Gattling Cannon change on Veterancy?

I don't think so. No red tracers at least.

xezon:
I tested Gattling Tank CONTINUOUS_FIRE_FAST on Vet 1 with RATE_OF_FIRE 120% and Vet 3 with RATE_OF_FIRE 160%, and the muzzle flashes have exact same interval as with Vet 0 with RATE_OF_FIRE 100%. So the barrel spin remains a match. I think there is nothing more to do here.

PR #1360, opened 2022-10-10
Change: Increase exshell texture visibility

xezon:
This change fixes Alpha channel of exshell texture. Opacity was raised from 240 to 255. It is used for Machine Gun particle effects.


xezon:
I tested this more and noticed that the bullets are now much fatter on smaller mips due the increased alpha. Closing.

PR #1367, opened 2022-10-15
Fix: Strategy Center Fan Animation Broken On Night Maps Unless Damaged

commy2:
### 1.04

strat center

- These fans are not animated on Night maps (both Summer and Winter), except when the building is damaged.

### patch

- The fans are always animated.


xezon:
On _DS USA Command Center there is also one Fan not spinning, but on _ES it is spinning again. I suspect identical setup issue to this.

xezon (approved):

commy2:
Command Center fan not spinning is a model issue, not fixable by INI changes. It also happens on summer maps.

PR #1370, opened 2022-10-15
Fix: USA Helipad Model Uses Wrong Texture In Badly Damaged State On Day Summer Maps

commy2:
### 1.04

- USA Helipad on Summer Day maps uses the same texture in damaged and badly damaged states.

### patch

- USA Helipad uses a matching texture in all damage states on all maps.

Helipad 1 04


commy2:
This is not an issue on Night or Snow maps:

<details><summary>click to expand</summary>



</details>

xezon (approved):

PR #1371, opened 2022-10-15
Fix: Remove Boneless Effects From USA Reactor, Rename Bones And Remove Misplaced Spark Effects

commy2:
### 1.04

Cold Fusion Reactor 1 04

- There are many effects coming out of the center of the reactor when damaged, because the bones used in INI do not exist in the model.
- The chimney smoke no longer comes from the chimney when damaged, but from the center of the building instead. 

### patch

Cold Fusion Reactor patched

- Fixed chimney smoke.
- Smoke effects from center of building are removed. Fire and smoke effects only emerge from places where they belong.


commy2:
- [x] There are still issues with the sparks.

commy2:
With this ^ change, all bones keep their names across the models. Some spark effect sources are removed, e.g. those spawning inside the building or at the center of the model (thus being invisible).

commy2:
Further issues with USA Power Plant: Wires disappear on Night maps (Summer and Winter) in badly damaged state. This does not happen in Day maps.

xezon:
Did you mean to "Merge branch 'main' into fix-usa-reactor-effects"? It means if we Rebase and Merge, it will be in commit history.

commy2:
I just fixed a merge conflict. This PR should be squash-and-merged.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -8724,348 +8726,222 @@ Object AirF_AmericaPowerPlant
...
+      ParticleSysBone    = Smoke01 SteamVent
     End
     ConditionState       = DAMAGED
       Model              = ABPWRPLANT_D
-      ParticleSysBone    = Smoke01 SmolderingSmoke
-      ParticleSysBone    = Smoke02 SteamVent
-      ParticleSysBone    = Smoke03 SmolderingSmoke
-      ParticleSysBone    = Fire01 SmolderingFireLarge
+      ParticleSysBone    = Smoke01 SteamVent
       ParticleSysBone    = Sparks01 LiveWireSparks
```
xezon: on DAMAGED there are sparks, but on REALLYDAMAGED are none. I think it would make more sense if it was consistent. Sparks on both or none.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -8724,348 +8726,222 @@ Object AirF_AmericaPowerPlant
...
       Model              = ABPWRPLANT_E
-      ParticleSysBone    = Smoke01 SmolderingSmoke
-      ParticleSysBone    = Smoke02 SteamVent
-      ParticleSysBone    = Smoke03 SmolderingSmoke
-      ParticleSysBone    = Smoke04 SmolderingSmoke
-      ParticleSysBone    = Smoke05 SmolderingSmoke
-      ParticleSysBone    = Smoke06 SmolderingSmoke
-      ParticleSysBone    = Fire01 SmolderingFireLarge
-      ParticleSysBone    = Fire02 SmolderingFireLarge
-      ParticleSysBone    = Fire03 SmolderingFireLarge
```
xezon, line 8751: I am a bit confused about this one. All the smoke and fire effects are removed here, but we still see smokes and fires on the damaged models, which of course is desired. But how is this possible?

commy2: There are 0 to multiple Fire{s}{xx} and Smoke{s}{xx} bones in any building model (where s is any of S, M, L and xx is a two digit number). These are coded to spawn effects "automatically" (See: [gamedata.ini](https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesEdited/Data/INI/GameData.ini#L142-L159)).

It is highly unusual for any building to spawn damage effect particles via ParticleSysBone, unless it is something specialized like sparks. What is usually spawned via ParticleSysBone are effects like chimney smokes / steam etc.

That these exist at all for some buildings must be an oversight during development. Left over from before the AutoFire/SmokeParticleSmallPrefix system was put in place. Now most of the bones don't exist and cause effects to be spawned at the center of the building. And sometimes the damage effects interact with the "normal" effects like chimney steams or Overcharge particles.

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
commy2: From memory, there is no bone for this in reallydamaged. Sometimes this is because EA forgot, sometimes it is because that part of the structure is entirely destroyed in E. Would have to check ingame to see what is the case here.

xezon: In E the power lines are still present.

commy2: `Sparks` (not the be confused with `Spark` bone does not exist in E models.

xezon: So, remove it from D then?

xezon (approved):
Very good!

PR #1372, opened 2022-10-15
Fix: Control Rods Upgrade Icon Missing From Boss Reactor

commy2:
The Boss Reactor can be upgraded with Control Rods, but it has no upgrade icon like all the xUSA Reactors have.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -1438,6 +1438,9 @@ Object Boss_PowerPlant
   SelectPortrait         = SAPowerPlant_L
   ButtonImage            = SAPowerPlant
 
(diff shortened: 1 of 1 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
```diff
@@ -1438,6 +1438,9 @@ Object Boss_PowerPlant
   SelectPortrait         = SAPowerPlant_L
   ButtonImage            = SAPowerPlant
 
+  ; Patch104p @bugfix commy2 3/10/2021 Add missing upgrade icon for Subliminal Messaging.
```
xezon: Date and message is not matching.

commy2: fixed

xezon (approved):

PR #1374, opened 2022-10-15
Change: Use New nbptower_e Texture In Propaganda Center

commy2:
See: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1239


xezon (approved):

commy2:
NBPCenter_ES.W3D
NBPCenter_ENS.W3D

should probably use nbptower_es.tga in future.

PR #1375, opened 2022-10-15
Fix: USA Power Plant Construction Fence Texture On Snow Night Maps

commy2:
### 1.04

Power Plant Fence 1 04

- The construction fence is very dark.

### patch

- The construction fence of the Power Plant uses the same texture as any other USA building construction fence.


xezon (approved):

xezon:
Not sure how this is meant to be committed.

Issue #1259, opened 2022-09-23
USA Powerplant model issues


xezon:
- [ ] Snow model meshes show no alpha texture, but texture has alpha (Image 1, 2)
- [ ] Wires disappear in Rubble state, but only at night (Summer and Winter). Doesn't happen at day. Sparks fly when upgraded, so wires should be added to the model (Image 3)
- [ ] Snow model uses non-snow textures

## Image 1

![shot_20220923_103957_1](https://user-images.githubusercontent.com/4720891/191923522-b01d84f2-f258-4694-8e6e-dc6b03d63e56.jpg)

## Image 2

![shot_20220923_104039_2](https://user-images.githubusercontent.com/4720891/191923543-ee9101d4-cb18-4bf8-979e-0b61fe9879a9.jpg)

## Image 3

![Cold Fusion Reactor damaged night](https://user-images.githubusercontent.com/6576312/196003018-e85a4af3-ec51-4293-8f79-e8a4d76618a3.jpg)

commy2:
Wires disappear in Rubble state, but only at night (Summer and Winter). Doesn't happen at day. Sparks fly when upgraded, so wires should be added to the model.

![Cold Fusion Reactor damaged night](https://user-images.githubusercontent.com/6576312/196003018-e85a4af3-ec51-4293-8f79-e8a4d76618a3.jpg)

xezon:
_S model does not show alpha texture channel at all. Possibly mesh material setup issue. Not a problem on _DS and _ES.

![shot_20221019_213639_1](https://user-images.githubusercontent.com/4720891/196787649-86aa8d24-8b4a-4893-8e07-d5331b56dbb1.jpg)

xezon:
Snow model(s) use non-snow textures for some meshes.

PR #1376, opened 2022-10-15
Fix: USA Power Plant Missing Animation When Damaged On Day Maps

commy2:
### 1.04

- Reactor cylinders are not animated on Summer and Winter Day maps when damaged. They are however animated in other states, including at Night.

### patch

- Cylinders are always animated.

This does not fix the black texture (no alpha channel?) mentioned here: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1259. That bug only happens on Snow (Day?) maps when UNDAMAGED.


xezon:
I can look into the textures.

xezon:
_S texture is ok. This appears to be a problem with mesh materials.

xezon (approved):

PR #1377, opened 2022-10-16
Change: Increase Beacon Limit From 3 To 5

### LINKED ISSUE #1363: Beacons have limit of 3 per player
Not sure if it's due to performance reasons, but the limit of 3 is way too low and it can mess up team play when you need to beacon quickly. Now to mark SW's for yourself you could set bookmarks but it's not as precise.

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1363

Self explanatory.


xezon:
We need to test in Multiplayer that this actually works properly.

MTKing4:
can test with you

xezon:
Works.

shot_20221030_163458_2

xezon (approved):

PR #1379, opened 2022-10-16
Fix: Remove Boneless Effects From China War Factory

commy2:
### 1.04

China War Factory 1 04

- Many fire and smoke effects are emerging from the center of the building when damaged, because there are no bones for them in the model. The effects are invisible, because they are covered by the mesh.

### patch

- Effects without bones are removed.


commy2:
Skript used:

```py
import re

files = [
    "NBWarFact.W3D",
    "NBWarFact_S.W3D",
    "NBWarFact_N.W3D",
    "NBWarFact_NS.W3D",
    "NBWarFact_D.W3D",
    "NBWarFact_DS.W3D",
    "NBWarFact_DN.W3D",
    "NBWarFact_DNS.W3D",
    "NBWarFact_E.W3D",
    "NBWarFact_ES.W3D",
    "NBWarFact_EN.W3D",
    "NBWarFact_ENS.W3D",
]

bones = [
    rb"FIRE\d\d",
    rb"SMOKE\d\d",
]

patterns = [re.compile(b) for b in bones]

for file in files:
    print(file)

    with open(file, "rb") as f:
        data = f.read()

    for pattern in patterns:
        for match in pattern.finditer(data):
            start, end = match.span()
            print("\t" + data[start: end].decode())

    print()
```
Notes:

`Smoke01` turns out to be on top of the chimney in D, so the effect is kept. E chimney is fallen down, so no effect used for those models. `Smoke01` also exists in normal and Night, but is missing on Snow. The bone is placed below the chimney at z=0, so no effect was _added_ (there is none in 1.04n for undamaged).

Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
```diff
@@ -28786,30 +28788,9 @@ Object ChinaWarFactory
...
-      ParticleSysBone = Smoke04 SmolderingSmoke
-      ParticleSysBone = Smoke05 SmolderingSmoke
-      ParticleSysBone = Fire01  SmolderingFire
-      ParticleSysBone = Fire01  SmolderingFlameCore
-      ParticleSysBone = Fire02  SmolderingFire
-      ParticleSysBone = Fire02  SmolderingFlameCore
     End
     ConditionState    = REALLYDAMAGED RUBBLE
       Model           = NBWarFact_E
-      ParticleSysBone = Smoke01 SmolderingSmoke
```
xezon, line 28800: So this means there is a Smoke01 on _D but not in _E. Is this removal reasonable or does Smoke01 need to be added to the _E model?

commy2: Smoke01 sits at the top of the chimney in D. In E, the chimney has collapsed.

xezon:
Can you rebase this? Change History looks a bit wonky right now.

xezon:
Or you merged into this branch? Not sure what is going on.

commy2:
Yes, #1380 and #1381 were merged into this branch.

xezon (approved):

PR #1380, opened 2022-10-16
Fix: Remove Floating Basket From Undamaged China War Factory On Snow Maps At Night

commy2:
### 1.04

China War Factory 1 04 markings

- There is a floating basket inside the China War Factory on Snow Night maps when undamaged (marking 2). It disappears when damaged.

### patch

- No floating baskets.


xezon (approved):
Nice find

PR #1381, opened 2022-10-16
Fix: China War Factory Conveyor Belt Animation Issues

commy2:
This change makes:

- Removes the conveyor belt animation from badly damaged states (Snow, Night, Night+Snow), which solves problems with floating baskets above the deformed pole and baskets colliding with a stuck basket inside the furnace.
- The conveyor belt animation no longer reset when starting / finishing production
- Instead, the conveyor belt animation only advances while the factory is producing units/upgrades.


xezon:
I like this. Certainly a big improvement over the Original setup.

xezon (approved):

PR #1382, opened 2022-10-16
Fix: Factory Door Animation Damage Transitions

Stubbjax:
Fixed door animations resetting when transitioning between damage states for vehicle production buildings (GLA Arms Dealer + China War Factory + USA War Factory).

## 1.04:
When the War Factory transitions between damage states, the door animation state is reset.

https://user-images.githubusercontent.com/11547761/196031878-4fa989b2-c5ec-4a87-b479-92e106172ba3.mp4

## Patch:
When the War Factory transitions between damage states, the door animation state is maintained.

https://user-images.githubusercontent.com/11547761/196031884-4bd62cc6-1994-4db5-b4e4-f99c2f1ec3e0.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -15493,7 +15493,7 @@ Object AirF_AmericaWarFactory
       Model           = ABWarFact_A8
       Animation       = ABWarFact_A8.ABWarFact_A8
       AnimationMode   = ONCE
-      Flags           = START_FRAME_FIRST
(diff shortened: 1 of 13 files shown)


xezon:
Oh wow I love it. A bug that everyone has noticed.

What about USA, China Airfield doors? And China, USA Command Center. They might have same issue.

Stubbjax:
> Oh wow I love it. A bug that everyone has noticed.
> 
> What about USA, China Airfield doors? And China, USA Command Center. They might have same issue.

Yep, a bunch of buildings require the same treatment. I was going to do those in separate pull requests but I can append them to this one if you believe it is preferable?

xezon:
As you wish.

xezon (approved):

ExiLe-ZH:
Mega pro ++++++++++++

commy2 (changes requested):
GC Stealth and Tox missing.

commy2 (approved):
Very nice.

PR #1388, opened 2022-10-16
Change: Use USA War Factory Crane Logic For China War Factory Crane

commy2:
This means the crane no longer resets when starting a new unit.


xezon:
I tracked issue for the animation jump with
* TheSuperHackers/GeneralsGameCode#157

Other than that this change looks very good.

xezon (approved):

Issue #17, opened 2021-08-23
Crates dropped on USA Supply drop zone sometimes land next to the zone at random


xezon:
Check whether or not USA Supply drop zone drop offs should scatter as much as they currently do.

![post-199071-1635771399](https://user-images.githubusercontent.com/4720891/139690312-88e8b23b-27a3-4a75-85a8-3008d5b53542.jpeg)

djchikito:
You could correct the error that supply bonuses fall outside the delivery areas and have to be sent for manual collection.

ReLaX82:
If possible, yes.

Jundiyy:
There is benefit but at the same time it would also change the way team games are played when on low income. As you can pass money on and would be a lot harder or not possible to do if fixed.
Of course the other argument is that players have to manually collect, which is extra work.
But again, as USA you get the most money with secondary economy and even more after the upgrade.

RisingZH:
I would fix.

ExiLe-ZH:
would fix too

Stubbjax:
> 
> 
> There is benefit but at the same time it would also change the way team games are played when on low income. As you can pass money on and would be a lot harder or not possible to do if fixed.
> Of course the other argument is that players have to manually collect, which is extra work.
> But again, as USA you get the most money with secondary economy and even more after the upgrade.

If a player wants to give their ally money, they can still do so by selling off the SDZ as soon as the cargo plane spawns, which is cleaner as it is then based on more intentional / skillful gameplay and less a coincidence of faulty and unreliable behaviour.

MTKing4:
Fix it

ReLaX82:
Realistically is what we have on 1.04, but its unsecure 2nd eco.

So, fix it.

xezon:
Should check if this still happens in Patch.

xezon:
With change

* #1390

the chance of hit miss appears to have been removed or reduced. Boxes no longer bounce off nearby structures.

The new cargo boxes also hit the very center of the Supply Drop Zone more accurately as opposed to the legacy boxes, indicating that the reduced length of movement or reduced number of boxes is beneficial to the drop accuracy.

## Original

![shot_20230116_130915_2](https://user-images.githubusercontent.com/4720891/212677005-b8f9c2ac-c9e7-4c99-84d8-7ebf28fc35a3.jpg)

## Patched

[...]

PR #1390, opened 2022-10-18
Change: Reduce Supply Drop Zone crate count from 6 to 3

xezon:
This change reduces the Supply Drop Zone crate count from 6 to 3. The value per crate increases from 250 to 500. For gameplay this should make no difference. The Supply Drop is no player controlled event. The first crate of 3 will fall down a bit later than the first crate of 6, but the last crate of 3 drops earlier than the last crate of 6.

I calculated the new drop distance with

410 * (1 - (1 - ((500 * 3) / (350 * 6))) * 0.5) = 351.4

The reduced crate count spams less sounds and spawns less world objects, which incurs less performance cost.

## Original

https://user-images.githubusercontent.com/4720891/196511579-9352fbaf-3d4f-41e2-8a33-9edfaa058c1e.mp4

## Patched

https://user-images.githubusercontent.com/4720891/196511663-a1092549-dab2-4e56-9719-6184c6f10cb3.mp4

## Rationale

Less world objects impose less cost in the game world. As illustrated above, the reduced crate count causes boxes to intersect less and spawn less sounds. Originally the sounds could not even keep up with the amount, as some parachute and cash sounds would be swallowed by the short interval. Additionally the new 3 boxes fly more gentle to the center of the Supply Drop Zone than some of the original ones which may land on the outer edges.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
@@ -549,7 +549,7 @@ Object SupplyDropZoneCrate
   KindOf = PARACHUTABLE CRATE INERT
   Behavior = MoneyCrateCollide      ModuleTag_ForbiddenChanges
     ForbiddenKindOf = PROJECTILE
-    MoneyProvided = 250
+    MoneyProvided = 500 ; Patch104p @tweak from 250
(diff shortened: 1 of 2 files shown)


ExiLe-ZH:
That looks pretty clean!

xezon:
Looks good.

xezon:
This change appears to have removed or reduced the cargo box miss chance.

* See #17

PR #1391, opened 2022-10-18
Change: Reduce particle count of AmericaJetCargoPlane_SupplyDropZone

xezon:
This change removes the black smoke trails from all 4 propellers on Supply Drop Zone plane. Only the contrails on the wing tips remain. This will render it a bit cheaper.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
@@ -2513,6 +2513,7 @@ Object AmericaJetCargoPlane
 End
 
 ; Patch104p @performance hanfield commy2 06/08/2022 A harmless plane that can not be attacked.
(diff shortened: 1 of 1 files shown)


commy2:
Is this necessary? Is there a tangible benefit?

ExiLe-ZH:
Idk, I really like those visuals. Is it really worth to sacrifice visual quality for the performance gain?

commy2:
Same ^
Supply Plane doesn't seem so bad compared to Hackers. And the +$20 spam from BMs is also done with particles, and that happens all the time, not every 2 minutes.

xezon:
Did you test it? The particle trails behind the propellors are faint. I was unable to observe a noticable visual degradation with these removed. Removing these 4 particles saves 4 particles per plane, or 40 for 10 planes. I would expect this to give some performance gain, because nothing is cost free. Do you need this measured?

xezon:
Other USA cargo planes also have propeller smoke.

PR #1394, opened 2022-10-19
Change: Animate USA War Factory Crane When Repairing Instead Of When Constructing

commy2:
### 1.04

- Crane on War Factory moves while building a unit.

### patch

- Crane on War Factory moves while repairing a unit instead.
- You can still tell if a unit is being build, because it produces smoke from the chimney.


xezon (approved):
Looks good.

PR #1395, opened 2022-10-19
Fix: China Power Plant Particle Effect Issues And Remove Effects Without Bones

commy2:
### 1.04

China Power Plant 1 04

- Overcharging sparks are shown while badly damaged, even when not Overcharging (not visible in image). This also only happens on day maps.
- Smoke and fire effects at center of building in damaged and badly damaged states due to effects on non-existing bones.
- Chimney smoke reduced/very white in damaged states. Only happens on day maps and also stops while Overcharging.
- Chimney fire missing on night maps.
- Chimney fire missing in badly damaged states. This may be intentional, as another fire breaks out on the small building.

---

Nuke Power Plant 1 04

- Nuke variant has no smoke over small chimneys. Bones exist and smoke is shown in portrait art.

---

Existing bones:
```
All
    Smoke01/02
    Spark01-07

All D/E
    Smoke06

Nuke _/N/S/NS:
    Smoke03 (thin little chimney)
    Smoke04 (thick little chimney)

Nuke D/DN/DS/DNS:
    Smoke03 (thin little chimney)
```

<details><summary>*click for script*</summary>

```py
import re
from pathlib import Path

files = list(Path(".").glob("*.W3D"))

bones = [
    rb"FIRE\d\d",
    rb"SMOKE\d\d",
    rb"SPARK\d\d",
]

patterns = [re.compile(b) for b in bones]

for file in files:
    print(file)

    with open(file, "rb") as f:
        data = f.read()

    for pattern in patterns:
        for match in pattern.finditer(data):
            start, end = match.span()
            print("\t" + data[start: end].decode())

    print()
```

</details>

### patch

- Overcharge particles only shown when overcharging.
- No effects emerging from center of building.
- Big fire in damaged state shown on all maps. Smoke column not reduced on some maps either.
- No inexplicable differences between normal and Overcharging mode.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -18542,7 +18542,7 @@ Object ChinaPowerPlant
       Model           = NBPwrPlant_E
       ParticleSysBone = Smoke01 SteamVent
       ParticleSysBone = Smoke02 SteamVent
-      ParticleSysBone = Smoke06 SmolderingSmoke
+      ParticleSysBone = Smoke06 FireFactionMedium
       ParticleSysBone = Spark01 LiveWireSparks
       ParticleSysBone = Spark02 LiveWireSparks
       ParticleSysBone = Spark03 LiveWireSparks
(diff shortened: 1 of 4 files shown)


xezon:
I will test this.

xezon:
There are 2 smoke particles on the _E big chimney. They have different look and velocity. I suggest replacing the small smoke with a small fire.

shot_20221020_114001_2

commy2:
The fire in D is replaced by a generic smoke column in E. Same bone (position). It is mixed with the steam from the Reactor. Reactor steam already is two particle effects next to each other, presumably for volume. So there're actually three smoke pilars in that chimney in 1.04.

xezon:
Looks very nice now. One more improvement I can think of is making one of the big smokes in _D and _E black, so that it looks darker than in undamaged. But we can do that in separate change, as that is not directly related to this change.

shot_20221021_133453_1

shot_20221021_133607_2

xezon (approved):

PR #1396, opened 2022-10-19
Add: Smoke Pillars To Small Chimneys On Nuke General Power Plant

commy2:
- merge after https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1395

### 1.04

- Smoke from small chimneys on portrait art, but not on model

### patch

- Smoke on model and portrait

Notes:

Small thick chimney is missing the bone in damaged state. Assume it is "broken" (meaning: the building, not the model). Badly damaged state chimneys are missing or collapsed, so no effects either.


xezon:
I will test this.

xezon:
I like the sizes and colors. However, the smoke directions do not match. The big reactor smoke has wind simulation pushing it rightwards. The new smoke is flying upwards. They should fly into the same direction. On that note, we may want to tackle this in a separate change if this is broader issue affecting several smokes.

shot_20221020_115025_1

xezon:
Smoke drift addressed with
* #1402

This change on its own is good as is.

xezon (approved):

xezon:
This change needs merge conflicts resolved.

PR #1398, opened 2022-10-20
Fix: Command Center Door Animation Damage Transitions

Stubbjax:
Fixed door animations resetting when transitioning between damage states for the China and USA Command Centers. Also fixed the radars resetting their animations as well.

The below footage demonstrates the doors of both Command Centers seamlessly opening and closing while transitioning between damage states (+ USA satellite dish).

https://user-images.githubusercontent.com/11547761/196996231-6d1a7c85-00d4-4ea9-afbb-a9d6f762b77b.mp4

https://user-images.githubusercontent.com/11547761/196996251-9e6789d5-ad18-4723-a410-d982dc8bdea1.mp4

One strange thing to note is that it appears the USA Command Centers each have a duplicate door entry, which likely means they are being rendered / processed twice (and changing both of them was necessary to solve the respective animation issue, rather than one or the other). Further investigation may be required.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -7783,25 +7783,25 @@ Object AirF_AmericaCommandCenter
       Model           = ABBtCmdHQ_A7
       Animation       = ABBtCmdHQ_A7.ABBtCmdHQ_A7
       AnimationMode   = ONCE
-      Flags           = START_FRAME_FIRST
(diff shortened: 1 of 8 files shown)


xezon (approved):
Looks good.

commy2:
Video of this with upgraded China CC (Radar)?

Stubbjax:
> Video of this with upgraded China CC (Radar)?

https://user-images.githubusercontent.com/11547761/197647140-73f7de29-f400-4945-8fb2-1aa1de0599c9.mp4

commy2 (approved):

Issue #638, opened 2022-01-18
Reduce effectiveness of USA Airforce Point Defense Laser


xezon:
Explore feature for countering USA Airforce Point Defense Laser (PDL).

### Proposal 1
Add a new laser lock mode to USA Missile Defender, that works similar to traditional laser lock, but instead of sending damaging rockets it will just fire alot of dummy bullets to overwhelm the PDL ability of the targeted aircraft and/or vehicle. With such feature, USA can try to combine targeting and attacking Missile Defenders to combat against USA Airforce units. **Strengthens USA vs Airforce and Avengers.**

### Proposal 2
Increase the USA Airforce PDL reload time to reduce its effectiveness.
**> Weakens Airforce against all factions.**

### Proposal 3
Add PDL to SWG Chinooks, and maybe make them as cheap as AFG Chinooks too.
**> Strengthens SWG against all factions.**

### Proposal 4
Make USA Airforce Chinook and King Raptor PDL an upgrade from the Strategy Center.
**> Weakens Airforce against all factions in early game. Combat Chinook rushes and fast King Raptors are no longer as difficult to stop.**

### Proposal 5
Make Projectiles from Missile Defender immune to PDL.
**> Strengthens USA vs Airforce and Avengers.**

* #1106

### Proposal 6
King Raptor has two PDL modules assigned and is twice as affective as Chinook. Remove one of them and tweak the remainders desired effectiveness.
**> Weakens Airforce against all factions.**

* #1107

### Proposal 7
Make Projectiles from USA Missile Defender more resistant to PDL, so they take more zaps to be destroyed.
**> Strengthens USA against Airforce and Avengers**

* #1108

### Proposal 8
Make Projectiles from all Rocket Men more resistant to PDL, so they take more zaps to be destroyed.
**> Strengthens all factions against Airforce and Avengers**

### Proposal 9
Activate Aircraft PDL only after purchasing the Countermeasures Upgrade in the Airfield. Remove traditional countermeasures flares if required. Note: Countermeasures Upgrade does not apply Armor boost on King Raptor unlike normal Raptor.
**> Weakens Airforce against all factions in very early game. Combat Chinook rushes and fast King Raptors are no longer as difficult to stop.**

### Proposal 10
Make PDL a special ability like laser lock is a special ability of USA Missile Defenders. When PDL icon is pressed that unit activates its PDL ability for X seconds. After X seconds passed it will cooldown for Y seconds until it can be activated again.
[...]

Issue #108, opened 2021-08-29
China ECM tank behaves differently between China factions


xezon:
China ECM tank behaves differently between China factions. Figure out what to do about it.

commy2:
vChina does not auto-attack nearby enemies when idle, only when in Guard mode.

Tank/Nuke/Inf does auto attack the nearest enemy.

See: https://github.com/commy2/zerohour/commit/68befac03dabebf6dee270f63ef78d09bf538196

xezon:
There are different options:
- Make all ECM Tanks behave the same
- Explain this ECM Tank difference in UI Tooltip
- Give ECM Tank slightly different model to signal different functionality

Jundiyy:
I would prefer 2 or 3.
3 would be great actually although might possibly throw some people off? I'd love it though if all agree.

Jundiyy:
I'm not 100% sure as I'm really bad with modelling but I think it's really easy to add extra army colour (housecolour) to units, could add an extra line or 2 going across the front (and back). So that could be a simple 'new model' if we can't get anything else (and it is the way we choose).

commy2:
I'm for making all of them behave the same. Not auto attacking, since that is the one that makes the most sense for this unit, as attacking turns off the jammer.

xezon:
I also prefer to have them all behave the same for consistency sake. They could all behave like the China one, or the other one. Either behaviour would be reasonable.

RisingZH:
Should all behave the same, regardless of faction. Should auto attack enemy unit if idle and player has retaliation mode on (like any other unit would).

ExiLe-ZH:
All should be the same and I prefer china's V behavior, because I feel it's the intended behavior.

The focus of micro'ing a unit should be on what it should be doing, not preventing it from doing something. That's just annoying and takes time away of micro'ing other important units.

Also where "micro" is taken away, new micro needed when wanting to disable units, which can be crucial ones like incoming TT's or sneaky dragons (guard mode could be used). Also mass disabling units still need the same micro.

ZekeDlyoung:
I also vote for all of them acting the same as the vchina one as it offers the most flexibility:
- if you need a missile shield then park it somewhere and leave it alone
- if you need to disable something, just manually tell it to do so, like how artillery works

It will still auto-attack when guarding or doing an attack move anyway

Madrageeeeee:
Agree with Rising here. Auto disable with retaliation on makes sense. 

[...]

PR #1400, opened 2022-10-21
Change: Winged aircraft no longer returns to the Airfield after 10 seconds

xezon:
This change improves the control over air planes. Players can leave airplanes idle for an unlimited amount of time. Idle planes will no longer return to the airfield after 10 seconds unless ordered so or out of ammo. The guard mode functionality of airplanes is unchanged.

## Rationale

Players are no longer punished for the lack of plane babysitting. Instead players can invest their attention into more meaningful tasks in a given match.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -1778,7 +1778,7 @@ Object AirF_AmericaJetAurora
     AttackLocomotorPersistTime  = 100    ; we start slowing down almost immediately
     AttackersMissPersistTime    = 2000   ; but remain untargetable fer a bit longer
     ReturnForAmmoLocomotorType  = SET_SLUGGISH
-    ReturnToBaseIdleTime        = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime        = 0         ; if idle for this long, return to base, even if not out of ammo
   End
   Locomotor = SET_NORMAL      AuroraJetLocomotor
   Locomotor = SET_SUPERSONIC  AuroraJetSupersonicLocomotor
@@ -2020,7 +2020,7 @@ Object AirF_AmericaJetStealthFighter
     TakeoffDistForMaxLift     = 0%   ; larger numbers give more lift sooner when taking off
     TakeoffPause           = 500
     MinHeight              = 5
-    ReturnToBaseIdleTime   = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime   = 0         ; if idle for this long, return to base, even if not out of ammo
   End
 
   Behavior = ExperienceScalarUpgrade ModuleTag_08
@@ -17997,7 +17997,7 @@ Object AirF_AmericaJetRaptor
     TakeoffPause              = 500
     MinHeight                 = 5
     ParkingOffset             = 3             ; scooch it a little forward so the tail doesn't hit the doors
-    ReturnToBaseIdleTime      = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime      = 0         ; if idle for this long, return to base, even if not out of ammo
   End
   Locomotor = SET_NORMAL RaptorJetLocomotor
   Locomotor = SET_TAXIING BasicJetTaxiLocomotor
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaAir.ini
@@ -700,7 +700,7 @@ Object AmericaJetRaptor
     TakeoffPause              = 500
     MinHeight                 = 5
     ParkingOffset             = 3             ; scooch it a little forward so the tail doesn't hit the doors
-    ReturnToBaseIdleTime      = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime      = 0         ; if idle for this long, return to base, even if not out of ammo
   End
   Locomotor = SET_NORMAL RaptorJetLocomotor
   Locomotor = SET_TAXIING BasicJetTaxiLocomotor
@@ -940,7 +940,7 @@ Object AmericaJetAircraftCarrierRaptor
     TakeoffPause              = 500
     MinHeight                 = 5
     ParkingOffset             = 3             ; scooch it a little forward so the tail doesn't hit the doors
-    ReturnToBaseIdleTime      = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime      = 0         ; if idle for this long, return to base, even if not out of ammo
     AutoAcquireEnemiesWhenIdle = YES ATTACK_BUILDINGS
   End
   Locomotor = SET_NORMAL RaptorJetLocomotor
@@ -1363,7 +1363,7 @@ Object AmericaJetAurora
     AttackLocomotorPersistTime  = 100    ; we start slowing down almost immediately
     AttackersMissPersistTime    = 2000   ; but remain untargetable fer a bit longer
     ReturnForAmmoLocomotorType  = SET_SLUGGISH
-    ReturnToBaseIdleTime        = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime        = 0         ; if idle for this long, return to base, even if not out of ammo
   End
   Locomotor = SET_NORMAL      AuroraJetLocomotor
   Locomotor = SET_SUPERSONIC  AuroraJetSupersonicLocomotor
@@ -1598,7 +1598,7 @@ Object AmericaJetStealthFighter
     TakeoffDistForMaxLift     = 0%   ; larger numbers give more lift sooner when taking off
     TakeoffPause           = 500
     MinHeight              = 5
-    ReturnToBaseIdleTime   = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime   = 0         ; if idle for this long, return to base, even if not out of ammo
   End
 
   Behavior = ExperienceScalarUpgrade ModuleTag_08
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -15637,7 +15637,7 @@ Object Boss_JetAurora
     AttackLocomotorPersistTime  = 100    ; we start slowing down almost immediately
     AttackersMissPersistTime    = 2000   ; but remain untargetable fer a bit longer
     ReturnForAmmoLocomotorType  = SET_SLUGGISH
-    ReturnToBaseIdleTime        = 10000         ; if idle for this long, return to base, even if not out of ammo
+    ReturnToBaseIdleTime        = 0         ; if idle for this long, return to base, even if not out of ammo
   End
   Locomotor = SET_NORMAL      AuroraJetLocomotor
   Locomotor = SET_SUPERSONIC  AuroraJetSupersonicLocomotor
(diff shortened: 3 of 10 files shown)


xezon:
### Summaries from chat

Exile:
> I completely agree on the arguments about meaningful APM, but yet I have doubts about getting rid of auto-return. For example, would it not become too easy to defend Chinooks in USA Airforce mirrors?

Rising:
> I do not like this change. There are situations where I do not want them to circle indefinitely. Currently I have the option to circle and auto-return, so it is ideal for me. Often I will move migs for example onto a flank so they are nearby to attack but then do not end up using them. If they do not auto-return, then I would have to remember and spend the time flying them back to somewhere safe. Also the side issue as Exile mentioned, King Raptors just become floating Avengers now, which would be super op.

Stubbjax:
> One ideally wants to maximize meaningful APM and eliminate meaningless APM. Sure it would be easier to have PDL planes guard the air space, but one would be paying a substantial opportunity cost to have them idling around Chinooks for long enough to make a difference. There are far, far more situations where I would want planes to circle indefinitely instead of auto land after 10 seconds.

MTKing:
> I think it will be the best feature we have implemented yet. No one is going to leave a plane idle forever to defend a Chinook, otherwise that is just a waste of a unit. Not to mention that it idles in a circle, meaning it will not have full coverage to always block the missiles. I would rather have planes idle to prepare them to attack somewhere instead of getting distracted for a second and come back to see them gone landing.

kABUSE1:
I suggest:
1. Have normal guard only attack ground
2. Have air-guard only attack air
3. Keep air-guard command button exclusive to planes that can actually attack air

This would:
- feel 100% logical, command button-wise, no hidden features
- prevent migs and stuff to constantly get distracted by all types of air units (mainly usa drop zone cargo planes).
- not allow abuse like king raptor PDL idle protection, as mentioned by Stubbjax.
- not buff alphas

If you want to consistently prevent your planes 100% from unintentionally auto-attacking, I find it fair that you have to consistently micro them around also.

It is important to note that alpha auroras do not have air-guard, so you can't have them idle in the air. I think it is their only and their most crucial weakness that needs to be kept untouched. You can either constantly micro them, leave them on the airfield to die from a sneezing rebel or have them guard ground, risking blowing shit up. It's fair, considering how insanely strong their damage output is.

ExiLe-ZH:
This sounds pretty good yea, however still isn't 100% reliable at preventing "lazy idle-protection", like Raptors can now Ground-Guard Chinooks in Air Mirrors for example. I guess scaffolds can be used to against this though.

xezon:
Perhaps increase plane idle landing timeout to 20 or 30 seconds?

kABUSE1:
That would further buff alpha auroras, which are already the most overpowered unit type in the game. Unless you want to re-balance alpha auroras, which most players seem to strongly disagree with.

MTKing4:
placing scaffolds to force enemy bombers to shoot at the ground should be considered a scaffold abuse tbh and we should work to eliminate these quirks instead of balancing the game ontop of them and implement an actual balance solution for alphas

having the planes to idle will only increase the quality of the games, where there are more meaningful actions and thoughtful planned orders instead of mindless APM babysitting units just to prevent them from putting you at a disadvantage

basically, you're microing not to lose instead of microing to win, this is a huge red flag in any RTS game where you mistake a mindless action as Skill

kABUSE1:
I think you underestimate just how much stronger alpha auroras become if you allow them to idle in the air, especially against GLA. They are no longer in danger of dying to rebel ambush, tox bombs or sneak attack placement.

How would this idle-feature even work in combination with the out-of-fuel mechanics?

How would you go about preventing scaffold abuse? Guard mode no longer attacking scaffolds? That would be annoying when you take out an enemy base. And rebel ambush and sneak attack would still make the alphas blow stuff up.

You could potentially add a "max build radius" around workers and dozers to place scaffolds in but that would be insanely annoying on large maps.

MTKing4:
none of the above, alpha auroras will no longer be as threatening after supersonic nerf as you can shoot them out of the sky

the bomber will still land after bombing (which is a window for ambush and sneak attack, and more a clear sign to the player to do them unlike blind ambushes and hope the bombers are landed) then it can refuel, and guard mode already goes against out-of-fuel mechanics by guarding forever

you prevent scaffold abuse by allowing the player to idle their planes without having to guard mode

xezon:
Let us assume for a moment that planes originally would not auto return to the airfield after 10 seconds when idle. They would only return after emptying their ammo or manually returning by user command. If I would then propose to add auto landing after 10 idle seconds, would that spark a strong agreement in player community? Would players then say _"Yea that is logical. I would like them to auto return after 10 seconds so that I need to babysit them for as long as I want them in the air."_ ?

In essence the plane auto return when idle for 10 seconds is mismatching with
1) Planes do not auto return when guarding for 10 seconds
2) Helicopters do not auto return when idling, guarding for 10 seconds
3) Ground units do not auto return when idling, guarding for 10 seconds

I do think that naturally a new player would expect a unit to stay where it was commanded to go, because the majority of units adhere to this behaviour. With planes this infinite idle behaviour is already possible by using guard mode, with the drawback of potentially attacking undesired targets. This in turn makes it a questionable design, as it limits the control over the plane. The player needs to babysit the plane, because the game controls do now give the option to command the plane to do passively what we would like it to do (for longer than 10 seconds).

I think this issue is comparable to the ECM Tank behaviour issue which we discussed and tackled. Some argued, that it is good design to be forced to pull the ECM Tank back and forth during combat so that its Missile Jammer would be active, instead of being passively active without babysitting. In the end we removed the babysitting requirement and made them behave consistently.

* #108
* #631

kABUSE1:
Can a plane idle without ammo? Is there even a way to prevent that? This decision impacts the out-of-fuel mechanics.

xezon:
> Can a plane idle without ammo? Is there even a way to prevent that? This decision impacts the out-of-fuel mechanics.

Planes will still auto land immediately when out of ammo as usual, as that works independently of the 10 seconds auto land feature.

ExiLe-ZH:
> Let us assume for a moment that planes originally would not auto return to the airfield after 10 seconds when idle. They would only return after emptying their ammo or manually returning by user command. If I would then propose to add auto landing after 10 idle seconds, would that spark a strong agreement in player community? Would players then say _"Yea that is logical. I would like them to auto return after 10 seconds so that I need to babysit them for as long as I want them in the air."_ ?

The most fair way to look at this is to assume nobody isn't used to anything yet, because as soon as that happens, it's always hard to go the other way and have people accept it.

There are pros and cons to both ways, but I do understand why the devs chose the current design at the time. It isn't implemented for no reason.

If we go with infinite idling, we need some replacement mechanics, because examples like _idle PDL defense_ takes all the skills out of Air mirrors. Currently it requires quite some management to be on point (offensively and defensively), which adds quite some fun admittedly. 

Manual PDL activation could be an idea, where you have to press a button/hotkey to activate PDL's for 10 seconds. The cooldown could be instant, but still requiring manual reactivation.

For Aurora's I have no idea honestly, infinite idling makes it much easier to keep them safe. Most fast harassment units can target ground only, like Techs or Bikes.

ExiLe-ZH:
If we plan on keeping the current mechanic we could make it more user-friendly and realistic by implementing fuel-reserves represented by a second HP bar in the UI. 

Aurora's could have lower fuel-reserves (because it needs to be light and fast) than Raptors for example.

We could generally make the time longer for all planes, but have it functional in all modes, idle, guard or air-guard.

MTKing4:
I think this whole alpha raptor thing being better is a bit more theoretical, we don't know if this really impact the balance so much, i think this can only be known with testing, so if you have no more issues with this design wise i suggest we merge it and see if it really impacts balance in actual games, if it does we can do something about it, otherwise this change is a no brainer to me.

xezon:
Plane fuel is a potentially interesting feature and something that can eventually be explored, but not at this time and not for our first release, because it is too big of a feature.

> but I do understand why the devs chose the current design at the time

Could you elaborate on this? It is actually not clear to me what purpose this feature serves other than making it easier for the player to restart his plane attacks from the airfield. That is because noobs tend to attack with planes from the airfield instead of from the air, because it requires less interactions and planning this way. So if they abort an attack by sending the planes to a random location, it will soon fly back to the Airfield by itself and the player can find his planes conveniently. If you pull a casual Red Alert 2 video on YouTube, you can see how they select planes from Airfield, click on enemy target, and let them fly over whole map. Therefore I suspect that the auto-landing was designed to accommodate this noob friendly play-style.

If you see other reasons I would like to read them.

> If we go with infinite idling, we need some replacement mechanics, because examples like idle PDL defense takes all the skills out of Air mirrors. Currently it requires quite some management to be on point (offensively and defensively), which adds quite some fun admittedly.

Yes. Makes PDL defense more convenient for sure. This underlines how good PDL is in practice, which we have tracked here
* #638

> For Aurora's I have no idea honestly, infinite idling makes it much easier to keep them safe. Most fast harassment units can target ground only, like Techs or Bikes.

In essence long passive idle can already be achieved in original game by using waypoints. With a programmable keyboard and high framerate you could use a macro that generates many waypoints very quickly and leaves the plane fly along these waypoints for a long time. It is just stupidly inaccessible because every player has to program it himself, but technically it is possible. A simple mouse click for infinite idle would be much better for player comfort.

kABUSE1:
> Let us assume for a moment that planes originally would not auto return to the airfield after 10 seconds when idle. They would only return after emptying their ammo or manually returning by user command. If I would then propose to add auto landing after 10 idle seconds, would that spark a strong agreement in player community?

I think that's a lazzy argument and could be used for everything. Imagine ambulances didn't heal vehicles and you'd change that. The outrage in the community would be huge, as humvees are already overpowered. Or imagine burton remote charges were currently silent but you added a beeping sound to them. People would call you insane over this.

I suggest instead we go by adding "air guard" to all stealth fighters and auroras and make it so that planes that are on "air guard" ONLY attack air. This way, we only need to find a way to balance out alpha auroras and it doesn't affect the other planes strength too much. It would also buff normal auroras, which are currently useless.

Finally, I think we are going back and forth here as long as we don't do some *thoroughly* competitive testing, as others have pointed out.

xezon:
> I think that's a lazzy argument

I disagree with this assessment. This method is great to assess whether or not a given feature (or bug) is reasonable or not. It takes away from the status quo bias we have.

> I suggest instead we go by adding "air guard" to all stealth fighters and auroras and make it so that planes that are on "air guard" ONLY attack air.

This does not appear like solid design to me, because Stealth Fighter and Aurora do not have air attack capabilities. Therefore the Air Guard button is misleading. Though perhaps it is possible to misappropriate the Air Guard functionality and disguise it under a new button art & text, but then why would some planes have it and some not? Adds inconsistent principles among planes which also does indicates bad design. Ideally all unit controls are streamlined - for ease of use.

ExiLe-ZH:
> Could you elaborate on this? It is actually not clear to me what purpose this feature serves other than making it easier for the player to restart his plane attacks from the airfield. That is because noobs tend to attack with planes from the airfield instead of from the air, because it requires less interactions and planning this way. So if they abort an attack by sending the planes to a random location, it will soon fly back to the Airfield by itself and the player can find his planes conveniently. If you pull a casual Red Alert 2 video on YouTube, you can see how they select planes from Airfield, click on enemy target, and let them fly over whole map. Therefore I suspect that the auto-landing was designed to accommodate this noob friendly play-style.
> 
> If you see other reasons I would like to read them.


This reason definitely sounds plausible, this case isn't as clear as some other cases so it's mostly speculations. To me it also sounds plausible that this mechanic was implemented to add a layer of management for the player to keep his planes safe and ready. Perhaps they thought Air/Ground-Guard was good enough for this utility and might not have realized the downsides (the unintentional attacks)? Though vice versa you could argue that the avoidance of unintentional attacks when using this mode also adds a layer of management.

> In essence long passive idle can already be achieved in original game by using waypoints. With a programmable keyboard and high framerate you could use a macro that generates many waypoints very quickly and leaves the plane fly along these waypoints for a long time. It is just stupidly inaccessible because every player has to program it himself, but technically it is possible. A simple mouse click for infinite idle would be much better for player comfort.

True, that's quite smart, but also kinda cheaty haha.

MTKing4:
Basically devs did intend to have units to idle indefinitely, but they failed at that job with these unintentional attacks, adding to the fact that idling forever can be achieved with pseudo hacks and tricks by everyone it makes total sense that we implement this feature officially to avoid all that mess

PR #1401, opened 2022-10-21
Add SteamVent effect variations for DAMAGED and REALLYDAMAGED conditions on Power Plants and War Factories

xezon:
This change adds color variations for SteamVent effects on USA, China Power Plant and USA Warfactory. On damaged states the big smoke becomes darker.

### TODO

- [ ] Complete USA Power Plant setup after its particle setup is fixed
- [ ] Add China Warfactory SteamVent effects when so available

### Rationale

A pure clean steam effect would have to be white. The SteamVent in its original state is grey. This means it has some dust particles in it. In the damaged states the dust particles increase because the damages cause fires in the building. It is logical that with more damage the exhaust becomes more dirty. It also sells the increasingly damaged look better.

### Image

255 to 128 to 64

shot_20221021_192715_1


commy2:
This is supposed to be steam, why would it turn this black? Grey looks okay, but only for E and because of the fire.

xezon:
We can make the color difference less apparent. My thinking here is that with damaged building, there is not just steam coming out, but also smoke with it, thus appears darkened. It sells the damaged look better.

xezon:
Here is less pronounced variation, 255 to 192 to 128

shot_20221022_000229_1

PR #1402, opened 2022-10-22
Fix: Streamline drift velocity of prominent smoke and steam particles

xezon:
Smoke and steam particle drift velocity is not consistent. I streamlined the prominent ones based on SteamVent velocity. It drift into similar direction of where clouds move.

Dozer smoke drifted way too much sideways.

Smoldering smoke drifted into opposite direction of Steam.

Some steams had no drift velocity.

There likely are still some inconsistent ones left over, but it is unclear to me how to find all relevant static ones.

shot_20221022_142251_2


xezon:
Reviewed. Fine tuned DozerSmokeLight, DozerSmokeHeavy, SteamMedium a bit more. Looks fine to me.

PR #1405, opened 2022-10-23
Fix: abbarrslab_d textures in abbarracks w3d

xezon:
This change fixes texture error(s) in abbarracks w3d. Python scripts included.

### ABBarracks_D, ABBarracks_DN

* Now uses ATBarrSlab_D.tga instead of ATBarrSlab_E.tga

### ABBarracks_DS, ABBarracks_DNS

* Now uses ATBarrSlab_DS.tga instead of ATBarrSlab_ES.tga


xezon:
Looks good.

PR #1406, opened 2022-10-23
Fix: atbarrslab textures

xezon:
This change fixes atbarrslab textures.

### atbarrslab_d

* Reduces amount of burns to look more distinct to atbarrslab_e

### atbarrslab_e

* Fixes errors on burns
* Reduces amount of burns near the flag pole to avoid high contrast edges

### atbarrslab_s

* Fixes all blue snow to look like regular snow
* Fixes missing snow on edges of some walls

### atbarrslab_ds

* Fixes snow drawing over burns, based on new atbarrslab_s, atbarrslab_d

### atbarrslab_es

* Fixes snow drawing over burns, based on new atbarrslab_s, atbarrslab_e

### Images

shot_20221023_152706_1

shot_20221023_152909_2


xezon:
I did one more pass on the look of snow. It now looks very similar to snow on Command Center. I used TXSnow05a.tga as a paint. I think this change can be concluded now.

PR #1407, opened 2022-10-23
Fix: atdoor textures

xezon:
This change fixes atdoor textures.

### atdoor_s

* Lowers height of snow on door a bit

### atdoor_ds

* Fixes door looking different than atdoor_s


xezon:
Not clear to me where this door is used. I inspected USA buildings but did not see it.

xezon:
Looks okay.

PR #1412, opened 2022-10-30
Change: Patriot Missile Battery standardisation

### LINKED ISSUE #897: USA Patriot Battery is rarely used

Stubbjax:
* Relevant to #1178
* Relevant to #1580
* Relevant to #1755

This change standardises a lot of inconsistencies between missile-based Patriot Batteries, and improves the standard Patriot Battery so that it shares the same price, range and shot delay as the EMP variant.

Other inconsistencies have also been addressed, such as the EMP assist weapon firing slower and both the assist and air weapons doing more or different damage. The scatter radius vs infantry and assist weapon idle reloads are now standardised as well.

The primary adjustments to the vanilla Patriot to consider are the -$100 price, +50 range, -0.15s shot delay, and +5 air/assist damage. And for the EMP variant; the -15 air damage, -10 assist damage and -0.15s assist shot delay. This set of changes should make standard Patriots a little more useful and bring both structures' outputs more in line with player expectations.

Before:

| | Standard Patriot | EMP Patriot
-- | -- | --
Build cost | 1000 | 900
Ground range | 225 | 275
Ground damage | 4 x 30 | 4 x 15
Ground reload | 4 x 0.25s | 4 x 0.1s
Air range | 350 | 400
Air damage | 4 x 25 | 4 x 30
Air reload | 4 x 0.25s | 4 x 0.1s
Assist range | 450 | 450
Assist damage | 4 x 25 | 4 x 25
Assist reload | 4 x 0.25s | 4 x 0.25s

After:

| | Standard Patriot | EMP Patriot
-- | -- | --
Build cost | 900 | 900
Ground range | 275 | 275
Ground damage | 4 x 30 | 4 x 15
Ground reload | 4 x 0.1s | 4 x 0.1s
Air range | 400 | 400
Air damage | 4 x 30 | 4 x 15
Air reload | 4 x 0.1s | 4 x 0.1s
Assist range | 450 | 450
Assist damage | 4 x 30 | 4 x 15
Assist reload | 4 x 0.1s | 4 x 0.1s

Closes #897.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -30625,7 +30625,7 @@ Object AmericaPatriotBattery
   Prerequisites
     Object = AmericaPowerPlant
   End
-  BuildCost        = 1000
+  BuildCost        = 900
   BuildTime        = 25.0           ; in seconds
   EnergyProduction = -3
   WeaponSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2052,7 +2052,6 @@ Weapon PatriotMissileAssistWeapon
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
-  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
@@ -6694,7 +6693,6 @@ Weapon Lazr_PatriotMissileAssistWeapon
   ClipSize                    = 3                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
-  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
@@ -7665,7 +7663,6 @@ Weapon SupW_PatriotMissileAssistWeapon
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
-  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1980,7 +1980,7 @@ Weapon PatriotMissileWeapon
   PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
-  AttackRange                 = 225.0
+  AttackRange                 = 275.0
   DamageType                  = EXPLOSION
   DeathType                   = EXPLODED
   WeaponSpeed                 = 1             ; locomotor specifies speed.
@@ -2009,7 +2009,7 @@ Weapon PatriotMissileWeaponAir
   PrimaryDamage               = 25.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
-  AttackRange                 = 350.0
+  AttackRange                 = 400.0
   DamageType                  = EXPLOSION
   DeathType                   = EXPLODED
   WeaponSpeed                 = 600
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2052,6 +2052,7 @@ Weapon PatriotMissileAssistWeapon
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
+  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
@@ -6693,6 +6694,7 @@ Weapon Lazr_PatriotMissileAssistWeapon
   ClipSize                    = 3                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
+  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
@@ -7663,6 +7665,7 @@ Weapon SupW_PatriotMissileAssistWeapon
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
+  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
   AntiAirborneVehicle         = Yes
   AntiAirborneInfantry        = Yes
   AntiGround                  = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1979,7 +1979,7 @@ End
 Weapon PatriotMissileWeapon
   PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 225.0
   DamageType                  = EXPLOSION
   DeathType                   = EXPLODED
@@ -2008,7 +2008,7 @@ End
 Weapon PatriotMissileWeaponAir
   PrimaryDamage               = 25.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 350.0
   DamageType                  = EXPLOSION
   DeathType                   = EXPLODED
@@ -2038,7 +2038,7 @@ End
 Weapon PatriotMissileAssistWeapon
   PrimaryDamage               = 25.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
   DamageType                  = EXPLOSION          ; ignored for projectile weapons
   DeathType                   = EXPLODED
@@ -7649,7 +7649,7 @@ End
 Weapon SupW_PatriotMissileAssistWeapon
   PrimaryDamage               = 25.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
   DamageType                  = EXPLOSION          ; ignored for projectile weapons
   DeathType                   = EXPLODED
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2006,7 +2006,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon PatriotMissileWeaponAir
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 400.0
@@ -2036,7 +2036,7 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 Weapon PatriotMissileAssistWeapon
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
@@ -7613,7 +7613,7 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon SupW_PatriotMissileWeaponAir
-  PrimaryDamage               = 30.0
+  PrimaryDamage               = 15.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 400.0
@@ -7647,7 +7647,7 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 Weapon SupW_PatriotMissileAssistWeapon
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 15.0
   PrimaryDamageRadius         = 5.0
   ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1989,7 +1989,7 @@ Weapon PatriotMissileWeapon
   FireSound                   = PatriotBatteryWeapon
   FireFX                      = FX_BuggyMissileIgnition
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 2000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
@@ -2018,7 +2018,7 @@ Weapon PatriotMissileWeaponAir
   FireFX                      = FX_BuggyMissileIgnition
   FireSound                   = PatriotBatteryWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 2000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
@@ -2048,7 +2048,7 @@ Weapon PatriotMissileAssistWeapon
   FireFX                      = FX_BuggyMissileIgnition
   FireSound                   = PatriotBatteryWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
@@ -7659,7 +7659,7 @@ Weapon SupW_PatriotMissileAssistWeapon
   FireFX                      = FX_BuggyMissileIgnition
   FireSound                   = PatriotBatteryWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -16525,7 +16525,7 @@ Object AirF_AmericaPatriotBattery
   Prerequisites
     Object = AirF_AmericaPowerPlant
   End
-  BuildCost        = 1000
+  BuildCost        = 900
   BuildTime        = 25.0           ; in seconds
   EnergyProduction = -3
   WeaponSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -12202,7 +12202,7 @@ Object Boss_PatriotBattery
   Prerequisites
     Object = Boss_Barracks
   End
-  BuildCost        = 1000
+  BuildCost        = 900
   BuildTime        = 25.0           ; in seconds
   EnergyProduction = -3
   WeaponSet


MTKing4:
looks good to me

xezon:
Could you please also compare damages, reload and ranges with USA Firebase, to have a better overview how these compare now?

Stubbjax:
> Could you please also compare damages, reload and ranges with USA Firebase, to have a better overview how these compare now?

Comparisons can be found [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/897#issuecomment-1225351340).

ExiLe-ZH:
This is quite lovely :)

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -7659,10 +7661,11 @@ Weapon SupW_PatriotMissileAssistWeapon
   FireFX                      = FX_BuggyMissileIgnition
   FireSound                   = PatriotBatteryWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 1000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
+  AutoReloadWhenIdle          = 2100    ; If I haven't fired in this long, I will reload on my own (rather than only after the last one is fired)
```
xezon: Looks like this was forgotten to be added with
* #1178

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -2018,7 +2018,7 @@ Weapon PatriotMissileWeaponAir
   FireFX                      = FX_BuggyMissileIgnition
   FireSound                   = PatriotBatteryWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
-  DelayBetweenShots           = 250                   ; time between shots, msec
+  DelayBetweenShots           = 100                   ; time between shots, msec
```
xezon, line 2021: Delay will be 133 now instead of 266. Maybe we can take it to 200, between the original Patriot and EMP Patriot, or is 133 de facto the best looker?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -2006,10 +2006,10 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon PatriotMissileWeaponAir
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
-  AttackRange                 = 350.0
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  AttackRange                 = 400.0
```
xezon, line 2012: How does this new AttackRange 400 compare with China Gattling Cannon and GLA Stinger?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -2006,10 +2006,10 @@ End
 
 ;------------------------------------------------------------------------------
 Weapon PatriotMissileWeaponAir
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 30.0
```
xezon, line 2009: It is not uncommon for Air weapons to deal less damage than the Ground weapon. Raising this can have undesired consequences. How does this compare with China Gattling Cannon and GLA Stinger(s)?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -2036,9 +2036,9 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 Weapon PatriotMissileAssistWeapon
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 450.0 ; at least Regular's range + regular's request assist range
```
xezon, line 2042: Judging by comment, assist AttackRange needs to be changed too if base AttackRange is changed?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -1979,8 +1979,8 @@ End
 Weapon PatriotMissileWeapon
   PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
-  AttackRange                 = 225.0
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
+  AttackRange                 = 275.0
```
xezon, line 1983: This looks like an outrageous ground AttackRange for base defense. Can you compare this with Stinger, Tunnel, Gattling Cannon, Tank Hunter in Bunker? I suspect this outranges them all by a large margin. Firebase also has no more range advantage now.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -1979,8 +1979,8 @@ End
 Weapon PatriotMissileWeapon
   PrimaryDamage               = 30.0
   PrimaryDamageRadius         = 5.0
-  ScatterRadiusVsInfantry     = 10.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
-  AttackRange                 = 225.0
+  ScatterRadiusVsInfantry     = 5.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
```
xezon, line 1982: Can you describe how this affects targeting infantry? At what distances will it hit walking undamaged and damaged infantry now vs before? Can a GLA Terror still run to Turret and hit it? To me this looks like it needs some testing and comparison with original to draw the right conclusions.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: [That was noted.](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1178#issuecomment-1242660040)

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: The idea is to match it with the EMP Patriot's fire rate to achieve consistency between the two.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: China Gattling Cannon and GLA Stinger Site are both 400.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: I just matched it to the EMP Patriot's values. Perhaps it needs looking at as 450 is less than the regular range (275) + request assist range (200).

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: It now matches EMP Patriots which players are already very familiar with.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: These value aren't new. This has essentially been tested countless times with EMP Patriots, which are some of the most frequently bombed / TCed base defences.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: All missile weapons, such as Stinger Sites, TOW Missiles, infantry missiles, aircraft missiles, etc. deal either the same or additional damage to aircraft. (And perhaps more bizarrely, Gattling Cannon and Gattling Tank air weapons deal `SMALL_ARMS` damage to aircraft, so it's not a direct comparison.)

In reality, Patriot Missile Systems are solely anti-air defences, so it is reasonable to expect that they are effective against aircraft. In addition, USA has fairly weak anti-air defences, and it makes sense to give the criminally underused Patriot some additional utility. It could help vs Combat Chinooks, stealth Comanches or Helixes, which can only be a good thing.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok. Wonder why commy did not act on it.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: I understand. But this now is a significant buff for Patriot fire rate. Consistency would also have been achieved by raising EMP Patriot delay from 100 to 250. Why was 100 chosen as the to-go option?

There are 3 options:

100 (equals 133)
200
250 (equals 266)

Stubbjax: The reason I preferred 100 over 250 is due to the ability for EMP Patriots to take down King Raptors under the right conditions. Increasing the delay significantly reduces the Patriot's ability to break through PDLs.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok from that comparison it looks reasonable then. However, does Search and Destroy add 20% range to defenses? If so, that would grow to 480 now, and 420 before. +80 more range than other faction defenses would be significant surplus.

Stubbjax: I don't believe buildings are affected.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok. Could we list missile damage types and values for Stinger for comparison?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax:
```
;Army bonuses granted by different battle plans
ValidMemberKindOf = INFANTRY CAN_ATTACK VEHICLE ;Battle plans affect any kind of these...
```

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: I understand it matches EMP Patriot, but is the gained consistency automatically justified? How does it compare with the other base defenses now? Does USA need such powerful bunker capabilities?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: `StingerMissileWeaponAir` deals 30 `EXPLOSION` damage to aircraft.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok this was not clear. In the first post the changed Scatter values are not present in the value table. Perhaps it should be appended for visibility.

Stubbjax: I did mention that was the case in the description. I didn't think it was worth repeating the value for every weapon as it only applies to the ground weapon and I standardised it for all of them.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok. I assume defenses also do not have CAN_ATTACK enum?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok. So damage output of one clip with Stinger and Patriot is identical then, no?

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Ok that is a good reason.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: I am still skeptical of range buff here. Patriot Weapon Assist will make ranges even better, as far back Patriots can assist forward Patriot. No other base defense can do this.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: Patriot Batteries fire 4 missiles per volley whereas Stinger Sites fire 3 missiles per volley (though the Stinger rockets are all fired at once and accelerate faster, resulting in greater PDL penetration). However, it must be noted that the delay between shots adds to the overall reload time, so the reload times would look something like this:

Stinger Site = 3 shots per 2000ms = 45 DPS

Patriot Battery (patched) = 4 shots per (2000ms + (4 × 100ms)) = 2400ms = 50 DPS (assuming the 4 × 100ms adds up to 2400ms and not 4 × ~133ms to ~2533ms)

Keep in mind that GLA gets the AP Rockets upgrade, which increases the Stinger Site's DPS even further.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: Base defences do have `CAN_ATTACK`, but I forgot to include the additional line below that one:
```
InvalidMemberKindOf = DOZER STRUCTURE AIRCRAFT DRONE
```
The 275 range is already observable with EMP Patriots in 1.04, so there's really no new behaviour to consider here.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Stinger shoots just 3 in the Air? There are 4 soldiers on the site, so I always thought all 4 shoot, no? Or do you mean if one is dead?

Stubbjax: There are three soldiers in the nest!

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: Well the issue is that vanilla Patriots are almost never used because the Firebase is almost always a better option, which lowers strategic diversity. The intent with this change is to also reduce the penalty / disadvantage incurred from using them without making them good enough to become the primary strategy. Doing so by standardising the values with the EMP Patriot seems like an acceptable way to do this, especially because it is effectively already well and truly tested. I would still expect them to be rarely used after this change, as players likely already consider the standard Patriot to be the same as the EMP one without the disabling effect (at least I did for a long time).

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Silly me. No idea why I though there are 4 inside.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
commy2: It makes assist bug worse, because now it likely fires the whole salvo instead of just what was left in the clip.

xezon: Stubbjax, did you test this? Did it work ok?

Stubbjax: From my testing, it seems that once the Patriot's assist weapon is activated, it will remain active until it has emptied its clip. This means that if the assist weapon fires only half of its clip, the Patriot will then fire the rest of its clip as soon as an enemy enters its assist range. This is likely why we sometimes see Patriots fire missiles incredibly far distances seemingly randomly. This behaviour is demonstrated in the footage below.

https://user-images.githubusercontent.com/11547761/199151654-fa78368c-776b-46d5-8725-9596e428d1fa.mp4

https://user-images.githubusercontent.com/11547761/199152517-3add7c97-5127-4394-9b9e-8aae3fb34d79.mp4

I was hoping that if the Patriot was allowed to reload its assist clip when idle, that it would then deactivate its assist weapon and return to its standard weapon like it does after a full volley. However, after testing with `AudoReloadWhenIdle` active, it appears that this is unfortunately not the case, and the assist weapon will still remain active despite the clip reloading when idle. The respective change will thus exacerbate the issue as @commy2 pointed out, as the entire assist clip may be fired after idling. This is evidently undesirable, so I will undo this change.

However, the good news is that the shorter assist weapon's delay between shots will slightly alleviate the issue, as the window during which the assist weapon can be interrupted is shortened.

xezon: Thanks for testing.

commy2:
Remove assist weapon.

ReLaX82:
Why change assist Reload time if it was equal before?

xezon:
> Why change assist Reload time if it was equal before?

So that there is no firing behaviour discrepancy between the various Patriot weapons. There are 6 in total, 3 for Patriot and 3 for EMP-Patriot.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -7647,9 +7647,9 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 Weapon SupW_PatriotMissileAssistWeapon
-  PrimaryDamage               = 25.0
+  PrimaryDamage               = 15.0
```
xezon, line 7650: Note that this is a hidden nerf for USA Super Weapon. With this reduction, 2 EMP Patriots will take longer to kill disabled vehicles than originally. Although this is a very logical approach, it is difficult to agree with this implementation, considering that USA Super Weapon General has a hard time against other factions. May I suggest to explore streamlining damage to a higher value for all EMP weapons? This way damage evens out a bit more. To illustrate problem, I will provide table:

| EMP Count | Damage Original | Damage Patched |
|-----------|-----------------|----------------|
| 1         | 15              | 15             |
| 2         | 50              | 30             |
| 3         | 75              | 45             |
| 4         | 100             | 60             |

As you can see the more Patriots assist, the less damage it deals.

The reduction of DelayBetweenShots will counter act this a bit, but not nullify the damage reduction.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Here are the ground ranges:

| Object                       | AttackRange | Area Covered |
|------------------------------|-------------|--------------|
| Original GLA Tunnel          | 175         | 96211        |
| Original GLA Stinger         | 225         | 159043       |
| Original China Gattling      | 225         | 159043       |
| Original China Rocket Bunker | 232         | 169093       |
| Original USA Firebase        | 275         | 237582       |
| Original USA Patriot         | 225         | 159043       |
| Patched USA Patriot (this)   | 275         | 237582       |

Image:

shot_20221103_182854_1

This change gives USA considerable defensive advantage over all other factions. Patriot also benefits from assisted Patriot group attacks. They will all fire earlier and kill quicker. This buff benefits USA Air Force, which is one of the best factions already.

Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
```diff
@@ -30625,7 +30625,7 @@ Object AmericaPatriotBattery
   Prerequisites
     Object = AmericaPowerPlant
   End
-  BuildCost        = 1000
+  BuildCost        = 900
```
xezon, line 30628: | Object                           | Price|
|----------------------------------|-------------|
| Original GLA Tunnel              | 800         |
| Original GLA Stinger             | 900         |
| Original China Gattling          | 1200        |
| Original China Bunker + 2 Rocket | 1100        |
| Original USA Firebase            | 1000        |
| Original USA Patriot             | 1000        |
| Patched USA Patriot (this)       | 900         |

USA Defense cost sits between China and GLA. Looks ok, but again will benefit USA Air Force here.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Visually looks proper. Logically it makes sense that it helps breaking PDL. This will also make it more effective against Avengers, not just against Air Force things.

This change on its own gives a DPS buff of 17%, because ClipReloadTime is not increased to counter balance it.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Together with reduced ClipReloadTime the Air and Assist Weapon yield +40% DPS over original. 

Comparison with Stinger:
 
| Object                          | Ground DPS | Air DPS | Assist DPS |
|---------------------------------|------------|---------|------------|
| Original GLA Stinger Site       | 30         | 45      |            |
| Original GLA Stinger Site Beta  | 30         | 45      |            |
| Original GLA Stinger Site Gamma | 45         | 60      |            |
| Original USA Patriot            | 43         | 36      | 36         |
| Patched USA Patriot (this)      | 50         | 50      | 50         |

I think this should be compared with all defenses to give more clarity.

This is significant buff. It is not clear to me whether or not it is desired. It is bad for China. Helix and Mig will die significantly quicker against this. The increased Anti Air range makes the effectiveness even greater.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: Changed ScatterRadiusVsInfantry does not appear to make much difference. Tested with GLA Terror running towards Patriots. Attack Outcome is very comparable.

What makes more of a difference right now is that in Patch, the Terror explodes and deals damage when killed by Patriot. But not in Original. I am not sure if this was planned. I will look into that.

xezon (changes requested):
This change does not look good in its current state. EMP Patriot is weaker than before and Regular Patriot is much better than the Original. This is a concern, because USA is already a competitive faction and USA Air Force is one of the best Generals. This change is bad news for China, who now have to deal with up to +40% stronger Patriots at significantly longer distances.

Additionally, the change description does not explain all the consequences this change entails. It does not list the DPS changes, and how they compare with DPS of other various Base Defenses. It also does not compare Attack Ranges against Ground and Air. This would be important to see all the impact this change has.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: I don't agree that this change is a problem nor that it plays even a minor role in Super Weapon General's prospects against other factions. It does not make sense to focus a faction's balance on such an obscure and rare mechanic, particularly one which is practically never seen in competitive games.

Moreover, giving EMP Patriots a damage boost would go against the faction and building's design philosophy and is unnecessary considering it's already one of the best defences in the game. Players are also very familiar with the current 4 × 15 damage output.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: What this change really does is provide more defensive options, particularly in terms of dealing with aircraft. It's still overall more beneficial to use Firebases due to the Patriot's power dependency, which is quite high for the respective factions, who would need an additional power plant or control rods upgrade to cover the consumption (covered [here](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/897#issuecomment-1225503815)). There isn't  really much advantage to be gained here; it's more of a disadvantage reduction.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
xezon: All match types are part of competitive player scene, including 1v1v1v1v1v1. Camping with EMP Patriot is common with Super Weapon General. Bundling EMP Patriots is effective, as long as Power levels can be maintained. It can be difficult to break through these defenses. If we take away from EMP Patriot capabilities, it will inevitable weaken USA Super Weapon in some capacity.

What is/are the goal(s) of this change?

I am asking because if the goals were to
* fix random discrepancies
* make Patriot more effective

then making EMP Patriot less effective would not fit into the goals.

xezon:
Ok, so what is the plan with this now? This change changes more than its advertised "standardisation": Patriot is strengthened by a lot, EMP Patriot is weakened a bit. There is no precise explanation of how these levels of strength are justified and logical.

If this cannot be addressed here, may I suggest to close this change and break it up in smaller chunks to go step by step? Perhaps it is too much to change all at once.

Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
Stubbjax: Games such as 1v1v1v1v1v1 are casual, where politics and luck are far more influential on the outcome of a game than a particular faction / matchup. I would also argue that any player who places enough EMP Patriots in a single area for this change to make any meaningful difference is not playing at a competitive level.

It's important to note that the EMP Patriot's efficacy lies in its disabling effect, not its damage. A delay of 2400ms (down from 3000ms) between volleys could be argued as an overall slight improvement to the assist weapon, despite the damage reduction, as more units will potentially be disabled. The damage reduction is also meaningless against aircraft, in which case the assist weapon is now more effective with its faster reload. Also don't discount the increased PDL penetration, which could make the difference between a group of Avengers or King Raptors remaining invulnerable or becoming entirely disabled / destroyed. This would directly affect USA Air Force, SWG's worst matchup.

Regardless, such a change is highly unlikely to have any impact in 99% of games.

The goals of this change are to:

- Reduce the disadvantage incurred from employing standard Patriot Batteries without making them more attractive than the current meta defence, the Firebase
- Streamline any inconsistencies between each Patriot's standard, air and assist weapons
- Streamline and standardise values and inconsistencies between the two Patriots while maintaining primary distinctions
- All of the above in the cleanest, simplest and subtlest way possible

Stubbjax:
> Ok, so what is the plan with this now?

You requested changes but provided no actionable items.

> This change changes more than its advertised "standardisation": Patriot is strengthened by a lot, EMP Patriot is weakened a bit. There is no precise explanation of how these levels of strength are justified and logical.

It was advertised as increasing the standard Patriot Battery's utility by standardising many of its attributes with the EMP Patriot. The EMP variant is not weakened in any significant or meaningful way, and could even be considered a minor improvement. Regardless, the impact is low. This is a direct response to #897 and much of the feedback from #965.

> If this cannot be addressed here, may I suggest to close this change and break it up in smaller chunks to go step by step? Perhaps it is too much to change all at once.

I would suggest merging it and acquiring feedback, as you have done and suggested with previous changes. It seems others are happy with this change as well. I expect standard Patriot Batteries to remain underused.

xezon:
I am skeptical about these points:

* Increased attack ranges
* Increased Patriot damages
* Decreased EMP Patriot damages

None of these buffs/nerfs are explained in the Rationale. It just mentions what is changed, but not how it is justified.>

Note that any other user looking at this change will ask the same question: Why is this changed to what it has changed? And we will not be able to give an accurate answer. I know this because I do not understand this change myself.

RisingZH:
I don’t like nerfing emp patts. Swg is one of the worst armies and emp pats are one of their few unique advantages.

Stubbjax:
> None of these buffs/nerfs are explained in the Rationale. It just mentions what is changed, but not how it is justified.>

The justification is that players are already used to the respective price / ranges / fire rates / attributes that the EMP Patriot uses, and naturally expect the same of the standard Patriot while maintaining the key factional differences, so it makes sense to use the same values for the standard Patriot.

> Note that any other user looking at this change will ask the same question: Why is this changed to what it has changed? And we will not be able to give an accurate answer. I know this because I do not understand this change myself.

I thought I made it clear that the standard Patriot's values are taken from the EMP Patriot because those values improve the structure's utility and are what players are already generally familiar with. There is not really much more to it. Is this justification not enough? I guess not if there is confusion. I will take a look at what information can be adjusted or added.

> I don’t like nerfing emp patts. Swg is one of the worst armies and emp pats are one of their few unique advantages.

Neither do I. This is a very carefully considered change, after which they could be technically considered better. Groups of them would be stronger against Air Force, and be able to disable more units.

ExiLe-ZH:
So the assisting EMP now basically fires faster, but with less damage? 

How does that work out vs incoming Scud missiles for example? Will they be stopped more or less reliably? 

Vs ground I can imagine this is a buff, because disabling stuff is what makes the EMP effective.

Stubbjax:
> So the assisting EMP now basically fires faster, but with less damage?
> 
> How does that work out vs incoming Scud missiles for example? Will they be stopped more or less reliably?

Probably at about the same rate. How many volleys a Patriot gets at a ballistic missiles is heavily dependent on the reload time. While a single EMP Patriot will get one and maybe two volleys off, an assisting EMP Patriot in 1.04 is quite unlikely to get two volleys off due to the slower fire rate / reload. Two EMP Patriots will stop a single SCUD missile before and after the change. Any more than one SCUD missile and the values are largely meaningless anyway.

MTKing4:
> I am skeptical about these points:
> 
> * Increased attack ranges
> * Increased Patriot damages
> * Decreased EMP Patriot damages
> 
> None of these buffs/nerfs are explained in the Rationale. It just mentions what is changed, but not how it is justified.>
> 
> Note that any other user looking at this change will ask the same question: Why is this changed to what it has changed? And we will not be able to give an accurate answer. I know this because I do not understand this change myself.

i'm not sure about this either, why not just standardize them to the same values and then worry about balance later? i think it needs further studying + testing, we don't know, we might buff SWG elsewhere, EMPs are already too good or too bad situationally and better not mess with that just yet

ExiLe-ZH:
what is the laser pat range?

I think it makes sense that all patriots have better range than other faction defenses, usa is technologically better. However it is strange that the firebase has no range advantage.

xezon:
```
Weapon Lazr_PatriotMissileWeapon
  AttackRange = 225.0
  
Weapon Lazr_PatriotMissileWeaponAir
  AttackRange = 350.0
```

xezon:
> I think it makes sense that all patriots have better range than other faction defenses, usa is technologically better. However it is strange that the firebase has no range advantage.

I do not understand the desire to make changes for the sake of faction diversity. Major differences will make it much more difficult to balance strengths as well. Weapon range is one of the most powerful weapon attributes, and changing it cannot be taken lightly.

ExiLe-ZH:
Hmmm, I'm not too sure now about the extra vPat range. Yes, you would expect the vPat to have the same range as the EMP, both are missile-based. But the sheer range advantage over Laserpats doesn't make too much sense, neither to have the same range as Firebases. We shouldn't touch Laserpats, EMP's or Firebases though.

Hard case tbh.

ExiLe-ZH:
> I do not understand the desire to make changes for the sake of faction diversity. Major differences will make it much more difficult to balance strengths as well. Weapon range is one of the most powerful weapon attributes, and changing it cannot be taken lightly.

And I don't get the desire to streamline everything between factions.

There shouldn't be a desire to do any of these extremes, but it should be clear that everything in this game is asymmetrically based on technology level, which is the primary faction differentiator. For example USA usually has the most firepower, range and precision, but is more expensive and often relies on power.

The only desire should be to balance things out with thematically sensible pros and cons to make the unit/building/upgrade fit a specific role.

In this case 275 feels a bit too extreme though, if it was up to me in 2003 I would never have given the EMP 275 in the first place, it doesn't make much sense compared to the Firebase. We shouldn't nerf the EMP though, but what about a middleroad for the VPat then? 
- Laserpat: 225, least range but most damage output. Strong vs everything.
- Vpat: 250, medium range and damage output. Strong vs light vehicles and aircraft, ok vs infantry and heavy armor.
- EMP: 275, most range but low damage output. Effectively disables all vehicles, bad vs infantry.

commy2:
The only thing that should change is EMP assist reload 0.25 -> 0.1, because it's a bug.

As far as I am concerned though, the entire assist feature should be scrapped until it is properly implemented in Thyme.

Stubbjax:
> why not just standardize them to the same values and then worry about balance later? i think it needs further studying + testing, we don't know, we might buff SWG elsewhere, EMPs are already too good or too bad situationally and better not mess with that just yet

Agreed that it should be tested so that it can be seen just how trivial of a difference the adjustment to the EMP Patriot's assist weapon is.

> I do not understand the desire to make changes for the sake of faction diversity. Major differences will make it much more difficult to balance strengths as well. Weapon range is one of the most powerful weapon attributes, and changing it cannot be taken lightly.

There are no changes in this PR that are 'for the sake of faction diversity'. On the contrary - this is more in line with faction homogeneity, as it standardises counterintuitive differences. Nothing is being taken lightly.

Why should EMP Patriots have greater range than standard Patriots? Because SWG's Humvees are expensive? (Ludicrous!) The EMP Patriot intuitively fires a different missile with a reduced damage yield in exchange for a crippling EMP effect. This is a concisely logical tradeoff and interesting design paradigm. The fact that the standard Patriot has a lower attack range and fire rate is counterintuitive to this paradigm, and pollutes the tradeoff. Standardising the attributes in such a way, conversely to your point, actually makes it _easier_ to balance and compartmentalise the factions as both designers and players have fewer (hidden) attributes to keep track of, which leads to a more streamlined design and experience.

> And I don't get the desire to streamline everything between factions.
> 
> There shouldn't be a desire to do any of these extremes, but it should be clear that everything in this game is asymmetrically based on technology level, which is the primary faction differentiator. For example USA usually has the most firepower, range and precision, but is more expensive and often relies on power.

> The only desire should be to balance things out with thematically sensible pros and cons to make the unit/building/upgrade fit a specific role.

I very much agree with this. Asymmetrical distinctions between factions should be intentional and logical, and are a core tenet of what makes the game interesting. None of the proposed changes negatively affect this paradigm. Regardless, I still disagree that any of the changes are extreme, nor that they will result in the standard Patriot becoming a more attractive option over the Firebase - primarily due to the high power requirement and cost, by extension. The +50 range is the difference of two Humvee lengths, which I would argue players are unlikely to notice due to the structure's overwhelming rarity. The difference is demonstrated below.

https://user-images.githubusercontent.com/11547761/200225170-423da986-c2bc-4a49-a435-17f0d67f03f0.mp4

> In this case 275 feels a bit too extreme though, if it was up to me in 2003 I would never have given the EMP 275 in the first place, it doesn't make much sense compared to the Firebase.

> We shouldn't nerf the EMP though, but what about a middleroad for the VPat then?
> 
>     * Laserpat: 225, least range but most damage output. Strong vs everything.
> 
>     * Vpat: 250, medium range and damage output. Strong vs light vehicles and aircraft, ok vs infantry and heavy armor.
> 
>     * EMP: 275, most range but low damage output. Effectively disables all vehicles, bad vs infantry.

Agreed, the EMP Patriot should never have been given 275 in the first place. It actually makes the least sense that the EMP Patriot has a greater range than the standard Patriot, as the disabling effect is far more punishing, and is uncharacteristic of ensnaring-type behaviour. It would make more sense if the ranges were flipped, and the most sense if they were equal (this change).

Regarding the Laser Turret; I think the fact that it is an entirely different (experimental) weapon provides more than enough context and leeway in delineating its distinction with the other Patriot Batteries. (It isn't even technically considered a Patriot Battery.) Many of the Laser Turret's properties are different, such as its immediate damage output, PDL / countermeasure immunity, effectiveness against infantry, and relatively lower power requirement. It is reasonable for the respective benefits to have alternative tradeoffs to counterbalance them, with range being one of them. It also makes logical and thematic sense that a laser cannot reliably travel as far as a missile.

Anyway, I can split this PR into two separate ones if the combination of changes is too much, but I really think it needs to be tested and contemplated as a combined change.

MTKing4:
I think the standardization should be a branch, and the nerf/buff a different branch, so that we have the freedom to choose one or both of the changes in the future

I also think we won't be able to see a clear result of the standardization if it was combined with the nerf/buff, that test is important, maybe if it goes well we don't need that nerf/buff

ExiLe-ZH:
Oh I always thought that the firebase range advantage was quite a bit more than demonstrated in that video. Honestly I never even knew that the EMP had the same range either.

I don't expect it to be OP then, just a more viable alternative to the Firebase. It's still power hungry so many players would still prefer Firebases most likely.

Let's test your current proposals.

Stubbjax:
> I think the standardization should be a branch, and the nerf/buff a different branch, so that we have the freedom to choose one or both of the changes in the future
> 
> I also think we won't be able to see a clear result of the standardization if it was combined with the nerf/buff, that test is important, maybe if it goes well we don't need that nerf/buff

But the buffs and standardisation kinda go hand in hand. I was thinking the standardisation between the EMP Patriot's primary, air and assist weapons could be one change, and the standardisation between the EMP Patriot and standard Patriot would be the other. But this still feels like overkill.

> Oh I always thought that the firebase range advantage was quite a bit more than demonstrated in that video. Honestly I never even knew that the EMP had the same range either.

Yes! This notion is a good example of how I imagine such a change would be perceived by most players - unknowingly!

Anyway, to address the "EMP is weakened" and "don't nerf EMP" concerns, below is a demonstration of the difference between 1.04's 250ms delay between shots and the proposed 100ms delay between shots for the EMP Patriot's assist weapon, which highlights several ways in which the change can technically be considered a buff. This will, of course, have little to no impact on your average game, and practically none in competitive games. The assist weapons are incredibly buggy in 1.04, which has a far greater impact on the game than a minor shot delay reduction. As @commy2 has repeatedly noted, the assist weapons / behaviour would be better off removed, at least until it can be properly fixed.

## Test 1 - 250ms vs 3 Avengers

1.04's values result in the EMP Patriots never breaking through the PDLs of the three Avengers. If they move a bit closer so that all three Patriots switch to their primary, faster-firing weapons, the Avengers will be in trouble.

https://user-images.githubusercontent.com/11547761/200353830-1cf9980d-05bd-4f68-a21e-5ef16a604d92.mp4

## Test 2 - 100ms vs 3 Avengers

The faster delay between shots of the assist weapons allows the EMP Patriots to break through the PDLs of the three Avengers with coordinated volleys.

https://user-images.githubusercontent.com/11547761/200354078-2a5934df-4b3b-4887-9a52-a4668bb803c2.mp4

## Test 3 - 100ms vs 4 Avengers

The faster-firing EMP Patriots have great difficulty trying to take down four Avengers. There is some elegance to this change as players can now more effectively determine how many Avengers they need to successfully block a Patriot barrage. If a player has fewer Avengers than there are Patriots, then they are in trouble, whereas if they have an equal amount of more, then they can attempt a push.

https://user-images.githubusercontent.com/11547761/200355649-242da3b6-2fd9-478f-997b-441067377572.mp4

## Test 4 - 250ms vs 2 King Raptors

The EMP Patriots could never break through the PDLs of the two King Raptors with 1.04's values. (No countermeasures.)

https://user-images.githubusercontent.com/11547761/200352509-0cbcfb95-b41b-49a2-bcf0-5e1dac5907c1.mp4

## Test 5 - 100ms vs 2 King Raptors

The faster delay between shots of the assist weapons allows the EMP Patriots to break through the PDLs of the two King Raptors if they get a good volley / coordination. It is a _very_ close call, but only one missile needs to get through and it's all over. (No countermeasures.)

https://user-images.githubusercontent.com/11547761/200352649-e083602f-b6c6-4688-8309-bba2299b62a4.mp4

The reduced delay between shots of the EMP Patriot's assist weapon demonstrates that not only can groups of EMP Patriots potentially disable larger armies / more targets, but they are more effective at taking down King Raptors and other PDL aircraft, thus increasing USA Super Weapon's prospects not only in general, but distinctly against USA Air Force.

commy2:
> Test 4

It's so dumb how assist mode makes the EMP Patriot shoot slower.

MTKing4:
> But the buffs and standardisation kinda go hand in hand. I was thinking the standardisation between the EMP Patriot's primary, air and assist weapons could be one change, and the standardisation between the EMP Patriot and standard Patriot would be the other. But this still feels like overkill.

Let me rephrase, i meant we have a branch that has all the changes combined (like this one) and other branch with only the streamlined values

xezon:
* Relevant to buff #1178
* Relevant to nerf #1580
* Relevant to buff #1755

Gdakgdak22 (approved):

Float1ngFree:
The worst thing about `PatriotMissileAssistWeapon` bug is that even if you set `AntiGround = No` and restrict it to be called only by `PatriotMissileWeaponAir` -- it still hits ground targets if any is in range before the assist clip is out.

Just another feature that was good on paper but underdeveloped and buggy. I'd disable it completely as it brings more imbalance than actual use.

PR #1414, opened 2022-10-31
Change: Remove Supply Drop Zone select sound from TechRepairPad, TechRepairBay, TechRadioStation

xezon:
This this change, all tech buildings except TechReinforcementPad have no select sound.


commy2:
This means they will just be silent when selected? How is that an improvement?

xezon:
Yes they will just be silent, like Artillery Platform, Oil Derrick and Oil Refinery. Do you prefer to keep the Drop Zone sound?

xezon:
Team wants to add sounds, not remove sounds
* #1448

PR #1417, opened 2022-11-02
Fix: Dragon Tanks can now force fire close Allies, Neutral targets with their primary weapon

Stubbjax:
* Closes TheSuperHackers/GeneralsGamePatch#1589

## 1.04:
The projectile travels through units in close proximity.

https://user-images.githubusercontent.com/11547761/199525027-88273bd9-89b8-4d18-9a06-8cf769b60880.mp4

## Patch:
The projectile collides with units in close proximity.

https://user-images.githubusercontent.com/11547761/199525042-589595a5-3dd9-4508-b82f-5d76e0bff464.mp4

There is also a discrepancy between the minimum range of the Black Napalm and standard flame weapons, which is a question of design and needs further discussion.
```
Weapon DragonTankFlameWeapon
  MinimumAttackRange = 0; 10.0
End
```
```
Weapon DragonTankFlameWeaponUpgraded
  MinimumAttackRange = 10.0
End
```

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
@@ -3321,14 +3321,16 @@ Locomotor PatriotMissileLocomotor
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @bugfix Stubbjax 03/11/2022 Changes MaxThrustAngle from 1 to fix flame projectiles not hitting targets on different terrain elevation.
+; Patch104p @bugfix Stubbjax 03/11/2022 Changes TurnRate from 5 to prevent flames from curving to followed targets.
 Locomotor DragonTankFlameLocomotor
   Surfaces = AIR
   Speed = 300               ; in dist/sec
   MinSpeed = 120            ; in dist/sec. (THRUST items must have nonzero minspeeds!)
   Acceleration = 2160       ; in dist/(sec^2)
   Braking = 0               ; in dist/(sec^2)
-  TurnRate = 5            ; in degrees/sec
-  MaxThrustAngle = 30       ; in degrees (NOT degrees/sec) ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets on different terrain heights / angles.
+  TurnRate = 0            ; in degrees/sec
+  MaxThrustAngle = 30       ; in degrees (NOT degrees/sec) 
   AllowAirborneMotiveForce = Yes
   Appearance = THRUST
 End
--- a/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
@@ -3328,7 +3328,7 @@ Locomotor DragonTankFlameLocomotor
   Acceleration = 2160       ; in dist/(sec^2)
   Braking = 0               ; in dist/(sec^2)
   TurnRate = 5            ; in degrees/sec
-  MaxThrustAngle = 1       ; in degrees (NOT degrees/sec)
+  MaxThrustAngle = 30       ; in degrees (NOT degrees/sec) ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets on different terrain heights / angles.
   AllowAirborneMotiveForce = Yes
   Appearance = THRUST
 End
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -4046,7 +4046,7 @@ Object DragonTankFlameProjectile
     AllowCollideForce = No  ; flames collide, but never apply forces when they do so
   End
   Behavior = MissileAIUpdate ModuleTag_05
-    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets in close proximity.
+    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 From No to fix flame projectiles not hitting targets in close proximity.
     FuelLifetime = 350
     DetonateOnNoFuel = Yes
     InitialVelocity = 120                ; in dist/sec
@@ -4141,7 +4141,7 @@ Object DragonTankFlameProjectileUpgraded
     AllowCollideForce = No  ; flames collide, but never apply forces when they do so
   End
   Behavior = MissileAIUpdate ModuleTag_05
-    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets in close proximity.
+    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 From No to fix flame projectiles not hitting targets in close proximity.
     FuelLifetime = 350
     DetonateOnNoFuel = Yes
     InitialVelocity = 120                ; in dist/sec
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -4046,7 +4046,7 @@ Object DragonTankFlameProjectile
     AllowCollideForce = No  ; flames collide, but never apply forces when they do so
   End
   Behavior = MissileAIUpdate ModuleTag_05
-    TryToFollowTarget = No
+    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets in close proximity.
     FuelLifetime = 350
     DetonateOnNoFuel = Yes
     InitialVelocity = 120                ; in dist/sec
@@ -4141,7 +4141,7 @@ Object DragonTankFlameProjectileUpgraded
     AllowCollideForce = No  ; flames collide, but never apply forces when they do so
   End
   Behavior = MissileAIUpdate ModuleTag_05
-    TryToFollowTarget = No
+    TryToFollowTarget = Yes ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets in close proximity.
     FuelLifetime = 350
     DetonateOnNoFuel = Yes
     InitialVelocity = 120                ; in dist/sec


xezon:
Oh wow this is great. I do wonder, could there be any reason EA did this? Could the fixed behaviour have unintended side effects?

commy2:
Does this also solve https://github.com/TheSuperHackers/GeneralsGameCode/issues/20 ?

I think neither normal nor Black Napalm should have minimum range. Similar thing as the Toxin Truck. Obviously a bug (you can tell by the comment that minimum range was removed at some point during development), and it makes no sense for upgrades to regress certain aspects of a weapon.

commy2:
This should be tested vs fast moving Technicals moving past the Dragon before and after change.

MTKing4:
This is great, i wonder if this fixes the jankiness of dragon tanks flame walls as well, as a flame wall less than minimum range could be the reason why it starts moving like a dumbass

Stubbjax:
> Does this also solve TheSuperHackers/GeneralsGameCode#20 ?

After coincidentally encountering some enemy red guard upon a hill while trying to test this and realising the flames weren't affecting them either, I believe I might have the solution to this. The key is the `MaxThrustAngle` in the `DragonTankFlameLocomotor`, which is set to a measly value of 1.

## 1.04:
A `MaxThrustAngle` of `1` prevents the projectile from hitting targets that are on a different height / angle to the Dragon Tank.

https://user-images.githubusercontent.com/11547761/199674100-9406518a-81ec-4807-a07d-791a79e7d52f.mp4

## Patch:
A `MaxThrustAngle` of `30` allows the projectile to hit targets at a different height / angle to the Dragon Tank.

https://user-images.githubusercontent.com/11547761/199674127-275a265a-3d46-414e-8753-720f01d37867.mp4

I also tried applying values of `20` and `25` but not all the Red Guard were effectively burned.

xezon:
Nice find. I assume 90 would be the max, but would then allow the Dragon Tank flame to shoot up a cliff possibly? Though I wonder, wouldn't the `MinTargetPitch`, `MaxTargetPitch` range prevent such an attack before `MaxThrustAngle` has any relevance? Does Dragon Tank have `MinTargetPitch`, `MaxTargetPitch` setup?

Stubbjax:
> Nice find. I assume 90 would be the max, but would then allow the Dragon Tank flame to shoot up a cliff possibly? Though I wonder, wouldn't the `MinTargetPitch`, `MaxTargetPitch` range prevent such an attack before `MaxThrustAngle` has any relevance? Does Dragon Tank have `MinTargetPitch`, `MaxTargetPitch` setup?

The `MinTargetPitch` and `MinTargetPitch` values strangely result in the Dragon Tank doing nothing if the target is outside the pitch range. I tried adding `MaxTargetPitch = 10` and this was the outcome:

https://user-images.githubusercontent.com/11547761/199686470-c36cf4f8-7362-4e70-9130-35cca45a5138.mp4

xezon:
Other tanks also have this setup and will not be able to shoot up and down more than 15 degrees. This can be very awkward on steep terrain and gives Rocket Men a big advantage.

xezon:
# Drive by test

## Original Fast Drive By

Technical takes no damage when speeding past attacking Dragon Tank.

https://user-images.githubusercontent.com/4720891/199827127-387e454c-4eac-406e-a089-632a40bc5586.mp4

## Patched Fast Drive By

Technical takes around 50% damage consistently when passing Dragon Tank.

https://user-images.githubusercontent.com/4720891/199827147-64749361-19bc-48c4-9944-60548bb554af.mp4

## Original Slow Drive By

Gattling takes no damage when speeding past attacking Dragon Tank.

https://user-images.githubusercontent.com/4720891/199828668-f36da4ec-3606-4d4b-bf0f-4a0bea410095.mp4

## Patched Slow Drive By

Gattling takes 100% damage when speeding past attacking Dragon Tank.

https://user-images.githubusercontent.com/4720891/199828710-6fa9cf0e-a29d-4dfb-95cf-638327bbe502.mp4

## Conclusion

This change would make Dragon Tank much better against ground units. Opponent would need to keep safe distance to Dragon Tank at all times. In Original game Dragon Tank cannot keep up with quick vehicles. This is why players often try to use Flame Wall ability to try hit something.

commy2:
I think the 1.04 Dragon Tank is very janky. I also think this change makes it too strong. It already is a mass produced unit in some matchups. Especially Stealthgen with the lack of Scorpions may get in trouble with this.

xezon:
The `TryToFollowTarget = Yes` part is responsible for the increased strength. Perhaps this change should be split into 2 or 3 separate ones. I suspect `MaxThrustAngle = 30` will not be as controversial.

xezon:
This needs split.

xezon:
New Pulls
* TheSuperHackers/GeneralsGamePatch#1588
* TheSuperHackers/GeneralsGamePatch#1589

xezon:
In `Behavior = MissileAIUpdate` there is another setting that is responsible for the way the Dragon Tank projectile behaves the way it does: `DistanceToTargetForLock = 0`. Removing this setting will make it behave identical to the GLA Toxin Tractor Stream. The Toxin Stream can hit targets much better. Increasing the value of `DistanceToTargetForLock` to 5, 10 or beyond is also an option, however it performs similarly to the `TryToFollowTarget = Yes` setting and in turn changes the way the Flame weapon works fundamentally. Setting `DistanceToTargetForLock` to around 10 also comes with its own issues, where the flames no longer uniformly hit their target, which is caused by the sweep of the turret.

I conclude that Dragon Tank weapon cannot be changed by INI without fundamentally changing the entire weapon performance. I recommend to leave it as is.

This is how the weapon performs right now. It is unable to hit close targets, but it overshoots and can hit targets further away than it can actually target to. This opens nice possibility for Micromanagement, as the effective weapon range can be optimized by firing near the actual target.

https://user-images.githubusercontent.com/4720891/213935621-ab407d15-854a-4163-a0d5-d6efa7e59c64.mp4

xezon:
To be closed if there is no more input for this.

MTKing4:
Why close? I thought we agreed it's a good change

xezon:
Changing the Projectile behavior of Dragon Tank makes it a quite different unit.

MTKing4:
it still is the same unit to me just less frustrating to use, besides, if we agreed it's a good change what does it matter then?

from what i see is that the angle fix essensial to fix TheSuperHackers/GeneralsGameCode#20 , and the the closer range just makes sense, it doesn't even buff it all that much it just makes it equal with all the other better units that could shoot that close, i don't see this change risky at all.

xezon:
This change now is entirely about "Dragon Tanks can now hit close targets with their primary weapon". Changing this makes Dragon Tank a different unit.

Stubbjax:
> This change now is entirely about "Dragon Tanks can now hit close targets with their primary weapon". Changing this makes Dragon Tank a different unit.

I've tested this pretty thoroughly. The minimum attack range standardisation has little to no impact on gameplay. How does this make it a different unit?

xezon:
> I've tested this pretty thoroughly. The minimum attack range standardisation has little to no impact on gameplay. How does this make it a different unit?

That change you mention is isolated with TheSuperHackers/GeneralsGamePatch#1588. This pull is just about the projectile (aka `TryToFollowTarget = Yes` & co).

ExiLe-ZH:
The only thing is that the minimum range prevented friendly fire quite a lot, can sometimes be hard controlling this with bigger armies, or when it's a little laggy.

Stubbjax:
> The only thing is that the minimum range prevented friendly fire quite a lot

How? Friendly fire doesn't hit close-proximity allies in 1.04, as demonstrated in the first video. That's all this change does if you take out the other fixes.

xezon:
> The only thing is that the minimum range prevented friendly fire quite a lot, can sometimes be hard controlling this with bigger armies, or when it's a little laggy.

I think you are refering to Fire Wall. That is Secondary weapon. Primary weapon was touched only.

commy2:
I agree to not changing the mechanics of primary weapon Dragon tanks, except that I would remove the minimum range from the Napalm versions, because the weapon has no minimum range before the upgrade, and having upgrades introduce regressions is weird.

Stubbjax:
Is everyone fine with closing this? Fire will continue to travel through explicitly targeted close-proximity allies without damaging them.

MTKing4:
I'm not ok with this, but it seems that it's getting closed regardless of our opinions

xezon:
> I conclude that Dragon Tank weapon cannot be changed by INI without fundamentally changing the entire weapon performance. I recommend to leave it as is.

This refers to changes to `TryToFollowTarget`, `DistanceToTargetForLock`. What this change does. Touching this will fix the issue yes, but it will also significantly buff the Dragon Tank in other ways (unintended).

If you find a way to fix the close proximity issue without changing the missile behaviour in other ways, feel free to open a Pull for it.

Stubbjax:
> > I conclude that Dragon Tank weapon cannot be changed by INI without fundamentally changing the entire weapon performance. I recommend to leave it as is.
> 
> This refers to changes to `TryToFollowTarget`, `DistanceToTargetForLock`. What this change does. Touching this will fix the issue yes, but it will also significantly buff the Dragon Tank in other ways (unintended).

In what ways? I observed no fundamental changes in performance or significant buffs.

xezon:
For example when a Quad drives by the Dragon Fire. In Original, Quad would take no damage. With this change, Quad dies. See above "Drive by Test".

Stubbjax:
> For example when a Quad drives by the Dragon Fire. In Original, Quad would take no damage. With this change, Quad dies. See above "Drive by Test".

There is also no damage if you force fire stationary targets as demonstrated by the original video. That's what this change is supposed to fix.

xezon:
I think I am unable to explain it properly. Maybe commy2 can explain it better.

Stubbjax:
> This is how the weapon performs right now. It is unable to hit close targets, but it overshoots and can hit targets further away than it can actually target to. This opens nice possibility for Micromanagement, as the effective weapon range can be optimized by firing near the actual target.

This is exactly how it performs after this change as well. This change fixes the flames going through targets that you _explicitly_ force fire upon. Targeting the ground is identical before and after.

xezon:
I cannot confirm this. Fire weapon worked vastly differently in my tests.

MTKing4:
I've reread the whole conversation now and understood the code behind it

basically `TryToFollowTarget = Yes` is forcing the projectile to hit the Targeted unit even if the turret is overshooting according to the game logic and visuals (damage is applied at the end of the flame effect) but it is also affecting moving units, forcing the flame to be on top of the moving unit rather than behind it (which was dictated by the slow turret movment and was shooting at where the quad was and not where it will be)

Yes it is a buff, but i still don't see it a big buff, dragon tank is really weak on that front, which is why people are defaulting to use firewall to hit moving units instead of normal attack, which is a cumbersome and not very user friendly at all, but it achieves the same result this change does, so it's really just an improvement of an already possible Senario

Stubbjax:
> I cannot confirm this. Fire weapon worked vastly differently in my tests.

https://user-images.githubusercontent.com/11547761/214313742-fa384ecd-5600-49cb-8291-b0fd25171ea1.mp4

> basically `TryToFollowTarget = Yes` is forcing the projectile to hit the Targeted unit even if the turret is overshooting according to the game logic

Pretty much. `TryToFollowTarget` will essentially cause the projectile to collide with the targeted unit, irrespective of the projectile's collision flags.

> Yes it is a buff

It's not really a buff because the flame projectile already collides with enemy units.

xezon:
The main outcome of this change is that for opponents it will be much harder to outrun the Dragon Tank, while the initial intention was to simply fix the proximity issue (see title of this change). This in turn would enable China to simply spam Dragon Tanks vs ground units and using Attack Move command to enemy base. It would yield much better success rate than Originally. And that is totally unrelated to fixing proximity.

Stubbjax:
> The main outcome of this change is that for opponents it will be much harder to outrun the Dragon Tank

It's not much harder at all. I've tested this thoroughly and it does not make any meaningful difference whatsoever.

https://user-images.githubusercontent.com/11547761/214323396-649453f6-613c-4312-ab23-a17d2d212f2e.mp4

https://user-images.githubusercontent.com/11547761/214323415-b44bd0d0-368b-4a9c-acd2-ee20ef4edb64.mp4

xezon:
Try driving past Dragon Tank not as close.

MTKing4:
Dragon tank has a very short range in comparison to other units so i doubt it would make it a _main battle unit_, as for being spammed, it is already spammed in 1.04

Also, we can test some 1v1s with this change to see if it really become so great, but my initial thoughts is that with this change the unit will become less janky at least, which is a good thing

Stubbjax:
> Try driving past Dragon Tank not as close.

https://user-images.githubusercontent.com/11547761/214544317-7a5147f7-ce88-4446-aae2-94153c432a7d.mp4

https://user-images.githubusercontent.com/11547761/214544332-fa0feb66-56e8-4419-bb6a-353594ed6ec4.mp4

xezon:
Interesting. That does not match the results of my own tests of this change. In my test vehicle takes different amount of damages.

Videos are here:
https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1417#issuecomment-1302639737

Stubbjax:
> Interesting. That does not match the results of my own tests of this change. In my test vehicle takes different amount of damages.

Your conclusion does not match your video demonstration. You are suggesting that `TryToFollowTarget = Yes` will make them perform better against opponents while only testing against friendly units.

xezon:
Ok, can you give an accurate overview of what `TryToFollowTarget` does vs enemy and friendly units?

The change description simply says "The projectile collides with units in close proximity."

But that appears to be not the full picture.

Stubbjax:
> Ok, can you give an accurate overview of what `TryToFollowTarget` does vs enemy and friendly units?

`TryToFollowTarget` allows the projectile to collide with the targeted unit, irrespective of the projectile's collision flags. As Dragon Tank flames don't collide with allies, `TryToFollowTarget` will allow collision if they are explicitly targeted.

xezon:
I think this is not whole picture. Dragon Flame is a missile aka `MissileAIUpdate`. Missile lock behaviour is originally disabled via `DistanceToTargetForLock = 0`. Added setting `TryToFollowTarget = Yes` overrides this setting and makes the missile lock on target. Dragon Flame will then follow the target, if the target changes position after the missile has launched.

I do not expect that Missile Projectile performs differently depending on target ally state. Whether or not this behaviour is indeed different on ENEMY and ALLIES I can test. If it is different, then this would be upsetting as well.

Stubbjax:
> I think this is not whole picture. Dragon Flame is a missile aka `MissileAIUpdate`. Missile lock behaviour is originally disabled via `DistanceToTargetForLock = 0`. 

A value of 0 is not disabled.

> Added setting `TryToFollowTarget = Yes` overrides this setting and makes the missile lock on target. Dragon Flame will then follow the target, if the target changes position after the missile has launched.

That is correct, but the projectile has such high acceleration and speed combined with low lifetime and angular movement that it makes no observable difference. The locomotor has a `TurnRate` of 5 degrees per second, and the projectile only lasts 350ms. So it may track by 1 or 2 degrees. If this is somehow of high concern, the `TurnRate` can be changed to 0.

> I do not expect that Missile Projectile performs differently depending on target ally state. Whether or not this behaviour is indeed different on ENEMY and ALLIES I can test. If it is different, then this would be upsetting as well.

It is different. As explained, the projectile does not collide with allies. `TryToFollowTarget` overrides this for explicit targets.

xezon:
> A value of 0 is not disabled.

Not clear what that means. Setting this to 1 or higher enables missile tracking. I was unable to observe any tracking in 0 state, so it did work as advertised.

> That is correct, but the projectile has such high acceleration and speed combined with low lifetime and angular movement that it makes no observable difference. The locomotor has a TurnRate of 5 degrees per second, and the projectile only lasts 350ms. So it may track by 1 or 2 degrees. If this is somehow of high concern, the TurnRate can be changed to 0.

It does make difference when force firing allied units. The flame does actually curve! Setting TurnRate to 0 is good idea.

> It is different. As explained, the projectile does not collide with allies. TryToFollowTarget overrides this for explicit targets.

Where is this Projectile Non-Allies collision setting?

Stubbjax:
> > A value of 0 is not disabled.
> 
> Not clear what that means. Setting this to 1 or higher enables missile tracking. I was unable to observe any tracking in 0 state, so it did work as advertised.

I never changed that field and the flame projectile follows targets.

> It does make difference when force firing allied units. The flame does actually curve! Setting TurnRate to 0 is good idea.

I don't believe the projectile curves. It may appear that way from the sweeping behaviour of the turret.

> Where is this Projectile Non-Allies collision setting?
```
DragonTankFlameWeapon
  ProjectileCollidesWith      = ENEMIES STRUCTURES WALLS SHRUBBERY
End
```

MTKing4:
I don't understand this whole thing of flames not colliding with allies, i don't see a difference between allies and enemies, can you demonstrate in a video what do you mean?

Your video here clear shows that it doesn't collide with enemies either unless force fired, which is the same with friendly units

> https://user-images.githubusercontent.com/11547761/214313742-fa384ecd-5600-49cb-8291-b0fd25171ea1.mp4

MTKing4:
So you're saying, if this was an enemy unit, flames will collide with it if you shoot behind it?

> https://user-images.githubusercontent.com/11547761/214313742-fa384ecd-5600-49cb-8291-b0fd25171ea1.mp4

xezon:
Ok. I have now retested this and found that the `TryToFollowTarget = Yes` setting is not responsible for the flame curving. Instead change TheSuperHackers/GeneralsGamePatch#1589 is. These 3 Dragon Tank changes are now split, so they can be tested and reviewed separately.

I would like a 3rd party to confirm that this change is good now.

Stubbjax:
> So you're saying, if this was an enemy unit, flames will collide with it if you shoot behind it?

That is correct.

> These 3 Dragon Tank changes are now split, so they can be tested and reviewed separately.

Okay, but please never rewrite git history like this unless absolutely necessary.

xezon:
There is one more thing needed to be tested here:

Do `MaxThrustAngle=1` and `TurnRate=5` do anything in the presence of `TryToFollowTarget = No`? If the answer is no, then `MaxThrustAngle` and/or `TurnRate` need to be set to zero with this change.

xezon:
~~I tested `MaxThrustAngle=90` + `TurnRate=90` + `TryToFollowTarget=No`. Flames still curve. This means we leave `MaxThrustAngle` and `TurnRate` in locomotor as is.~~

https://user-images.githubusercontent.com/4720891/215221720-76f6fc3a-b17c-4c53-8d39-7212da631c60.mp4

xezon:
~~This change looks good to me. Any other opinions on this?~~

xezon:
I did a mistake in test from 10 minutes ago. Forgot to save one of the edits before test. With proper correct test, situation is even more strange:

`MaxThrustAngle=90` + `TurnRate=90` + `TryToFollowTarget=No` will make flames NOT CURVE when targeting, but will allow killing near ALLIES, and near terrain, far terrain accurately.

But it also changes flame to no longer overshoot, thus decreases effective range.

https://user-images.githubusercontent.com/4720891/215223375-cad12214-3a22-47e5-aec1-ba0741545757.mp4

To me it looks as if flame curve only occurs when combining higher `MaxThrustAngle` with `TryToFollowTarget=Yes`. So we need to make a decision. Either change `MaxThrustAngle` or set `TryToFollowTarget=Yes`, not both.

xezon:
> > I cannot recall whether or not I tested that.
> > What happens with `MaxThrustAngle=30`, `TryToFollowTarget=Yes`, `TurnRate=0`?
> 
> This seems to be a good combo. The projectile does not follow units with a turn rate of 0.

I updated this change to `MaxThrustAngle=30`, `TryToFollowTarget=Yes`, `TurnRate=0`.

With this setup, close Allies can be hit, and flames will not curve, but targets on hills cannot be hit.

So it still does not work like this.

xezon:
I think we need to reduce this fix to the `TryToFollowTarget=Yes` only. We cannot have `MaxThrustAngle=30` without introducing vast flame targeting ability changes.

MTKing4:
I keep forgetting what goal is this pull trying to achieve, this is to only fix friendly minimum range force firing right? and it's not achievable? it doesn't seem as important as first advertised, TheSuperHackers/GeneralsGamePatch#1589 has higher priority to fix imo, as that also not achievable?

also, what is the undesirable side effects with the change? i thought this video showed that there's no undesirable change

> 
https://user-images.githubusercontent.com/11547761/214544317-7a5147f7-ce88-4446-aae2-94153c432a7d.mp4

https://user-images.githubusercontent.com/11547761/214544332-fa0feb66-56e8-4419-bb6a-353594ed6ec4.mp4

PR #1425, opened 2022-11-05
Change: Remove duplicate fields

Stubbjax:
Removed duplicate fields from various objects / definitions. Only assessed parent fields, not children. The former field was always removed, unless the value was identical, in which case the field with the worse / less consistent formatting was removed.

Note: Did not address the duplicate `ExperienceValue` fields of SpectreGunship, SpectreGunship1, SpectreGunship2, SpectreGunship3 and A10Thunderbolt. Also did not address duplicate `RadarPriority` fields of AngryMobNexus. The original values for the respective fields look more correct than the duplicate / used values and are better addressed separately.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -297,7 +297,6 @@ Armor ToxinTruckArmor ;TruckArmor that is immune to poison
   Armor = SMALL_ARMS        50%
   Armor = GATTLING          50%    ;resistant to gattling tank
   Armor = COMANCHE_VULCAN   50%
-  Armor = INFANTRY_MISSILE  50%
   Armor = POISON            0%    ;IMMUNE! It spews poison :)
   Armor = RADIATION         0%
   Armor = MICROWAVE         0%
@@ -321,7 +320,6 @@ Armor DragonTankArmor
   Armor = COMANCHE_VULCAN   25%
   Armor = FLAME              0%
   Armor = RADIATION         50%      ;Radiation does less damage to tanks.
-  Armor = MICROWAVE          0%
   Armor = POISON            25%    ;Poison does a little damage, just for balance reasons.
   Armor = MICROWAVE         0%
   Armor = SNIPER            0%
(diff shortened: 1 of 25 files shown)


commy2:
I know it is extra work, but can we split stuff like this into individual PRs? The changes are all unrelated to each other, e.g. there is no connection between the Spectre armor values vs Flashbangs and Black Lotus' command set. Would make it easier to review and look up stuff later. Also, some values are not merely duplicates, but different values that overwrite each other. This still documents intend that is lost by just deleting the lines. Are we sure some tokens can't be used twice? E.g. research sound of Angry Mob does contain a voice line and a cheer. What's up with the first effect?

Stubbjax:
> I know it is extra work, but can we split stuff like this into individual PRs? The changes are all unrelated to each other, e.g. there is the connection between the Spectre armor values vs Flashbangs and Black Lotus' command set. Would make it easier to review and look up stuff later.

Sure, though I'm not sure it'd make it easier as each commit can be assessed just as easily as separate pull requests.

> Also, some values are not merely duplicates, but different values that overwrite each other. This still documents intend that is lost by just deleting the lines. Are we sure some tokens can't be used twice? E.g. research sound of Angry Mob does contain a voice line and a cheer. What's up with the first effect?

I tested most of them and they are straight-up overridden. I looked at the angry mob upgrade sound and it is silent - the audio it references does not exist.

xezon:
The change can be committed as "Rebase and Merge", then all commits will be preserved on Main. `(#1425)` must be appended to all commit descriptions in that case.

xezon:
How did you search and find these duplicates? With a script?

commy2:
> the audio it references does not exist.

Okay, then it can be removed without replacement.

Same for true duplicate entries (twice the same value). Can be removed as is.

But when two conflicting values are used anywhere, then maybe it should be noted what the other value was, even if it did not apply.

> How did you search and find these duplicates? With a script?

Searching these by script would be very helpful. I don't think the changes in this PR are anywhere near complete. Another reason it should be split up imo.

Stubbjax:
> How did you search and find these duplicates? With a script?

Yep, with a script.

> Searching these by script would be very helpful. I don't think the changes in this PR are anywhere near complete. Another reason it should be split up imo.

It is true that there are likely to be more. I will split the changes into multiple PRs.

PR #1426, opened 2022-11-05
Change: Add garrison audio to the Barracks

Stubbjax:
Injured infantry units that would enter the Barracks in 1.04 played no audio whatsoever. This change adds enter / exit sound effects in order to provide feedback to the player and standardises the enter / exit feedback with other structures in the game.

https://user-images.githubusercontent.com/11547761/200110800-88175919-fb07-4895-b9e5-27a8a8a5863d.mp4


commy2:
No sound is played when a unit is produced from a Barracks. Why would a unit leaving the quarters play a sound in one case, but not the other?

xezon:
Thinking about it, when a unit spawns, it already plays a creation sound. In same cases even a level up sound. Garrison sound on top would not be necessary. I am ok with this change as is.

xezon (dismissed):

xezon:
According to this video, a stealth unit already makes a sound when entering barracks. This means that this change gives no additional intelligence on stealth units.

https://user-images.githubusercontent.com/4720891/201519267-f45b255d-756f-46aa-90d8-81930495947c.mp4

xezon (approved):

PR #1436, opened 2022-11-05
Change: Removed a duplicate attack field from the Aurora Bomber creation audio event.

Stubbjax:
The `gradio1g` audio that is referenced by the former value does not exist.


xezon:
Is this not a bug then, meaning sometimes no sound plays because the sound is not found?

xezon (approved):

Stubbjax:
> Is this not a bug then, meaning sometimes no sound plays because the sound is not found?

It would be if the former value was used.

xezon:
Ah right, it is duplicate line

PR #1440, opened 2022-11-05
Change: Removed a duplicate command set item field from the GLA Stealth General's Battle Bus.

Stubbjax:
This button is overridden, and also nonfunctional if placed in a valid slot such as `10`.


xezon (approved):

MTKing4:
You're referring to the GC Stealth General and not the normal stealth general, right?

Stubbjax:
> You're referring to the GC Stealth General and not the normal stealth general, right?

No, this change is referencing the Stealth General's Battle Bus, which uses `Slth_GLAVehicleBattleBusCommandSet`. The GC variant uses `GC_Slth_GLAVehicleBattleBusCommandSet`, which is a distinct command set.

MTKing4:
Ah ok, i was wondering what that nonfunctional command set item was for stealth general's bus, because i know that GC Stealth bus has one of those

PR #1442, opened 2022-11-05
Change: Removed duplicate fields from various civilian / system objects.

Stubbjax:
The former field is completely overridden.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
@@ -16953,7 +16953,7 @@ Object MogadishuWell
   End
 
   ; *** ENGINEERING Parameters ***
-  KindOf                = STRUCTURE SELECTABLE IMMOBILE
+  KindOf                = STRUCTURE SELECTABLE IMMOBILE CAN_SEE_THROUGH_STRUCTURE
   Body            = ActiveBody ModuleTag_02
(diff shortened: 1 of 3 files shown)


xezon:
Do any flags of the duplicate field make any sense for the respective objects?

Stubbjax:
> Do any flags of the duplicate field make any sense for the respective objects?

It's basically a question whether `CAN_SEE_THROUGH_STRUCTURE` should apply to any of the following objects:

image

xezon:
CAN_SEE_THROUGH_STRUCTURE renders the silhouette through object? How are comparable structures setup?

Stubbjax:
> CAN_SEE_THROUGH_STRUCTURE renders the silhouette through object? How are comparable structures setup?

It allows units to shoot through them.

xezon:
I have no preference. Could make sense that unit can shoot over than fountain. If in doubt, keep original behavior.

Stubbjax:
> I have no preference. Could make sense that unit can shoot over than fountain. If in doubt, keep original behavior.

It's funny that you can blow it up. Only then are you allowed to shoot over it. It seems reasonable to adjust this one.

https://user-images.githubusercontent.com/11547761/200130870-234102c0-a0c0-4ffc-8d4a-468c5c3365e1.mp4

xezon:
I suggest commit these 2 changes individually, to keep it separate. I created task for the CAN_SEE_THROUGH_STRUCTURE.

* #1446

xezon (approved):

PR #1445, opened 2022-11-05
Fix: atventwall02, atventwall03, atventwall04 textures

xezon:
This change fixes atventwall textures. They are used on various USA structures.

### atventwall02_d, atventwall02_e, atventwall03_d, atventwall03_e, atventwall04_d, atventwall04_e

* Fixes burns drawing over vent shafts
* Repaints some burns to avoid repeating patterns


xezon:
Looks good to me.

Issue #676, opened 2022-07-05
Non-controllable planes will fly in circles over certain objects


xezon:
Non-controllable planes will fly in circles over certain objects.

See Replay at 09:43 minutes.

[carpet_bomber_circles.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/9043587/carpet_bomber_circles.zip)

Goldfish:
> The reason the plane bugged is because of the planes on the ground. I made a test map with a wall of these and the a10's got stuck above them flying circles for minutes.

![post-884445-1651762453](https://user-images.githubusercontent.com/4720891/177259146-058e9b65-84fa-4f79-b7c9-d7965b9d3d78.jpg)

More structures have been identified to trigger this supposed "collision avoidance" behaviour:

![post-884445-1651769921](https://user-images.githubusercontent.com/4720891/177259483-49fed9e4-6d94-4be2-9582-3c32d7e5f776.jpg)

![post-884445-1651769931](https://user-images.githubusercontent.com/4720891/177259485-1370ae89-9b78-4304-b704-333d14f6c2cf.jpg)

![post-884445-1651769940](https://user-images.githubusercontent.com/4720891/177259490-ab5341a5-7f87-44f3-8bfd-02603ef8b51c.jpg)

![post-884445-1651769949](https://user-images.githubusercontent.com/4720891/177259492-371abf73-6e2a-47b9-94d6-a0b76412ee72.jpg)

ECBVovin:
These have KindOf AIRCRAFT_PATH_AROUND. Which make aircraft path around. Can cause weird issues with things going in circles. Wouldn't suggest using on maps.

xezon:
AsianMegaHotel01: AIRCRAFT_PATH_AROUND

Damn.

Maybe it would be no harm to remove all these flags and just let planes clip through these mega hotels and smaller things.

Jundiyy:
Perhaps maybe a carbon copy and have the code in the copy? So they can still be used in fun maps?
This would mean all pvp maps would be auto fixes, but also non pvp maps (in case someone wanted it in). 
I don't think anyone would miss them but at least they can still use them if they wish.

Could also try editing locomotor, maybe it could fix the circling, don't know how much is needed.

ExiLe-ZH:
Make aircraft and buildings go boom? 

9/11? :O

commy2:
map.ini can add 

```ini
Object AsianMegaHotel01
[...]

PR #1447, opened 2022-11-06
Change: Remove AIRCRAFT_PATH_AROUND from small non cinematic structures

xezon:
Partly addresses
* #676

This change removes AIRCRAFT_PATH_AROUND KindOf from non cinematic structures. This avoids plane collision avoidance logic.

The following structures are affected

* BaghdadVictoryMemorial01
* BaghdadVictoryMemorial02
* ~~EuropeanMuseum~~
* ~~AsianTVStation~~
* EnglishTVStatio
* ~~AsianBank~~
* StanWatchTowerTall
* WaterPlant
* ~~EuroHighrise~~
* ~~BaikonurRocketPad~~
* ~~BaikonurRocketPad_C~~
* ~~OceanConventionCenter~~
* ~~AsianSkyScraper01~~
* ~~AsianMegaHotel01~~
* ~~AsianOffice02~~
* ~~InternationalHotel01~~
* ~~BioRocketPad_C~~

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianBuilding.ini
@@ -222,7 +222,8 @@ Object BaghdadVictoryMemorial01
   End
 
   ; *** ENGINEERING Parameters ***
-  KindOf                = STRUCTURE SELECTABLE IMMOBILE AIRCRAFT_PATH_AROUND
+  ; Patch104p @tweak Remove AIRCRAFT_PATH_AROUND to avoid circling planes.
+  KindOf                = STRUCTURE SELECTABLE IMMOBILE
   Body                  = ActiveBody ModuleTag_02
     MaxHealth       = 2000.0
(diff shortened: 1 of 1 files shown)


Stubbjax:
Does this mean aircraft will visually fly through the respective structures if they are tall enough? Does it affect controlled aircraft?

xezon:
Scripted planes will fly through/over tall structure.

xezon:
Good to go?

MTKing4:
Good for me

Stubbjax (approved):

ExiLe-ZH:
maybe similar code would work for cargo planes to avoid any enemy building 😂

commy2:
I think the planes should still try to avoid buildings that are (almost) as large as high the planes fly.

ExiLe-ZH:
By flying higher then? 

We can't keep this if it means genpower planes don't reach the target destination because of it.

xezon:
I looked at all structures listed above. Only 5 of them look safe for removal of AIRCRAFT_PATH_AROUND flag. So I adjusted this change accordingly.

PR #1450, opened 2022-11-06
Fix: atcemblock02 textures

xezon:
This change fixes atcemblock02 textures.

## atcemblock02_e

* Repaints burns

## atcemblock02_es

* Repaints burns


xezon:
I do not know where this texture is used. Texture looks ok however.

Issue #1212, opened 2022-09-15
USA's Bombardment and HoldTheLine Battleplans are rarely used


ExiLe-ZH:
These are pretty much only used with AFG when he flies, since it doesn't affect Aircraft. The Bombardment Cannon then helps out defensively and HoldTheLine makes the Strat center a tougher target. 

Other than that there's basically no reason to initiate these battleplans as they are much inferior with Rock/TowVee spam. Probably same story for Tank (+ Tommy) spam.

### Proposal 1

Increase Bombardment bonus to +25%, same as AP Bullets/Rockets, Uranium Shells, Chain Guns etc. 

### Proposal 2

Increase HoldTheLine bonus to +25% HP (0.8), to cancel out AP Bullets/Rockets.

Might need to fix the double Composite Armor bug.

### Proposal 3

Increase Bombardment Gun range, damage or radius.

xezon:
```
  WeaponBonus = BATTLEPLAN_BOMBARDMENT       DAMAGE        120%
  WeaponBonus = BATTLEPLAN_SEARCHANDDESTROY  RANGE         120%

Object AmericaStrategyCenter
    HoldTheLinePlanArmorDamageScalar      = 0.9 ;Armor damage bonus scalar -- LESS is better!
    SearchAndDestroyPlanSightRangeScalar  = 1.2 ;Sight range bonus -- more is better!
    StrategyCenterSearchAndDestroySightRangeScalar = 2.0
    StrategyCenterSearchAndDestroyDetectsStealth   = Yes
    StrategyCenterHoldTheLineMaxHealthScalar       = 2.0
End

Armor StructureArmor
  Armor = DEFAULT           100%
  Armor = Surrender         0%
  Armor = SMALL_ARMS        50%
  Armor = GATTLING          10%
  Armor = COMANCHE_VULCAN   50%
  Armor = RADIATION         0%
  Armor = MICROWAVE         0%
  Armor = SNIPER            0%
  Armor = POISON            1%
  Armor = MELEE             0%
  Armor = LASER             0%
  Armor = HAZARD_CLEANUP    0%
  Armor = INFANTRY_MISSILE  50%
  Armor = PARTICLE_BEAM     200%
  Armor = KILL_PILOT        0%
  Armor = SURRENDER         0%
  Armor = AURORA_BOMB       250%
  Armor = LAND_MINE         0%
  Armor = FLAME             50%
[...]

PR #1457, opened 2022-11-08
Change: Reduce the Strategy Center's stealth detection range and S&D vision bonus

Stubbjax:
As all heroes (now) have a sight range of 400, options for reacting to a Strategy Center's presence before being detected are incredibly limited. Not only this, but the structure's extensive stealth detection range over the owner's entire base effectively renders heroes useless - on top of the substantial range bonus that Search and Destroy provides. Reducing the stealth detection range is a reasonable way to counterbalance the incredible effectiveness of S&D without directly affecting the bonus itself.

A range reduction could also have some additional benefits / implications, such as the increased importance and skill involved in Strategy Center placement - it makes sense that there would be some _strategy_ involved in placing the Strategy Center, after all. In 1.04, players can practically place the structure down anywhere and be covered, but with a reduced stealth detection range, the building's location might have to be more carefully considered. This would also somewhat increase the prospect of stealth / guerilla tactics for opponents and diversify strategies a bit more. It may even increase the utility of Sentry Drones (and Spy Drones - for opposing USAs as well), and it could certainly benefit / reward players who actively pursue stealth detection / awareness, which is far more meaningful than having it handed to them on a silver platter. A stealth detection range of 350 provides enough leeway for heroes to react accordingly, and affords heroes and other stealth units greater manoeuvrability around an opponent's base.

Another thing to note is the Strategy Center's extreme shroud-clearing range of 400 → 800 (×2) under the effects of Search and Destroy, and is another unnecessary advantage that discounts map awareness. The ×2 scalar has been reduced to ×1.5, resulting in a more palatable range of 400 → 600. This was preferred over reducing the default vision of 400 to 300 due to the Bombardment cannon's weapon range of 400.

To summarise, the proposed changes are:
- Reduce the Strategy Center's S&D stealth detection range from 500 to 350
- Reduce the Strategy Center's S&D vision bonus from ×2 (800) to ×1.5 (600)

I believe both of these changes would be fairly subtle and unlikely to be noticed by the majority of players.

image

Originally discussed in #1212.

# Tests

After some extensive testing, it seems as though there is a mismatch between the detection and shroud reveal ranges. Despite Black Lotus's shroud range of 400 and the Strategy Center's stealth detection range of 500, it seems as though the two are roughly equal during gameplay, as seen in the footage below. It is unclear whether this should impact the suggested range of 350, as a large component of the change is the resultant enhanced manoeuvrability and strategical flexibility in an opponent's base.

## Test 1 - 500 detection range vs Black Lotus
The Strategy Center is pretty much revealed at the same time as Black Lotus is detected. No counterplay potential - especially when factoring in network latency.

https://user-images.githubusercontent.com/11547761/200460543-e689627a-59cc-4f1e-ab95-8dacb03ff022.mp4

## Test 2 - 400 detection range vs Black Lotus
There is now a reasonable amount of time to react to the presence of a Strategy Center, and improved manoeuvrability within its proximity. Allows greater potential for heroes to access other areas of the opponent's base without detection.

https://user-images.githubusercontent.com/11547761/200461751-f4e386dc-3945-404f-b9b3-d97d2088b368.mp4

## Test 3 - 350 detection range vs Black Lotus
Even more time to react to a Strategy Center's presence, and further manoeuvrability prospects around an opponent's base. Provides heroes with further access to other areas of the opponent's base without detection.

https://user-images.githubusercontent.com/11547761/200467805-67e1abbb-0fd5-47e9-9307-85384115c5b2.mp4

## Comparison

image

Common stealth detection ranges for reference:

| | Range |
| - | - |
Scout Drone | 150 |
Tunnel | 150 |
Troop Crawler | 175 |
Gattling Cannon | 200 |
King Raptor | 200 |
Pathfinder | 200 |
Patriot Battery | 200 |
Stinger Site | 200 |
Sentry Drone | 225 |
Radar Van | 250 |
Spy Drone | 250 |
Black Lotus | 300 |
Outpost | 300 |
Strategy Center (this) | 350 |
Strategy Center (1.04) | 500 |

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -11824,7 +11824,7 @@ Object AirF_AmericaStrategyCenter
     ;***NOTE*** WEAPON bonuses for army are specified in GameData.ini file!
 
     ;Building bonuses granted based on battle plan mode.
-    StrategyCenterSearchAndDestroySightRangeScalar = 2.0
+    StrategyCenterSearchAndDestroySightRangeScalar = 1.5
     StrategyCenterSearchAndDestroyDetectsStealth   = Yes
     StrategyCenterHoldTheLineMaxHealthScalar       = 2.0
     StrategyCenterHoldTheLineMaxHealthChangeType   = PRESERVE_RATIO
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -11857,7 +11857,7 @@ Object AirF_AmericaStrategyCenter
   Behavior = StealthDetectorUpdate ModuleTag_16
     DetectionRate       = 500   ; how often to rescan for stealthed things in my sight (msec)
     InitiallyDisabled   = Yes   ; only will be active when search & destroy plan active.
-    DetectionRange     =  500    ;Dustin, enable this for independant balancing!
+    DetectionRange     =  350    ;Dustin, enable this for independant balancing!
     CanDetectWhileGarrisoned  = No ;Garrisoned means being in a structure that you units can shoot out of.
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -11824,7 +11824,7 @@ Object AirF_AmericaStrategyCenter
     ;***NOTE*** WEAPON bonuses for army are specified in GameData.ini file!
 
     ;Building bonuses granted based on battle plan mode.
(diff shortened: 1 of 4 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -11824,7 +11824,7 @@ Object AirF_AmericaStrategyCenter
     ;***NOTE*** WEAPON bonuses for army are specified in GameData.ini file!
 
     ;Building bonuses granted based on battle plan mode.
-    StrategyCenterSearchAndDestroySightRangeScalar = 2.0
+    StrategyCenterSearchAndDestroySightRangeScalar = 1.5 ; Patch104p @tweak Stubbjax 08/11/2022 Reduced from 2.0 to 1.5.
```
xezon, line 11827: `Reduced from 2.0` or `From 2.0` would be enough information. Same with the other change descriptions.

xezon (approved):
Change makes sense. Good change description. China Propaganda Center also has small area of effect.

MTKing4:
Good change, 350 makes perfect sense

ExiLe-ZH:
Agreed with 350, the player should have some chance to scout the base without being detected. 50 is a pretty small margin so the player has to pay attention.

PR #1459, opened 2022-11-08
Fix: atmetal02 textures

xezon:
This change fixes and enhances atmetal02 textures. This texture is used on USA Particle Cannon.

## atmetal02_d, atmetal02_e

* Repaints some burns

## atmetal02_s

* Fixes semi-transparent snow

## atmetal02_ds, atmetal02_es

* Fixes semi-transparent snow
* Fixes semi-transparent burns
* Repaints some burns


xezon:
This texture is used on the USA Particle Cannon. I optimized it for this use case and reduced the tiling. Looks good.

PR #1463, opened 2022-11-11
Change: Adjusted Battle, Hellfire and Scout Drone armour and health.

Stubbjax:
Battle, Hellfire and Scout Drones provide USA with a huge advantage, as they distract enemy anti-air units and act as an additional decoy and damage sponge that significantly and disproportionately extends the longevity of USA vehicles. They are fairly cheap and build incredibly fast, which exacerbates the problem even further, and are easily and readily replaced. They are also incredibly frustrating to play against, as a player's anti-air units will prioritise the drones over vehicles. It is clear that drones have a far greater impact on the game than they have any right to, and are part of an incredibly flawed design that should really be addressed[.](https://user-images.githubusercontent.com/11547761/201368326-66025352-a350-412d-af98-2176196b5a64.png)

To lessen the severity of the respective issue(s), the following changes have been applied:

- Replaced drone `TankArmor` with `AirplaneArmor`
- Doubled drone hit points from 100 to 200
- Doubled Drone Armor hit points boost from 25 to 50 to match the doubled hp

With the above changes, the three drones are now standardised with the Spy Drone, which already uses `AirplaneArmor` with 200 hit points. This provides a nice reference, as players are already familiar with the Spy Drone and its attributes. As much as these changes may make sense, it's important to note that drones are used incredibly often, and so a great deal of care and consideration must be taken when altering them. In an ideal world, these drones would have minimal armour and die in half the time they take to create, but that would have greater implications and is beyond the scope of this patch.

The key differences between the two armour types are displayed in the table below.

| | TankArmor | AirplaneArmor
-- | -- | --
SMALL_ARMS | 25% | 120%
GATTLING | 10% | 120%
FLAME | 25% | 100%
INFANTRY_MISSILE | 100% | 120%
JET_MISSILES | 100% | 25%
RADIATION | 50% | 25%

The above `TankArmor` properties - particularly the low `25% SMALL_ARMS` and ` 10% GATTLING` modifiers - are the primary offenders here, and the reason drones are so hard for Quad Cannons, Gattling Tanks and Avengers to shoot down.

In 1.04, a Quad Cannon deals 5 damage per round with its anti-air weapon. `TankArmor`'s 25% modifier for `SMALL_ARMS` damage reduces this down to a pitiful 1.25 damage per round against drones. With 100 hit points, this means a Quad Cannon must fire 80 rounds to take down a single drone. At one round every 100ms (~133ms) - roughly 7.5 rounds per second - a drone takes ~10.5 seconds to destroy. This is an outrageously long time, especially when a Humvee takes 48 rounds to destroy and a King Raptor takes just 40 rounds (half of a drone). Quite ludicrous!

Gattling Tanks are a bit trickier to compare due to the barrel spin logic, so let's assume a fully spun up state. A Gattling Tank deals 12 damage per round with its anti-air weapon. Due to the odd fact that Gattling Tanks deal `SMALL_ARMS` damage with their anti-air weapons, this damage is also multiplied by `TankArmor`'s 25% modifier to 3 damage per round against drones. With 100 hit points, a drone takes 34 rounds to kill. At one round every 100ms (~133ms), a drone takes ~4.4s for a Gattling Tank to destroy. A King Raptor takes just 17 rounds to destroy, for reference.

It is also important to note that an Ambulance's healing aura can provide drones with an additional 5 hit points per second, which is equivalent to 4 Quad Cannon Rounds and ~1.67 Gattling Tank rounds. This effectively doubles the time it takes for a Quad Cannon to destroy a drone to ~21.7s. (It takes the same time to destroy a Cold Fusion Reactor.) This bonus will be effectively halved after the change, as drones will take twice as long to heal up to 200 hit points.

With these crazy numbers in mind, let's compare the results of the change. The number of shots to destroy a drone are as follows:

| | 1.04 | This | Change
-- | -- | -- | --
Quad Rounds | 80 | 34 | ×2.35
Gattling Rounds | 34 | 14 | ×2.43
Infantry Missiles | 3 | 5 | ×0.6
King Raptor Missiles | 1 | 7 | ×0.14

The Quad Cannon and Gattling Tank's prospects seem much more reasonable against the `AirplaneArmor`, and will destroy a drone (200hp) slightly faster than a King Raptor (240hp) when both use the same armour.

An unfortunate consequence of the `AirplaneArmor` change is that it (strangely) protects strongly against `JET_MISSILES`. This is an existing problem in 1.04, as a King Raptor cannot take down a simple Spy Drone in one volley (6 × 125 × 0.25 = 187.5). Whether `AirplaneArmor`'s damage multiplier for `JET_MISSILES` should be slightly increased to offset this anomaly is another question and ideally separate from this change as it would impact most aircraft.

There is also the issue of some unspecified damage types effectively doing half damage after the change (due to the ×2 hit points), though other damage types are typically powerful enough and never really a problem in the original game. Regardless, a subsequent follow-up alteration would likely be to create a dedicated armour class for drones and apply additional damage for the respective damage types to keep them consistent with 1.04 or adjust hp where it makes sense.

## 1.04:
A dedicated anti-air unit struggles to take down three pitiful $100 drones.

https://user-images.githubusercontent.com/11547761/201356064-54cfbb49-b82b-4ca1-aaba-37332e44b282.mp4

## Patch:
The three drones are dispatched in a much more reasonable time frame.

https://user-images.githubusercontent.com/11547761/201356053-ed71930f-329c-4c16-a04e-e67c7b303e00.mp4

With the proposed changes, drones no longer stall anti-air units for ridiculously long periods of time, and their reduced durability aligns much more logically with their prices, build speed and function. Ultimately, USA will effectively have to pay more to maintain their drones.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -291,6 +291,26 @@ Armor AvengerArmor
   Armor = SUBDUAL_BUILDING     0%
 End
 
+Armor DroneArmor
+  Armor = SMALL_ARMS       120%
+  Armor = GATTLING         120%
+  Armor = EXPLOSION        200%
+  Armor = INFANTRY_MISSILE 120%
+  Armor = JET_MISSILES     200%
+  Armor = RADIATION         25%
+  Armor = POISON            25%
+  Armor = KILL_PILOT         0%
+  Armor = SUBDUAL_VEHICLE    0%
+  Armor = SUBDUAL_MISSILE    0%
+  Armor = SUBDUAL_BUILDING   0%
+  Armor = MELEE              0%
+  Armor = SURRENDER          0%
+  Armor = SNIPER             0%
+  Armor = MELEE              0%
+  Armor = LASER              0%
+  Armor = HAZARD_CLEANUP     0%
+  Armor = MICROWAVE          0%
+End
 
 Armor ToxinTruckArmor ;TruckArmor that is immune to poison
   Armor = CRUSH             50%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -292,10 +292,10 @@ Armor AvengerArmor
 End
 
 Armor DroneArmor
-  Armor = SMALL_ARMS       120%
-  Armor = GATTLING         120%
+  Armor = SMALL_ARMS       100%
+  Armor = GATTLING         100%
   Armor = EXPLOSION        200%
-  Armor = INFANTRY_MISSILE 120%
+  Armor = INFANTRY_MISSILE 100%
   Armor = JET_MISSILES     200%
   Armor = RADIATION         25%
   Armor = POISON            25%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -5948,8 +5948,8 @@ Object AirF_AmericaVehicleBattleDrone
   RadarPriority       = UNIT
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE CAN_ATTACK DRONE NO_SELECT
   Body = ActiveBody ModuleTag_03
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 160.0
+    InitialHealth   = 160.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_04
@@ -5972,7 +5972,7 @@ Object AirF_AmericaVehicleBattleDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_06
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 40.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
@@ -6101,8 +6101,8 @@ Object AirF_AmericaVehicleScoutDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE DRONE NO_SELECT
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 120.0
+    InitialHealth   = 120.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -6138,7 +6138,7 @@ Object AirF_AmericaVehicleScoutDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_07
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 30.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
@@ -6354,7 +6354,7 @@ Object AirF_AmericaVehicleSpyDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   VisionRange     = 250
@@ -6379,8 +6379,8 @@ Object AirF_AmericaVehicleSpyDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SCORE DRONE SELECTABLE INERT NO_SELECT IGNORES_SELECT_ALL ; Patch104p @bugfix commy2 03/09/2021 No longer select Spy Drone with Q-shortcut.
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 120.0
+    InitialHealth   = 120.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -6434,7 +6434,7 @@ Object AirF_AmericaVehicleSpyDrone
   End
   Behavior = MaxHealthUpgrade ModuleTag_10
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 30.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -5919,7 +5919,7 @@ Object AirF_AmericaVehicleBattleDrone
   End
   ArmorSet
     Conditions        = None
-    Armor             = AirplaneArmor
+    Armor             = DroneArmor
     DamageFX          = SmallTankDamageFX
   End
 
@@ -6074,7 +6074,7 @@ Object AirF_AmericaVehicleScoutDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 100
@@ -6217,7 +6217,7 @@ Object AirF_AmericaVehicleHellfireDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 500
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -5919,7 +5919,7 @@ Object AirF_AmericaVehicleBattleDrone
   End
   ArmorSet
     Conditions        = None
-    Armor             = TankArmor
+    Armor             = AirplaneArmor
     DamageFX          = SmallTankDamageFX
   End
 
@@ -5948,8 +5948,8 @@ Object AirF_AmericaVehicleBattleDrone
   RadarPriority       = UNIT
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE CAN_ATTACK DRONE NO_SELECT
   Body = ActiveBody ModuleTag_03
-    MaxHealth       = 100.0
-    InitialHealth   = 100.0
+    MaxHealth       = 200.0
+    InitialHealth   = 200.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_04
@@ -6074,7 +6074,7 @@ Object AirF_AmericaVehicleScoutDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = TankArmor
+    Armor           = AirplaneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 100
@@ -6101,8 +6101,8 @@ Object AirF_AmericaVehicleScoutDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE DRONE NO_SELECT
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 100.0
-    InitialHealth   = 100.0
+    MaxHealth       = 200.0
+    InitialHealth   = 200.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -6138,7 +6138,7 @@ Object AirF_AmericaVehicleScoutDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_07
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 25.0
+    AddMaxHealth  = 50.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
@@ -6217,7 +6217,7 @@ Object AirF_AmericaVehicleHellfireDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = TankArmor
+    Armor           = AirplaneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 500
@@ -6247,8 +6247,8 @@ Object AirF_AmericaVehicleHellfireDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE CAN_ATTACK DRONE NO_SELECT
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 100.0
-    InitialHealth   = 100.0
+    MaxHealth       = 200.0
+    InitialHealth   = 200.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -6277,7 +6277,7 @@ Object AirF_AmericaVehicleHellfireDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_07
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 25.0
+    AddMaxHealth  = 50.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
@@ -872,8 +872,8 @@ Object AmericaVehicleBattleDrone
   RadarPriority       = UNIT
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE CAN_ATTACK DRONE NO_SELECT
   Body = ActiveBody ModuleTag_03
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 160.0
+    InitialHealth   = 160.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_04
@@ -896,7 +896,7 @@ Object AmericaVehicleBattleDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_06
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 40.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
@@ -1025,8 +1025,8 @@ Object AmericaVehicleScoutDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE DRONE NO_SELECT
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 120.0
+    InitialHealth   = 120.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -1062,7 +1062,7 @@ Object AmericaVehicleScoutDrone
 
   Behavior = MaxHealthUpgrade ModuleTag_07
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 30.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
@@ -1277,7 +1277,7 @@ Object AmericaVehicleSpyDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   VisionRange     = 250
@@ -1302,8 +1302,8 @@ Object AmericaVehicleSpyDrone
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SCORE DRONE SELECTABLE INERT NO_SELECT IGNORES_SELECT_ALL ; Patch104p @bugfix commy2 03/09/2021 No longer select Spy Drone with Q-shortcut.
 
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 200.0
-    InitialHealth   = 200.0
+    MaxHealth       = 120.0
+    InitialHealth   = 120.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_03
@@ -1357,7 +1357,7 @@ Object AmericaVehicleSpyDrone
   End
   Behavior = MaxHealthUpgrade ModuleTag_10
     TriggeredBy   = Upgrade_AmericaDroneArmor
-    AddMaxHealth  = 50.0
+    AddMaxHealth  = 30.0
     ChangeType    = ADD_CURRENT_HEALTH_TOO   ;Choices are PRESERVE_RATIO, ADD_CURRENT_HEALTH_TOO, and SAME_CURRENTHEALTH
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
@@ -843,7 +843,7 @@ Object AmericaVehicleBattleDrone
   End
   ArmorSet
     Conditions        = None
-    Armor             = AirplaneArmor
+    Armor             = DroneArmor
     DamageFX          = SmallTankDamageFX
   End
 
@@ -998,7 +998,7 @@ Object AmericaVehicleScoutDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 100
@@ -1139,7 +1139,7 @@ Object AmericaVehicleHellfireDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = AirplaneArmor
+    Armor           = DroneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 500
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
@@ -843,7 +843,7 @@ Object AmericaVehicleBattleDrone
   End
   ArmorSet
     Conditions        = None
-    Armor             = TankArmor
+    Armor             = AirplaneArmor
     DamageFX          = SmallTankDamageFX
   End
 
@@ -872,8 +872,8 @@ Object AmericaVehicleBattleDrone
   RadarPriority       = UNIT
   KindOf = PRELOAD CAN_CAST_REFLECTIONS VEHICLE SELECTABLE CAN_ATTACK DRONE NO_SELECT
   Body = ActiveBody ModuleTag_03
-    MaxHealth       = 100.0
-    InitialHealth   = 100.0
+    MaxHealth       = 200.0
+    InitialHealth   = 200.0
   End
 
   Behavior = AIUpdateInterface ModuleTag_04
@@ -998,7 +998,7 @@ Object AmericaVehicleScoutDrone
   TransportSlotCount = 0                 ;how many "slots" we take in a transport (0 == not transportable)
   ArmorSet
     Conditions      = None
-    Armor           = TankArmor
+    Armor           = AirplaneArmor
     DamageFX        = SmallTankDamageFX
   End
   BuildCost       = 100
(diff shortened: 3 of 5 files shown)


xezon:
On the table for "number of shots to destroy a drone" you can also note the percentage of change. Quad can now shoot down drone 2.35 times quicker than before. Are we confident that such massive Nerf/Buff is okay for starters?

Note also, you do not need to use `AirplaneArmor`. You can create a `VehicleDroneArmor` right from the get go.

ExiLe-ZH:
could a new armor fix the weirdness like jet missiles?

commy2:
> the weirdness like jet missiles

Elaborate.

ExiLe-ZH:
Stubbjax explained the issue already.

xezon:
Is this change a step one for a step two? If yes what will step two look like and why is it decoupled from this change?

I am asking, because while I do agree that vehicle drone armor is odd to be as strong as a tank, I am not sold on the presented setup. They assume that `AirplaneArmor` + 200 health are a good fit because the Spy Drone has this setup. But the Spy Drone is not a battle unit, so it is unclear to me how Spy Drone is referenced as the main justification for this specific setup.

Spy Drone is stealth. Vehicle drones are not.

Stubbjax:
> On the table for "number of shots to destroy a drone" you can also note the percentage of change. Quad can now shoot down drone 2.35 times quicker than before. Are we confident that such massive Nerf/Buff is okay for starters?

When you really consider just how imbalanced / broken the original values are, it is more than reasonable. The primary question I asked myself while contemplating the changes was whether a drone should take longer to kill than a King Raptor or Humvee as they do in 1.04, which is completely nonsensical and no longer the case with this change. Drones would ideally die twice as fast or even faster (and not have the ×2 hit points), but adjusting it just enough so that they die only slightly faster than the aforementioned units seems like a reasonable compromise and (if necessary) starting point.

A drone will now die slightly faster than a King Raptor, as demonstrated in the **shots to kill** table below.
| | Drone | King Raptor | Difference
-- | -- | -- | --
Quad (1.04) | 80 | 40 | 50% ↓
Quad (this) | 34 | 40 | 18% ↑
Gattling (1.04) | 34 | 17 | 50% ↓
Gattling (this) | 14 | 17 | 21% ↑

> Is this change a step one for a step two? If yes what will step two look like and why is it decoupled from this change?

I was contemplating adding a new armour type as a second step because it would affect the Spy Drone as well. I was wary of including it in the same PR after what happened with #1412. A new armour type also has far greater design considerations.

> They assume that `AirplaneArmor` + 200 health are a good fit because the Spy Drone has this setup. But the Spy Drone is not a battle unit, so it is unclear to me how Spy Drone is referenced as the main justification for this specific setup.

Is that such an unreasonable assumption to make? Using `AirplaneArmor` made sense because the numbers already exist and are very familiar, with the added bonus of removing some inconsistencies by aligning the values with the Spy Drone. The extra benefit of the ×2 hit points is already explained as halving the (massive) impact of healing, which is an important element to consider when Humvees are almost always accompanied by an Ambulance. A `HEALING` modifier of 50% might make more sense with a dedicated armour type.

Nevertheless, this was originally a test branch just to see how the standardised values would feel (which do feel much better). The more I thought about it, the more it made sense, and so opened a PR for visibility and discussion on the current setup. If you have a more reasonable suggestion and/or justification, then by all means.

xezon:
Thank you for detailed response. Before going more in detail, could you add a few more target units to the "shots to kill" table if you have the time? It would be useful to see the original setup for the following:

* Chinook
* Comanche
* Helix
* Raptor
* Mig
* Humvee
* Avenger
* Tomahawk
* Crusader

Stubbjax:
Here you go, ordered from least shots to most shots to kill:

image

ExiLe-ZH:
I think you very nicely proved two things Stubbjax:

1. Original setup makes zero sense. It might even be the root of all balance problems and you made me see that now. I never noticed before how much armor they had, but it's ridiculous that they have more resistance to quad/gat rounds than the common host vehicle.

2. The new setup makes perfectly sense and is still very generous, they still take more shots than 10-20x more expensive units like raps, migs and aurora's. Scout and spy drones are exact same unit, the only difference is the stealth property (spy drone would be useless without).

Ofcourse it still needs testing, but I predict spamming them as pure cannon fodder isn't nearly as viable as before and it shouldn't be. Spy drones will still be viable in limited quantities for their intended purpose.

One more thing, keep in mind the armor can still upgraded from the Strategy Center.

xezon:
Can you also show in first post how this change affects `Upgrade_AmericaDroneArmor`? I suspect ratio remains same?

Stubbjax:
> Can you also show in first post how this change affects `Upgrade_AmericaDroneArmor`? I suspect ratio remains same?

| Drone | 1.04 | Upgrade | This | Upgrade
-- | -- | -- | -- | --
Battle Drone | 100 → 150 | +50 (50%) | 200 → 250 | +50 (25%)
Hellfire Drone | 100 → 125 | +25 (25%) | 200 → 250 | +50 (25%)
Scout Drone | 100 → 125 | +25 (25%) | 200 → 250 | +50 (25%)

xezon:
Drone Armor, Chain Guns and AP Bullets would cancel each other out then I expect.

I found the above videos and numbers are a bit deceiving when observing the total amount of time and shots to kill a drone, because standard Quad Cannon has low fire rate, but can in practice increase its fire rate and or damage on scrap, veteran or other modifiers. And the China Gattling Cannon needs to spin up fully. In optimal circumstances, Quad Cannon and Gattling Tank will perform similarly.

| Object                  | Shots to kill | Optimal fire rate | Frames to kill | Seconds to kill |
|-------------------------|---------------|-------------------|----------------|-----------------|
| Quad vs Drone (orig)    | 80            | 2                 | 160            | 5.3             |
| Quad vs Drone (this)    | 34            | 2                 | 68             | 2.3             |
| Gattling vs Drone (orig)| 34            | 4                 | 136            | 4.5             |
| Gattling vs Drone (this)| 14            | 4                 | 56             | 1.8             |

Drones are not Airplanes. The way drones approach enemy targets is not comparable with a plane. Planes will approach at high velocity, unload missiles, and then turn around to reload at home. Naturally the exposure to enemy fire is minimal with such an attack. A drone does not suffer from a return-home-to-reload penalty, but it will typically not engage targets as swiftly as a plane. The drone hovers around a slower host vehicle, so it is bound to its limitations in some capacity. Consequently enemy forces can shoot on vehicle drones much longer than on air planes.

Drones are not vehicles either. They are not trainable and will therefore get no perks. They cannot be shot at by tanks, but they can be shot at by Anti Air guns from much bigger distances than ground units. This means drones will, like air planes, draw fire when engaging ground targets.

The drones come at different prices and serve different functions. It is very reasonable to say that a Scout Drone for 100 cost should not be able to tank a lot of damage from enemy forces. But a Hellfire Drone for 500 cost naturally must withstand much more enemy fire to be useful. If it dies too quickly in situations where it was originally expected to be useful, then it may become a risky investment with this patch and cause major frustration.

| Object         | Price | Purpose        | Preferred Location | Does need Armor? |
|----------------|-------|----------------|--------------------|------------------|
| Scout Drone    | 100   | Scout          | near host vehicle  | Little           |
| Battle Drone   | 300   | Repair, Defense| near host vehicle  | Some             |
| Hellfire Drone | 500   | Attack         | near enemy forces  | Much             |

I do think Drone Armor setup needs more nuance to account for the different drone types.

xezon:
Hmm something does not add up with the numbers. 5 seconds to kill per drone would make it 15 seconds, but video shows 35 seconds. There is error somewhere in the numbers.

Stubbjax:
> Drone Armor, Chain Guns and AP Bullets would cancel each other out then I expect.
> 
> I found the above videos and numbers are a bit deceiving when observing the total amount of time and shots to kill a drone, because standard Quad Cannon has low fire rate, but can in practice increase its fire rate and or damage on scrap, veteran or other modifiers. And the China Gattling Cannon needs to spin up fully. In optimal circumstances, Quad Cannon and Gattling Tank will perform similarly.

I don't understand or agree with this sentiment at all. The data has been presented in the least deceiving way it can be. Default attributes - before any further upgrades, ranks, propaganda, battle plans, bonuses, laser targeting or salvage - should always serve as the base comparison. It is far easier to extrapolate circumstantial values and modifiers in a single direction. I also don't see why you've listed 2 as the Quad Cannon's 'optimal' fire rate when it fires every 4 frames like the Gattling Tank. The Quad Cannon values you've listed are much more misleading.

> The drones come at different prices and serve different functions. It is very reasonable to say that a Scout Drone for 100 cost should not be able to tank a lot of damage from enemy forces. But a Hellfire Drone for 500 cost naturally must withstand much more enemy fire to be useful. If it dies too quickly in situations where it was originally expected to be useful, then it may become a risky investment with this patch and cause major frustration.

It's just as reasonable to say that a dedicated anti-air unit should be super effective against a small, relatively cheap, airborne utility drone. Nobody should be building Hellfire Drones against anti-air anyway - you'll typically see them relatively early in a game against Technicals, Troop Crawlers, sneaky Dragons and Chinook drops or to harass Dozers and Supply Trucks; situations where it makes sense and is a viable investment. It seems reasonable to say that this change does not really impact the utility of the Hellfire Drone. It's already an incredibly risky investment in 1.04 when any type of missile weapons are involved, which is somewhat lessened with this change in particular (5 infantry missiles vs 3). It may even technically be an overall improvement, as Hellfire Drones are often involved in taking out missile infantry, especially early on against a Chinook drop, Tunnel, or Tunnel Defenders dropped from a Technical.

Anyway, if you think the armour should differ between the different drones, I'd love to see some numbers.

ExiLe-ZH:
I agree with the differentiation between drones xezon.

The Scout Drone changes are a good starting point, but the Battle and Hellfire Drones should be a bit more durable in battle because they're designed for it. 

They shouldn't be more durable than actual battle units however. Taking down a swarm of Drones should't take longer than taking down a swarm of Comanches (with Counter measures). This doesn't make any sense.

Stubbjax:
Should the Spy Drone match the Scout Drone?

Stubbjax:
I mean, we could take some health away from the Scout / Spy Drone (200 → 160) to make the other two distinct? But then that'd be an even greater change, which I was trying to avoid. Adding health (200 → 220) to the Hellfire and Battle Drone and widening the gap with Raptors / MiGs just feels wrong.

image

ExiLe-ZH:
> image

I'd like to see a good overview of RPGs/Stingers vs Drones aswell.

xezon:
> Hmm something does not add up with the numbers. 5 seconds to kill per drone would make it 15 seconds, but video shows 35 seconds. There is error somewhere in the numbers.

Ah. My own table had me confused. Base Quad has 1 shot per 4 frames so it will take 10.6 seconds to kill 1 drone. This is very long time, especially against Scout Drone. We agree on that. As for misleading numbers: It is wrong to assume that Quad will only ever take 10.6 seconds to kill a drone. There are far more states where Quad will take closer to 5.3 seconds than 10.6 seconds. That is, by gaining veterancy, driving over scrap, getting AP Bullets, getting modifiers such as Propaganda (Team Games).

> I mean, we could take some health away from the Scout / Spy Drone (200 → 160) to make the other two distinct? But then that'd be an even greater change, which I was trying to avoid. Adding health (200 → 220) to the Hellfire and Battle Drone and widening the gap with Raptors / MiGs just feels wrong.

To me that looks like it goes into the right direction. Again, Vehicle Drones are not vehicles, nor planes. It is not reasonable to treat them as such, because they are different units. One must check how they perform and see how strong they should be. This change highlights the abundant armor of the original drones and reduces strength very much. I take issue with that, because we make China stronger through many other avenues already, so USA will already have more trouble as is. Plus GLA is very good against USA in late game as is, so USA by no means has an easy time even with strong drones.

Yes armor against missiles is increased a bit, but armor against bullets is decreased much more and the healing rate is decreased as well. I do actually like the redistribution of bullet armor to missile armor. The original missile armor is too little.

Stubbjax:
> I'd like to see a good overview of RPGs/Stingers vs Drones aswell.

Your typical infantry missile deals 40 `INFANTRY_MISSILE` damage. This deals 100% damage to `TankArmor` and 120% damage against `AirplaneArmor`. The guys in the Stinger Site deal 20 `EXPLOSION` damage. This damage is uncommonly modified, and in this case deals 100% to each.

A 200hp drone with `AirplaneArmor` vs a 40 damage infantry missile = 200 / (40 * 1.2) = 4.1 = 5 shots.
A 100hp drone with `TankArmor` vs a 40 damage infantry missile = 100 / (40 * 1.0) = 2.5 = 3 shots.

A 200hp drone with `AirplaneArmor` vs a 20 damage stinger missile = 200 / (20 * 1.0) = 10.0 = 10 shots.
A 100hp drone with `TankArmor` vs a 20 damage stinger missile = 100 / (20 * 1.0) = 5.0 = 5 shots.

So yeah, like I said at the bottom of the description, some unspecified damage types effectively deal half damage now. But we can rectify that by introducing a dedicated armour type.

Stubbjax:
> It is wrong to assume that Quad will only ever take 10.6 seconds to kill a drone. There are far more states where Quad will take closer to 5.3 seconds than 10.6 seconds. That is, by gaining veterancy, driving over scrap, getting AP Bullets, getting modifiers such as Propaganda (Team Games).

That was never an assumption. The assumption is that the base numbers / default state are enough to extrapolate to other circumstantial modifiers. Otherwise it would be tedious and time-consuming to list out every combination in every which way and only make things more complicated and harder to read.

> I take issue with that, because we make China stronger through many other avenues already, so USA will already have more trouble as is. Plus GLA is very good against USA in late game as is, so USA by no means has an easy time even with strong drones.

So what do you suggest?

ExiLe-ZH:
@xezon 

> Again, Vehicle Drones are not vehicles, nor planes. It is not reasonable to treat them as such, because they are different units. One must check how they perform and see how strong they should be.

I'm not sure what you're saying with this, are you saying their resistance vs anti-air weapons can't be compared to other air units or the host vehicles?

xezon:
> I'm not sure what you're saying with this, are you saying their resistance vs anti-air weapons can't be compared to other air units or the host vehicles?

It cannot be compared to both. It is not a tank, it is not a plane. It is a slave drone with very unique behavior.

I agree with overall direction of this change, but would like to see less nerf.

Nerf:

1. Decrease healing rate by about 50%
2. Decrease Battle Drone armor bonus from 50% to 25%
3. Decrease Scout Drone Bullet Armor by 66%
4. Decrease Battle Drone Bullet Armor by 58%
5. Decrease Hellfire Drone Bullet Armor by 50%

Buff:

1. Increase Battle Drone Missile Armor by 50%
2. Increase Hellfire Drone Missile Armor by 100%

With this change, Drones are overall worse. Reduced healing means all drones are likely to die quicker.

Scout Drone dies significantly quicker against bullets.

Battle Drone is no longer as good when upgraded. It has better resistance against bullets and missiles than Scout Drone.

Hellfire Drone can tank still as much as before, but the armor capabilities have been redistributed: Bullet armor is halved, missile armor is doubled. This gives Hellfire drone more balanced armor attributes. It has better armor than Battle Drone.

xezon:
Using AirplaneArmor, the values would then be

Scout Drone Health 160
Battle Drone Health 200
Hellfire Drone Health 240

That should come close to proposed setup. Correct me if wrong.

Stubbjax:
> Using AirplaneArmor, the values would then be
> 
> Scout Drone Health 160 Battle Drone Health 200 Hellfire Drone Health 240
> 
> That should come close to proposed setup. Correct me if wrong.

Oh, are you simply suggesting health changes? I thought you meant a specific armour set for each drone with different multipliers?

xezon:
Only if really necessary. For starters, a new "VehicleDroneArmor" would be good, that borrows its setup from "AirplaneArmor" and "TankArmor".

Stubbjax:
Here's a combination that fixes the `JET_MISSILES` discrepancy and gives `EXPLOSION` (Patriot, Stinger, TOW Missile, Laser Lock, etc.) a bit more punch.

image

ExiLe-ZH:
Drones are simply air units since they fly and are (in)vurnerable to the same type of weapons as other airplanes.

Therefore I think airplane armor does make sense aslong as that doesn't limit us in achieving a balanced and logical setup.
If that can't be achieved then they should be put in a dedicated sub category.

This doesn't make the comparison to other vehicles invalid though, it's still fair to compare their anti-air resistance to their host or other vehicles. For example: Does it make sense that a Drone has more resistance to QuadCannon-fire than the Host Humvee?

MTKing4:
I think we should find a medium between the two armor types by making a new armor type with moderate numbers.

i mean right now with this change a stinger needing 10 shots just to kill a `AirplaneArmor` drone makes zero sense, just as much it made zero sense when quads needed 80 shots to kill a `TankArmor` drone

So i suggest `DroneArmor` damage ratios as follows:

| Weapon | Damage |
-- | -- |
SMALL_ARMS | 60%
GATTLING | 60%
INFANTRY_MISSILE | 100%
JET_MISSILES | 100%

Basically taking half the `AirplaneArmor` damage for quads/gats while retaining rocket damage the same from `TankArmor` for rocket weapons
That way we buff quads and gats only against drones, and not as big as the 120% buff, so that using drones wouldn't be completely useless, but still, i think the best judge here is testing.

As for health, I'm not sure, but I'm thinking 150

Stubbjax:
> I think we should find a medium between the two armor types by making a new armor type with moderate numbers.
> 
> i mean right now with this change a stinger needing 10 shots just to kill a `AirplaneArmor` drone makes zero sense, just as much it made zero sense when quads needed 80 shots to kill a `TankArmor` drone

That's why I had `EXPLOSION` on the new dedicated `DroneArmor` set to 150%, so that it would be 7 shots (which would equal 3 volleys). But you're right, it probably should be 200% to match the doubled health.

> So i suggest `DroneArmor` damage ratios as follows:
> SMALL_ARMS 	60%
> GATTLING 	60%
> INFANTRY_MISSILE 	100%
> JET_MISSILES 	100%
> 
> Basically taking half the `AirplaneArmor` damage for quads/gats while retaining rocket damage the same from `TankArmor` for rocket weapons That way we buff quads and gats only against drones, and not as big as the 120% buff, so that using drones wouldn't be completely useless, but still, i think the best judge here is testing.
> 
> As for health, I'm not sure, but I'm thinking 150

Do the drones look completely useless from the test footage in the PR description? They still take a relatively substantial amount of damage before dying.

I applied your modifiers to the damage table. Do you believe the drones should be able to absorb more hits than all those other aircraft?

image

MTKing4:
> Do the drones look completely useless from the test footage in the PR description? They still take a relatively substantial amount of damage before dying.

I can't tell from the video as it's alternating between all three drones making it hard tell how fast they die, but my point is that AA units will have much less time to spend on killing drones from a distance, which could possibly mean you will spend money on drones and die before paying off, so we might see them get used less often if they become to easy to kill


> I applied your modifiers to the damage table. Do you believe the drones should be able to absorb more hits than all those other aircraft?

`50` and `21` is a lot better than `80` and `34` for a start
If not enough we can dial it further, that's why i think we need testing, eyeballing numbers can only do so much

We could start by testing your PR as well np, but I don't think it's optimal currently, at least maybe until after adjusting the rocket weapons

Stubbjax:
> We could start by testing your PR as well np, but I don't think it's optimal currently, at least maybe until after adjusting the rocket weapons

I've added and applied the dedicated `DroneArmor` definition as described [here](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1463#issuecomment-1312793390) and adjusted the `EXPLOSION` and `JET_MISSILES` modifiers to 200% for your testing pleasure.

MTKing4:
> I've added and applied the dedicated DroneArmor definition as described here and adjusted the EXPLOSION and JET_MISSILES modifiers to 200% for your testing pleasure.

Great! Ty
So that means these values now give the same result as if it was unchanged, currect?

And what about INFANTRY_MISSILE?

Stubbjax:
> > I've added and applied the dedicated DroneArmor definition as described here and adjusted the EXPLOSION and JET_MISSILES modifiers to 200% for your testing pleasure.
> 
> Great! Ty So that means these values now makes give the same result as if it was unchanged, currect?
> 
> And what about INFANTRY_MISSILE?

That is correct.

A 200hp drone with `DroneArmor` vs a 20 damage stinger missile = 200 / (20 * 200%) = 5.0 = 5 shots.
A 100hp drone with `TankArmor` vs a 20 damage stinger missile = 100 / (20 * 100%) = 5.0 = 5 shots.

I added missiles to the table (`INFANTRY_MISSILE` = 120%):

image

MTKing4:
Ok but infantry missiles used to take 3 shots, and now is 5, Why do we need this changed?

Stubbjax:
> Ok but infantry missiles used to take 3 shots, and now is 5, Why do we need this changed?

To stop this kinda nonsense!

https://user-images.githubusercontent.com/11547761/201653736-2fd30756-6bee-44ea-94c8-33967360c1e4.mp4

But really, it was more of a coincidental change of matching the armour with the Spy Drone's `AirplaneArmor`. And the more I thought about it, the more it felt like a nice way to offset some of the lost resistance to bullets and lessen the negative impact of the change.

MTKing4:
> To stop this kinda nonsense!

Fair Enough, i think it's reasonable to give a little buff to drones vs rockets and a little nerf vs quads/gats, because both are bad extremes, as to jets, I'm not sure, but i think that a full jet volly should be able to kill a drone, makes no sense why it needs two, this goes to Raptors, KRaptors and MiGs

> But really, it was more of a coincidental change of matching the armour with the Spy Drone's `AirplaneArmor`.

I'm aware of that, but it needed addressing as to why should it be that way

Stubbjax:
> Fair Enough, i think it's reasonable to give a little buff to drones vs rockets and a little nerf vs quads/gats, because both are bad extremes, as to jets, I'm not sure, but i think that a full jet volly should be able to kill a drone, makes no sense why it needs two, this goes to Raptors, KRaptors and MiGs

How little are we talking here? Where would you place them in relation to other vehicles / aircraft? It's tough because giving drones higher durability than your average fighter jet makes absolutely no sense; but at the same time, any values in the same range or lower will likely have a much greater impact.

Here's an alternative with 100% modifiers for `GATTLING`, `SMALL_ARMS` and `INFANTRY_MISSILE`, and each drone having hit points in relation to its price (a separation of a 40 damage infantry missile).

image

MTKing4:
I do agree with Xezon that they can't be compared to jets and vehicles, they aren't controllable and usually just stay under enemy fire for longer unlike aircrafts where they just hit and run, and vehicles don't get targetted with AA weapons so can't be compared either, that's why i believe nerfing them too much might not be the best solution here

That said, i do wanna test that last table and how it fares ingame as I don't have a clue on how much worse they are now

Stubbjax:
> I do agree with Xezon that they can't be compared to jets and vehicles, they aren't controllable and usually just stay under enemy fire for longer unlike aircrafts where they just hit and run, and vehicles don't get targetted with AA weapons so can't be compared either, that's why i believe nerfing them too much might not be the best solution here

This makes no sense to me. They absolutely can be compared. Drones are airborne units just like any other aircraft. Why should their autonomy and battle proximity have any bearing on their damage resistance? This is not an intuitive relationship at all. Players have no reason to expect a small utility drone to absorb more anti-air rounds than a dedicated fighter jet or attack helicopter. The fact that players are well-accustomed to this flawed design is what makes this so tricky to solve.

Feel free to propose a better solution!

> That said, i do wanna test that last table and how it fares ingame as I don't have a clue on how much worse they are now

I have pushed a commit with the [adjusted values](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1463#issuecomment-1313687626). You can check out the prior commit if you want to test [the previous values](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1463#issuecomment-1313494877). I'm always happy to test!

xezon:
> This makes no sense to me. They absolutely can be compared. Drones are airborne units just like any other aircraft. Why should their autonomy and battle proximity have any bearing on their damage resistance?

They can be compared because they are faction units. But Drones are not Airplanes, nor Ground Vehicles. They are Drones. They have a different name, behaviour, utility, control, spawn method, size, damage, armor. Using an Aurora Bomber is vastly different from using a Scout Drone. They are different kind of units. Therefore why should their armor be the same if they are different things?

Stubbjax:
> > This makes no sense to me. They absolutely can be compared. Drones are airborne units just like any other aircraft. Why should their autonomy and battle proximity have any bearing on their damage resistance?
> 
> They can be compared because they are faction units. But Drones are not Airplanes, nor Ground Vehicles. They are Drones. They have a different name, behaviour, utility, control, spawn method, size, damage, armor. Using an Aurora Bomber is vastly different from using a Scout Drone. They are different kind of units. Therefore why should their armor be the same if they are different things?

You are misrepresenting my point. All units obviously have a different name, behaviour, utility, movement, size, damage, health, etc. Autonomy and battle proximity are not attributes that invalidate the comparison, nor should they play any major role in a unit's damage resistance.

xezon:
Would you say that a slow melee combat unit can do with just as much damage resistence as an agile long range combat unit? And would you say drones and airplanes are the same in terms of combat range and agility?

MTKing4:
> This makes no sense to me. They absolutely can be compared. Drones are airborne units just like any other aircraft. Why should their autonomy and battle proximity have any bearing on their damage resistance? This is not an intuitive relationship at all. Players have no reason to expect a small utility drone to absorb more anti-air rounds than a dedicated fighter jet or attack helicopter. The fact that players are well-accustomed to this flawed design is what makes this so tricky to solve.

It makes sense from gameplay prospective, if you're thinking about it logically, Realistically and by comparing to units with different roles then of course you're going to have troubles accepting it

> Feel free to propose a better solution!

The solution is to adjust the values to make them not frustrating to kill while not eliminating their viability by being too weak and not worth the investment. That is still undecided because we don't know which values are optimal yet.

> I have pushed a commit with the adjusted values. You can check out the prior commit if you want to test the previous values. I'm always happy to test!

Will do, although I'm not sure how to get the files from the older commits

Stubbjax:
> Would you say that a slow melee combat unit can do with just as much damage resistence as an agile long range combat unit?

What sane army would place their bowmen on the front-line? A better question to ask: Why would a player expect a utility drone to have more hp / armour than a fighter jet / attack helicopter?

> It makes sense from gameplay prospective, if you're thinking about it logically, Realistically and by comparing to units with different roles then of course you're going to have troubles accepting it

It really doesn't, though. It's completely counterintuitive.

> The solution is to adjust the values to make them not frustrating to kill while not eliminating their viability by being too weak and not worth the investment. That is still undecided because we don't know which values are optimal yet.

Which has been the intent of this change. It feels far more reasonable, with no eliminated viability, from my own tests.

> Will do, although I'm not sure how to get the files from the older commits

You can checkout individual commits.

image

xezon:
> A better question to ask: Why would a player expect a utility drone to have more hp / armour than a fighter jet / attack helicopter?

Because due its decreased agility and combat range it will naturally receive more damage than an airplane. You can compare this with Crusader Tank vs Tomahawk Tank. Crusader is at the front. Takes damage. Tomahawk stays back, can shoot far away. Crusader Tank has more armor than Tomahawk. Reversing or equalling the armors on both these tanks would not be justified, because their attack ranges differ.

Stubbjax:
> Because due its decreased agility and combat range it will naturally receive more damage than an airplane. You can compare this with Crusader Tank vs Tomahawk Tank. Crusader is at the front. Takes damage. Tomahawk stays back, can shoot far away. Crusader Tank has more armor than Tomahawk. Reversing or equalling the armors on both these tanks would not be justified, because their attack ranges differ.

So your argument is that close-combat units should generally have more health?

ExiLe-ZH:
I predict that, when we're going to test Stubbjax's changes that even at those lower resistance levels, that mass HellfireVees will still totally dominate mass Quads/Gats, the difference however is that the Quads/Gats will atleast be able to trade a good number of drones for it and earn some XP.

ExiLe-ZH:
> > A better question to ask: Why would a player expect a utility drone to have more hp / armour than a fighter jet / attack helicopter?
> 
> Because due its decreased agility and combat range it will naturally receive more damage than an airplane. You can compare this with Crusader Tank vs Tomahawk Tank. Crusader is at the front. Takes damage. Tomahawk stays back, can shoot far away. Crusader Tank has more armor than Tomahawk. Reversing or equalling the armors on both these tanks would not be justified, because their attack ranges differ.

This logic is correct, but it doesn't prove Drones need a mega ton of armor just because of the combat range or lack of agility. The thing is that these Drones only have a very few counters in the field and the majority can't even target the Drones at all. The units that are left (Gats/Quads) can't even take them out efficiently either! This is outrageous broken design tbh.

xezon:
> The units that are left (Gats/Quads) can't even take them out efficiently either! This is outrageous broken design tbh.

Gattlings, Quads, Rocket men, Tow-Missile Humvees are core faction forces. Currently it does take a lot of firing power, but is the nerf of -58% bullet armor justified? If yes, why by this much exactly? And why all drones equally as much? I do agree with reduction in armor, but not with how much as presented by the first iteration of this change. It is just too much for a first step. We have questioned other changes for much much less impact than this one.

Stubbjax:
> Currently it does take a lot of firing power, but is the nerf of -58% bullet armor justified? If yes, why by this much exactly?

The answers to this question should exist in the original description. Future iterations and further discussion / contemplation has resulted in a more thorough understanding of just how broken the values of 1.04 are, and why the values make more sense. It feels very unreasonable to give drones more hp / armour than a King Raptor or Humvee. Either way, it serves as a good starting point for testing.

> And why all drones equally as much? I do agree with reduction in armor, but not with how much as presented by the first iteration of this change. It is just too much for a first step.

All drones were originally changed equally as much because all drones in 1.04 have the same hp / armour. Regardless, [this](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1463#issuecomment-1313687626) is the latest iteration after your suggestion to distribute different health based on the unit cost / value.

ExiLe-ZH:
> > The units that are left (Gats/Quads) can't even take them out efficiently either! This is outrageous broken design tbh.
> 
> 
> 
> Gattlings, Quads, Rocket men, Tow-Missile Humvees are core faction forces. Currently it does take a lot of firing power, but is the nerf of -58% bullet armor justified? If yes, why by this much exactly? And why all drones equally as much? I do agree with reduction in armor, but not with how much as presented by the first iteration of this change. It is just too much for a first step. We have questioned other changes for much much less impact than this one.

Percentage before/after is irrelevant really, it's whatever values are needed to make it more balanced. If that proves to be alot then it only shows how broken it was before.

Stubbjax already brought this up; if you want to fight anti-air units, for what reason would you mass invest in airborne units? If you want an edge in tank vs tank combat, or vs light vehicles (vees vs techs) it makes total sense. But empty Hellvees shouldn't dominate Gats/Quads like they currently do.

Anyways yes, I believe it makes sense that Hellfire Drones are more durable than Scout Drones in battle. I think the earlier proposed 160, 200, 240 values are a good start. If that's too little, lets crank it a bit. That's not what I predict however https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1463#issuecomment-1315160802.

xezon:
Here my feedback on setup. Green ok, red not so ok.

Ok:
* Increased drone health makes drones repair slower. Hellfire drone has biggest penalty with halved repair time. I would not mind if health values were even higher.
* Hellfire Drone has halved bullet armor
* Battle Drone has sufficient Gattling armor
* Scout Drone has sufficient Missile armor (same as original)

Not so ok:
* Hellfire Drone has halved bullet armor, but not doubled missile armor to compensate nerf
* Battle Drone has less Quad armor than expected. Also has less Missile armor than expected
* Scout Drone has overall less armor than expected.
* Spy Drone does not need to be changed. It is a General Promotion. It cannot be spammed. It does not need to be as weak as Scout drone. Spy Drone visually is larger than Scout Drone.

drones3534764572146

Stubbjax:
So your layout of good vs bad indicates that it is impossible to align the values without each drone having its own armour type. Is this what you would prefer? There is no way to keep those health values while satisfying the other points otherwise. The acceptability of the Battle Drone's Quad shots vs Gattling shots can't be different because the damage modifier of `SMALL_ARMS` applies to both.

Also, the Spy Drone is actually identical in size to the Scout Drone. The preferred height might give the illusion that they are different.

xezon:
I began splitting this change into smaller ones. This way we can hopefully reach good setup more effortlessly.

* #1469
* #1470

Stubbjax:
I think proceeding incrementally will just complicate things further as each change won't be a direct comparison to 1.04 but also to all preceding changes. It really needs to be tested and compared / contemplated as a composite change. Fixing the Battle Drone armour upgrade inconsistency does make sense as a standalone change, but it has little to no bearing on the complexity of this change.

Besides that, the entire contents of the healing change are already superseded by this PR and any future health changes, and will possibly be replaced with a `HEALING X%` modifier in the drone armour definition for simplicity once the other attributes are locked down. It's also not guaranteed that we want a healing reduction when the other values are still up in the air. It may not even be necessary to reduce healing if we go with lower hit points, as it may be a good way to offset a lower durability in relation to other aircraft / vehicles. Additionally, the doubled hit points and damage multipliers are not the optimal approach (values should be normalised to 100%) and were just the result of standardising the values with the Spy Drone.

xezon:
It means we can focus at one thing at a time and lock in a particular change step.

MTKing4:
> You can checkout individual commits.

i use GitHub Desktop, don't know if that's possible there, will have to try

also i've been been busy the past two weeks, i might not be available for testing as much for the time being, but will try and catch up with the latest changes

PR #1484, opened 2022-11-26
Fix: zhca_aihero, zhca_aihero2 textures

Stubbjax:
Removed the unsightly house colour from Colonel Burton's ears.

image

### TODO

- [ ] Fix uv mapping


ExiLe-ZH:
I think army tags aren't necesarry when it's not balance related

Stubbjax:
> I think army tags aren't necesarry when it's not balance related

I did consider this and thought it could be useful if someone wanted to e.g. search for all USA art adjustments. Feel free to remove it.

xezon:
I must have missed the alpha channel part.

MTKing4:
I thought we agreed that that was his AirPods?

xezon:
I noticed some body parts did not blend so well on Burton. I made another pass on texture and improved face to blend better with ears and arms blend better with shoulders.

shot_20221127_014409_1

He is still not a perfect looker, but from further away looks good.

shot_20221127_014416_2

xezon (approved):

Stubbjax:
Yeah, I would have fixed more but the UV is a total mess and it would make more sense to fix that first.

image

xezon:
Yeah looks quite wild

PR #1489, opened 2022-11-30
Change: Added scorch effects to nuclear explosions

Stubbjax:
Added scorch effects to the following:

- China Power Plant explosion
- Nuke Cannon shell explosion
- Nuke Carpet Bomber explosion
- Nuclear Missile Silo explosion

https://user-images.githubusercontent.com/11547761/204748812-ddce94db-440d-4132-badc-9d8369f7e023.mp4

Before (1.04) vs after (this):

image

### TODO

- [x] Remove scorch from FX_ChinaVehicleNukeCannonDeathExplosion (Added to FX_ChinaVehicleNukeCannonDeathExplosionInitial)

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -5153,6 +5153,11 @@ FXList FX_ChinaPowerPlantDeath
     Type = SEVERE
   End
 
+  TerrainScorch ; Patch104p @tweak Stubbjax 30/11/2022 Added scorch effect.
+    Type = RANDOM
+    Radius = 60
+  End
+
   Sound
     Name = ExplosionMiniNuke
   End
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -4927,6 +4927,11 @@ FXList FX_NukeGLA
     Type = SEVERE
   End
 
+  TerrainScorch ; Patch104p @tweak Stubbjax 30/11/2022 Added scorch effect.
+    Type = RANDOM
+    Radius = 100
+  End
+
   Sound
     Name = ExplosionMiniNuke
   End
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -8662,6 +8662,11 @@ FXList Nuke_WeaponFX_NukeCannon
     Type = SEVERE
   End
 
+  TerrainScorch ; Patch104p @tweak Stubbjax 01/12/2022 Added scorch effect.
+    Type = RANDOM
+    Radius = 40
+  End
+
   Sound
     Name = ExplosionMiniNuke
   End
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -8838,7 +8838,7 @@ FXList Nuke_WeaponFX_NapalmMissileDetonation
   End
   TerrainScorch
     Type = RANDOM
-    Radius = 15
+    Radius = 40
   End
   LightPulse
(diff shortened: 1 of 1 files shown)


xezon:
The Nuke Cannon blast scorch allows to paint big scorches on terrain. Do we want this?

shot_20221130_223518_4

Other than that, I noticed that the death scorch for the Nuke Cannon is a bit small. After all, it does die with a bigger nuclear explosion.

shot_20221130_223654_5

And, the Nuke Battlemaster death leaves 2 small scorches on kill for some reason. It would be better if it just left 1 with similar size as for Nuke Cannon.

shot_20221130_223439_3

Stubbjax:
> The Nuke Cannon blast scorch allows to paint big scorches on terrain. Do we want this?

I mean, why not? The Firebase leaves scorches and you don't see anyone painting with those. It could actually be handy and help a player gauge the maximum / minimum firing range.

> Other than that, I noticed that the death scorch for the Nuke Cannon is a bit small. After all, it does die with a bigger nuclear explosion.

~Yeah, I think this would be better addressed as a separate change, because I believe it should also deal damage in a similar way to the Nuclear Battlemaster when it explodes (many players still think it actually does damage).~

> And, the Nuke Battlemaster death leaves 2 small scorches on kill for some reason. It would be better if it just left 1 with similar size as for Nuke Cannon.

I did briefly look at this and thought it was some standard logic with every unit, as all vehicles seem to leave a scorch mark upon destruction, and I didn't like the idea of adding another scorch on top. But if the nuclear explosion is already creating another scorch, then that can be adjusted to suit.

I've pushed a change that adds a scorch to the Nuke Cannon's death and a larger scorch for the Nuclear Battlemaster.

image

xezon:
> I mean, why not? The Firebase leaves scorches and you don't see anyone painting with those. It could actually be handy and help a player gauge the maximum / minimum firing range.

I suggest to make the scorch smaller then. It looks quite big and blurry.

ExiLe-ZH:
Yes should leave marks, but bigger for both the BM and warhead.

xezon:
Nuke Cannon death also spawns 2 scorches. Can be seen here. Notice the smaller scorch in the bigger one.

shot_20221201_201602_1

xezon:
I would say, reduce WeaponFX_NukeCannon, Nuke_WeaponFX_CarpetBomb from 60 to 40. That should be plenty.

xezon:
Regarding the 2 scorches at different position of Nuke Battlemaster, I found it is caused by mispositioned Nuklear Explosion. I fixed it with #1493. But we still should remove the double scorch if possible.

MTKing4:
I think we should keep in mind that there's a limit to how many scorches can exist at a time, so having to many things causing scorches we might see some goofy stuff sometimes like a big nuke scorch disappears right after killing a powerplant nearby

Also particle cannons use up a lot of scorch marks as well, so that issue might become even more apparent

xezon:
A comment in INI:
```
    ;Each scorch mark creates an object in the world, so while making more looks better
    ;it becomes a performance issue. The scalar will calculate the size of the scorch
    ;mark to make based on the current size of the beam (which changes dynamically). The
    ;number also matches the number of times the "GroundHitFX" is called.
```

commy2:
Nuclear explosions already leave radiation fields. I think those make big craters superfluous.

xezon:
Scorches also stack on top of each other. I think we need to be considerate about performance.

commy2:
What explosions etc. trigger scorch marks in the base game?

Stubbjax:
I don't think performance is a concern here at all, considering destroyed vehicles make up the overwhelming majority (I'd estimate >95%) of scorch marks. That is often hundreds of scorches in competitive 3v3s and even thousands in casual spam games.

xezon:
I spent several hours with this today. I fine tuned the terrain scorch sizes for nuke explosions. They are smaller now. This way they are not such a big looker and do not spam the terrain as much.

Also added missing scorches to regular Carpet Bombs. Artillery already has fitting scorches.

Also added missing scorches to Mig Nuke Missiles + Helix Nuke Bomb.

xezon (approved):
Looks good to me.

PR #1490, opened 2022-11-30
Change: Improved death explosion of Nuclear Missile Silo

Stubbjax:
Improved the explosion / death of the Nuclear Missile Silo. More damage; greater area.

## 1.04:
The Nuclear Missile Silo's explosive destruction applies damage in a very confusing and counterintuitive way due to sharing its death weapon with the Nuclear Reactor while having a much larger profile. It is extremely rare that a unit will take full damage from the explosion, as demonstrated in the footage below.

https://user-images.githubusercontent.com/11547761/204782424-a03b3445-2535-4049-9652-c870027cafdc.mp4

## Patch:
The Nuclear Missile Silo now applies harsher damage over a greater area, which feels much more in line with player expectations when considering the size and intensity of the explosion effect.

https://user-images.githubusercontent.com/11547761/204783349-64b3befa-ef34-4753-b9dc-4de4ae507a83.mp4

All units can attack from a distance outside of the primary blast radius. Only Overlords and elite+ Crusaders and Paladins can survive the primary blast. Heroes, Rangers and some elite+ infantry can survive the secondary blast. Colonel Burton will run to just the right distance after deploying a bomb to escape the secondary blast unscathed.

Values are compared in the table below:

| | Nuclear Reactor | Nuclear Missile Silo | Change |
| - | - | - | - |
| Primary Damage | 400 | 600 | +50% |
| Primary Damage Radius | 50 | 75 | +50% |
| Secondary Damage | 100 | 150 | +50% |
| Secondary Damage Radius | 100 | 125 | +25% |

And here is a rough diagram of the difference:

image

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -8096,7 +8096,7 @@ Object Boss_NuclearMissileLauncher
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_12
-    DeathWeapon = ChinaPowerPlantDeathWeapon
+    DeathWeapon = ChinaNuclearMissileLauncherDeathWeapon ; @tweak Stubbjax 30/11/2022 Added dedicated + improved death weapon.
     StartsActive  = Yes
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -14959,7 +14959,7 @@ Object ChinaNuclearMissileLauncher
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_12
-    DeathWeapon = ChinaPowerPlantDeathWeapon
+    DeathWeapon = ChinaNuclearMissileLauncherDeathWeapon ; @tweak Stubbjax 30/11/2022 Added dedicated + improved death weapon.
     StartsActive  = Yes
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/InfantryGeneral.ini
@@ -6863,7 +6863,7 @@ Object Infa_ChinaNuclearMissileLauncher
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_12
-    DeathWeapon = ChinaPowerPlantDeathWeapon
+    DeathWeapon = ChinaNuclearMissileLauncherDeathWeapon ; @tweak Stubbjax 30/11/2022 Added dedicated + improved death weapon.
     StartsActive  = Yes
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
@@ -8653,7 +8653,7 @@ Object Nuke_ChinaNuclearMissileLauncher
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_12
-    DeathWeapon = ChinaPowerPlantDeathWeapon
+    DeathWeapon = ChinaNuclearMissileLauncherDeathWeapon ; @tweak Stubbjax 30/11/2022 Added dedicated + improved death weapon.
     StartsActive  = Yes
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
@@ -6642,7 +6642,7 @@ Object Tank_ChinaNuclearMissileLauncher
   End
 
   Behavior = FireWeaponWhenDeadBehavior ModuleTag_12
-    DeathWeapon = ChinaPowerPlantDeathWeapon
+    DeathWeapon = ChinaNuclearMissileLauncherDeathWeapon ; @tweak Stubbjax 30/11/2022 Added dedicated + improved death weapon.
     StartsActive  = Yes
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2914,6 +2914,25 @@ Weapon ChinaPowerPlantDeathWeapon
 End
 
 ;------------------------------------------------------------------------------
+Weapon ChinaNuclearMissileLauncherDeathWeapon
+  PrimaryDamage               = 600.0
+  PrimaryDamageRadius         = 75.0
+  SecondaryDamage             = 150.0
+  SecondaryDamageRadius       = 125.0
+  AttackRange                 = 100.0
+  DamageType                  = EXPLOSION
+  DeathType                   = EXPLODED
+  WeaponSpeed                 = 99999.0
+  FireFX                      = FX_ChinaPowerPlantDeath
+  FireOCL                     = OCL_RadiationFieldMedium
+  RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
+  DelayBetweenShots           = 0
+  ClipSize                    = 1
+  ClipReloadTime              = 0
+  AutoReloadsClip             = No
+End
+
+;------------------------------------------------------------------------------
 Weapon TomahawkMissileWeapon
   PrimaryDamage               = 150.0
   PrimaryDamageRadius         = 10.0


MTKing4:
This may be a bit too much, no other super weapon Damages(and kills) units nearby so why should the nuke?

Personally I don't think this change is necessary nor is needed

xezon:
I tested this change. The original damage radius and amount indeed makes no sense. Tanks close to destroyed Nuke Silo receive a minuscule amount of damage. The new primary damage radius makes more sense when compared to the Nuclear Reactor blast. Since Nuclear Reactor already applies this principle of nearby units getting killed, it is a familiar concept for the Nuclear Silo.

~~I would classify this change mostly as Nerf, not a Buff. Mainly because this change makes Silo repairs more difficult. A repairing Dozer would always die nearby.~~

The radius increase appears comparable to Nuclear Reactor. It did seem a bit too short in practice, but judging by the bbox numbers, the +25 radius should be ok.

```
Silo
  GeometryMajorRadius = 45.0
  GeometryMinorRadius = 55.0

Reactor
  GeometryMajorRadius = 27.0
  GeometryMinorRadius = 34.0
```

I was skeptical about the tripled secondary damage. I tested the amount on Dozer and it is approx 60% damage. That looks ok.

I did find that nearby Dozer does not always die in blast. I do not know why. Perhaps the Dozer that actively repairs the Superweapon has some sort of immunity. If so, then this change has less of a detrimental impact.

I am ok with this change, but it needs to be applied to all Nuke Silos. Currently is applied to Nuklear General only.

## Dozer & Reactor

https://user-images.githubusercontent.com/4720891/204909626-a6031abf-7b91-4db8-ad97-e05bd9b53823.mp4

## Dozers & Silo

https://user-images.githubusercontent.com/4720891/204909966-98e9abd7-a04b-4d42-871f-428d0bbe4893.mp4

Stubbjax:
> This may be a bit too much, no other super weapon Damages(and kills) units nearby so why should the nuke?

> Personally I don't think this change is necessary nor is needed

Each of the three Super Weapons is unique - that a particular behaviour is not shared between them is not a very strong justification. The structure blows up with a huge fireball explosion - it's completely reasonable to expect this to deal proper damage; especially when players are already very familiar with the explosive behaviour of the Nuclear Reactor. Does 1.04's silo explosion make sense? I suggest you try it out for yourself.

> ~I would classify this change mostly as Nerf, not a Buff. Mainly because this change makes Silo repairs more difficult. A repairing Dozer would always die nearby.~

Yeah, I thought about this and concluded that it would also require more care from opponents in avoiding collateral damage or casualties, especially in tight spaces. It ultimately raises the skill ceiling for both sides and provides another dynamic and greater avenue for players to demonstrate skill / knowledge (or lack thereof).

> I was skeptical about the tripled secondary damage. I tested the amount on Dozer and it is approx 60% damage. That looks ok.

Well it's not quite tripled, but yeah, the original secondary damage just tickles nearby units and it makes sense to increase it somewhat in relation to the blast, which looks a lot more intense / powerful than the Nuclear Reactor's.

> I did find that nearby Dozer does not always die in blast. I do not know why. Perhaps the Dozer that actively repairs the Superweapon has some sort of immunity. If so, then this change has less of a detrimental impact.

Hmm, could this be an executable bug? Is it only while repairing? The dozer that looked to be repairing the silo in your last video died, so it may be unrelated to the active repairer.

> I am ok with this change, but it needs to be applied to all Nuke Silos. Currently is applied to Nuklear General only.

Fixed!

xezon:
> Hmm, could this be an executable bug? Is it only while repairing? The dozer that looked to be repairing the silo in your last video died, so it may be unrelated to the active repairer.

It appears it is always one Dozer that survives. I suspect the one in the video that was closest the structure was actually repairing, not the one that I manually controlled. I could be wrong.

Perhaps it is deliberate that an interacting Worker unit cannot be damaged by the structure as a way to prevent unfavourable suicide scenarios.

ExiLe-ZH:
Makes sense yea.

How much does it affect nearby buildings?

Stubbjax:
> Makes sense yea.
> 
> How much does it affect nearby buildings?

Here's the damage applied to some common structures if a wise player decides to place them right next door.

https://user-images.githubusercontent.com/11547761/205083239-cf53e49c-1446-465b-80f7-2d943a22c9c1.mp4

ExiLe-ZH:
Nice, amount of damage looks fair to me.

xezon:
I tested a few more Dozer repairs + kill, and I am pretty confident to say that the repairing Dozer receives no damage. Why that is, I do not know, but I think that is good when judging the impact of this change. This way there is no penalty for trying to repair Nuke Silo.

Oddly enough however, a repairing GLA Worker and USA Dozer dies in the blast. But 1 nearby China Dozer that does not repair will not die then. This anomaly makes no sense to me. It is weird.

There is this INI comment
```
  ArmorSet
    Conditions        = None
    Armor             = DozerArmor ;Special Explosion-Proof Armor
    DamageFX          = TankDamageFX
  End
```

xezon:
Should this effect also be applied to 
* SupW_AmericaNuclearMissileLauncher
* SupW_AmericaCruiseMissileLauncher
?

MTKing4:
> Each of the three Super Weapons is unique - that a particular behaviour is not shared between them is not a very strong justification. The structure blows up with a huge fireball explosion - it's completely reasonable to expect this to deal proper damage; especially when players are already very familiar with the explosive behaviour of the Nuclear Reactor. Does 1.04's silo explosion make sense? I suggest you try it out for yourself.

The Nuke Death Explosion was set this way internationally i bet, making it pretty much cosmetic to not have an impact on the gameplay

And it's punishing really for the player killing the nuke, imagine losing 10 quads in a sneak attack while killing the nuke, how's that rewarding for the gla player? He probably lost more cash value than 5K in that scenario

And yes we can compare it to other super weapons because it's not really a unique behavior, check particle cannon death explosion or scud deaths, neither do damage (except for the toxins)

If we're really going by what makes logical sense over making sense gameplay wise then you might as well add explosion damage to each particle, scud and oil derricks, because that make logical sense, right? And not to mention that the latter has always been this way in previous and newer C&C games, being oils do explosion damage.

I really think we should stop making gameplay changes for the satisfaction of making the game more realistic and sacrificing gameplay for it.

Stubbjax:
> I tested a few more Dozer repairs + kill, and I am pretty confident to say that the repairing Dozer receives no damage. Why that is, I do not know, but I think that is good when judging the impact of this change. This way there is no penalty for trying to repair Nuke Silo.

I cannot seem to reproduce this strange behaviour. I've destroyed dozens of silos and my dozers have all died every time. Would you maybe create a map where the issue can be reproduced?

> Should this effect also be applied to
> SupW_AmericaNuclearMissileLauncher
> SupW_AmericaCruiseMissileLauncher
> ?

I saw this but wasn't sure it was worth doing because the structures are unused. I don't have a preference either way.

> The Nuke Death Explosion was set this way internationally i bet, making it pretty much cosmetic to not have an impact on the gameplay

The intent was clearly to have it deal damage upon death, or it would not be assigned the Nuclear Reactor's death weapon and have a corresponding effect. Regardless, intent takes a backseat over bugs, poor design and counterintuitive gameplay.

> And it's punishing really for the player killing the nuke, imagine losing 10 quads in a sneak attack while killing the nuke, how's that rewarding for the gla player? He probably lost more cash value than 5K in that scenario

It's not supposed to be rewarding; this is precisely the point. It's the same risk as destroying a Nuclear Reactor. It's also a risk that a skilled / knowledgeable opponent should be aware of and be able to mitigate when destroying the structure. Players already tend to do this because they expect the explosion to deal damage. Again, you should try it out for yourself.

> And yes we can compare it to other super weapons because it's not really a unique behavior, check particle cannon death explosion or scud deaths, neither do damage (except for the toxins)

It _is_ a unique behaviour. The Nuclear Reactor explodes but the Cold Fusion Reactor does not. The same logic applies here.

> If we're really going by what makes logical sense over making sense gameplay wise then you might as well add explosion damage to each particle, scud and oil derricks, because that make logical sense, right? And not to mention that the latter has always been this way in previous and newer C&C games, being oils do explosion damage.

> I really think we should stop making gameplay changes for the satisfaction of making the game more realistic and sacrificing gameplay for it.

Where is the sacrifice? This is not about realism at all. The game already employs a massive artistic license when it comes to nuclear physics. Improving gameplay and meeting player expectations are the primary motivators here.

ExiLe-ZH:
> And it's punishing really for the player killing the nuke, imagine losing 10 quads in a sneak attack while killing the nuke, how's that rewarding for the gla player? He probably lost more cash value than 5K in that scenario

It's nothing but a mistake to park your quads next to it while killing it, same with a Nuclear Reactor or any China building because of mines. Maybe it could happen once to someone who doesn't know about the change because it never been like that, but that shouldn't be a reason it doesn't do damage, logically it makes sense that it does.

commy2:
I get why you would want to increase the radius, because it doesn't work for a buildings larger than the Reactor.

But why increase the damage? Makes no sense to me.

Stubbjax:
> Maybe it could happen once to someone who doesn't know about the change because it never been like that

I actually don't think most players would even realise anything has changed.

> But why increase the damage? Makes no sense to me.

Because the explosion is larger and most players would expect a larger explosion to deal more damage.

xezon:
I used this map to test in Skirmish

https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Maps/Any/!TestAllFactions

MTKing4:
> The intent was clearly to have it deal damage upon death, or it would not be assigned the Nuclear Reactor's death weapon and have a corresponding effect. Regardless, intent takes a backseat over bugs, poor design and counterintuitive gameplay.

I get that it needs to do damage, but this is too much, killing two lines of enemy vehicles around it is too much, it should be less than that

> It's not supposed to be rewarding; this is precisely the point. It's the same risk as destroying a Nuclear Reactor. It's also a risk that a skilled / knowledgeable opponent should be aware of and be able to mitigate when destroying the structure. Players already tend to do this because they expect the explosion to deal damage. Again, you should try it out for yourself.

How is killing the enemy super weapon, the most dangerous weapon on the battlefield with the highest price and hp not rewarding? I don't get that, are you suggesting to disincentivize players to kill the enemy's strongest building with the biggest army the fastest way possible? And not have to worry about army killing explosion, like we'd have to leave 1 unit behind to do the bit of damage, and it might die before killing it changing the outcome of that game, this has greater impact on gameplay than what is being implied here, but ok, I'll test it myself as well

> It is a unique behaviour. The Nuclear Reactor explodes but the Cold Fusion Reactor does not. The same logic applies here.

That is true, but this whole nuke explosion was never consistent to begin with, see nuke cannon's explosion, also i think reactor and nuke serve a different purpose so there's no issue if they behaved differently, besides i never said to remove damage, but making it less will be better imo

> Where is the sacrifice? 

Affecting balance 

> Players already tend to do this because they expect the explosion to deal damage. Again, you should try it out for yourself.

I've never seen anyone excepting anything at all when killing the nuke and I've been playing the game for years now, i even bet some people think that it doesn't even do damage, so this change will surprise them for sure

ExiLe-ZH:
> I get that it needs to do damage, but this is too much, killing two lines of enemy vehicles around it is too much, it should be less than that

> How is killing the enemy super weapon, the most dangerous weapon on the battlefield with the highest price and hp not rewarding? I don't get that, are you suggesting to disincentivize players to kill the enemy's strongest building with the biggest army the fastest way possible? And not have to worry about army killing explosion, like we'd have to leave 1 unit behind to do the bit of damage, and it might die before killing it changing the outcome of that game, this has greater impact on gameplay than what is being implied here, but ok, I'll test it myself as well

Let me remind you of 2 facts:

1. It only kills 1 row of units, the ones that are directly hugging the nuke. The 2nd row survives with half hp, check the video.
2. You shouldn't get too close to China buildings anyway, there could be mines. Stay out of mine range and your units will be safe.

Sorry, but you're massively overdrawing the possible impact on balance.


> I've never seen anyone excepting anything at all when killing the nuke and I've been playing the game for years now, i even bet some people think that it doesn't even do damage, so this change will surprise them for sure

My previous reply addresses this:
> It's nothing but a mistake to park your quads next to it while killing it, same with a Nuclear Reactor or any China building because of mines. Maybe it could happen once to someone who doesn't know about the change because it never been like that, but that shouldn't be a reason it doesn't do damage, logically it makes sense that it does.

xezon:
The radius is perfectly fine. The only reason there are a few places where 2 dragons die is because they are packed so close, likely in World Builder. In regular match situation there will only ever be one row of dead tanks.

kill

I tested ranges in game and it is comparable to Nuclear Reactor.

The dealt damage looks perfectly fine. Note that this damage should also destroy mines. So in a way this change is a buff and a nerf at the same time. How this plays out in a given match is situational.

MTKing4:
> My previous reply addresses this:

Right okay, it does makes sense, i didn't see the mines comment earlier 

Still, I'm not very convinced why it's necessary to increase range and damage, maybe only one of them would've been fine, but I'll see how it looks ingame, maybe it's fine.

> The dealt damage looks perfectly fine. Note that this damage should also destroy mines. So in a way this change is a buff and a nerf at the same time. How this plays out in a given match is situational.

Pretty sure old explosion also kills mines doesn't it?

xezon:
> Still, I'm not very convinced why it's necessary to increase range and damage

This change is not necessary, nor motivated by balancing. It is there to make similar events have similar consequences - achieving consistent behaviour. The goal of such change is to make gameplay more interesting, and in turn make the game more fun. Blowing up things is precisely the main motivator for combat.

xezon:
> I tested a few more Dozer repairs + kill, and I am pretty confident to say that the repairing Dozer receives no damage.

Also tested in Network Match. Same result there.

xezon (approved):
Looks good to me. May need to revisit, if range is too large or damage is too high in practice.

PR #1493, opened 2022-12-01
Fix: Nuke Battlemaster Detonation Effect Has Offset

xezon:
This change fixes detonation offset of Nuke Battlemaster. The fix uses same principles as
* #1122

### Original

https://user-images.githubusercontent.com/4720891/205148273-7354f628-6c53-49f1-a694-a4e61052663a.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -5100,6 +5100,35 @@ Object NukeCannonNeutronShellDetonationDummy
   End
 End
 
+; Patch104p @bugfix xezon 01/12/2022 Dummy to spawn Nuke Battlemaster detonation.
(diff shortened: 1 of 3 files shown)


Stubbjax:
Does this mean it will be harder for Nuclear Battlemaster explosions to hit Humvees?

ExiLe-ZH:
Was thinking the same thing, it'll be a pretty huge nerf if becomes easier for units to stay out of the explosion radius of rushing BMs.

commy2:
The explosion damage location is not changed by this PR. The effect will be displayed where the damaged is actually dealt in 1.04: at the center of the wreck.

~~What is moved back however is the radiation field.~~

Correction: the radiation field is created at the same position before and after the patch. This makes this change entirely visual.

commy2:
I see another visual glitch. The muzzle fire is displayed when the tank is blown up.

commy2 (approved):

PR #1494, opened 2022-12-04
Change: Tweak scorch sizes

xezon:
This change increases some small scorches, and decreases some big scorches.

Merge with Rebase.


xezon:
Looks okay. There will be another change buildings on top of this. May need to fine tune scorches a bit further once all new scorches are in place.

PR #1495, opened 2022-12-05
Fix: exscorch01 texture

xezon:
This change fixes and enhances exscorch01 texture.

### exscorch01

* 2x AI upscale
* Fixes alpha channel to match opacity of 4 scorches
* Fixes shape of scorch number 4

shot_20221231_125738_1


xezon:
I added a few more tweaks. Looks very good to me.

PR #1504, opened 2022-12-15
Add appropriately sized scorches to all major faction explosions

### LINKED ISSUE #1501: Demo_GLAVehicleBattleBusDeathWeaponDummySuicide does not work
Demo_GLAVehicleBattleBusDeathWeaponDummySuicide does not work. It applies no damage and no particle effects.

xezon:
* Fixes #1501
* Follow up for: #102

This change adds appropriately sized scorches to all major faction explosions.

* All explosions drop just one scorch
* China  TNT bombs now have smaller effects than C4 bombs
* China Nuclear Tank explosions are no longer displaced
* Demo GLA vehicles emits different sized scorches depending on Demo Upgrade
* Demo GLA Battle Bus now just emits one death explosion instead of 2
* Demo GLA Battle Bus suicide damage now works

Merge with Rebase


xezon:
I found Fire Storm scorch is quite massive. I reduced its size from 100 to 80 to match the core size of the flame better.

xezon:
I tested everything I could and it all looked good. Many nice scorches. I am a bit skeptical about the Tomahawk, SCUD Launcher scorches. But they are big explosions, so does fit in principle. We will keep an eye on scorches in actual matches. If too many scorches are drawn, then weapon scorches are good candidates for removal.

What we could remove already is Jet Missile scorches. Other small missiles from Rocket Men, Scorpion, Patriot, Comanche, ... also do not paint scorches.

xezon (commented):
Looks good to me.

Stubbjax:
Most vehicle scorches now appear too large. They were fine the way they were. Listening Outposts still create two scorches via `SlowDeathBehavior ModuleTag_07`. There are two `FX_GattlingExplosionOneFinal` definitions.

And again, why was this merged without approval, consensus or even discussion? This is terrible practice and needs to stop. And how do you reasonably expect anyone to review `+2,028 -332` changes? This should have been several separate PRs (like how #1425 became 16 PRs) and it should have been made clear that this alters gameplay in addition to its presentation as a visuals-only change.

xezon:
I will check Listening Outpost.

As for too large scorch, can you please show example?

xezon:
I addressed Stubbjax' points in follow up changes

* #1535
* #1536

PR #1514, opened 2023-01-01
Change: Normalize Placement view angle

xezon:
**Merge with Rebase**

This change normalizes placement view angles of all faction structures.

* All base defenses have matching placement view angle
* ~~All faction tech buildings have matching placement view angle~~ #1534
* ~~All faction barracks have matching placement view angle~~ #1534

## Original Base Defenses

shot_20230101_114435_2

## Patched Base Defenses

shot_20230101_113601_4

## Original Tech Buildings

shot_20230101_114544_3

## ~~Patched Tech Buildings~~

shot_20230101_113728_5

## Original Barracks

shot_20230101_114337_1

## ~~Patched Barracks~~

shot_20230101_113404_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -14455,7 +14455,7 @@ Object AirF_AmericaBarracks
   ; *** ART Parameters ***
   SelectPortrait         = SABarracks_L
   ButtonImage            = SABarracks
-  PlacementViewAngle     = -45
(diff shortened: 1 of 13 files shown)


xezon:
I think the change for USA Barracks has potential to cause upset, because for regular players this is very noticeable change. However, logically it makes most sense to adjust it this way.

ExiLe-ZH:
Don't think anyone will notice

MTKing4:
I think they might feel that something is different, but they won't know what it is unless they keep looking

ExiLe-ZH:
I mean the amount you have to rotate already differs in four ways, based on orientation of the map and spot you have. Very few would notice I'm sure. Especially when you don't notice right away, you will probably never know how it was before. So change is fine I think.

xezon:
No risk no fun.

PR #1520, opened 2023-01-06
Fix: Missile particle gliches

xezon:
Merge with Rebase.

This change fixes and improves missile particles:

* Fixes skipping frames on several missile flares (1)
* Fixes particle life time inconsistencies (2)
* Fixes heroic missile shape of China Tank Hunter, GLA Tunnel Defender (3)
* Fixes missile texture (3), size of USA Missile Defender (4)

## Details

(1) Skipped frames means lens flares spawned every 2nd frame only. This gives the illusion of stutter.

(2) Some missile particles had life times exceeding their max draw time (Color, Alpha). Life time can be matched with max draw time.

(3) Original heroic red lens flare shape (EXLenzflare4.dds) does not match non-heroic yellow lens flare shape (EXLenzflare2.dds). Therefore texture was changed to EXLenzflare6.dds, which is red and matches shape of yellow flare.

(4) Original Missile Defender non-heroic exhaust trail and lens flare has large size. And original heroic exhaust trail + lens flare has very tiny size. Non-heroic and heroic effects where changed to have same size, just 20% smaller than original non-heroic effects. They are still larger than effects from GLA Tunnel Defender and China Tank Hunter, as is illustrated in video below.

## Original 1

https://user-images.githubusercontent.com/4720891/211104619-d82ac124-7c3b-4e01-9fd5-a3c3adcbfdc1.mp4

## Patched 1

https://user-images.githubusercontent.com/4720891/211104651-bba92730-4dae-4a6d-b721-380401eadfee.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/ParticleSystem.ini
@@ -2455,13 +2455,13 @@ ParticleSystem HeroicMissileDefenderMissileExhaust
   AngularDamping = 1.00 1.00
   VelocityDamping = 0.99 0.99
   Gravity = 0.00
-  SlaveSystem = InfantryStingerMissileLenzflare
+  SlaveSystem = HeroicInfantryStingerMissileLenzflare ; Patch104p @tweak from InfantryStingerMissileLenzflare
   SlavePosOffset = X:0.00 Y:0.00 Z:0.00
   Lifetime = 30.00 30.00 ; Patch104p @tweak from 35.00 35.00
   SystemLifetime = 0
-  Size = 0.05 0.05
+  Size = 0.20 0.20 ; Patch104p @tweak from 0.05 0.05 to match non heroic particle size
   StartSizeRate = 0.00 0.00
-  SizeRate = 0.10 0.10
+  SizeRate = 0.00 0.30 ; Patch104p @tweak from 0.10 0.10 to match non heroic particle size
   SizeRateDamping = 0.93 0.94
   Alpha1 = 0.00 0.00 0
   Alpha2 = 0.00 0.00 0
@@ -19516,7 +19516,7 @@ ParticleSystem HeroicInfantryStingerMissileLenzflare
   IsOneShot = No
   Shader = ADDITIVE
   Type = PARTICLE
-  ParticleName = EXLnzFlar2.tga
+  ParticleName = EXLnzFlar6.tga ; Patch104p @tweak from EXLnzFlar2.tga
   AngleZ = 0.00 0.00
   AngularRateZ = 0.00 0.00
   AngularDamping = 1.00 1.00
@@ -35874,7 +35874,7 @@ ParticleSystem MissileDefenderMissileExhaust
   SlavePosOffset = X:0.00 Y:0.00 Z:0.00
   Lifetime = 30.00 30.00 ; Patch104p @tweak from 35.00 35.00
   SystemLifetime = 0
-  Size = 0.25 0.25
+  Size = 0.20 0.20 ; Patch104p @tweak from 0.25 0.25 to reduce size a bit
   StartSizeRate = 0.00 0.00
   SizeRate = 0.00 0.30
   SizeRateDamping = 0.93 0.94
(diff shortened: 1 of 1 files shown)


Stubbjax:
Missiles should not have double exhaust flares like this; it just looks bizarre.

image

This is also the case with SCUD missiles and Tomahawk missiles, which also look strange with the continuous flares spawned as part of their trail effects.

Laser Lock exhausts are a welcome change, though I feel like they should have the trails as well. Should probably be in a separate PR.

There needs to be more elaboration on some changes, such as `Fixes missile texture, size of USA Missile Defender`.

xezon:
The double exhaust only looks this way from above. From regular view angle (37.5) it looks proper. See second video. Note: This flare is as per Original style. It spawns at 100% visibility and then fades to 50% next frame and then to 0% in next frame.

I tested Laser Lock flares with exhaust trail and it looks odd because of the laser. There are just too much effects happening that way. Perhaps it would be ok if the trails were thinner, like they are for GLA Tunnel Defender and China Tank Hunter. I could make follow up change after this with further experiments.

I appended new detailed information in first post to explain changes.

xezon:
Is this change good?

Stubbjax:
> The double exhaust only looks this way from above. From regular view angle (37.5) it looks proper.

It's quite visible from the default view too. Is there a way to get the smoother flare without the doubling up effect? A still frame from the original looks much nicer with the single flare.

image

> I tested Laser Lock flares with exhaust trail and it looks odd because of the laser. There are just too much effects happening that way.

Did you happen to record this? It'd be good to see what it looks like.

xezon:
> Did you happen to record this? It'd be good to see what it looks like.

I have no record of this. You can test it by changing
```
  ProjectileExhaust = MissileDefenderLaserGuidedMissileExhaust
  VeterancyProjectileExhaust = HEROIC HeroicMissileDefenderLaserGuidedMissileExhaust
```
to
```
  ProjectileExhaust = MissileDefenderMissileExhaust
  VeterancyProjectileExhaust = HEROIC HeroicMissileDefenderMissileExhaust
```

xezon:
> It's quite visible from the default view too. Is there a way to get the smoother flare without the doubling up effect? A still frame from the original looks much nicer with the single flare.

The game update is 30 frames a second. We are limited by this. If velocity of effect is quick, then artifacts like this are inevitable. The only thing we can do is decrease particle life time to one frame. Then the trailing particle is effectively gone. ~~I doubt that this is preffered~~.

xezon:
Laser Lock Missile with trail. Is this preferred?

https://user-images.githubusercontent.com/4720891/212470900-a4065bc4-ecea-4afb-9f69-f4582a36e54f.mp4

xezon:
Trailing second flare removed from all missiles.

shot_20230114_134954_1

shot_20230114_135025_2

Stubbjax:
Oh wow, the latest update is awesome. The singular flare is perfect, assuming that's as close as it can get to the bottom of the missile.

Regarding Laser Lock, I think the trail looks okay, though it does obscure the laser a bit (though maybe not enough to cause any confusion). Perhaps the laser effect could be made more red to better distinguish it from the trail and better reflect the icon? Regardless, it may be safer to leave it and consider it in a separate change.

ExiLe-ZH:
> Laser Lock Missile with trail. Is this preferred?
> generals.2023-01-14.13-05-31-24.mp4

The rockets look slower? It's probably just me.

xezon:
> The singular flare is perfect, assuming that's as close as it can get to the bottom of the missile.

Closer to rocket would look like this.

shot_20230114_133740_1

> Regarding Laser Lock, I think the trail looks okay, though it does obscure the laser a bit (though maybe not enough to cause any confusion). Perhaps the laser effect could be made more red to better distinguish it from the trail and better reflect the icon? Regardless, it may be safer to leave it and consider it in a separate change.

It is a separate commit right now (the 3rd last commit of this pull).

xezon:
> The rockets look slower? It's probably just me.

Rocket speed is not changed.

Stubbjax (approved):

ExiLe-ZH:
Last one is a bit too close, I guess in between is not possible due to framerate? Then the one before that is best.

MTKing4:
The laser trail looks fine to me

xezon:
Laser Guided missile effect changes have been removed from this Pull.

ExiLe-ZH:
Can do laser lock fx in separate pull maybe, think I like it, although I like original aswell and it's kinda big change.

PR #1523, opened 2023-01-07
Change: Decrease China Neutron Mines cost to 400, time to 15

### LINKED ISSUE #1250: Neutron Mines are underused; Normal Mines can help better against early GLA agression
Neutron Mines take long and also add another 500 to the usual mines.

Without having the exact times in mind, i think it is 600/30s for normal mines and 500/40s for neutral mines, correct me if I am wrong.

It all comes down to what is beeing buffed for China vs GLA in the end.
Still reducing the price and build time can be a good option against early GLA harassment.

Currently, thinking about all the other changes up to today, I would vote for:
--> Normal mines for $500/-30% and neutron mines $400/-40%

xezon:
* Fixes #1250

This change slightly decreases build cost and time of China Neutron Mines.

| Mines                          | Cost   | Time  |
|--------------------------------|--------|-------|
| Original Mines                 | 600    | 20    |
| Original Neutron Mines         | 500    | 25    |
| Original Mines + Neutron       | 1100   | 45    |
| Patched Neutron Mines (this)   | 400    | 15    |
| Patched Mines + Neutron (this) | 1000   | 35    |

# Rationale

China Neutron Mines are almost never used in competitive matches. Reducing the cost and time to research them can make them a bit more attractive for occasional investment.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
@@ -242,8 +242,8 @@ End
 Upgrade Upgrade_ChinaEMPMines
   DisplayName        = UPGRADE:EMPMine
   Type               = OBJECT
-  BuildTime          = 25.0
-  BuildCost          = 500
+  BuildTime          = 15.0 ; Patch104p @balance from 25.0
+  BuildCost          = 400 ; Patch104p @balance from 500
   ButtonImage        = SNEMPMine
   ResearchSound      = MineFieldPlaced
 End


Stubbjax (approved):
Seems reasonable. I'd still like to see a build time reduction on normal mines. (And it would be even nicer if the build times matched.)

xezon:
Make Mines build time 15 perhaps?

ExiLe-ZH:
I would be for it. With the new door times, unit production is even more preferred over the mine research delay than before.

PR #1527, opened 2023-01-07
Prepare sound for Oil Derrick and Oil Refinery

xezon:
Prepares sound for Oil Derrick and Oil Refinery. The new sounds are not used and commented out.


xezon:
I do not like the Oil Derrick sound.

Stubbjax (approved):
We need to be very careful with what sounds end up being added here.

PR #1531, opened 2023-01-08
Fix: excontrail texture

Stubbjax:
Fixed the alpha of the contrail effect texture so that the edges would not stretch over the entire trail plane.

Old vs fixed:

trail


xezon (dismissed):

xezon:
The original excontrail.dds has mip maps. The new one does not. Is this intentional?

Stubbjax:
> The original excontrail.dds has mip maps. The new one does not. Is this intentional?

Forgot to tick the mip maps box!

xezon:
The best way is to use TGA or PSD and let Mod Builder create the right DDS format. Simply drop it in GenerateMip folder.

xezon (approved):

PR #1532, opened 2023-01-08
Fix Helix Commandset

### LINKED ISSUE #1513: China Helix Nuke/Napalm Bomb has identical icon for different buttons
China Helix Nuke/Napalm Bomb has identical icon for different buttons.

shot_20221231_111638_1

The Infantry China Helix is the only Helix that shows only one Napalm Icon at a time because its Command Buttons are cramped.

shot_20221231_112144_3

The proper fix is to show a slightly different icon for equipped Helix Bomb.

xezon:
This change was partially reverted with

* #1538

Merge with Rebase.

* ~~Fixes #1513~~

~~This change puts Helix Bomb Upgrade and Helix Bomb Drop buttons on same Command position for all Helix variants.~~ Additionally, Helix Bomb Upgrade is now greyed out while another upgrade is in production.

### No upgrades

shot_20230108_143327_1

### Bomb purchased

shot_20230108_143418_3

### Bomb Upgrade greyed out

shot_20230108_143340_2

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -3022,11 +3022,12 @@ CommandButton Command_UpgradeChinaHelixBattleBunker
   UnitSpecificSound       = HelixVoiceModeBunker
 End
 
+; Patch104p @bugfix Adds NOT_QUEUEABLE to grey out upgrade icon while other upgrade is in production.
 CommandButton Command_UpgradeChinaHelixNapalmBomb
   Command                 = OBJECT_UPGRADE
   Upgrade                 = Upgrade_HelixNapalmBomb
   TextLabel               = CONTROLBAR:UpgradeHelixNapalmBomb
-  Options                 = OK_FOR_MULTI_SELECT
+  Options                 = OK_FOR_MULTI_SELECT NOT_QUEUEABLE
   ButtonImage             = SNHelixUp04
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
   DescriptLabel           = CONTROLBAR:ChinaUpgradeHelixNapalmBomb
@@ -6478,9 +6479,10 @@ CommandButton Nuke_Command_UpgradeChinaFusionReactors
   DescriptLabel           = CONTROLBAR:TooltipUpgradeChinaNuclearTanks
 End
 
+; Patch104p @bugfix Adds NOT_QUEUEABLE to grey out upgrade icon while other upgrade is in production.
 CommandButton Nuke_Command_UpgradeChinaHelixNukeBomb
   Command                 = OBJECT_UPGRADE
-  Options                 = OK_FOR_MULTI_SELECT
+  Options                 = OK_FOR_MULTI_SELECT NOT_QUEUEABLE
   Upgrade                 = Nuke_Upgrade_HelixNukeBomb
   TextLabel               = CONTROLBAR:UpgradeHelixNukeBomb
   ButtonImage             = SNHelixUp05
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -645,7 +645,26 @@ CommandSet ChinaVehicleHelixCommandSet
   8 = Command_UpgradeChinaHelixPropagandaTower
  10 = Command_UpgradeChinaHelixGattlingCannon
   ;---------
-  7 = Command_UpgradeChinaHelixNapalmBomb
+  9 = Command_UpgradeChinaHelixNapalmBomb
+  ;---------
+ 11 = Command_AttackMove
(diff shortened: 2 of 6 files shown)


ExiLe-ZH:
This is nice

ExiLe-ZH (approved):

Stubbjax (approved):
This is great. The `NOT_QUEUEABLE` is a nice addition as well; that always bugged me.

MTKing4:
I think this adds a new confusion + Problem

The confusion is that i can no longer know if i have the bomb or the upgrade by looking at the icon (i know they have orange and green outlines but it's not very visible, and yes cameo is there but still, less ways to know can be annoying)

The problem is that if i select multiple helixes with 1 having the upgrade i can no longer order a dropping and upgrade Separately

I'm asuming when multiple helixes are selected I'll be forced to buy them all upgrades for the commandset to change to drop, this is a problem

xezon:
The confusion stems from 2 identical icons doing different things. One icon upgrades, the other icon drops bomb. Both look identical and sit right next to each other.

You are correct. With this change applied, Bomb Upgrade and Bomb Drop can no longer be used with shortcut when 2 Helixes have different Command Set active.

Revert this change?

Note that Infantry Helix has identical behavior in original game, due limited command slots.

shot_20230109_205514_2

MTKing4:
> The confusion stems from 2 identical icons doing different things. One icon upgrades, the other icon drops bomb. Both look identical and sit right next to each other.

Yes but the severity of the confusion was less before as you could at least tell you have the upgrade when the upgrade commandset gets a darker tone

I think revert, and stick to the different icon solution, i think a `Red Bomb facing down` icon or `Red Target Crosshair` icon will look nice and quite straight forward and good looking imo

MTKing4:
something like this:

Fun 1

xezon:
I agree with revert, though it is upsetting that it will be inconsistent with Infantry Helix.

ExiLe-ZH:
I'm a simple user, I just tested Inf Lix, the interface, hotkeys and functionality all works perfectly.
 
I don't understand for what reason two buttons is preferred? Or even different icons? That only makes things more inconsistent and cluttered.

Stubbjax:
It is already the case if a player tries to do the same with the Bunker / Gattling Cannon / Speaker Tower upgrades. The same behaviour applies to the Overlord. I disagree that this change adds any meaningful level of confusion at all, and it may even reduce confusion as there is no longer a behavioural difference between Helixes - and further, between upgrades. Standarding the behaviour with the Infantry Helix is a cleaner solution in this case.

However, the inability to drop bombs if Helixes without the upgrade are in the unit selection is a serious problem that needs to be addressed.

xezon:
> However, the inability to drop bombs if Helixes without the upgrade are in the unit selection is a serious problem that needs to be addressed.

This is the culprit of this change. It cuts functionality.

As for the Attachment upgrades, I am trying to fix it, but so far unsuccessfully. It works correctly on Humvees with Drone Upgrades.

ExiLe-ZH:
Ok sorry, I missed that issue. Personally never ever used it with multiple Lixes at the same time in all of my playtime.

An old idea, what about making the upgrade global with all the same parameters as Rocket Pods? :)

xezon:
> An old idea, what about making the upgrade global with all the same parameters as Rocket Pods? :)

This would solve all Command Set design problems.

Stubbjax:
> This is the culprit of this change. It cuts functionality.

If that is the case then I would say revert - we do not want to lose functionality.

> Personally never ever used it with multiple Lixes at the same time in all of my playtime.

This is also my experience, so the impact may be low and the benefits would likely outweigh the drawbacks. But that may change if the upgrade becomes more appealing.

> An old idea, what about making the upgrade global with all the same parameters as Rocket Pods? :)

This would be far too controversial!

ExiLe-ZH:
> This would be far too controversial!

Compared to other controversial changes already in the patch? Lol nah.

It's a nice little improvement for a shitty upgrade noone uses.

Stubbjax:
> > This would be far too controversial!
> 
> Compared to other controversial changes already in the patch? Lol nah.
> 
> It's a nice little improvement for a shitty upgrade noone uses.

What other changes? I think it certainly would be. Especially because there is associated dialogue that would never be heard again.

ExiLe-ZH:
> What other changes?

Seems you missed alot

But I can make another controversial upsetting public topic to see what people think.

Stubbjax:
> > What other changes?
> 
> Seems you missed alot
> 
> But I can make another controversial upsetting public topic to see what people think.

I am interested to know. Probably best to move this discussion to Discord.

xezon:
> Seems you missed alot

This is not a helpful answer to the question. A comparable controversial change was asked for.

I agree with Stubbjax that it would be highly controversial. If my memory is not clouded we have not yet made a change that has changed a unit ability in such a way.

We have report for it here:
* #1014

MTKing4:
I don't recall us merging any controversial change without the majority agreeing to it, that being said, I'd like it if we could agree about adding that bomb upgrade, maybe we can gather public opinion on the matter to see if it's really that controversial

PR #1533, opened 2023-01-08
Change: Decrease China Mines build time by 2 seconds, Neutron Mines build time by 7 seconds

xezon:
* Successor to #1523

This change decreases build time of China Mines. Regular Mines and Neutron Mines now have identical build time.

| Mines                              | Time  |
|------------------------------------|-------|
| Original Mines                     | 20    |
| Original Neutron Mines             | 25    |
| Original Mines + Neutron           | 45    |
| Patched Neutron Mines (1)          | 15    |
| Patched Mines + Neutron (1)        | 35    |
| **Patched Mines (this)**           | 18    |
| **Patched Neutron Mines (this)**   | 18    |
| **Patched Mines + Neutron (this)** | 36    |

# Rationale

With this change, regular Mines build 2 seconds faster than original. There needs to be a balance between too quick and too slow. Mines should not become subject of abuse by purchasing them while enemy forces are located near a China structure. On the other hand, the production of Mines should not delay production queue for too long. Even though it is just a small buff, China will be happy with each buff it can get for free.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Upgrade.ini
@@ -232,8 +232,7 @@ End
 Upgrade Upgrade_ChinaMines
   DisplayName        = UPGRADE:Mines
   Type               = OBJECT
-  BuildTime          = 20.0
-; BuildTime          = 1.0  ; for testing purposes
+  BuildTime          = 18.0 ; Patch104p @balance from 20.0
   BuildCost          = 600
   ButtonImage        = SSMineBunker
   ResearchSound      = MineFieldPlaced
@@ -242,7 +241,7 @@ End
 Upgrade Upgrade_ChinaEMPMines
   DisplayName        = UPGRADE:EMPMine
   Type               = OBJECT
-  BuildTime          = 15.0 ; Patch104p @balance from 25.0
+  BuildTime          = 18.0 ; Patch104p @balance from 25.0
   BuildCost          = 400 ; Patch104p @balance from 500
   ButtonImage        = SNEMPMine
   ResearchSound      = MineFieldPlaced


ExiLe-ZH (approved):
looks good

Stubbjax:
> Mines should not become subject of abuse by purchasing them while enemy forces are located near a China structure.

Why do you consider this abuse and not counterplay? Having mines finish researching while enemies are on top is currently a feat that is almost impossible to reliably pull off with intent, and it feels cheap for the opponent and not as rewarding for the owner that such a potentially game-changing action / event almost always comes down to pure chance rather than skill.

Reducing the research time somewhat to place a bit more emphasis on reactionary / predictive play would surely increase the skill ceiling, better align with the action's consequences, and feel more rewarding when a player pulls the feat off while reducing the feeling of cheapness for the opponent.

ExiLe-ZH:
Main discussion was here: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1250

xezon:
> Why do you consider this abuse and not counterplay? Having mines finish researching while enemies are on top is currently a feat that is almost impossible to reliably pull off with intent, and it feels cheap for the opponent and not as rewarding for the owner that such a potentially game-changing action / event almost always comes down to pure chance rather than skill.

> Reducing the research time somewhat to place a bit more emphasis on reactionary / predictive play would surely increase the skill ceiling, better align with the action's consequences, and feel more rewarding when a player pulls the feat off while reducing the feeling of cheapness for the opponent.

You make good points.

I wrote this because it appears odd if China can defend in such ways that Mines could be used for adhoc defense tactics. Let us say, just for the illustration of extremes, the research time was 1 second. Then China player would never again need to invest in Mines as a way to gamble for success, or as intimidation, but as reactionary means. So when enemy attacks, then purchase Mines, otherwise do not. This would be significant buff for China, as it will carry intimidation of Mines on all structures and enemy player needs to move with much more caution, greatly slowing down many ground attacks against China. This in turn can harm enemy ability to harrass China effectively, if structures are placed in ways where passage gaps are narrow. It will also render common suicide attack strategies of GLA useless.

Therefore there needs to be a balance between too short and too long research time.

Stubbjax:
> Let us say, just for the illustration of extremes, the research time was 1 second. Then China player would never again need to invest in Mines as a way to gamble for success, or as intimidation, but as reactionary means. So when enemy attacks, then purchase Mines, otherwise do not.

> Therefore there needs to be a balance between too short and too long research time.

This is exactly how I looked at it as well. A balance is definitely required, and the original time of 20s feels very skewed to one side, whereas a value of 15s might strike a perfect balance between the original value and providing players with that skill potential.

Of course, I also have a slight bias to 15s as it is a multiple of 5, which every other upgrade in the game follows.

xezon:
Would you aim for 15 Mines + 15 Neutron?

ExiLe-ZH:
I wouldn't mind giving China better defensive capabilities, investing in those slows down your offensive capabilities (how games are usually won, the faster you can push the better), but ofcourse it's sometimes necesarry to invest in order to minimize damage. This indirectly affects your offense again, especially when it pays off.

My point is, I don't feel it's too much of buff, spamming mines doesn't suddenly become attractive or good. But it could potentially slow down the enemies offense.

One big issue I'm having with this "strategy" is that you're donating XP by letting the enemy destroy your buildings, you can't sell them because the mines will dissappear. I feel this often is more gamechanging and hurting than selling.

ExiLe-ZH:
> One big issue I'm having with this "strategy" is that you're donating XP by letting the enemy destroy your buildings, you can't sell them because the mines will dissappear. I feel this often is more gamechanging and hurting than selling.

If we want to make it an actual solid strategy then we could make mines last until they're manually cleared.

xezon:
I had a brief look at
```
  Behavior GenerateMinefieldBehavior
    MineName
    UpgradedMineName
    UpgradedTriggeredBy
    GenerationFX
    DistanceAroundObject
    MinesPerSquareFoot
    GenerateOnlyOnDeath
    BorderOnly
    SmartBorder
    SmartBorderSkipInterior
    AlwaysCircular
    Upgradable
    RandomJitter
    SkipIfThisMuchUnderStructure
    TriggeredBy
    ConflictsWith
    RemovesUpgrades
    FXListUpgrade
    RequiresAllTriggers
```
And there appears to be no such option natively. I suspect we would have to spawn dummy objects for the mines to have such functionality.

However this could lead to other issues, such as mines never disappearing on their own after structure sell/destruction.

ExiLe-ZH:
> However this could lead to other issues, such as mines never disappearing on their own after structure sell/destruction.

Well that's kind of the point, after selling too. You don't want to give the enemy XP, he will have vetted units, have promotions sooner and you will have less money. 

Unless we want to do less of a buff, where mines do disappear when selling, but stay in situations where you can't sell in time anymore.

In any case (whatever is even possible), after surrendering the mines should instantly explode, unless you have a teammate ofcourse.

xezon:
```
  Behavior MinefieldBehavior
    DetonationWeapon
    DetonatedBy
    StopsRegenAfterCreatorDies
    Regenerates
    WorkersDetonate
    CreatorDeathCheckRate
    ScootFromStartingPointTime
    NumVirtualMines
    RepeatDetonateMoveThresh
    DegenPercentPerSecondAfterCreatorDies
    CreationList
```

xezon:
I think it would be sufficient if the mine field would stay 30 seconds after structure sell. Just like it does on death.

MTKing4:
> I think it would be sufficient if the mine field would stay 30 seconds after structure sell. Just like it does on death.

That would be nice, yes

I like Exile's idea as well, might be the last needed buff for chinas to be as good as other factions is if mines don't disappear with sell at least

penfriendz:
Yes, not having mines instantly disappear on sell would be great.

ExiLe-ZH:
> I think it would be sufficient if the mine field would stay 30 seconds after structure sell. Just like it does on death.

A little more low profile, but also nice.

Stubbjax:
> Would you aim for 15 Mines + 15 Neutron?

Yes. 15 + 15 seems much better.

I think deciding whether to sell and lose the mines to deny the enemy xp is an interesting strategic decision; just like evacuating a Humvee to try and deal a bit of extra damage at the risk or likelihood of the enemy earning more xp.

ExiLe-ZH:
> > Would you aim for 15 Mines + 15 Neutron?
> 
> Yes. 15 + 15 seems much better.
> 
> I think deciding whether to sell and lose the mines to deny the enemy xp is an interesting strategic decision; just like evacuating a Humvee to try and deal a bit of extra damage at the risk or likelihood of the enemy earning more xp.

When you still can sell in time, it's an absolute no brainer for all the reasons I mentioned already. Therefor I think xezons suggestion is much better and also quite low profile.

It's also fair, because you don't get half your money back either from the mines.

PR #1536, opened 2023-01-08
Change: Fine tune tank scorch sizes

xezon:
Merge with Rebase.

* Successor to #1494

This change fine tunes tank scorch sizes. They are 10-20% smaller than before, but larger than Original.

This change also removes the duplicate FX_GattlingExplosionOneFinal definition. And spawns the Nuke Cannon death scorch on the secondary explosion instead of the first one, like other China tanks.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -1915,6 +1915,11 @@ FXList FX_ChinaVehicleNukeCannonDeathExplosion
     Offset = X:0.0 Y:0.0 Z:35.0
   End
 
+  TerrainScorch
+    Type = RANDOM
+    Radius = 30
+  End
+
 End
 
 ;---------------------------------------------------------------------------------------
@@ -2268,10 +2273,6 @@ FXList FX_ChinaVehicleNukeCannonDeathExplosionInitial
   End
   ViewShake
     Type = STRONG
-  End
-  TerrainScorch
-    Type = RANDOM
-    Radius = 30
   End
   Sound
     Name = TankDie
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -1656,7 +1656,7 @@ FXList FX_BombExplosion
 End
 
 ; ----------------------------------------------
-FXList FX_TankDieExplosion
+FXList FX_TankDieExplosion ; Unused
   ViewShake
     Type = NORMAL
   End
@@ -2153,7 +2153,7 @@ FXList FX_BattleMasterExplosionOneInitial
 End
 
 ; -----------------------------------------------------------------------------
-FXList FX_GattlingExplosionOneInitial
+FXList FX_GattlingExplosionOneInitial ; Unused
   ParticleSystem
     Name = HotPillarArms
     Offset = X:0 Y:0 Z:10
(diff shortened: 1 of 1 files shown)


xezon:
Adjusted scorch sizes are ok?

Stubbjax:
What was the reason for increasing their size? Whenever I look at a normal screenshot of a battle's aftermath in 1.04, the vehicle scorch mark sizes seem perfect to me. Maybe it's just because it's what I'm used to, but it doesn't seem like a necessary change. I'd probably even temper them somewhat so that they appear more like a scorch / burn marks instead of splotches.

SCORCHES

xezon:
The reason for tweaking scorch sizes was in Original Game, some scorches are bigger than explosion or damage radius, and other scorches are smaller than explosion or damage radius.

The way to review this is look at the actual unit + its death explosion. To give obvious example, check Overlord death scorch and compare it with Overlord model size and its explosion size.

Stubbjax:
> The reason for tweaking scorch sizes was in Original Game, some scorches are bigger than explosion or damage radius, and other scorches are smaller than explosion or damage radius.
> 
> The way to review this is look at the actual unit + its death explosion. To give obvious example, check Overlord death scorch and compare it with Overlord model size and its explosion size.

I've checked every single vehicle. Overlord, Nuke Cannon and Bomb Truck are good, but everything else seems fine as it was in 1.04.

xezon:
> Overlord, Nuke Cannon and Bomb Truck are good, but everything else seems fine as it was in 1.04.

I do not agree with this assessment. To illustrate, here are images of Battlemaster death, which is exemplary for all tanks in the game.

The scorch radius is 18. That radius is just 20% larger than the original (15).

shot_20230112_211721_8

shot_20230112_211722_10

shot_20230112_211725_13

shot_20230112_211739_14

The new scorch size is on the small end of what I would aim for, solely because you asked them to be smaller. Fact is, the tank model and the explosion is larger than the scorch.

To give example of a realistic scorch, here is exploded tank. As is evident, the scorch radius of 18 is still small.

Stubbjax:
> > Overlord, Nuke Cannon and Bomb Truck are good, but everything else seems fine as it was in 1.04.
> 
> I do not agree with this assessment. To illustrate, here are images of Battlemaster death, which is exemplary for all tanks in the game.
> 
> The scorch radius is 18. That radius is just 20% larger than the original (15).

The scorch protrudes too far out from the chassis. Vehicles not only explode upwards, but are raised up from the ground by their wheels / treads, so any explosion's impact on the ground is substantially reduced. General vehicle explosions in 1.04 are also harmless, so it makes less sense to have a larger scorch.

> To give example of a realistic scorch, here is exploded tank. As is evident, the scorch radius of 18 is still small.

This is not how a scorch from a vehicle explosion would look at all. That is clearly the result of a prolonged fire. Any such explosion of the vehicle itself would never result in such a scorch.

Nevertheless, regardless of how physics work, this is clearly a matter of personal taste and thus should be excluded / reverted on those grounds alone (and a separate case made for the change, if desired). Having the game feel the same as 1.04 is a priority, and unless it is a unanimously objective improvement over 1.04, the change should be avoided (and certainly without any consensus or discussion at all). Changing the visual appearance of all these scorch marks, even by a small amount, will add up.

xezon:
Here is a freshly hit tank with a scorch much larger than what this change does.

https://www.youtube.com/watch?v=Z6hJb4Km-Ig

tank

MTKing4:
I think in real life, the size of the tank scorch also sometimes relies on the size of the rocket/bomb explosion itself

xezon:
Scorch size of WeaponFX_BombTruckHighExplosive decreased from 60 to 55.
Scorch size of FX_ScudLauncherExplosionOneFinal decreased from 25 to 22.

xezon:
Looks good to me.

PR #1537, opened 2023-01-09
Fix: Keep acquired China Helix attachment upgrade in command set

xezon:
This change keeps acquired China Helix attachment upgrade in command set. Applies to regular Helix and Nuke Helix. This way keyboard shortcut for said attachment will work when multiple Helixes are selected where some have it purchased and others do not yet. This behavior is consistent with Infantry Helix.

shot_20230109_211447_1

shot_20230109_211450_2

shot_20230109_211454_3


xezon:
Maybe we can do better than that... I see drone upgrades work better than that.

xezon:
Superseded by #1542

PR #1542, opened 2023-01-11
Fix: Keep disabled Helix attachment upgrade buttons in Command Set

xezon:
This change simplifies Helix attachment Command Set. Buttons will no longer disappear after research. This means shortcut key can be used on multi Helix selection to build an upgrade any time.

This behavior is consistent with USA vehicle drone upgrade buttons.

shot_20230111_220238_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -654,7 +654,7 @@ CommandSet ChinaVehicleHelixCommandSet
  14 = Command_Stop
 End
 
-CommandSet ChinaHelixGattlingCannonCommandSet
+CommandSet ChinaHelixGattlingCannonCommandSet ; Legacy
   1 = Command_TransportExit
   2 = Command_TransportExit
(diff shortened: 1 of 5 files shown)


Stubbjax (approved):
Good change. Any reason to keep the legacy command sets?

xezon:
I kept the legacy command sets in case some mod map uses them.

Stubbjax:
> I kept the legacy command sets in case some mod map uses them.

Surely this patch already contains or will end up containing changes that result in similar incompatibilities? Are we limited to changes that won't affect some abandoned mod map from a decade ago? Cleaner / more manageable ini files feels more beneficial to me.

xezon:
Even if we have some changes that make mod map already incompatible, we can still curb removals, because then there are less things mod maps need to change to adapt to patch.

MTKing4:
What legacy command sets? I thought we agreed to revert the bomb buttons

xezon:
Helix Command Set count was reduced from 4 to 1.

MTKing4:
Count of What? I don't see any difference, can you explain?

PR #1544, opened 2023-01-12
Fix: China Overlord, Emperor command sets

xezon:
**Merge with Rebase.**

This change fixes and improves China Overlord, Emperor command sets.

### TODO

- [ ] Remake this change for optional use

## Overlord

* Upgrade buttons now have same position as Helix upgrade buttons
* Upgrade buttons now have same order as Helix upgrade buttons, ordered left to right from least expensive upgrade to most expensive upgrade
* Upgrade icons now have same order as Helix, Emperor upgrade icons - matching order of upgrade buttons
* Evacuate button in Bunker Command Set now has same position as Listening Outpost, Troopcrawler evacuate button
* Disabled upgrade buttons remain visible in Command Set - allowing use of shortcut key to research upgrade on multi unit selection

## Emperor

* Upgrade buttons now have same position as Helix upgrade buttons
* Disabled upgrade buttons remain visible in Command Set - allowing use of shortcut key to research upgrade on multi unit selection

## Original Overlord Command Set & Cameo

shot_20230112_195805_1

## Patched Overlord Command Set & Cameo

shot_20230112_193923_1

## Original Overlord Bunker Command Set

shot_20230112_195815_2

## Patched Overlord Bunker Command Set

shot_20230112_195436_1

## Original Emperor Command Set

shot_20230112_195903_3

## Patched Emperor Command Set

shot_20230112_195655_2

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -605,25 +605,30 @@ CommandSet ChinaTankOverlordDefaultCommandSet
   14 = Command_Stop
 End
 
+; Patch104p @tweak Keeps upgrade buttons in all command sets.
+; Patch104p @tweak Moves Command_Evacuate from 6 into same position as Outpost, Troopcrawler evacuate.
 CommandSet ChinaTankOverlordBattleBunkerCommandSet
   1 = Command_TransportExit
   2 = Command_TransportExit
   3 = Command_TransportExit
   4 = Command_TransportExit
   5 = Command_TransportExit
-  6 = Command_Evacuate
+  6  = Command_UpgradeChinaOverlordBattleBunker
+  8  = Command_UpgradeChinaOverlordPropagandaTower
+  9  = Command_Evacuate
+  10 = Command_UpgradeChinaOverlordGattlingCannon
   11 = Command_AttackMove
   13 = Command_Guard
   14 = Command_Stop
 End
 
-CommandSet ChinaTankOverlordGattlingCannonCommandSet
+CommandSet ChinaTankOverlordGattlingCannonCommandSet ; Legacy
   11 = Command_AttackMove
   13 = Command_Guard
   14 = Command_Stop
 End
 
-CommandSet ChinaTankOverlordPropagandaTowerCommandSet
+CommandSet ChinaTankOverlordPropagandaTowerCommandSet ; Legacy
   11 = Command_AttackMove
   13 = Command_Guard
   14 = Command_Stop
@@ -4731,13 +4736,13 @@ CommandSet Tank_ChinaVehicleBattleMasterCommandSet
   14 = Command_Stop
 End
 
-CommandSet Tank_ChinaTankOverlordGattlingCannonCommandSet
+CommandSet Tank_ChinaTankOverlordGattlingCannonCommandSet ; Legacy
   11 = Command_AttackMove
   13 = Command_Guard
   14 = Command_Stop
 End
 
-CommandSet Tank_ChinaTankOverlordPropagandaTowerCommandSet
+CommandSet Tank_ChinaTankOverlordPropagandaTowerCommandSet ; Legacy
   11 = Command_AttackMove
   13 = Command_Guard
   14 = Command_Stop
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -594,11 +594,12 @@ CommandSet ChinaTankBattlemasterCommandSet
   14 = Command_Stop
 End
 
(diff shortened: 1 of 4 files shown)


xezon:
Marked as controversial, because button positions have changed. At a glance it did not feel weird, because there is a lot of void around the Overlord upgrade buttons that make the reposition more forgiving as opposed to a more crowded command set. Changing Overlord upgrade button art later should also help improve visibility.

Stubbjax:
This is a great change functionality-wise, but the movement of the buttons is a massive counterpoint that makes this change very difficult to consider.

Is there a reason the evacuate button is in a different location to the Helix's? Perhaps it should be beside the transport slots? It might be a good idea to have a look at how the evacuate slots are positioned for other units and pursue some kind of consistency.

xezon:
> This is a great change functionality-wise, but the movement of the buttons is a massive counterpoint that makes this change very difficult to consider.

It is either that or legacy behaviour. Unfortunately the legacy upgrade positions collide with the garrison slots.

> Is there a reason the evacuate button is in a different location to the Helix's?

Yes. Because the Helix Evacuate button is not where Outpost, Troopcrawler Evacuate button is located.

> Perhaps it should be beside the transport slots?

That is legacy position.

> It might be a good idea to have a look at how the evacuate slots are positioned for other units and pursue some kind of consistency.

This is what this change does. It puts Evacuate button at same position as Evacuate button of Outpost, Troopcrawler. Since Outpost, Troopcrawler are ground vehicles like the Overlord, I have choosen their Evacuate button position to take precedent over the Helix Evacuate button position.

Stubbjax:
> This is what this change does. It puts Evacuate button at same position as Evacuate button of Outpost, Troopcrawler. Since Outpost, Troopcrawler are ground vehicles like the Overlord, I have choosen their Evacuate button position to take precedent over the Helix Evacuate button position.

Okay, that is good. It is now shared with Battle Bus, Technical, Humvee and Ambulance.

It is strange that the Listening Outpost and Troop Crawler use a different command button for their evacuate behaviours (the icon has a vehicle instead of a building), despite the same functionality. This prevents the action from being used when those units are mixed with others. It would be good to standardise this.

Stubbjax:
There is also the question of whether USA vehicle drone button positions should be taken into consideration.

xezon:
> There is also the question of whether USA vehicle drone button positions should be taken into consideration.

I was thinking about this, but concluded that because USA is a different faction and China Helix already has a reasonable setup, it is sufficient to streamline Overlord button setup with Helix and other China vehicle button setups.

> It is strange that the Listening Outpost and Troop Crawler use a different command button for their evacuate behaviours (the icon has a vehicle instead of a building), despite the same functionality. This prevents the action from being used when those units are mixed with others. It would be good to standardise this.

Oh interesting observation. I was already wondering why the Evacuate button was not working across different unit types, Overlord + Outpost. It gave me the impression that it is general limitation.

MTKing4:
A few points here: 

- Does the position of the evac button really have to be the same on different units for it to function? Because if not i think it's best to keep the evac button near the slots, it's quite the standard in many units having the evac after the slots
- while i do understand and like the upgrades position change to increase functionality, they don't have to be exactly like the helixes', i think shifting them to the right by 1 button will be optimal especially if we're putting the Evac button after the slots
- i don't like the changed order of the upgrades, it serves no purpose and messes with player's muscle memory, again, no need to be streamlined with helix
- the same goes of the order of icons/Cameos
- i think for emperor, it's a different unit, so i think it's fine to put the gat upgrade at slot 1, it looks weird putting it at like in the middle of an empty command set

xezon:
> Does the position of the evac button really have to be the same on different units for it to function? Because if not i think it's best to keep the evac button near the slots, it's quite the standard in many units having the evac after the slots

They need to be on same position, otherwise shortcut key does not work.

> while i do understand and like the upgrades position change to increase functionality, they don't have to be exactly like the helixes', i think shifting them to the right by 1 button will be optimal especially if we're putting the Evac button after the slots

There is limitation in freedom due the 5 Bunker slots.

> i don't like the changed order of the upgrades, it serves no purpose and messes with player's muscle memory, again, no need to be streamlined with helix

Fixed button order is good for new players.

> the same goes of the order of icons/Cameos

Fixed icon order is good for readability.

> i think for emperor, it's a different unit, so i think it's fine to put the gat upgrade at slot 1, it looks weird putting it at like in the middle of an empty command set

Perhaps. On that note: Bomb Truck upgrade positions are odd.

MTKing4:
> There is limitation in freedom due the 5 Bunker slots

You can't use the slot that's under the attack move button?

xezon:
> It is strange that the Listening Outpost and Troop Crawler use a different command button for their evacuate behaviours (the icon has a vehicle instead of a building), despite the same functionality. This prevents the action from being used when those units are mixed with others. It would be good to standardise this.

* Fixed with #1545

xezon:
I think it is possible to make the button position changes optional. Then it won't be controversial.

xezon:
Superseeded by #1896 and #1894

PR #1545, opened 2023-01-13
Fix evacuate buttons

xezon:
**Merge with Rebase.**

This change fixes evacuate buttons. Result:

* All vehicles now use the Evacuate button for vehicles
* ~~All helicopters use the Evacuate button for vehicles~~
* All structures use the Evacuate button for structures

The following faction unit buttons are affected:

* USA Humvee, Ambulance, ~~Chinook, Combat Chinook~~
* China Overlord, Helix
* GLA Technical, Combat Bike, Battle Bus

With this change it is now possible to evacuate China Outpost and Troopcrawler in group selection with other vehicles, such as Humvee and Battle Bus.

## Original

shot_20230113_202435_3

## Patched

shot_20230113_203739_3

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -649,7 +649,7 @@ CommandSet ChinaVehicleHelixCommandSet
   9 = Command_ChinaHelixDropNapalmBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -665,7 +665,7 @@ CommandSet ChinaHelixGattlingCannonCommandSet ; Legacy
   9 = Command_ChinaHelixDropNapalmBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -681,7 +681,7 @@ CommandSet ChinaHelixPropagandaTowerCommandSet ; Legacy
   9 = Command_ChinaHelixDropNapalmBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -697,7 +697,7 @@ CommandSet ChinaHelixBattleBunkerCommandSet ; Legacy
   9 = Command_ChinaHelixDropNapalmBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -3320,7 +3320,7 @@ CommandSet Nuke_ChinaVehicleHelixCommandSet
   9 = Nuke_Command_ChinaHelixDropNukeBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -3336,7 +3336,7 @@ CommandSet Nuke_ChinaHelixGattlingCannonCommandSet ; Legacy
   9 = Nuke_Command_ChinaHelixDropNukeBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -3352,7 +3352,7 @@ CommandSet Nuke_ChinaHelixPropagandaTowerCommandSet ; Legacy
   9 = Nuke_Command_ChinaHelixDropNukeBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -3368,7 +3368,7 @@ CommandSet Nuke_ChinaHelixBattleBunkerCommandSet ; Legacy
   9 = Nuke_Command_ChinaHelixDropNukeBomb
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -4171,7 +4171,7 @@ CommandSet Infa_ChinaVehicleHelixCommandSet
   10 = Infa_Command_UpgradeChinaHelixBattleBunker
   ;---------
  11 = Command_AttackMove
- 12 = Command_Evacuate
+ 12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
@@ -4189,7 +4189,7 @@ CommandSet Infa_ChinaHelixBombCommandSet
   9 = Command_ChinaHelixDropNapalmBomb
   10 = Infa_Command_UpgradeChinaHelixBattleBunker
   11 = Command_AttackMove
-  12 = Command_Evacuate
+  12 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
   13 = Command_Guard
   14 = Command_Stop
 End
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -105,7 +105,7 @@ CommandSet AmericaTransportCommandSet
   6 = Command_TransportExit
   7 = Command_TransportExit
   8 = Command_TransportExit
-  9 = Command_Evacuate
+  9 = Command_EmptyCrawler ; Patch104p @tweak from Command_Evacuate, so that all vehicles use same evacuate command button.
  13 = Command_Guard
  14 = Command_Stop
 End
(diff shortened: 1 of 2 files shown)


Stubbjax:
How come you used the Chinook's version of the Evacuate command for the Helix? While the stylised air-specific icon may be more interesting, it's also a lesser form of consistency than matching the Chinook's Evacuate command with the Helix's and arguably more confusing.

If anything, I think the Chinook's Combat Drop ability and Evacuate button icons are easier to mix up than any other combination of buttons in the game.

xezon:
We can make follow up change and turn Chinook, Helix evac buttons to vehicle evac icon.

Stubbjax:
> We can make follow up change and turn Chinook, Helix evac buttons to vehicle evac icon.

For clarity, the follow up change should just be concerning the Chinook and not include a revision of a prior merge as well.

xezon:
Why?

Stubbjax:
> Why?

I wonder if there is a book for this stuff. These are what I believe to be fundamental principles of collaborative projects.

## Solution 1:

You commit a simple change to this branch, undoing a potential point of controversy. The Helix Evacuate button remains visually the same as 1.04, and the other Evacuate buttons have improved usability with negligible risk. Future changes can be discussed independently as one or more standalone topics. The change is an objective improvement. Everyone is happy and the PR can be merged without issue.

## Solution 2:

You ignore feedback and merge a controversial change mixed with other usability improvements. There is now less clarity as to the validity of the change as a whole. A modification point is needlessly added to the history. Likelihood of conflicts is increased. The promised future revision may never happen.

Not only that, but there are potential problems if the future revision is combined with another controversial change - i.e. changing the Helix's Evacuate button back may be agreed upon, but changing the Chinook's Evacuate button to use the same icon may not be agreed upon.

To break it down even further:

Easier PR: Change Chinook Evacuate button to look like every other Evacuate button.
Easier PR: Change Helix Evacuate button to look like the Chinook's Evacuate button.

Harder PR: Change Chinook Evacuate button to look like every other Evacuate button and revert the Helix Evacuate button icon to 1.04's state.

Does this seem reasonable?

xezon:
> If anything, I think the Chinook's Combat Drop ability and Evacuate button icons are easier to mix up than any other combination of buttons in the game.

I agree with this sentiment. At the same time I am surprised you raise this just 24 hours after claiming Overlord upgrades with 96% similarity add no confusion at all.

Stubbjax:
> I agree with this sentiment. At the same time I am surprised you raise this just 24 hours after claiming Overlord upgrades with 96% similarity add no confusion at all.

I made no such claims! It is also a different problem. In the case of the Overlord upgrades, each icon is an explicit identifier of the upgrade. In the case of the Chinook, the Combat Drop icon looks more representative of an evacuation than the Evacuate button; which is likely the primary source of confusion. The difference between the Combat Drop and Evacuate icons themselves is very easy to distinguish.

xezon:
Quote:

> Nobody is a beginner player anymore and everybody is familiar with the current buttons. Whether the current buttons are confusing is very questionable. Changing this now would only cause more confusion.

The Bunker Overlord and Propaganda Overlord look very similar on the icons. There is a measurable pixel difference of just 4%. The reason you think these buttons are okay is because you are trained to recognize them.

This is the same principle as Chinook Evacuate and Chinook Rappel buttons. The buttons look similar and do not do a good job of telling them apart. The reason you take issue with them is because you are trained less to differentiate them, for whatever reason.

I am trying to look at these icons in a non biased way by default.

Bottom line, I agree with your Chinook button feedback.

xezon:
~~Updated helicopter evacuation buttons~~

**IS NO LONGER PART OF THIS PULL**

shot_20230114_141153_4

shot_20230114_141155_5

Stubbjax:
> The Bunker Overlord and Propaganda Overlord look very similar on the icons. There is a measurable pixel difference of just 4%. The reason you think these buttons are okay is because you are trained to recognize them.

And humans are highly adept at discerning that 4%.

> This is the same principle as Chinook Evacuate and Chinook Rappel buttons. The buttons look similar and do not do a good job of telling them apart.

But it's an entirely different principle. The buttons are easy to tell apart. These two icons apparently have a similarity of 39.51% - and there are only two of them to distinguish - and yet I have a harder time discerning which of these two buttons does what than differentiating between the Overlord attachment upgrades.

> The reason you take issue with them is because you are trained less to differentiate them, for whatever reason.
> 
> I am trying to look at these icons in a non biased way by default.

"The reason you think these buttons are okay is because you are trained to recognize them." - this is a very biased assumption and not accurate. How have I been trained to recognise the Overlord upgrades? Have I not also been trained to recognise the Chinook button icons?

> Updated helicopter evacuation buttons

I see you have chosen the dark side; solution 2! While I agree with this direction of standardising the Chinook's Evacuate icon, it should really be proposed as a separate change.

xezon:
> And humans are highly adept at discerning that 4%.

I cannot tell them apart at a glance. When you talk about humans you clearly exclude me and everyone like me.

xezon:
> I see you have chosen the dark side; solution 2! While I agree with this direction of standardising the Chinook's Evacuate icon, it should really be proposed as a separate change.

They are separate commits, so it is fine. It is ok to have multiple commits in one PR if they are closely related.

MTKing4:
> If anything, I think the Chinook's Combat Drop ability and Evacuate button icons are easier to mix up than any other combination of buttons in the game.

I agree, in fact when i first saw it i was going to ask why did we use the combat drop icon for the evac button 😅

@Stubbjax didn't you say there was a specific air evac button? Which one is that?

> Updated helicopter evacuation buttons

For helix it's as it was, better than having the nook icon

For the nook weirdly enough i kinda liked the original icon 😅 even though it adds confusion, i don't seem to get confused when i see both evac and combat drop at the same time because the combat drop icon is quite telling

I'd like to see that air evac icon stubjax was talking about, maybe that will fix everything

Stubbjax:
> > And humans are highly adept at discerning that 4%.
> 
> I cannot tell them apart at a glance. When you talk about humans you clearly exclude me and everyone like me.

Of course, I'm talking in general. There is a bell curve we are working within. Sure, the icons are similar, but they seem to be well within the realm of readability for the overwhelming majority of players. It's not ideal to make changes that notably affect everyone just to satisfy the extreme outliers.

If it can be done in a subtle and elegant way, such as a minor zoom in on the current icons without significantly altering the appearance or losing quality, then that could be potentially acceptable. As far as priorities go, however, I'd likely put this right at the bottom of the list.

> They are separate commits, so it is fine. It is ok to have multiple commits in one PR if they are closely related.

It's not fine because this PR as a whole is no longer an objective improvement. I would approve it in a heartbeat if it only changed `Command_Evacuate` → `Command_EmptyCrawler`. But `Command_ChinookUnload` → `Command_EmptyCrawler` pollutes the change.

> @Stubbjax didn't you say there was a specific air evac button? Which one is that?

Oh nah, I was just reading into the intent of the change. It seemed as though the Chinook's evacuate icon was applied to the Helix as a kind of 'air-based' evacuate icon, similarly to the vehicle and building variants being used for all vehicles and buildings, respectively.

xezon:
> It's not fine because this PR as a whole is no longer an objective improvement.

This Pull covers all Evacuate buttons. It is not necessary to split it in smaller parts. Just adds more work for me.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -119,7 +119,7 @@ CommandSet AmericaVehicleChinookCommandSet
   6 = Command_TransportExit
   7 = Command_TransportExit
   8 = Command_TransportExit
-  9 = Command_ChinookUnload
+  9 = Command_EmptyCrawler ; Patch104p @tweak from Command_ChinookUnload, so that all vehicles use same evacuate command button.
```
Stubbjax: You can just commit the suggestion. Hardly any work.

```suggestion
  9 = Command_ChinookUnload
```

Stubbjax (dismissed):

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
xezon: No. This PR is designed to be Merged with Rebase.

Stubbjax: Care to explain further? We cannot make commits to a branch?

xezon: Merge with Rebase means the commits are applied to main branch as is. Adding random commits onto the commit stack would break the clean stack.

Stubbjax: Well that is not an effective workflow and should be rectified.

xezon: It has some disadvantages but also advantages, aka I can create one change and see all changes work together and test them properly.

Stubbjax: Creating one change and seeing all changes work together...? What does this mean?

ExiLe-ZH:
Whatever is done shouldn't bother anyone who wants to optimize their reaction speed, those use hotkeys either how.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
xezon: I mean creating one branch with all changes to see how they all work together is more valuable in some cases. For example when making particle changes.

xezon:
Is this change good to go or should anything be changed?

Stubbjax (changes requested):
> Is this change good to go

No! The Chinook's Evacuate button icon is incorrectly altered.

> or should anything be changed?

Yes! Please revert the Chinook's Evacuate button to use `Command_ChinookUnload` like it does in 1.04.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
Stubbjax: I still don't understand. Nothing about what you're saying has any relation to making amendment commits to address feedback.

xezon:
I don't get it. You raised that Chinook evac icons are confusing. So they got fixed. Why do you want this reverted?

Stubbjax:
> I don't get it. You raised that Chinook evac icons are confusing. So they got fixed. Why do you want this reverted?

My initial assumption was that this Pull Request was a direct response to my comment [here](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1544#issuecomment-1381880745), and would either change both the Listening Outpost and Troop Crawler's Evacuate icons to match the other vehicles, or vice versa. As `Command_Evacuate` looks so similar to `Command_EmptyCrawler`, and because these two vehicles are the only outliers, either way is fine and carries almost no risk. The change would have been approved with no issues had that been the extent of the change.

However, you made an additional design decision to change the Helix and replace its Evacuate icon with the Chinook's, which I expect was intended to achieve some level of consistency with air-based transport Evacuate icons. While consistency is obviously the entire purpose of this change, the Chinook's Evacuate icon looks significantly different to the other Evacuate icons and is nowhere near as objectively clean and risk-free as the former change. A larger change like this needs to face a lot more scrutiny and discussion before being accepted - ideally as a standalone change so that it does not affect or become impacted by the other objective improvements.

I got ahead of myself and highlighted why this additional design decision was the wrong approach - a premature attempt to effectively start that discussion - rather than pointing out that it should have been excluded from this change. However, I did immediate clarify this with two follow-up comments, which were seemingly ignored or disregarded.

Unfortunately, we now have the opposite problem as a result of that discussion where the Chinook's Evacuate icon has been changed to match the Helix's, which is just as big of a change, and should be excluded and proposed as a separate change for the aforementioned reasons.

ExiLe-ZH:
aesthetically I like the old nook button, would be shame if it was gone.  It's also clear as is, screw consistency :P

xezon:
> aesthetically I like the old nook button, would be shame if it was gone. It's also clear as is, screw consistency :P

The drawback of keeping different button art is that we will never get to make Helix and Chinook shortcut compatible, which is a rare but real functionality discrepancy.

ExiLe-ZH:
Was aware of that, but like you said, rare. 

In my 40k+ games I've never faced this situation/issue.

xezon:
Chinook evac icon change commit split into separate pull
* #1565

Stubbjax (approved):

PR #1547, opened 2023-01-14
Change: Decrease USA Superweapon Power Plant build cost from 900 to 800

### LINKED ISSUE #1503: Balancing USA powerplants
| Faction | Unit Cost | Upgrade Cost | Power | Upgraded Power |
|---------|----------:|-------------:|------:|---------------:|
| USA     | 800       | 500          | 5     | 10 (5+5)             |
| AFG     | 800       | 500          | 5     | 10 (5+5)             |
| Laser   | 700       | 500          | 8     | 16 (8+8)             |
| SWG     | 900       | 500          | 5     | 20 (5+15)             |

It doesn't seem entirely fair for SWG, her base reactor costs more while not generating extra power. Ofcourse Laser needs the extra power for his Laser Crusader production, but SWG doesn't have this option at all. She relies alot on her defenses since she got more expensive Vees aswell. Furthermore Laser is allowed to make a Laser Patriot on two Supplies without needing the Control Rods Upgrade, while SWG can not.

### Proposals

1. Increase SWG base power production (8 or 10).
2. Reduce SWG Reactor price (optionally increase Upgrade price to compensate).

xezon:
* Fixes #1503

This change decrease USA Superweapon Power Plant build cost from 900 to 800.

| Object                                    | Cost | Upgrade Cost | Power | Upgraded Power |
|-------------------------------------------|-----:|-------------:|------:|---------------:|
| AmericaPowerPlant                         | 800  | 500          | 5     | 10             |
| AirF_AmericaPowerPlant                    | 800  | 500          | 5     | 10             |
| Lazr_AmericaPowerPlant                    | 700  | 500          | 8     | 16             |
| SupW_AmericaPowerPlant                    | 900  | 500          | 5     | 20             |
| **Patched SupW_AmericaPowerPlant (this)** | 800  | 500          | 5     | 20             |

## Rationale

The original USA Superweapon Power Plant price model is problematic. When an Upgraded Power Plant is lost early, then it is difficult to recover, because it takes a large investment of cash and time to rebuild and upgrade the Plant.

Reducing the initial cost by $100 will help offset the initial $850 Humvee cost without impacting the feel of the game. This will reduce the total cost per unit of power from $70 to $65, which is a more meaningful gap between Laser General's $75 per unit of power.

Additionally it adds consistency with USA and Air Force Power Plants, and widens the gap with Laser.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -17270,7 +17270,7 @@ Object SupW_AmericaPowerPlant
   EditorSorting    = STRUCTURE
   Prerequisites
   End
-  BuildCost        = 900
+  BuildCost        = 800 ; Patch104p @balance from 900 to cost just as much as USA, AFG.
   BuildTime        = 10.0           ; in seconds
   EnergyProduction = 5
   EnergyBonus      = 15


ExiLe-ZH:
I think I personally prefer more power output over a price reduction. Like this you keep the distinction intact between factions in a meaningful way, more expensive but also better.

xezon:
Close this?

Stubbjax:
> I think I personally prefer more power output over a price reduction. Like this you keep the distinction intact between factions in a meaningful way, more expensive but also better.

This change has far fewer implications than a power output alteration. The expense is intended to be in the units, not on the structure that fuels SWG's supposed proficiency.

MTKing4:
I too think that a power output change is more meaningful for SWG early game than a price reduction

xezon:
If we cannot agree on this change I would prefer to close this, as this means this change is not good enough.

ExiLe-ZH:
The proposal is good and I'm convinced with my experience that a sufficient amount of test games will show positive hard data on this (higher winrates). 

Unless you don't want to **effectively** help SW early game.

xezon:
Are Power and Price changes mutually exclusive?

ExiLe-ZH:
Oh I thought I was replying in other (issue) topic.

xezon:
Close or merge?

Stubbjax (approved):
I think it's a very safe change.

MTKing4:
looks good

PR #1551, opened 2023-01-14
Change: Add missile lens flare particle to guided missiles of USA Missile Defender

xezon:
* Closes #1552

This change adds missile lens flare particle to guided missiles of USA Missile Defender. Nothing else.

https://user-images.githubusercontent.com/4720891/212485148-da408a7c-3839-48bd-ae47-f50c58ae6b74.mp4


MTKing4:
I think this might be sufficient without the exhaust trail, i don't really like changing the laser color

xezon:
The upside of this change is that it is minimal change and close to the original with the benefit of increased visibility of rockets.

The downside compared to the other change is that the lack of rocket exhaust is not plausible. Additionally, the green flare at either side of the laser beam has a different color than the red beam and the red flare on the button icon.

PR #1530, opened 2023-01-07
Change: Increase USA Comanche Armor by up to 23%, Comanche Armor with Countermeasures by up to 9%

xezon:
This change makes Comanche Armor and Countermeasures by up to 23% stronger.

## Stats

| Type             | Original Armor | Patched Armor (this) | Change % |
|------------------|---------------:|---------------------:|---------:|
| DEFAULT          | 100%           | 100%                 | 0%       |
| SMALL_ARMS       | 120%           | 100%                 | -17%     |
| GATTLING         | 120%           | 100%                 | -17%     |
| EXPLOSION        | 130%           | 100%                 | -23%     |
| INFANTRY_MISSILE | 120%           | 100%                 | -17%     |

| Type             | Original Countermeasures | Patched Countermeasures (this) | Change % |
|------------------|-------------------------:|-------------------------------:|---------:|
| DEFAULT          | 100%                     | 100%                           | 0%       |
| SMALL_ARMS       | 75%                      | 75%                            | 0%       |
| GATTLING         | 75%                      | 75%                            | 0%       |
| EXPLOSION        | 110%                     | 100%                           | -9%      |
| INFANTRY_MISSILE | 100%                     | 100%                           | 0%       |
| HEALING          | 100%                     | 100%                           | 0%       |

# Rationale

Regular Comanches are rarely used. This changes improves the utility of regular Comanches.

PR #1553, opened 2023-01-15
Change: Decrease USA Airforce Comanche Armor with Countermeasures by up to 21%

xezon:
* Follow up for #1530
* Related to #1135

This change makes AFG Comanche Countermeasures by up to 21% weaker.

## Stats

| Type             | Original AFG Armor | Patched AFG Armor (this) | Change % |
|------------------|-------------------:|-------------------------:|---------:|
| DEFAULT          | 100%               | 100%                     | 0%       |
| SMALL_ARMS       | 100%               | 100%                     | 0%       |
| GATTLING         | 100%               | 100%                     | 0%       |
| EXPLOSION        | 100%               | 100%                     | 0%       |
| INFANTRY_MISSILE | 100%               | 100%                     | 0%       |

| Type             | Original AFG Countermeasures | Patched AFG Countermeasures (this) | Change % |
|------------------|-----------------------------:|-----------------------------------:|---------:|
| DEFAULT          | 90%                          | 100%                               | +11%       |
| SMALL_ARMS       | 62%                          | 75%                                | +21%     |
| GATTLING         | 62%                          | 75%                                | +21%     |
| EXPLOSION        | 90%                          | 100%                               | +11%       |
| INFANTRY_MISSILE | 90%                          | 100%                               | +11%       |
| HEALING          | 90%                          | 100%                               | +11%     |

## Rationale

This change takes a bit of the edge off of Air Force Comanches by amplifying the impact of user-error that often comes as a result of direct anti-air engagement, as well as improving the counter-play abilities of opponents.

This is achieved by reducing inconsistencies instead of adding them. The regular Comanche and Air Force Comanche now have identical armor sets. The 90% `HEALING` modifier is no longer present.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1553_afg_comanche_armor.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1553_afg_comanche_armor.yaml
@@ -1,0 +1,23 @@
+---
+date: 2023-01-15
+
+title: Decrease armor of USA Airforce Comanche with Countermeasures by up to 21%
+
+changes:
+  - tweak: The armor of the USA Airforce Comanche with Countermeasures against Quad Cannons and Gattlings is now set to 75% instead of 62% (smaller is better).
+  - tweak: The armor of the USA Airforce Comanche with Countermeasures against Explosions and Infantry Missiles is now set to 100% instead of 90%.
+  - fix: The USA Airforce Comanche with Countermeasures no longer heals 11% slower than without Countermeasures.
+
+labels:
+  - controversial
+  - design
+  - major
+  - nerf
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1553
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -1104,9 +1104,9 @@ Armor DemoGenTerroristArmor ; Special Boost for Demo Generals Terrorist
 
 End
 
-;*******************************************************************
+;***********************************************************************
 ;***IF YOU CHANGE THESE, CHANGE AFG_CountermeasuresComancheArmor TOO!***
-;*******************************************************************
+;***********************************************************************
 Armor AFG_ComancheArmor
   Armor = DEFAULT           100%    ;this sets the level for all nonspecified damage types
   Armor = SMALL_ARMS        100%    ;gives Quad and gattling a little more punch.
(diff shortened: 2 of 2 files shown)


Stubbjax:
Can we simplify / normalise `DEFAULT`, `EXPLOSION` and `INFANTRY_MISSILE` to 100% to match non-Countermeasures and standard Comanche armour? Otherwise there's this strange double-up where some rockets are dodged but the rockets that do hit strangely deal less damage. Not to mention laser turrets, fire, jet missiles, particle cannons, bombs, tanks, other comanches, humvees, sentry drones, angry mobs, etc. counterintuitively doing less damage.

See [here](https://github.com/TheSuperHackers/GeneralsGamePatch/compare/main...comanche-survability-balance) for reference.

xezon:
That would mean Counter Measures for Regular Comanche will be weaker too accordingly. Is this desired?

* See #1530

ExiLe-ZH:
I'm not 100% sure about this yet, depends on what we're going to do with the experience values.

Stubbjax:
> I'm not 100% sure about this yet, depends on what we're going to do with the experience values.

I would suggest keeping the experience values the same. If the armour is standardised, then the only functional difference between the two variants becomes the additional stealth upgrade, and they are otherwise functionally identical. It would be counterintuitive if they were to yield more xp based on upgrade potential. There is no precedent for upgrades being considered when it comes to xp yields.

xezon:
Any further input on setup?

Stubbjax:
> Any further input on setup?

Aside from normalising the other modifiers to 100%, I'd also suggest leaving `AirforceGeneral.ini` unchanged to maintain backwards compatibility.

ExiLe-ZH:
> > I'm not 100% sure about this yet, depends on what we're going to do with the experience values.
> 
> I would suggest keeping the experience values the same. If the armour is standardised, then the only functional difference between the two variants becomes the additional stealth upgrade, and they are otherwise functionally identical. It would be counterintuitive if they were to yield more xp based on upgrade potential. There is no precedent for upgrades being considered when it comes to xp yields.

Well it's a fact that Commies vet up really quickly because they can pick off buildings really fast (and don't have passengers). It could be that the CM nerf is enough overall, but otherwise I would suggest adjusting the experience for both AF and Vanilla Commies.

Stubbjax:
> Well it's a fact that Commies vet up really quickly because they can pick off buildings really fast (and don't have passengers). It could be that the CM nerf is enough overall, but otherwise I would suggest adjusting the experience for both AF and Vanilla Commies.

Oh, you mean experience requirements? I believe that's definitely worth considering, especially when compared to the poor Helix that has the same requirements but abysmal chances of ever destroying anything with its tiny gun.

xezon:
Setup adjusted according to Stubbjax recommendations.

xezon:
I am worried that this could be too detrimental for Air Force General. Thoughts?

xezon:
Brooks:
> I think these changes may not be as big of a deal as hp has limited bearing on the 'correct' usage of Comanches. It makes them harder to use but definitely not unusable and I like the standardisation.

ExiLe-ZH:
I don't get the extra vurnerability vs gatling type weapons though, do Quads then fall under small arms?

xezon:
Quad Cannons use SMALL_ARMS yes.

ExiLe-ZH:
Ok.

So does this change + #1530 make AFG Commie and vCommie similar then?

xezon:
Yes

> The regular Comanche and Air Force Comanche now have identical armor sets.

With this change the only difference would be the Stealth Upgrade.

commy2:
Gatling tank also deals SMALL_ARMS when firing at airborne targets.

xezon:
> Gatling tank also deals SMALL_ARMS when firing at airborne targets.

Not anymore.

PR #1554, opened 2023-01-15
Change: Increase USA unit armor bonus of HOLD THE LINE Battle Plan from 11.1% to 20.0%

### LINKED ISSUE #1212: USA's Bombardment and HoldTheLine Battleplans are rarely used

xezon:
* Fixes #1212

This change increases the USA unit armor bonus of HOLD THE LINE Battle Plan from 11.1% to 20.0%. Armor bonus is synonymous for damage resistance.

| Battle Plan                      | Damage Bonus | Range Bonus | Armor Bonus |
|----------------------------------|-------------:|------------:|------------:|
| Original BOMBARDMENT             | +20%         |             |             |
| Original SEARCH AND DESTROY      |              | +20%        |             |
| Original HOLD THE LINE           |              |             | +11.1%      |
| **Patched HOLD THE LINE (this)** |              |             | +20%        |

Infantry, vehicles ~~and planes~~ benefit from armor bonus. Combat samples:

| Battle to death              | Shots | Shots with HOLD THE LINE | Bonus |
|------------------------------|------:|-------------------------:|------:|
| Crusader vs Crusader         | 8     | 10                       | 20%   |
| Crusader vs Missile Defender | 17    | 21                       | 19%   |
| Quad vs Missile Defender     | 9     | 12                       | 25%   |
| Gatling vs Raptor            | 12    | 12                       | 0%    |

HOLD THE LINE ~~applies armor bonus to all units, whereas~~ BOMBARDMENT and SEARCH AND DESTROY apply damage and range bonus to infantry and vehicles only.

| Battle Plan        | Infantry Bonus | Vehicle Bonus | Planes Bonus | Structure Bonus | Strategy Center Perks |
|--------------------|----------------|---------------|--------------|-----------------|-----------------------|
| BOMBARDMENT        | Yes            | Yes           | No           | No              | Yes                   |
| SEARCH AND DESTROY | Yes            | Yes           | No           | No              | Yes                   |
| HOLD THE LINE      | Yes            | Yes           | No           | No              | Yes                   |

All Battle Plan perks in a nutshell:

### BOMBARDMENT
* gives ground units +20% damage.
* gives Strategy Center a gun.

### SEARCH AND DESTROY
* gives ground units an attack range bonus of 20%
* gives ground units a sight range bonus of 20%
* gives Strategy Center a sight range bonus
* gives Strategy Center stealth detection

### HOLD THE LINE
* gives ground units an armor bonus of 20%
* doubles the health of the USA Strategy Center

# Rationale

This change makes the use of the HOLD THE LINE Battle Plan more attractive for all USA factions. ~~It is especially attractive for USA strategies that involve planes, such as spamming Comanche, Aurora and Raptor.~~ Traditionally USA favors the SEARCH AND DESTROY Battle Plan above the others by a large margin.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -11823,7 +11823,7 @@ Object AirF_AmericaStrategyCenter
     ValidMemberKindOf                   = INFANTRY CAN_ATTACK VEHICLE ;Battle plans affect any kind of these...
     InvalidMemberKindOf                 = DOZER STRUCTURE AIRCRAFT DRONE   ;...but make sure they don't have any of these
     BattlePlanChangeParalyzeTime          = 5000 ;Subjected to paralyzation whenever a battle plan is changed.
-    HoldTheLinePlanArmorDamageScalar      = 0.9 ;Armor damage bonus scalar -- LESS is better!
+    HoldTheLinePlanArmorDamageScalar      = 0.8333333 ;Armor damage bonus scalar -- LESS is better! ; Patch104p @balance from 0.9 to increase damage resistance from 11.1% to 20.0%
     SearchAndDestroyPlanSightRangeScalar  = 1.2 ;Sight range bonus -- more is better!
     ;***NOTE*** WEAPON bonuses for army are specified in GameData.ini file!
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/FactionBuilding.ini
@@ -7312,7 +7312,7 @@ Object AmericaStrategyCenter
     ValidMemberKindOf                   = INFANTRY CAN_ATTACK VEHICLE ;Battle plans affect any kind of these...
     InvalidMemberKindOf                 = DOZER STRUCTURE AIRCRAFT DRONE   ;...but make sure they don't have any of these
     BattlePlanChangeParalyzeTime          = 5000 ;Subjected to paralyzation whenever a battle plan is changed.
-    HoldTheLinePlanArmorDamageScalar      = 0.9 ;Armor damage bonus scalar -- LESS is better!
+    HoldTheLinePlanArmorDamageScalar      = 0.8333333 ;Armor damage bonus scalar -- LESS is better! ; Patch104p @balance from 0.9 to increase damage resistance from 11.1% to 20.0%
     SearchAndDestroyPlanSightRangeScalar  = 1.2 ;Sight range bonus -- more is better!
     ;***NOTE*** WEAPON bonuses for army are specified in GameData.ini file!
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -11508,7 +11508,7 @@ Object Lazr_AmericaStrategyCenter
     ValidMemberKindOf                   = INFANTRY CAN_ATTACK VEHICLE ;Battle plans affect any kind of these...
     InvalidMemberKindOf                 = DOZER STRUCTURE AIRCRAFT DRONE   ;...but make sure they don't have any of these
     BattlePlanChangeParalyzeTime          = 5000 ;Subjected to paralyzation whenever a battle plan is changed.
-    HoldTheLinePlanArmorDamageScalar      = 0.9 ;Armor damage bonus scalar -- LESS is better!
+    HoldTheLinePlanArmorDamageScalar      = 0.8333333 ;Armor damage bonus scalar -- LESS is better! ; Patch104p @balance from 0.9 to increase damage resistance from 11.1% to 20.0%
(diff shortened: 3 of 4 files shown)


xezon:
~~I am a bit skeptical about this change in regards to USA Airforce, because it benefits from this buff more than the other USA factions. But on the other hand, if USA Airforce does not use SEARCH AND DESTROY, then this could also be a chance for other faction to overwhelm it with Anti Air weapons. Typically Airforce is strong with SEARCH AND DESTROY Humvees + Planes. It is not entirely clear if HOLD THE LINE Planes are better on their own. Or if HOLD THE LINE Humvees + Planes are better.~~

On top of that, we do have proposed change for reduction of AFG Comanche Armor
* #1553

And we do have discussion about decreasing effectiveness of Point Defense Laser (PDL).
* #638

~~These things would essentially combine with this change.~~

~~For USA Superweapon, this change is a welcome addition to help survivability of Alpha Auroras.~~

ExiLe-ZH:
Wow I didn't know HTL applied to aircraft, that sounds very OP with Commies.

Possible to remove Aircraft?

xezon:
> Possible to remove Aircraft?

~~It is one setting for all. I think it is just a bug that BOMBARDMENT and SEARCH AND DESTROY does not apply to planes.~~

ExiLe-ZH:
I always believed the Strat switch thing made it impossible to apply this stuff to aircraft, but that logic doesn't stand anymore.

MTKing4:
Sounds good, now are we going to revert the double composite armor for crusader and paladin to compensate? Think they'll be insane with all that armor combined

xezon:
Selecting more Armor means less Damage and Range. +20% Damage and +20% Range are good perks too. It is questionable that +20% Armor is superior.

ExiLe-ZH:
It's gonna be too much with Commies, but I guess AFG's Counter Measures nerf makes up for that? But that change on it's own needs to be discussed more and tested.

penfriendz:
All else equal, I don't really care about hold the line if search and destroy stays at 20%. It effectively makes your units invulnerable in many situations which beats any amount of armour bonus. And all else equal (e.g. no nerf to countermeasures), this is a major buff for afg.

S&D could have a 10% range bonus and this would still make it the best choice, just the micro would have to be one level better than brain dead.

ExiLe-ZH:
After thinking more about it I'm actually quite fond of the idea of combining the HTL buff with the CM nerf, when they cancel each other out pretty much. It means the nerf only applies early game/pre-tech, giving the enemy a slightly easier time to secure himself before the Commies become OP like usual.

Stubbjax:
If aircraft are somehow gaining battle plan bonuses then it is a bug. Aircraft do _not_ gain any battle plan bonuses in 1.04. 
```
Behavior = BattlePlanUpdate ModuleTag_14
  ValidMemberKindOf = INFANTRY CAN_ATTACK VEHICLE ;Battle plans affect any kind of these...
  InvalidMemberKindOf = DOZER STRUCTURE AIRCRAFT DRONE ;...but make sure they don't have any of these
End
```

> Infantry, vehicles and planes benefit from armor bonus. Combat samples:

How did you get those test results? If a Raptor is dying in 10 shots then something is wrong. A Raptor (160 health) dies to 12 Gattling shots (12 × 12 damage × 120%). I sanity checked this in-game as well.

xezon:
I tested on !TestAllFactions in Skirmish. With USA and China units. By friendly fire on self player units.

https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Maps/Any/!TestAllFactions

Stubbjax:
> I tested on !TestAllFactions in Skirmish. With USA and China units. By friendly fire on self player units.

Do you happen to have the replay?

xezon:
I can make one later. This behaviour is 100%.

xezon:
I spotted error in my test. I tested with BOMBARDMENT against HOLD THE LINE by mistake, not realizing it does apply damage bonus to the Gatling Tank.

I have corrected all mistakes above accordingly.

ExiLe-ZH:
Then its a safe change

xezon:
Yes, this change should be safe change.

Should probably add core perk descriptions to tool tips in follow up change.

xezon:
Localization updated.

Stubbjax (approved):

xezon:
Merge #1559 before this one.

xezon:
Texts updated. Looks good.

ExiLe-ZH:
I still would like to know the exact performance in the field, there might still be a difference between BB and HTL. I predict BB still performs slightly better, but this is actually a good thing because HTL would be preferred if the field performance is the same, since the Structure Armor boost is probably more valuable than the Cannon.

PR #1559, opened 2023-01-17
Fix: USA Battle Plan tool tip text details

xezon:
This change fixes wrong and inaccurate information in USA Battle Plan tool tip texts.

### TODO
- [x] Fix comment slashes when so possible


commy2:
It's a better description, but "All ground units" is technically wrong too. It doesn't increase the firepower of Drones like the Sentry Drone. It also doesn't increase the damage caused by Demogen Workers suicide ability, while it works on all other GLA unit's suicide ability.

xezon:
Do we consider drones as ground units? They are floating above ground, so it seems they are excluded from "ground".

The Demogen Worker damage is a bug?

Stubbjax:
> Do we consider drones as ground units?

Sentry Drones.

Just rewrite it to "Vehicles and Infantry". Drones are not vehicles and Workers are not Infantry.

commy2:
> The Demogen Worker damage is a bug?

Oversight I would say. He's a DOZER kind, but DOZER is on the forbidden list for Strat Center bonuses. Just removing DOZER from the forbidden list would make Hold The Line give extra HP to the USA Dozer though. 

Just ignore that as a quirk.

xezon:
Fixed.

Stubbjax (commented):
The collocation "Infantry and Vehicles" feels unnatural. In English, the semantically larger "vehicles" would typically come first, e.g. "Vehicles and Infantry", and the lighter word often comes first as well.

xezon:
Fixed. Also added change comments to str file. I made a mistake in gametextcompiler tool and comments start with `\\` instead of `//`. I will correct this mistake in future version.

Stubbjax:
> I made a mistake in gametextcompiler tool and comments start with `\\` instead of `//`. I will correct this mistake in future version.

What are the implications of this? Can't you just commit a fix?

xezon:
I would need to hook up new gametextcompiler version to Mod Builder first.

Stubbjax:
> I would need to hook up new gametextcompiler version to Mod Builder first.

But it is fine to be in the main branch and won't cause issues?

xezon:
This change as is right now does work with the backslashes.

Stubbjax (approved):

xezon:
Comments now use forward slashes. Everything works.

PR #1565, opened 2023-01-19
Change: Use vehicle evacuation button for USA Chinook evacuation

xezon:
* Successor to #1545

This change streamlines evacuate buttons. Result:

* All helicopters use the Evacuate button for vehicles

The following faction unit buttons are affected:

* USA Chinook, Combat Chinook

With this change it will be possible to make Chinooks and Helixes evacuate in a group selection. Otherwise impossible.

## Patched

shot_20230114_141155_5

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1565_chinook_evacuation_button.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1565_chinook_evacuation_button.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-01-19
+
+title: Replaces USA Chinook specific evacuation button with generic vehicle evacuation button
+
+changes:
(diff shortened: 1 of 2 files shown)


xezon:
Any comments on this?

MTKing4:
Doesn't bother me,  functionality > Visuals

xezon:
Merge? The next change after this would be to move Evac of Chinook and Helix to same position, if possible. This will allow them to evacuate in group.

MTKing4:
It would look silly on the Chinook having evac and combat drop not next to each other but I don't think we have a choice

You have my ok on merge

xezon:
We can make it optional change, then it is safer.

xezon:
Missing Combat Chinook button tweak added. Made optional change. Documentation added. Tested and worked.

PR #1199, opened 2022-09-14
Change: Increase GattlingTankGun ContinuousFireCoast from 1000 to 2000 ms

xezon:
* Fixes #1185

This change increases GattlingTankGun ContinuousFireCoast from 1000 to 2000 ms. This matches the setup from Generals (CCG) and also the setting from the Zero Hour (ZH) GattlingTankGunAir.

| Object                          | ContinuousFireOne | ContinuousFireTwo | ContinuousFireCoast |
|---------------------------------|-------------------|-------------------|---------------------|
| Original CCG GattlingTankGun    | 3                 | 6                 | 2000                |
| Original CCG GattlingTankGunAir | 3                 | 6                 | 2000                |
| Original ZH GattlingTankGun     | 2                 | 6                 | 1000                |
| Original ZH GattlingTankGunAir  | 3                 | 6                 | 2000                |
| Patched GattlingTankGun (this)  | 2                 | 6                 | 2000                |

# Rationale

In Zero Hour it is difficult to use Pre Spin effectively before engaging enemy units, because the Spin Up time only last 1 second. With this change, Gattling Tank gun can be microed more generously to squeeze out the most damage potential.

| Object                         | Spin Level 1 DPS | Spin Level 2 DPS | Spin Level 3 DPS |
|--------------------------------|------------------|------------------|------------------|
| Original Quad Cannon (Vet 0)   | 75               |                  |                  |
| Original Quad Cannon (Vet 1)   | 165              |                  |                  |
| (#1055) Quad Cannon (Vet 1)    | 120              |                  |                  |
| Original Gattling Tank (Vet 0) | 37.5             | 75               | 112.5            |
| Original Gattling Tank (Vet 1) | 49.5             |                  | 123.75           |

Issue #1175, opened 2022-09-09
Gattling Tank EXP Requirement Balancing


ExiLe-ZH:
Unit | Regular | Veteran | Elite | Hero
--  | -- | -- | -- | --
Gatling Tank | 0 | 80 | 200 | 300
Quad Cannon | 0 | 100 | 150 | 300
Humvee | 0 | 100 | 150 | 300
Toxin Tractor | 0 | 100 | 150 | 300
Dragon Tank | 0 | 80 | 200 | 300

Now the balance between Gats and Quads should massively improve with the Quad Damage Distribution change
* #1055

To further improve this the Experience Requirements can't be overlooked. When we compare the requirements for Gats and Quads the balance doesn't add up very logically. Here a few facts:
- Vanilla Gats have a 20xp advantage in reaching Veteran status.
- Veteran Quad have a 50xp advantage in reaching Elite status.
- Quads can kill buildings to gain experience really fast.
- Quads can collect scraps to upgrade it's gun.

With these facts in mind we can conclude that Quads have a much better potential at getting strong fast. The only advantage Vanilla Gats have is the 20xp reduction for Veteran level.

Few more things to note:
- The Gat Veteran requirement of 80 is an inconsistency, for every other unit except Dragons it's 100xp.
- Gat Veteran nerf to 100 would be a buff for Tank, a nerf for VChina/Nuke.
- Gats need advantages across the board to compete?

### Proposal 1
First proposal would be to give Gats some advantage at each level:
`0 80 140 280`

### Proposal 2
To play it safer bring down Gat Elite Requirement to 150 (from 200):
`0 80 150 300`

### Proposal 3
Quad Elite Requirement to 200:
`0 100 200 300`

Gat Elite Requirement to 180 (and Hero to 280):
`0 80 180 300` or `0 80 180 280`

Jundiyy:
Would prefer 2 over 1, although don't know if I prefer anything at all at the moment. Been reading the recent topics and we are changing quite a few of these things, need to see what adds up.

ExiLe-ZH:
Which ones make China OP?

Stubbjax:
Although the Gattling Tank's transition from veteran to elite is rather slow / awkward, such a change would mostly benefit China Tank over vChina and Nuke - which is really the opposite of what is needed.

ExiLe-ZH:
Tank could also use it, it has a fairly low winrate overall.

[...]

PR #1567, opened 2023-01-19
Change: Increase spin up time of China Gattling Tank Ground Gun by 6 frames

xezon:
* Successor to #1199
* Closes #1691

This change increases barrel spin up time of China Gattling Tank Ground Gun by 6 frames. An alternative to #1691.

| Weapon                          | ContinuousFireOne | ContinuousFireTwo | ContinuousFireCoast |
|---------------------------------|-------------------|-------------------|---------------------|
| Original CCG GattlingTankGun    | 3                 | 6                 | 2000                |
| Original CCG GattlingTankGunAir | 3                 | 6                 | 2000                |
| Original ZH GattlingTankGun     | 2                 | 6                 | 1000                |
| Original ZH GattlingTankGunAir  | 3                 | 6                 | 2000                |
| Patched GattlingTankGun (1)     | 2                 | 6                 | 2000                |
| **Patched GattlingTankGun (this)** | 3              | 6                 | 2000                |

### Gattling Tank Ground Gun firing sequence in frames

Exclamation mark (!) represents a shot fired in timeline.

```
Original 12          6     6     6     3  3  3  3  3  3
         !...........!.....!.....!.....!..!..!..!..!..!

Patched  12          12          6     6     3  3  3  3  3  3
         !...........!...........!.....!.....!..!..!..!..!..!
```

## Rationale

Original `GattlingTankGun` and `GattlingTankGunAir` have inconsistent spin up times. With this change they are consistent.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1120,6 +1120,9 @@ Weapon RedguardBayonet
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @tweak from ContinuousFireCoast=1000 to prolong barrel spin timeout (same as GattlingTankGun of Generals 1.08, GattlingTankGunAir)
+; Patch104p @tweak from ContinuousFireOne=2 to prolong spin up time (same as GattlingTankGun of Generals 1.08, GattlingTankGunAir)
+;------------------------------------------------------------------------------
 Weapon GattlingTankGun
   PrimaryDamage         = 15.0
   PrimaryDamageRadius   = 0.0       ; 0 primary radius means "hits only intended victim"
@@ -1135,10 +1138,9 @@ Weapon GattlingTankGun
   DelayBetweenShots     = 400               ; time between shots, msec
   ClipSize              = 0                    ; how many shots in a Clip (0 == infinite)
   ClipReloadTime        = 0              ; how long to reload a Clip, msec
-  ContinuousFireOne     = 2 ; How many shots at the same target constitute "Continuous Fire"
+  ContinuousFireOne     = 3 ; How many shots at the same target constitute "Continuous Fire"
   ContinuousFireTwo     = 6 ; How many shots at the same target constitute "Continuous Fire Two"
   ContinuousFireCoast   = 2000 ; msec we can coast without firing before we lose Continuous Fire
-                               ; Patch104p @tweak from 1000 to prolong max damage time
   WeaponBonus           = CONTINUOUS_FIRE_MEAN RATE_OF_FIRE 200% ; When the object achieves this state, this weapon gets double the rate of fire
   WeaponBonus           = CONTINUOUS_FIRE_FAST RATE_OF_FIRE 300% ; This is not cumulative, so with Delay of 40, and values of 2 and 4 for these bonuses, you shoot at (40, 20, 10)
   WeaponBonus           = PLAYER_UPGRADE DAMAGE 125%     ; ChainGun upgrade


xezon:
Not clear to me if we want this. Perhaps ok if Gattling Tank is stronger in other ways.

MTKing4:
I'm not sure we want this either

ExiLe-ZH:
I don't understand the terms, values and results, outsiders/newbies probably neither when reading this.

Can someone explain this is noob language?

xezon:
It means the spin up takes 6 frames longer than before, equalling spin up of Zero Hour Gattling Air Gun, Generals Gattling Guns.

Stubbjax:
Perhaps a bit more rationale would be nice. I feel like such a change would drastically impact Gatts vs kiting Humvees. Towvee spam would probably be a lot more devastating. Denying terror techs or chasing down super techs would be harder, etc. There are lots of implications here.

xezon:
I do not know rationale for this, other than EA changing this value in an attempt to NERF the Gattling Tank, when this has the exact opposite effect. That is why I put a question mark to invite someone else to put a rationale.

ExiLe-ZH:
To me it all sounds counter intuitive, the only thing that happens is that it takes longer before the damage output maximizes? Sounds like nothing but a nerf?

Is this the exact change that made ZH Gats worse than CCG Gats? So how does that work? Does longer spin ups mean less "momentum loss" when switching between targets? (Or switching from force-fire to an actual target?)

xezon:
EA nerfed Gattling Tank. So they reduced its Damage, and its Range, but at the same time buffed the spin up by 6 frames.

See relevant comment by commy2 here:
https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1175#issuecomment-1242670353

So Weapon Damage was decreased from 20 to 15 per bullet and Weapon Range from 170 to 150, which are much more consequential nerfs than buffing the spin up time by a measly 6 frames.

ExiLe-ZH:
So conclusion is this particular change is nothing but a nerf?

Is there another way to keep more momentum so that pre-forcefiring becomes more effective but nothing else changes? So basically only changing the "spin-down" time?

xezon:
ContinuousFireOne should never have been changed the way it was.

Stubbjax:
> Is there another way to keep more momentum so that pre-forcefiring becomes more effective but nothing else changes? So basically only changing the "spin-down" time?

This has already been adjusted [here](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1199).

xezon:
Close this?

penfriendz:
Actually, I'd be fine with changing this given the coast time buff. But this will nerf noobs who don't force fire.

xezon:
`ContinuousFireCoast` from 1000 to 2000 is a consequential buff not just for those who want to micro manually, but also for regular battle engagements where a Gattling Tank would shoot and kill Ground Target A, and then take more than 1 second to engage next Ground Target B.

Here is comparison of former `ContinuousFireCoast` buff (#1199) vs this `ContinuousFireOne` nerf.

```
Original                    12          6     6     6     3  3  3  3  3  3
                            !...........!.....!.....!.....!..!..!..!..!..!
Patched ContinuousFireCoast 3  3  3  3  3  3  3  3  3  3  3  3  3  3  3  3
                            !..!..!..!..!..!..!..!..!..!..!..!..!..!..!..!
Diff                        ...!..!..!.....!.....!.....!..................
```
```
Original                    12          6     6     6     3  3  3  3  3  3
                            !...........!.....!.....!.....!..!..!..!..!..!
Patched ContinuousFireOne   12          12          6     6     3  3  3  3
                            !...........!...........!.....!.....!..!..!..!
Diff                        ..................!..............!............
```

Given this oversimplified comparison, the `ContinuousFireCoast` buff is 3 times better than the `ContinuousFireOne` nerf is bad. Advantages and disadvantages are situational, but assuming a somewhat even distribution of battle situations, the `ContinuousFireCoast` will make Gattling Tank better than originally, and the `ContinuousFireOne` nerf is much less (3 times less) detrimental in comparison.

**Correction: 3 times better, not 6 times better as previously stated.**

commy2:
GattlingTankGun and GattlingTankGunAir should use the same values for this.

MTKing4:
Close

xezon:
Alternative setup:

* #1691

xezon:
No consensus.

PR #1572, opened 2023-01-20
Change: Cancel Construction command button border color from Blue to Yellow

xezon:
This changes Cancel Construction command button border color from Blue to Yellow.

🟦 Blue = BUILD
🟩 Green = ACTION
🟧 Orange = UPGRADE
🟨 Yellow = SYSTEM

The Yellow border color matches better the color scheme of similar buttons, such as Yellow STOP button for units, and Yellow SELL button for structures.

## Patched

shot_20230120_232357_2

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -2238,11 +2238,13 @@ CommandButton Command_AmbulanceCleanupArea
 End
 
 ; Dozer construction commands ---------------------------------------------------------------------
+
+; Patch104p @tweak ButtonBorderType from BUILD.
 CommandButton Command_CancelConstruction
   Command       = DOZER_CONSTRUCT_CANCEL
   TextLabel     = CONTROLBAR:CancelBuild
   ButtonImage   = SSStop
-  ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
+  ButtonBorderType        = SYSTEM ; Identifier for the User as to what kind of button this is
   DescriptLabel           = CONTROLBAR:ToolTipCancelConstruction
 End
 


xezon:
Any opinions on this?

commy2:
Looks weird for some reason. "Stop" here has a different meaning (cancel construction) than "Stop" on units (stop moving/firing), so they don't necessarily have to use the same colour I suppose.

Stubbjax:
> Looks weird for some reason. "Stop" here has a different meaning (cancel construction) than "Stop" on units (stop moving/firing), so they don't necessarily have to use the same colour I suppose.

I was going to say, one could also argue that the blue border distinguishes the Cancel Build command as a separate function from the Stop command. As far as players are concerned, all construction-related command borders are blue. Since Cancel Build is a construction-related command, the current colour scheme makes sense in its current state and requires no change.

xezon:
> Since Cancel Build is a construction-related command, the current colour scheme makes sense in its current state and requires no change.

Wouldn't the same principle then apply for Structure Sell Command too? Sell is Yellow.

Stubbjax:
> > Since Cancel Build is a construction-related command, the current colour scheme makes sense in its current state and requires no change.
> 
> Wouldn't the same principle then apply for Structure Sell Command too? Sell is Yellow.

Perhaps, but this is the way it has always been. Selling isn't technically construction-related, despite the in-game mechanic appearing as such. Intuitively, people don't associate selling with construction. At least I don't expect my house to be bulldozed when I sell it.

MTKing4:
I think it looks fine as yellow and it makes sense, both stop commands are literally the same icon and everything, so why would they have different borders? The stop building argument doesn't make much of a sense really, since sell is actually building related according to the game logic and is yellow, so this one should be yellow just like other stop buttons

commy2:
> both stop commands are literally the same icon and everything, so why would they have different borders

Because they do different things.

xezon:
If they do different things, then they should not be the same icon.

commy2:
Invent _time machine_ and tell EALA about it.

penfriendz:
Well we can make a new button image right?

MTKing4:
I wouldn't like creating a new image, current image is discriptive and clear as it is, also it's very _Iconic_ (see what i did there? ^^ ) and people like the current icon very much

PR #1573, opened 2023-01-20
Fix: Remove all Special Power Shortcut command button border colors

xezon:
This change removes all Special Power Shortcut command button border colors consistently.

## Original

shot_20230120_230240_5

shot_20230120_230416_6

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -124,13 +124,14 @@ CommandButton Command_FireParticleUplinkCannon
   InvalidCursorName = GenericInvalid
 End
 
+; Patch104p @fix Disable ButtonBorderType on all Generals Powers Shortcuts.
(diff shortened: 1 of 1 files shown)


Stubbjax:
Any particular reason to comment them out instead of delete?

xezon:
Yes, because they are commented out by EA in those cases where they did remove the borders.

Stubbjax:
So you suggest the rule should be:

- If EA deletes entries, we also delete them
- If EA comments entries, we also comment them

xezon:
I do not think it has to be a strict rule per se. I just noticed that original devs have consistently commented in on the other buttons, so I did the same. We can also do it other way. I do not think it is important.

Stubbjax:
It seems like sometimes they deleted them; that's why I'm clarifying. I think it's cleaner to remove entries that do not apply, but that's subjective.

Stubbjax (approved):

xezon:
If I remember correctly it was consistent in all *FromShortcut buttons. That were the only ones I looked at.

PR #1579, opened 2023-01-21
Fix: Add Stop command button to all worker units

xezon:
**Merge with Rebase**

This change adds Stop command button to all worker units (except Boss). This way button can be pressed to stop worker units, including in group selection with other units.

To put the Stop button at its expected position, the Clear Mines, GLA Fake Structures and USA Airfield buttons had to be moved.

### Advantages

* Stop button added
* Clear Mines button now at same Position as Guard Mode button
* USA Airfield button is closer to other structures, and in same row as China Airfield

### Disadvantages

* Clear Mines, GLA Fake Structures, USA Airfield buttons moved
* GLA Fake Structures button is closer to Arms Dealer button

### TODO

- [ ] Replicate these changes in optional CommandSet.ini file

## Current

shot_20230121_131853_1

shot_20230121_131859_3

shot_20230121_131902_4

## Patched

shot_20230121_133026_2

shot_20230121_133030_3

shot_20230121_133037_4

All worker units selected.

shot_20230121_132855_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -42,17 +42,18 @@ End
 ; Dozer Command Sets ----------------------------------------------------------
 
 ; Patch104p @fix Add Stop button and move DisarmMinesAtPosition from 14 to 13, Airfield from 13 to 10 to make room for it.
+; Patch104p @tweak Move Airfield from 10 to 2, StrategyCenter from 2 to 4, SupplyDropZone from 4 to 6, ParticleCannonUplink from 6 to 8, CommandCenter from 8 to 10.
 CommandSet AmericaDozerCommandSet
   1  = Command_ConstructAmericaPowerPlant
-  2  = Command_ConstructAmericaStrategyCenter
+  2  = Command_ConstructAmericaAirfield
   3  = Command_ConstructAmericaBarracks
-  4  = Command_ConstructAmericaSupplyDropZone
+  4  = Command_ConstructAmericaStrategyCenter
   5  = Command_ConstructAmericaSupplyCenter
-  6  = Command_ConstructAmericaParticleCannonUplink
+  6  = Command_ConstructAmericaSupplyDropZone
   7  = Command_ConstructAmericaPatriotBattery
-  8  = Command_ConstructAmericaCommandCenter
+  8  = Command_ConstructAmericaParticleCannonUplink
   9  = Command_ConstructAmericaFireBase
-  10 = Command_ConstructAmericaAirfield
+  10 = Command_ConstructAmericaCommandCenter
   11 = Command_ConstructAmericaWarFactory
   13 = Command_DisarmMinesAtPosition
   14 = Command_Stop
@@ -1808,17 +1809,18 @@ CommandSet AirF_SpecialPowerShortcutUSA
 END
 
 ; Patch104p @fix Add Stop button and move DisarmMinesAtPosition from 14 to 13, Airfield from 13 to 10 to make room for it.
+; Patch104p @tweak Move Airfield from 10 to 2, StrategyCenter from 2 to 4, SupplyDropZone from 4 to 6, ParticleCannonUplink from 6 to 8, CommandCenter from 8 to 10.
 CommandSet AirF_AmericaDozerCommandSet
   1  = AirF_Command_ConstructAmericaPowerPlant
-  2  = AirF_Command_ConstructAmericaStrategyCenter
+  2  = AirF_Command_ConstructAmericaAirfield
   3  = AirF_Command_ConstructAmericaBarracks
-  4  = AirF_Command_ConstructAmericaSupplyDropZone
+  4  = AirF_Command_ConstructAmericaStrategyCenter
   5  = AirF_Command_ConstructAmericaSupplyCenter
-  6  = AirF_Command_ConstructAmericaParticleCannonUplink
+  6  = AirF_Command_ConstructAmericaSupplyDropZone
   7  = AirF_Command_ConstructAmericaPatriotBattery
-  8  = AirF_Command_ConstructAmericaCommandCenter
+  8  = AirF_Command_ConstructAmericaParticleCannonUplink
   9  = AirF_Command_ConstructAmericaFireBase
-  10 = AirF_Command_ConstructAmericaAirfield
+  10 = AirF_Command_ConstructAmericaCommandCenter
   11 = AirF_Command_ConstructAmericaWarFactory
   13 = Command_DisarmMinesAtPosition
   14 = Command_Stop
@@ -3515,17 +3517,18 @@ CommandSet SupW_SpecialPowerShortcutUSA
 END
 
 ; Patch104p @fix Add Stop button and move DisarmMinesAtPosition from 14 to 13, Airfield from 13 to 10 to make room for it.
+; Patch104p @tweak Move Airfield from 10 to 2, StrategyCenter from 2 to 4, SupplyDropZone from 4 to 6, ParticleCannonUplink from 6 to 8, CommandCenter from 8 to 10.
 CommandSet SupW_AmericaDozerCommandSet
   1  = SupW_Command_ConstructAmericaPowerPlant
-  2  = SupW_Command_ConstructAmericaStrategyCenter
+  2  = SupW_Command_ConstructAmericaAirfield
   3  = SupW_Command_ConstructAmericaBarracks
-  4  = SupW_Command_ConstructAmericaSupplyDropZone
+  4  = SupW_Command_ConstructAmericaStrategyCenter
   5  = SupW_Command_ConstructAmericaSupplyCenter
-  6  = SupW_Command_ConstructAmericaParticleCannonUplink
+  6  = SupW_Command_ConstructAmericaSupplyDropZone
   7  = SupW_Command_ConstructAmericaPatriotBattery
-  8  = SupW_Command_ConstructAmericaCommandCenter
+  8  = SupW_Command_ConstructAmericaParticleCannonUplink
   9  = SupW_Command_ConstructAmericaFireBase
-  10 = SupW_Command_ConstructAmericaAirfield
+  10 = SupW_Command_ConstructAmericaCommandCenter
   11 = SupW_Command_ConstructAmericaWarFactory
   13 = Command_DisarmMinesAtPosition
   14 = Command_Stop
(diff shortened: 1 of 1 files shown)


commy2:
Seems strange that Airfield appears last now instead of between WF and SC where it belongs in the tech tree.

xezon:
Should put Airfield at the left bottom, and push the rest 1 field to the right?

Or put behind Strategy Center? Then it would be at same spot as China Airfield.

penfriendz:
Put it left bottom. If you want consistency you could swap china airfield and internet centre.

If we made internet centre strictly require propaganda then we could put internet centre after propaganda and this would make the most sense.

MTKing4:
> Should put Airfield at the left bottom, and push the rest 1 field to the right?
> 
> Or put behind Strategy Center? Then it would be at same spot as China Airfield.

Can we see both to see which is better? I think either of them will be better than current order

xezon:
Repositioned USA Airfield buttons

## Version 1

shot_20230128_003028_1

## Version 2

shot_20230128_003205_1

penfriendz:
How about 1, and reposition the china buildings?
Wf, airfield, prop, internet centre?
This has the advantage of being consistent with USA and also consistent in tech tree. You may point out that internet centre requires wf, but it now can also require prop, and it's still after wf, just like how prop needs wf but isn't directly after wf.

xezon:
> How about 1, and reposition the china buildings? Wf, airfield, prop, internet centre? This has the advantage of being consistent with USA and also consistent in tech tree. You may point out that internet centre requires wf, but it now can also require prop, and it's still after wf, just like how prop needs wf but isn't directly after wf.

With that approach, Airfield, Tech Center and Economy Structure would be on same spot for all factions.

commy2:
Demo Trap = GLA Airfield. This confirms what we always knew.

xezon:
Ok. Command Set buttons adjusted according to latest conversations.

Advantages:
* Boss Barracks, Supply Center shares same position as regular factions

Disadvantages:
* More buttons moved around, most notably USA Airfield, now sitting at the opposite end

## Moved buttons with China

shot_20230128_095927_2

## Moved buttons with USA

shot_20230128_095932_3

## Moved buttons with Boss

shot_20230128_100644_4

MTKing4:
Please don't change buttons unnecessarily, it just adds to the list of things that players need to relearn, we don't need that, it's better to keep changes minimal.

My suggestion is keep china untouched

And go with Version 1 for USA as it's the better version and call it a day

xezon:
I can imagine that short term it could cause some confusion for some players, especially with the USA Airfield reposition.

But long term I expect this to be a profit, because STOP button is added + Worker unit button layouts are much more consistent across all factions, which is good for new starters, and anyone playing longer than just a few matches. I would expect it to take no more time getting used to than to Control Bar Pro.

I need to play USA a bit to see how it feels.

Stubbjax:
I agree with @MTKing4 on this. This is quite a disruptive change and completely goes against the principle of keeping visible / conspicuous changes as minimal as possible.

Though the initial GLA and China changes are fairly reasonable, USA tarnishes the deal. Players are used to the existing layout of the construction command sets. Changing these layouts would cause far more confusion than any minuscule amount you'd save (I'd argue it wouldn't save any confusion for new players at all). While I understand and appreciate the consistency argument, I believe the disadvantages greatly exceed the advantages in this case.

penfriendz:
I don't think it causes much disruption. Most people use keyboard shortcuts. Those who click aren't focused on muscle memory saving them 300 milliseconds. The visibility argument is valid so it's visibility vs consistency. Personally I prefer to have something good for perpetuity rather than sticking to something less good because that's how it's always been done.

xezon:
I expect that there will be more clashing points in Command Bar in the future. I am inclined to create an optional Command Bar big file with all controversial changes, such as this one, overlord upgrade icons and so on. This way, legacy players can use traditional layouts & images, and new players can opt in the optimized layouts & images.

The only downside to this - duplicated INI setups - is additional maintenance overhead for us, as we will need to replicate changes in 2 sets of files.

MTKing4:
So the whole stop button/clear mines will not be added/streamlined? It feels bad that we're missing out on functionality because of this, for me it was an acceptable compromise to have the buttons added and USA goes with Version 1 and that's it, no unnecessary reordering, the latter can be in the optional package

xezon:
There is most certainly no way to properly add Stop button without moving Clear Mines, GLA Fake Structures and USA Airfield buttons.

Stubbjax:
> So the whole stop button will not be added? It feels bad that we're missing out on functionality because of this, for me it was an acceptable compromise to have the buttons added and USA goes with Version 1 and that's it, no unnecessary reordering, that can be in the optional package

Reordering the construction command set is a huge change though. This would literally affect every player. And I don't see how it would really benefit many players, even new ones. Stop is probably one of if not the first hotkey everyone learns because it's necessary when you're in the heat of battle and don't have time to be manually clicking buttons. I just don't think the benefits outweigh the risks.

xezon:
> I'd argue it wouldn't save any confusion for new players at all

I disagree with this assessment. There is a very simple way to verify the validity of changes like this one: Assume the reverse of the status quo and think about whether or not it would make sense to remove the Stop buttons from Worker units and move similar building buttons of different factions to different Command Set positions. If the answer is "No", then there is merit to the change.

The only argument against this change, as far as I can tell, is its difference to the status quo. Other than that, it is a completely logical setup over a slightly less logical and less functional setup in the Original and therefore should not be ignored.

MTKing4:
> There is most certainly no way to properly add Stop button without moving Clear Mines, GLA Fake Structures and USA Airfield buttons.

And that is fine with me as i said, but no other buildings changed

MTKing4:
> > So the whole stop button will not be added? It feels bad that we're missing out on functionality because of this, for me it was an acceptable compromise to have the buttons added and USA goes with Version 1 and that's it, no unnecessary reordering, that can be in the optional package
> 
> Reordering the construction command set is a huge change though. This would literally affect every player. And I don't see how it would really benefit many players, even new ones. Stop is probably one of if not the first hotkey everyone learns because it's necessary when you're in the heat of battle and don't have time to be manually clicking buttons. I just don't think the benefits outweigh the risks.

It's a compromise over functionality, and isn't as big if it's only one building moved, frankly, if you have showed me the new USA command bar and told me this is how it always been I'd believe you, as i don't have building spots memorized, and don't think many people do either, but when more buildings get moved and unnecessarily then yes people will notice

Stubbjax:
> I disagree with this assessment. There is a very simple way to verify the validity of changes like this one: Assume the reverse of the status quo and think about whether or not it would make sense to remove the Stop buttons from Worker units and move similar building buttons of different factions to different Command Set positions. If the answer is "No", then there is merit to the change.

That is not an effective way of determining whether an interface is confusing or counterintuitive for new players, which seems to be the primary argument that has been put forward for this change. I also never suggested the change was without merit or invalid. Sure, 1.04's layout is not perfect, but it is well within the realm of usability, and nobody has seemingly had any issue with it for two decades. While that does not immunize the design to change, it does inflict a very high burden on proving that the change is necessary.

If there is a significant likelihood for players to experience "breakdown" (the user becomes consciously aware of and is detached from the embodied interface) as a result of being unable to visually find the Stop command, then you have a more solid case. But it's important to note that there is no data or evidence to suggest this, and it doesn't seem like anyone has any associated memories of this occurring. The game will likely never have more new players than active players + returning veterans, so such a change may only increase confusion potential rather than reduce it.

> The only argument against this change, as far as I can tell, is its difference to the status quo. Other than that, it is a completely logical setup over a slightly less logical and less functional setup in the Original and therefore should not be ignored.

That is certainly the only argument (and I don't believe anyone suggested otherwise), but it is an incredibly powerful argument.

Nothing is being ignored. As you said, _slightly_ less logical. I don't believe slightly less logical is enough to overpower the status quo, which is really the instrument with which all changes should be measured. While consistency is nice, its pursuit must be nuanced. All of these deviations from 1.04 will add up and if the initial perception comes with a high degree of confusion and unfamiliarity, then players will be put off and you will end up with a dual economy or irreparably harming the playerbase. Controversial changes like this that would not only affect everyone, but would objectively introduce confusion should only be made if extremely necessary and the benefits overwhelmingly outweigh the risks. I do not believe they do in this case.

xezon:
> That is not an effective way of determining whether an interface is confusing or counterintuitive for new players, which seems to be the primary argument that has been put forward for this change.

Personally I found this to be a reasonable way to help judge a change with. You may try it on any future change and may be pleasantly suprised.

> Sure, 1.04's layout is not perfect, but it is well within the realm of usability, and nobody has seemingly had any issue with it for two decades. While that does not immunize the design to change, it does inflict a very high burden on proving that the change is necessary.

The interface of the Original is usable yes. There is a reason however many pro players nowadays use Control Bar Pro instead of the Original Control Bar. For evidence, open any POV Zero Hour livestream of a pro player on YouTube, and check whether or not they use a Control Bar Addon or the Original one. Many players are open to improvements, and reasonably so.

As for "nobody has seemingly had any issue with it for two decades" - I believe it is a mostly meaningless statement. Some players will tell you today that the game is perfect as is and there is no need to change anything about it, all while there is a large amount of critical problems in the form of crashes, mismatches, exploits and breaking bugs. Some players have limiting abilities to accurately imagine what they like or don't like, until they experience it. When big issues are overlooked or ignored or downplayed by players, then so are small issues by even more players.

> That is certainly the only argument (and I don't believe anyone suggested otherwise), but it is an incredibly powerful argument.

It is reasonable argument, and that is why I am inclined to make this a purely optional change, that a user can manually opt-in for. The same way we have optional audio and art fixes.

But before that, I need to test the USA interface to experience it.

Stubbjax:
> Personally I found this to be a reasonable way to help judge a change with. You may try it on any future change and may be pleasantly suprised.

I may have even tried it in the past if you recall. I never suggested it was not a reasonable way to help judge a change, but that it is not an effective way of determining whether an interface is confusing or counterintuitive for new players.

> The interface of the Original is usable yes. There is a reason however many pro players nowadays use Control Bar Pro instead of the Original Control Bar. For evidence, open any POV Zero Hour livestream of a pro player on YouTube, and check whether or not they use a Control Bar Addon or the Original one. Many players are open to improvements, and reasonably so.

Sure, many players are open to improvements. This is undoubtedly reasonable. However, it is critically important to consider the cost and subjectivity of such improvements. In this case, the cost is high and the improvement is tenuous.

Control Bar Pro cannot be used as a precedential case to green-light any proposal / change like you seem to be suggesting, as it is optional and purely visual - the context is completely different. It would be a _very_ different story if Control Bar Pro were forced upon everyone.

> As for "nobody has seemingly had any issue with it for two decades" - it is a mostly meaningless statement. Some players will tell you today that the game is perfect as is and there is no need to change anything about it, all while there is a large amount of critical problems in the form of crashes, mismatches, exploits and breaking bugs. Some players have limiting abilities to accurately imagine what they like or don't like, until they experience it. 

How is it meaningless? I am highlighting that this change has very little weight to it, and is likely something nobody has really considered or desired before. I can't imagine anyone defending crashes, lag and mismatches. Players often lament bugs, exploits and balance issues. But having a stop button for builders? This is not an issue on anybody's radar. Such a change is not only highly visible, but will objectively introduce disruption and confusion for all players. This is a substantial cost, and one that must be significantly outweighed by the benefits. I do not believe it is in this case.

> Given my experience in ZH community, it will be difficult to convince me otherwise, and that is why I say your statement is mostly meaningless.

I must say, I was not expecting such a courteous and receptive response. My overall concern is that if we start changing every little inconsistency or thing that doesn't make sense, then the summation of visible changes will be too drastic, the transition to the patch will be jarring, the game will not feel like 1.04 anymore, and the patch will fail. The initial transition needs to be as seamless as possible. The rearrangement of the construction command sets goes strongly against this notion, and carries too much risk in relation to the little benefit it may provide. It is surely out of scope in terms of player expectations.

> It is reasonable argument, and that is why I am inclined to make this a purely optional change, that a user can manually opt-in for. The same way we have optional audio and art fixes.

Feel free to do so. This is certainly preferable to forcing a change with clear consequences upon everyone. But I doubt it would be worthwhile creating and maintaining two layouts for the very few players that would prefer or benefit from such a change. It would be far easier for everyone if things were left the way they are.

commy2:
I don't think players are blindly clicking where the buttons used to be. At least not for tech beyond WF. It's probably easier to find the Airfield when it's filed in logically instead of put at the end.

xezon:
This pull has been superseeded. Changes have been split into separate pulls. Closing.

PR #1580, opened 2023-01-21
Fix: Increase reload time of USA Patriot missile assist weapons from 1000 to 2000

xezon:
**This change has been reverted with #2422**

* Related to #1755

~~This change increases the reload time of USA Patriot missile assist weapons from 1000 to 2000.~~

## Original

See the below footage for reference. In 1.04, the assisting Patriot Battery reloads its second (assist) volley faster than the left Patriot. (Note: The assist weapon's `DelayBetweenShots` is set to 100 here for consistency.)

https://user-images.githubusercontent.com/11547761/221337051-f083f66a-0d9f-4140-87f5-23a95a2522a2.mp4

---

First assist volley times:

00:11.25 [355]
00:11.29 [359]
00:12.03 [363]
00:12.07 [367]

Second assist volley times:

00:13.15 [405]
00:13.19 [409]
00:13.23 [413]
00:13.27 [417]

Conclusion:

The time between the first and second assist volleys in the above demonstration is only 38 frames. Normally there are exactly 60 frames (2000ms) between continuous non-assist volleys. This change removes this discrepancy and nerfs the assist weapon of the EMP Patriot by increasing its reload from 1000ms to 2000ms.

# Rationale

Patriot missile assist weapons no longer reload twice as fast as non-assisted Patriot missile weapons. This in turn prevents the Patriot to attack a new target after just 1000 ms after an assisted attack, which makes it consistent with non-assisted Patriot missile attacks. This likely is what player would naturally expect always.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1580_patriot_assist_reload_time.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1580_patriot_assist_reload_time.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-01-21
+
+title: Increases reload time of USA Patriot missile assist weapons from 1000 to 2000
+
+changes:
+  - fix: Increases the reload time of USA Patriot missile assist weapons from 1000 to 2000. This way reload time is consistent for all weapons of the Patriot Battery and there can be no scenarios where the turret can shoot on new targets after just 1 instead of 2 seconds.
+
+labels:
+  - bug
+  - controversial
+  - minor
+  - nerf
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1580
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2053,6 +2053,7 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 ; Patch104p @fix xezon 21/01/2023 Enables automatic reload when idle.
+; Patch104p @fix xezon 21/01/2023 Changes clip reload time from 1000 to match non-assist Patriot weapons.
 Weapon PatriotMissileAssistWeapon
   PrimaryDamage               = 25.0
   PrimaryDamageRadius         = 5.0
@@ -2068,7 +2069,7 @@ Weapon PatriotMissileAssistWeapon
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
   DelayBetweenShots           = 250                   ; time between shots, msec
   ClipSize                    = 4                        ; how many shots in a Clip (0 == infinite)
-  ClipReloadTime              = 1000               ; how long to reload a Clip, msec
+  ClipReloadTime              = 2000               ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
   AutoReloadWhenIdle          = 2100
   AntiAirborneVehicle         = Yes
@@ -6727,6 +6728,7 @@ End
 ;------------------------------------------------------------------------------
 ; For use with the Assisted Targeting Update.  No Assist Listing and longer range
 ; Patch104p @fix xezon 21/01/2023 Enables automatic reload when idle.
+; Patch104p @fix xezon 21/01/2023 Changes clip reload time from 1000 to match non-assist Patriot weapons.
 Weapon Lazr_PatriotMissileAssistWeapon
   PrimaryDamage               = 40.0
   PrimaryDamageRadius         = 3.0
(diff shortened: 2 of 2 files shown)


commy2:
This should be thoroughly tested before merge.

Stubbjax:
Isn't this change a nerf? I can hear echoes "This is bad because it nerfs SWG therefore we should not merge it".

xezon:
I was not able to observe difference. Assisting Patriots already reloaded in same times. Perhaps because of `ShareWeaponReloadTime = Yes`.

```
  WeaponSet
    Conditions          = None
    Weapon              = PRIMARY     PatriotMissileWeapon
    Weapon              = SECONDARY   PatriotMissileAssistWeapon
    Weapon              = TERTIARY    PatriotMissileWeaponAir
    AutoChooseSources   = PRIMARY     FROM_PLAYER FROM_AI FROM_SCRIPT
    AutoChooseSources   = SECONDARY   NONE
    AutoChooseSources   = TERTIARY    FROM_PLAYER FROM_AI FROM_SCRIPT
    PreferredAgainst    = PRIMARY     VEHICLE INFANTRY
    PreferredAgainst    = SECONDARY   BALLISTIC_MISSILE AIRCRAFT
    PreferredAgainst    = TERTIARY    BALLISTIC_MISSILE AIRCRAFT
    ShareWeaponReloadTime = Yes
  End
```

xezon:
Anything to add?

commy2:
There is no reason to make this change if there is no observable difference.

xezon:
I do not know if it does or does not do something.

MTKing4:
if we don't know what it does, we probably shouldn't be messing with it, could break something we're unaware of

xezon:
I prefer this potential problem to be tested and figured out instead of being ignored.

MTKing4:
sure, we can find out, then we can know what to do with it

Stubbjax:
> I prefer this potential problem to be tested and figured out instead of being ignored.

Why was this not done before opening the PR?

xezon:
> Why was this not done before opening the PR?

Because author does not know how to test this properly.

Stubbjax:
> > Why was this not done before opening the PR?
> 
> Because author does not know how to test this properly.

Then why open the PR?

xezon:
It invites someone else to test it. In this case commy2 assigned it to himself, indicating that he would want to test it at some point in time.

Stubbjax:
> It invites someone else to test it. In this case commy2 assigned it to himself, indicating that he would want to test it at some point in time.

A PR should be a change that you have already tested and assured of quality and want additional eyes to confirm such quality and catch issues before merging. This doesn't seem to be the case here.

Stubbjax:
I went and tested this anyway. This change does have an impact on the reload time when a Patriot switches assist targets after having fired.

See the below footage for reference. In 1.04, the assisting Patriot Battery clearly reloads its second (assist) volley faster than the left Patriot. (Note: The assist weapon's `DelayBetweenShots` is set to 100 here for consistency.)

https://user-images.githubusercontent.com/11547761/221337051-f083f66a-0d9f-4140-87f5-23a95a2522a2.mp4

---

First assist volley times:

00:11.25 [355]
00:11.29 [359]
00:12.03 [363]
00:12.07 [367]

Second assist volley times:

00:13.15 [405]
00:13.19 [409]
00:13.23 [413]
00:13.27 [417]

Conclusion:

The time between the first and second assist volleys in the above demonstration is only 38 frames. Normally there are exactly 60 frames (2000ms) between continuous non-assist volleys. This change removes this discrepancy and nerfs the assist weapon of the EMP Patriot by increasing its reload from 1000ms to 2000ms.

_I can hear those echoes again._

xezon:
Thank you for investigation. Now everything is clear. I have updated change description and have split this change into two: this one for the nerfed reload time, and #1755 for buffed auto reload when idle.

MTKing4:
Now that we know what this change does, we should close it :D 

We have no real consensus to do thie change, Normal Patriots are bad and don't need a nerf, and SWG is quite weak and relies a lot on it's EMPs

xezon:
I think this should be merged, along with the buff #1755. The annoying thing about quirks like this is that it makes setups more complicated and more difficult to tweak accordingly because things do not work the way one would expect them to work.

penfriendz:
If it makes it easier and you know the effects, I think it's okay. I bet no one would notice anyway.

xezon:
Looks good.

commy2:
Why?

MTKing4:
Indeed why

We expressed that this isn't a good change yet you have merged it anyway

xezon:
This is merged because this is the way it should be. Whether or not this is an ever so slight Nerf is inconsequential. Patriot Battery was already buffed with Auto Reload added and there is this #1412 which would Buff Patriot a thousand times more than this Nerf.

MTKing4:
> This is merged because this is the way it should be.

According to what basis? That is a poor reason to make such balance altering change which the majority have agreed not to merge 

I don't see any convincing argument do this change, there's no consensus here, you can't just change something just so that it looks good to you, normal attack and assist don't need to have the same rate

xezon:
If we went with your logic, we would need to revert the following 2 bug fixes too: #294, #1334. This issue is literally the same thing, just a little less extreme than the other 2.

MTKing4:
Those are not the same thing, those are double fire and are due to having two weapons iirc and that reads more like a bug, where this one shows some sign of intent to be this way, for being assisted, also we the majority agreed on it, this one we didn't, this is not my concern alone

commy2:
It's not a bug. Assisting Patriot fires faster and aims at farther away targets. That's how they're supposed to snowball in groups.

xezon:
The fire rate of the assisting Patriot is actually bound to its master Patriot, 2000 ms. It only fires faster when switching from an old Master to a new Master, min 1000 ms. I consider this a bug, because it can effectively halve its fire rate by switching targets, and only then. In every other scenario the Battery will reload in 2000 ms, also when assisting.

MTKing4:
Wait so assisting the same pat will have them shoot at 2000ms? Hmm

xezon:
> Wait so assisting the same pat will have them shoot at 2000ms? Hmm

Yes. Because they wait for the Master Patriot to shoot. Then they shoot at the same time.

PR #1587, opened 2023-01-22
Fix: Change damage type of Gattling Air weapons from SMALL_ARMS to GATTLING

xezon:
This change fixes damage types of Gattling Air guns to match their Gattling Ground guns counterparts. Instead of `SMALL_ARMS` they now use `GATTLING`.

All Armor Sets for air units, except `AirDroneArmor`, use identical armor modifiers for `SMALL_ARMS` and `GATTLING`. `AirDroneArmor`'s `GATTLING` was adjusted accordingly. Therefore this change does not modify unit damages.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Armor.ini
@@ -1012,12 +1012,14 @@ Armor FireBaseArmor
   Armor = SUBDUAL_BUILDING    100%
 End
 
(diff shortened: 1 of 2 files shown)


xezon:
Any comments on this?

penfriendz:
This nerfs range buggers, so it's good for me.

xezon:
It is supposed to nerf or buff nothing.

Stubbjax (approved):
Should have been two separate commits but otherwise seems fine.

xezon:
Split into two commits.

PR #1588, opened 2023-01-22
Fix: Standardize the China Dragon Tank's upgraded primary weapon's minimum range with its non-upgraded weapon

xezon:
This change standardizes the China Dragon Tank's upgraded primary weapon's minimum range with its non-upgraded weapon.

~~Gameplay wise this is inconsequential, because the Dragon Tank cannot hit targets at those small distances with its primary weapon anyway.~~

shot_20230122_180543_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -3091,7 +3091,7 @@ Weapon DragonTankFlameWeaponUpgraded
   SecondaryDamageRadius       = 10.0
   ScatterRadiusVsInfantry     = 0.0     ;When this weapon is used against infantry, it can randomly miss by as much as this distance.
   AttackRange                 = 75.0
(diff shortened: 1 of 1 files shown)


MTKing4:
I don't understand this change, does a Black Napalm Dragon tank have a different minimum range set?

commy2:
In 1.04, the Dragon tank has no minimum range before Black Napalm.

After Black Napalm, it erroneously gains a minimum range.

This makes Black Napalm a sideways-grade instead of a direct upgrade.

xezon:
An alternative would be to set a reasonable minimum attack range, to where the Dragon Tank will actually hit its target properly.

ExiLe-ZH:
A minimum range would mean the unit turns around and moves back when you want to attack too upclose? (A situation that can also be caused by lag).

xezon:
Yes, Dragon Tank would then move away from target to acquire valid firing distance.

commy2 (approved):

PR #1589, opened 2023-01-22
Fix: China Dragon Tank can now hit targets at different elevations with the primary weapon

### LINKED ISSUE #20: Under certain conditions China Dragon Tank can shoot on bunker but does not clear its occupants
Under certain conditions China Dragon Tank can shoot on bunker but does not clear its occupants. Reproducible on Defcon 6 map with white bunkers in front of main bases. Does not happen always.

xezon:
* Fixes TheSuperHackers/GeneralsGameCode#20
* Related to TheSuperHackers/GeneralsGamePatch#1417

## Original

A `MaxThrustAngle` of `1` prevents the projectile from hitting targets that are on a different height / angle to the Dragon Tank.

https://user-images.githubusercontent.com/11547761/199674100-9406518a-81ec-4807-a07d-791a79e7d52f.mp4

## Patched

A `MaxThrustAngle` of `30` allows the projectile to hit targets at a different height / angle to the Dragon Tank. Values of `20` and `25` would not hit all the Red Guard effectively.

https://user-images.githubusercontent.com/11547761/199674127-275a265a-3d46-414e-8753-720f01d37867.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
@@ -3328,7 +3328,7 @@ Locomotor DragonTankFlameLocomotor
   Acceleration = 2160       ; in dist/(sec^2)
   Braking = 0               ; in dist/(sec^2)
   TurnRate = 5            ; in degrees/sec
-  MaxThrustAngle = 1       ; in degrees (NOT degrees/sec)
+  MaxThrustAngle = 30       ; in degrees (NOT degrees/sec) ; Patch104p @bugfix Stubbjax 03/11/2022 Fixed flame projectiles not hitting targets on different terrain heights / angles.
   AllowAirborneMotiveForce = Yes
   Appearance = THRUST
 End


xezon:
Is angle of 30 good as is? Does it need to be higher for some maps eventually? Is TheSuperHackers/GeneralsGameCode#20 confirmed fixed?

MTKing4:
It looks good for this situation, but we would have to test if that's sufficient to solve TheSuperHackers/GeneralsGameCode#20 first

xezon:
`MaxThrustAngle = 30` does not only allow projectile to curve up and down, but also left and right. This in turn makes the weapon much better at hitting moving targets.

https://user-images.githubusercontent.com/4720891/199827147-64749361-19bc-48c4-9944-60548bb554af.mp4

commy2:
Seems like a non-starter then.

xezon:
Yes. It appears we need an alternative way to fix this. Or close.

MTKing4:
What about `MaxThrustAngle = 20` ?

xezon:
Any `MaxThrustAngle` larger than the original 1 will increase flame (missile) performance.

MTKing4:
Sure but I'm trying to find a medium where it makes it better vs elevated Bunkers while not having a impactful Difference in performance

xezon:
I think this locomotor change needs more testing in isolation of `TryToFollowTarget = Yes`, TheSuperHackers/GeneralsGamePatch#1417. It is possible that locomotor change is preferred, but both changes together will make flame projectiles curve.

MTKing4:
Me and @ImTimK can test that

Stubbjax:
> `MaxThrustAngle = 30` does not only allow projectile to curve up and down, but also left and right. This in turn makes the weapon much better at hitting moving targets.

Did you test this only on allies again? The higher max thrust angle allows it to hit allies, which is what your video shows. I don't see any left / right curvature.

xezon:
The curving is illustrated in this hidden video. It only happens with these settings in combination.
https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1417#issuecomment-1407169341

Stubbjax:
> The curving is illustrated in this video. It only happens with these settings in combination. [#1417 (comment)](https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1417#issuecomment-1407169341)

I set the `MaxThrustAngle` to 90 and see no curvature to the projectile at all.

https://user-images.githubusercontent.com/11547761/215337413-b89834e1-a161-4d9b-8b80-f561d1bb2ab1.mp4

xezon:
You need to set `MaxThrustAngle=90` + `TryToFollowTarget=Yes`, then it curves.

Stubbjax:
> You need to set `MaxThrustAngle=90` + `TryToFollowTarget=Yes`, then it curves.

I thought we were just talking about this change?

xezon:
The important thing to note here is that changes to `MaxThrustAngle` and `TryToFollowTarget` cannot be combined, as that will fundamentally change missile performance. Therefore we need to choose one of them. `MaxThrustAngle` may be the better change, but it is not yet clear to me.

Stubbjax:
> The important thing to note here is that changes to `MaxThrustAngle` and `TryToFollowTarget` cannot be combined, as that will fundamentally change missile performance. Therefore we need to choose one of them. `MaxThrustAngle` may be the better change, but it is not yet clear to me.

But they both do entirely separate things. They can certainly be combined in the right way. For example, `TryToFollowTarget` will have no impact on projectile motion if the locomotor's `TurnRate` is set to 0.

xezon:
I cannot recall whether or not I tested that.

What happens with `MaxThrustAngle=30`, `TryToFollowTarget=Yes`, `TurnRate=0`?

I certainly tested `MaxThrustAngle=30`, `TryToFollowTarget=Yes`, `TurnRate=5` and that did curve.

Stubbjax:
> I cannot recall whether or not I tested that.
> 
> What happens with `MaxThrustAngle=30`, `TryToFollowTarget=Yes`, `TurnRate=0`?

This seems to be a good combo. The projectile does not follow units with a turn rate of 0.

xezon:
> This seems to be a good combo. The projectile does not follow units with a turn rate of 0.

Does not work. Breaks the hill fix, because flame can no longer turn up-/downhill.

I think fix requires new Locomotor setting:

`MaxThrustPitch = 30` as opposed to `MaxThrustAngle = 30`. Then it should work.

xezon:
```
MaxThrustPitch = 30
ThrustRoll = 0
```
Does not prevent sideways curve either.

xezon:
I think we need to close this unfortunately. Unless someone has a clever idea.

MTKing4:
just do the `MaxThrustAngle=30` and keep `TryToFollowTarget=No`

meaning we only fix the angle and leave minimum range as is, will that work?

xezon:
This change makes flame weapon curve in all directions. That is why I proposed to add a new `MaxThrustPitch` locomotor setting.

Closing.

MTKing4:
maybe restore the branch for further testing? because i see conflicting test results here and i like to test for myself

> I set the `MaxThrustAngle` to 90 and see no curvature to the projectile at all.


> https://user-images.githubusercontent.com/11547761/215337413-b89834e1-a161-4d9b-8b80-f561d1bb2ab1.mp4

xezon:
It is unclear to me why there is observation of non curving missiles. On my machine `MaxThrustAngle = 30` makes fire curve sideways. When combining it with `TurnRate = 0` then it does not curve sideways, but also no longer up and down hill.

xezon:
Branch is restored. Feel free to test.

Stubbjax:
> It is unclear to me why there is observation of non curving missiles. On my machine MaxThrustAngle = 30 makes fire curve sideways.

Where does this sideways curving behaviour occur? I have not experienced any such behaviour across dozens of tests and we are yet to be presented with any evidence of such.

All `MaxThrustAngle = 30` does is allow the projectile to fire within a 30 degree pitch of the barrel. This means the projectile can be fired more directly at the target. It doesn't make the projectile shoot out sideways if that's what you're suggesting.

Which of the below projectiles do you perceive curving?

https://user-images.githubusercontent.com/11547761/221389264-ad3ab765-201e-4a20-924e-9634cf86b61c.mp4

https://user-images.githubusercontent.com/11547761/221389267-48c787e9-eacc-4dfc-ae76-f5ce015defda.mp4

xezon:
> Which of the below projectiles do you perceive curving?

In the bottom video the projectile curves towards the vehicle. In the top video it does not.

I am confused as to why no one else is able to see it.

xezon:
Here is illustration of curving observation to try highlight the problem once more.

In top projectile, missiles trail behind the white line because the target is moving while the projectiles fly. In bottom projectile, missiles curve to object before the white line when the target is moving.

This enhances combat performance of Dragon Tank in unexpected way. That is why I suggested to wait for Thyme and implement a "MaxThrustPitch" locomotor setting to fix this properly.

curving

MTKing4:
Just tested, Xezon is Correct, Dragon tank now shoots at where the unit will be in the next frame rather than where it was, I think i can understand that `MaxThrustAngle = 30` is like an Aim Cone where it makes rockets hit anything inside that angle or Cone, so it does make it shoot from a wider angle, the but it calling Curving is misleading imo, as we expected the rocket path was Curving, this is not the case, it's not really curving, the rocket path is Still Straight, it's just shooting from a wider angle, therefore applying more damage

the Test Below Shows the Difference Very Clearly, I put up the Patched Video First to highlight the regression

## Patched


https://user-images.githubusercontent.com/83613147/221415372-7c1ed2ba-18c7-44ae-803f-8e2d73db838f.mp4


## Original

https://user-images.githubusercontent.com/83613147/221415332-f28c0aa5-58a0-46c2-be59-d65d96355a2a.mp4

MTKing4:
I think we should change the title of this Pull and call it a buff, in case we want such a buff for the dragon tank in the future.

I gotta say I like the new dragon tank, there could be a reason to buff it this way as it sucks vs most attacking units

you can keep this closed for now.

xezon:
I think curving is pretty accurate description for this. The missile does actually deviate from its forward trajectory with this change. But regardless, good to see that the observation has been replicated on another client.

Stubbjax:
> Here is illustration of curving observation to try highlight the problem once more. In top projectile, missiles trail behind the white line because the target is moving while the projectiles fly. In bottom projectile, missiles curve to object before the white line when the target is moving.

You've drawn lines over projectiles that each have a different target position / trajectory, which is unhelpful. The first projectiles have a target position that is further away, whereas the second are a lot closer. All this talk about curvature is misleading. What this change does is allow the projectiles to more precisely hit the target location. I've tested this at great lengths where it is extremely obvious that the projectiles hit precisely where the target is / was. There is no sideways curvature of the projectile whatsoever.

> the Test Below Shows the Difference Very Clearly, I put up the Patched Video First to highlight the regression

This test is inconclusive because it was performed against allies, which do not collide with the projectile and would instead be greatly affected by this change. It's also two separate games with different inputs / actions which makes changes harder to read. However, it's evident that the projectile is more precise in the patched footage as the impact effects are visibly reduced.

> I think i can understand that MaxThrustAngle = 30 is like an Aim Cone where it makes rockets hit anything inside that angle or Cone, so it does make it shoot from a wider angle, the but it calling Curving is misleading imo, as we expected the rocket path was Curving, this is not the case, it's not really curving, the rocket path is Still Straight, it's just shooting from a wider angle, therefore applying more damage

This is only half-correct. The cone is strictly vertical, not horizontal.

> I think curving is pretty accurate description for this. The missile does actually deviate from its forward trajectory with this change. But regardless, good to see that the observation has been replicated on another client.

I think curving is a very inaccurate description. What is actually happening is the projectile is aimed more towards the base of the vehicle instead of directly at / past it. This results in the projectile travelling a shorter distance and colliding with the ground much closer to the vehicle instead of evaporating in the air further away when it misses. It can be likened to firing rockets against other players in games like Quake or Team Fortress 2, where you ideally target the enemy's feet instead of shooting directly at them so that the blast still impacts them if the rocket misses and hits the ground. Firing towards the base of the unit is what allows units on elevated terrain to be hit. Projectile trajectories are otherwise very clunky when it comes to elevated terrain.

Nevertheless, it seems as though weapon / projectile logic involving elevated terrain is a [complete mess](https://user-images.githubusercontent.com/11547761/221446411-3a4d5b23-d975-4c2e-9e4b-7dedb0a1f140.mp4) anyway, and there may not be a suitable or elegant solution to the respective issue from ini scripting alone.

There is also a [bug in 1.04 where the flame projectile detonates twice](https://user-images.githubusercontent.com/11547761/221444262-40bcfa20-9996-451e-9a77-204843146fb5.png) - once when the projectile runs out of fuel and a second time when the dead projectile hits the ground further away. This bug causes damage to be applied in both locations. The increased max thrust angle would fix this as well.

xezon:
It is disappointing to read that my feedback on this change is viewed as 'unhelpful', when all I try to raise is a legitimate concern that no one else was able to spot.

xezon:
> I think curving is a very inaccurate description. What is actually happening is the projectile is aimed more towards the base of the vehicle instead of directly at / past it.

I took a look at the corresponding game code.

https://github.com/xezon/Thyme/blob/develop/src/game/logic/object/locomotor.cpp#L1362-L1374

```c++
void Locomotor::Move_Towards_Position_Thrust(
    Object *obj, PhysicsBehavior *physics, const Coord3D &goal_pos, float on_path_dist_to_goal, float desired_speed)
{
...
    float max_turn = Get_Max_Turn_Rate(damage);
    Vector3 dir;
    Calc_Direction_To_Apply_Thrust(obj, physics, pos, max_accel, dir);
    float max_thrust_angle;

    if (max_turn > 0.0f) {
        max_thrust_angle = m_template->m_maxThrustAngle;
    } else {
        max_thrust_angle = 0.0f;
    }

    Vector3 actual_dir;
    float f2 = Try_To_Rotate_Vector_3D(max_thrust_angle, cur_dir, dir, actual_dir);
...
}
```
Calculates direction to target object, and then rotates current projectile direction to new projectile direction by the amount of max thrust angle. This is in line with what I have been explaining.

Stubbjax:
> In top projectile, missiles trail behind the white line because the target is moving while the projectiles fly. In bottom projectile, missiles curve to object before the white line when the target is moving.

> Calculates direction to target object, and then rotates current projectile direction to new projectile direction by the amount of max thrust angle. This is in line with what I have been explaining.

Would you kindly explain what you mean by 'curving'? Below I have drawn a line from the turret to the target as you have demonstrated. Do you consider any projectiles that don't align with the drawn line as curving?

image

xezon:
Curving, aka turning, means the projectile takes a turn like a car would do on a curved road. This is evident in the math of `Locomotor::Move_Towards_Position_Thrust`, and also by observation.

In your above image, the original behaviour I would not label curving, because trajectory is mostly straight forward. If the outer projectiles were closer the the white line or on the other side, then trajectory is curved.

ExiLe-ZH:
wouldn't (some degree of) 'tracking' be a more accurate term?

xezon:
I do not think that terminology is the fundamental issue here. It just so happened that MaxThrustAngle was/is not properly understood. I raised this issue 4 months ago already with this post:

https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1417#issuecomment-1302639737

"Tracking" works for me. Code refers to this as "turning", "thrust angle". I called it curving, because this is what happens when a thing takes a turn. Like the car turning on the curved road. It should not cause confusion, because all these words refer to the same thing.

Stubbjax:
> Curving, aka turning, means the projectile takes a turn like a car would do on a curved road. This is evident in the math of `Locomotor::Move_Towards_Position_Thrust`, and also by observation.

Are you suggesting that the Dragon Tank's projectile will laterally turn (or 'curve') towards the targeted unit?

image

MTKing4:
I'm Going to Restore this Branch for testing purposes

MTKing4:
Ok i've had this test prepared since the last time we discussed this but got very busy so couldn't finish it, but at last, here it is

> This test is inconclusive because it was performed against allies, which do not collide with the projectile and would instead be greatly affected by this change. It's also two separate games with different inputs / actions which makes changes harder to read. However, it's evident that the projectile is more precise in the patched footage as the impact effects are visibly reduced.

At the time it felt to me that this was an argument for the sake of argument because from my test I couldn't see why being against allies or enemies makes any difference here especially since you agree that the dragon is more precise now which was the main point, but regardless, I made a conclusive test, against AI, Scripted Movement in a circle with Identical waypoints So no room For Variation and Human Error, with multiple Units with different speeds so that the difference now is crystal clear.

> This is only half-correct. The cone is strictly vertical, not horizontal.

if it's only vertical why is it making the tank shoot at a wider angle horizontally? 

anyway here are the results:

# Rocket Buggy

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/9c41d64d-0ebd-4c7e-9f1e-f1896770bd76


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/828fd0fd-69dc-4215-9580-e51bbafe667d

MTKing4:
# Emperor

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/89baddac-bfe5-4a91-811d-2c554c514a9d


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/a3b0fc1f-a3d0-4362-a296-ed3727e1a712

MTKing4:
# Nuke Overlord

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/0f469087-4187-480f-80f8-c05e28920cb1


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/2448044c-905b-475b-8835-7e1e2921046f

MTKing4:
# Quad

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/51cb169d-260f-4fa3-8cba-10d182106d38


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/cb406384-5d38-4547-8acc-755ade19f6b5

MTKing4:
# Technical

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/aaf3ee79-9c79-4fe6-ab1f-42638cceff06


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/da1dd78d-4b9c-4d6c-b807-429d1def9b1b

MTKing4:
# Humvee

### Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/57fb4c33-e97c-4b7f-85a1-6dac5bd222c8


### Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/3836c527-a1ca-40d7-93bb-e808f8c3a435

MTKing4:
if you want any other Scripted Movement test maps let me know

xezon:
Delete branch again?

MTKing4:
I'm done testing, you can delete, unless someone wants to test further.

xezon:
I think this needs no further testing, because the results are very conclusive. It just needs to be understood :-)

ExiLe-ZH:
> > Curving, aka turning, means the projectile takes a turn like a car would do on a curved road. This is evident in the math of `Locomotor::Move_Towards_Position_Thrust`, and also by observation.
> 
> Are you suggesting that the Dragon Tank's projectile will laterally turn (or 'curve') towards the targeted unit?
> 
> image

So conclusion is it does this?

MTKing4:
More like the tank is now shooting at where the target will be instead of shooting directly at it, which makes it less likely to trail behind the moving unit

xezon:
> More like the tank is now shooting at where the target will be instead of shooting directly at it, which makes it less likely to trail behind the moving unit

This is not correct.

I have already provided concise accurate description above:

> Calculates direction to target object, and then rotates current projectile direction to new projectile direction by the amount of max thrust angle.

commy2:
I think to fix this without making the Dragon more effective against circling vees, this needs a MaxThrustElevation setting that only rotates the projectiles inclination and not azimuth.

xezon:
Faith has been restored :P

commy2:
tbf, I worked on projectile simulation in Arma 3, specifically weapon zeroing and tank/Panzerfaust fire control systems.

commy2:
Concise explanation would be, that MaxThrustAngle turns the Dragon tank flame projectiles into homing missiles. (resulting in anterior bent curves)

... while the unpatched version curves posterior, because particle lines are drawn between projectile(s) and turret muzzle.

MTKing4:
You've lost me at azimuth.

MTKing4:
What i couldn't accept about the idea that it's changing the direction of the projectiles mid air, aka tracking or 'homing' is that it can still miss fast moving vehicles like technical, if it's homing, shouldn't it always hit?

commy2:
> You've lost me at azimuth.

Compass direction, meaning, everything you observe on the horizon (inclination=0, azimuth=0-360 degree). If you looked at the north star, inclination would be 90 degree, if you looked at the core, it would be minus 90 degree. 

> if it's homing, shouldn't it always hit?

No, because it can only correct for, apparently 30 degree per time step (maxThrust = 30). If the target has an angular momentum > 30 degree, i.e. it is very close and/or very fast, the missile will still miss.

MTKing4:
Oh ok, that is much more complicated than i thought it was :D
Also you could've just said Vertical and horizontal :P

commy2:
horizontal, vertical and distal, I suppose. I'm going of the [Spherical coordinate system](https://en.wikipedia.org/wiki/Spherical_coordinate_system#Definition), which may describe `Try_To_Rotate_Vector_3D()` best, but there may also be optimizations/ simplifications considering this is an early 3d engine.

Here, have some peak German culture featuring a homing missile (and more):

https://www.youtube.com/watch?v=GAoL_Z3Q5Z4

PR #1597, opened 2023-01-23
Fix: Add bomb falling sounds to ClusterMinesBomb, EMPPulseBomb

### LINKED ISSUE #1592: China EMP and Mine Planes have no sound for the bomb drop
Bomb drop sound usually indicate something dangerous is dropping, while that does sound good for EMP, I'm not sure i like it on the mine drop, but it does make sense that the bomb should have a sound when dropping

xezon:
* Fixes #1592


Stubbjax (approved):
https://user-images.githubusercontent.com/11547761/215516710-5354c8c5-b5e0-4515-9e7a-40b68e2e55ef.mp4

xezon:
Thank you for posting video.

PR #1606, opened 2023-01-26
Fix: Double scorch on death of China Battlemasters and Overlords before Isotope Stability and after Nuclear Tanks upgrades

xezon:
**Merge with Rebase**

This change fixes double scorch on death of
* Nuke Battlemaster before Isotope Stability upgrade
* Nuke Overlord before Isotope Stability upgrade
* China/Tank/Boss/CINE Battlemaster after Nuclear Tanks upgrade
* China/Tank/Boss/CINE Overlord after Nuclear Tanks upgrade

- [x] Rename FX_Suicide_BattleMasterExplosionOneFinal

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
@@ -2726,6 +2726,38 @@ FXList FX_OverlordExplosionOneFinal
   End
 End
 
+; -----------------------------------------------------------------------------------
+; Patch104p, A copy of FX_OverlordExplosionOneFinal, scorchless variant.
+; A death weapon is supposed to spawn the terrain scorch.
+; -----------------------------------------------------------------------------------
+FXList ScorchlessFX_OverlordExplosionOneFinal
+  ParticleSystem
+    Name = OverlordExplosionSmoke
+  End
+  ParticleSystem
+    Name = OverlordSubExplosionSmoke
+    Radius = 15 30 UNIFORM
+    Height = 0 20 UNIFORM
+    InitialDelay = 167 667 UNIFORM
+  End
+  ParticleSystem
+    Name = OverlordExplosionDebris
+  End
+  ParticleSystem
+    Name = OverlordExplosionLenzflare
+    Offset = X:0.0 Y:0.0 Z:10.0
+  End
+  ParticleSystem
+    Name = OverlordExplosionShockwave
+  End
+  ViewShake
+    Type = STRONG
+  End
+  Sound
+    Name = TankDie
+  End
+End
+
 ; -----------------------------------------------------------------------------
 FXList FX_OverlordDamageTransition
   ParticleSystem
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
@@ -17305,13 +17305,30 @@ Object Nuke_ChinaTankOverlord
   End
 
   ; Just explode death
+  ; Patch104p @tweak xezon 02/02/2023 Use different death behaviour with different effect(s) when upgrade is acquired.
+  Behavior = StatusBitsUpgrade ModuleTag_StatusBitsUpgrade
+    TriggeredBy = Upgrade_ChinaIsotopeStability
+    StatusToSet = STATUS_RIDER1
+  End
+
   Behavior = SlowDeathBehavior ModuleTag_16
     ProbabilityModifier = 25
     DestructionDelay = 200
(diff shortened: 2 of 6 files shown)


Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
```diff
@@ -2750,6 +2750,38 @@ FXList FX_OverlordExplosionOneFinal
   End
 End
 
+; -----------------------------------------------------------------------------------
+; Patch104p, A copy of FX_OverlordExplosionOneFinal, suicide variant
+; A death weapon spawns the scorch.
+; -----------------------------------------------------------------------------------
+FXList FX_Suicide_OverlordExplosionOneFinal
```
commy2:
```suggestion
FXList Nuke_FX_OverlordExplosionOneFinal
```
so: 

- both appear when grepping `FX_OverlordExplosionOneFinal`
- it's obvious that this is a Nukegen exclusive variant (name convention).
- Suicide is the keyword for the Demogen ability, not for Nuke shenanigans

Why not make this effect spawn the (presumably bigger) scorch mark? Meaning, remove the scorch mark from the death weapon and keep that responsibility on SlowDeathBehavior for all vehicles (consistency).

Is the before Isotope upgrade scorch mark bigger?

Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
```diff
@@ -17069,14 +17069,30 @@ Object Nuke_ChinaTankBattleMaster
...
+    StatusToSet = STATUS_RIDER1
+  End
 
   Behavior = SlowDeathBehavior ModuleTag_12
+    DeathTypes = ALL -CRUSHED -SPLATTED
+    ProbabilityModifier = 25
+    DestructionDelay = 200
+    DestructionDelayVariance = 100
+    OCL = FINAL    OCL_ChinaTankBattleMasterDebris
+    FX  = FINAL    FX_Suicide_BattleMasterExplosionOneFinal
```
commy2: Doesn't exist yet?

xezon: `FX_Suicide_BattleMasterExplosionOneFinal` already exists from previous change.

Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
xezon: Yes we can change name.

I will need to check about spawning scorch via SlowDeath effect instead of death weapon. As for other units with death weapons, I also opted to spawn scorch with death weapon. I think reason being, that death weapon can eventually be delayed, so if scorch is tied to death weapon effects, then there is strong guarantee that it has proper timing.

xezon: I checked reason for scorchless slow death effect + weapon scorch vs the other way around.

There are 2 reasons:

1. `SlowDeathBehavior` can have `DestructionDelay`, but `FireWeaponWhenDeadBehavior` not.
2. An object can have different death weapons depending on upgrades, for example Bomb Truck.

For these reasons the death weapon bound scorch is good setup. I have done it this way for all the scorch changes and fixes.

Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
xezon: I can rename the `FX_Suicide_*` to `ScorchlessFX_*` in follow up change.

Patch104pZH/GameFilesEdited/Data/INI/FXList.ini
xezon: Effect renamed to `ScorchlessFX_OverlordExplosionOneFinal`

xezon:
Merge #1617 before this and then rename `FX_Suicide_BattleMasterExplosionOneFinal` accordingly.

Patch104pZH/GameFilesEdited/Data/INI/Object/NukeGeneral.ini
xezon: Renamed.

xezon:
I noticed that the same fix needs to be applied to the other Battlemasters and Overlords. With their Nuclear Tanks upgrade the situation is inverted, but the same principles apply.

New fixes pushed, split into 4 commits. Change description updated accordingly.

xezon:
Tested with China Battlemaster, Overlord + Nuclear Tanks, Nuke Battlemaster, Overlord + Isotope Stability. Worked.

xezon:
Looks good to me.

PR #1611, opened 2023-01-28
Streamline KindOf flags of hulk objects

xezon:
**Merge with Rebase**

This change attempts to fix some hulk flags misconfigurations.

* Hulks of USA Microwave, Humvee, China Troop Crawler(s), Listening Outpost have obsolete IMMOBILE flag removed
* All hulks can now cast reflections

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -134,7 +134,8 @@ Object AmericaVehicleTomahawkHulk
   EditorSorting       = DEBRIS
 
   ; *** ENGINEERING Parameters ***
-  KindOf              = NO_COLLIDE HULK
+  ; Patch104p @fix xezon 28/01/2023 Add flag for casting reflections.
+  KindOf = CAN_CAST_REFLECTIONS NO_COLLIDE HULK
 
   Behavior                  = PhysicsBehavior ModuleTag_03
     Mass                    = 10.0
@@ -168,7 +169,8 @@ Object ChinaVehicleNukeCannonHulk
   EditorSorting       = DEBRIS
 
   ; *** ENGINEERING Parameters ***
-  KindOf              = NO_COLLIDE HULK
+  ; Patch104p @fix xezon 28/01/2023 Add flag for casting reflections.
+  KindOf = CAN_CAST_REFLECTIONS NO_COLLIDE HULK
 
   Behavior                  = PhysicsBehavior ModuleTag_03
     Mass                    = 10.0
@@ -206,7 +208,8 @@ Object AmericaJetRaptorHulk
   EditorSorting       = DEBRIS
 
   ; *** ENGINEERING Parameters ***
-  KindOf              =  NO_COLLIDE HULK
+  ; Patch104p @fix xezon 28/01/2023 Add flag for casting reflections.
+  KindOf = CAN_CAST_REFLECTIONS NO_COLLIDE HULK
 
   Behavior                  = PhysicsBehavior ModuleTag_03
     Mass                    = 2.0
@@ -246,7 +249,8 @@ Object AmericaJetA10Hulk
   EditorSorting       = DEBRIS
 
   ; *** ENGINEERING Parameters ***
-  KindOf              =  NO_COLLIDE HULK
+  ; Patch104p @fix xezon 28/01/2023 Add flag for casting reflections.
+  KindOf = CAN_CAST_REFLECTIONS NO_COLLIDE HULK
 
   Behavior                  = PhysicsBehavior ModuleTag_03
     Mass                    = 2.0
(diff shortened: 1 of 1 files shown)


commy2:
UNATTACKABLE does nothing if no ActiveBody (or similar) module exists.

commy2:
What does "casting reflections" mean? I don't remember reflections in the game.

xezon:
> What does "casting reflections" mean? I don't remember reflections in the game.

`CAN_CAST_REFLECTIONS` means that an object can draw in mirrors. I do not know about such mirrors, but I see corresponding code in Thyme.

commy2:
I wouldn't change UNATTACKABLE and CAN_CAST_REFLECTIONS flags if these don't do anything.

xezon:
I will leave `CAN_CAST_REFLECTIONS`. If the vehicles cast reflections, then logically so should the hulks.

xezon:
I will revert the `UNATTACKABLE` if not needed.

xezon:
Added `UNATTACKABLE` flags are removed.

Stubbjax:
> All vehicle hulks can now bounce

What does this look like?

(This line is also a bit unclear, as the same can be said whether you changed one hulk or all of them.)

xezon:
The bounce change was affecting USA Microwave, Humvee, China Troop Crawler(s), Listening Outpost only.

I took another look at bouncing and was not able to observe difference with removed `IMMOBILE` flag, neither with `AllowBouncing` set to No.

So looks like it does nothing tangible. I renamed the commit accordingly.

## Original

https://user-images.githubusercontent.com/4720891/216760009-0243a08d-5ceb-4702-ba32-5685d1c45ef8.mp4

## Patched, Immobile flag removed

https://user-images.githubusercontent.com/4720891/216760024-9f7bec8b-4b76-48ed-ae26-cf8aa61a2d34.mp4

Stubbjax:
Hmm, there doesn't seem to be a lot of difference. The bounces might look a bit smoother on the second video, but it's hard to tell whether that's just coincidence. If there is no tangible benefit, maybe it is worth considering reapplying `IMMOBILE` to all hulks for performance reasons (assuming this flag negates some physics calculations)?

xezon:
The vast majority of hulks do not have IMMOBILE flag. I would need to get better understanding of IMMOBILE to judge whether or not hulks should have it.

The relevant code is in `PhysicsBehavior::On_Collide` function. It is not yet clear to me that code is 100% correct, because `immobile1` variable is unused.

https://github.com/TheAssemblyArmada/Thyme/blob/develop/src/game/logic/object/update/physicsupdate.cpp#L447-L576

Stubbjax:
From testing, it doesn't seem to have any effect on physics.

In the below footage, the bottom Overlord has `IMMOBILE` while the top Overlord does not. They both bounce around in the same manner.

https://user-images.githubusercontent.com/11547761/216765174-248bfa18-3c64-4697-8270-e0207940d295.mp4

xezon:
I tested Humvee and Ambulance hulks and put breakpoint in `PhysicsBehavior::On_Collide` function. Behaviour is exact same, because `AIUpdateInterface* update` will be null, and `PhysicsBehaviour* physics` will be not null. In this scenario, IMMOBILE flag has no relevance in code path.

Therefore IMMOBILE flag is obsolete and can be removed.

debug

Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
```diff
@@ -19,7 +19,9 @@ Object DeadMicrowaveHulk
   ; *** AUDIO Parameters ***
   ; *** ENGINEERING Parameters ***
   RadarPriority = UNIT
-  KindOf = CAN_CAST_REFLECTIONS IMMOBILE NO_COLLIDE HULK UNATTACKABLE
+
+  ; Patch104p @bugfix xezon 28/01/2023 Remove immobile flag to allow bouncing.
```
Stubbjax: These comments might not be so accurate. All `IMMOBILE` seems to do is prevent move orders. The `AllowBouncing = Yes` in the `PhysicsBehavior` seems likely to be responsible for the bouncing.

xezon:
I need to merge this soon because I am already working on next fixes and piling changes on unit hulks will make testing difficult.

Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
xezon: Fixed the comments.

xezon:
I have now isolated the NO_COLLIDE flags fix for Militia Tank from this Pull, because that is actually a tangible bug that should have its own Pull.

xezon:
This change looks good to me. Helicopters still have IMMOBILE flag. Maybe can remove too in another change.

Stubbjax:
> I need to merge this soon because I am already working on next fixes and piling changes on unit hulks will make testing difficult.

This is a poor reason to bypass the review process.

> Fixed the comments.

This branch was not approved for merge. You gave me just six minutes to approve after addressing my feedback. You also rewrote git history which is extremely dangerous and terrible practice. How am I supposed to review the change in relation to the review?

PR #1618, opened 2023-02-04
Fix premature hulk spawn

xezon:
**Merge with Rebase.**

* Related to #1163

This change fixes death wreck and debris spawn before model disappears.

Affects

* USA Microwave
* USA Laser Tank
* USA Tomahawk
* USA Sentry Drone
* China Nuke Cannon
* Militia Tank

Crusader Tank wreck was already fixed with #1163.

## Original

https://user-images.githubusercontent.com/4720891/216763444-5ec018bc-6862-45dd-a9ea-93a9713ce2da.mp4

PR #1619, opened 2023-02-04
Fix: Remove duplicate radiation field on China Nuke Cannon death

xezon:
This change removes the duplicate `RadiationFieldSmall` before the Nuke Cannon explodes. This means the small radiation field no longer spawns before the Nuclear Explosion.

## Patched

https://user-images.githubusercontent.com/4720891/216764144-0d3fb69d-dd7e-4fbb-a1dd-cdf86e5c5bce.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
@@ -1832,12 +1832,12 @@ Object CINE_ChinaVehicleNukeLauncher ; Alias CINE_ChinaVehicleNukeCannon
     Mass = 50.0
   End
 
+  ; Patch104p @fix xezon 04/02/2023 Remove duplicate initial small radiation field.
   Behavior = SlowDeathBehavior ModuleTag_06
     DeathTypes = ALL -CRUSHED -SPLATTED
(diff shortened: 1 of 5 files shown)


Stubbjax:
Seems like these changes are already covered by #1618.

xezon:
> Seems like these changes are already covered by #1618.

This is not correct. #1618 does not remove duplicate field.

Stubbjax:
> > Seems like these changes are already covered by #1618.
> 
> This is not correct. #1618 does not remove duplicate field.

So #1618 removes the `FINAL` radiation field and this change removes the `INITIAL` radiation field? Does that not result in there being no radiation field after both are merged?

xezon:
The Radiation field is added into the `OCL_ChinaVehicleNukeCannonDie`.

Stubbjax:
> The Radiation field is added into the `OCL_ChinaVehicleNukeCannonDie`.

Alright; but I'm still confused as to why the `FINAL` radiation field was removed as part of #1618 without any mention in the description or comments instead of this change?

xezon:
Change #1618 is not responsible for removing duplicate field. It is only concerned about fixing wreck spawns. Whereas this change removes duplicate field, in isolation of #1618.

xezon:
Looks good.

PR #1624, opened 2023-02-04
Fix: Describe China Nationalism and Patriotism upgrade bonuses more accurately

### LINKED ISSUE #810: Nationalism and Patriotism Work Without Horde
Nationalism and Patriotism are a +25% RATE_OF_FIRE bonus for all units that can horde (Battle Master, Red Guard, Minigunner, Tank Hunter) regardless of whether they are currently horded or not. A single Tank Hunter will fire faster once Nationalism has been upgraded, no horde needed.

xezon:
* Fixes #810


commy2:
Iirc there is one for Boss General (only mentions Tank Hunter, because no Red Guard for Boss).

"Fanaticism" (called Patriotism in game) perhaps should only mention the Minigunner and Tank Hunter, because he doesn't have Battle Masters.

xezon:
Fixed Boss Nationalism.
Fixed Patriotism.

commy2 (approved):

PR #1626, opened 2023-02-04
Fix: Add proper text to Civilian Tractor

xezon:
This change adds proper text to Civilian Tractor. It is no longer called "Backhoe".


commy2 (approved):
Fix is correct. The actual digger is still called "Backhoe".

PR #1631, opened 2023-02-05
Fix: Civilian FarmerTruck now has correct model

xezon:
* Closes #1634

Civilian FarmerTruck now has correct model. CVFrmrtk.w3d & Co do not exist.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -10615,16 +10615,19 @@ Object FarmerTruck
   ButtonImage            = SUTerrorist
 
   Draw = W3DModelDraw ModuleTag_01
+    ; Patch104p @fix xezon 05/02/2023 Change Model from CVFRMRTK.
     ConditionState = NONE
-      Model = CVFrmrtk
+      Model = CVFRMRTRK
     End
 
+    ; Patch104p @fix xezon 05/02/2023 Change Model from CVFRMRTK_D.
     ConditionState = REALLYDAMAGED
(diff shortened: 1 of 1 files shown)


Stubbjax:
What is the implication of this change? Can we see a before and after? Is this not a change that would be seen in-game?

xezon:
Originally FarmerTruck is invisible. With this change it is visible.

invisible

Stubbjax:
So this change doesn't just affect the world builder? The truck will be seen in-game now?

xezon:
That depends whether or not it was placed in any official map. Since we have no automated way to check if it was referenced, we do not know.

Stubbjax:
I'm curious why `FarmerTruck` exists when `TruckFarmer` seems to be the same thing? They are otherwise identical aside from the display name and geometry radius. Should the former just be deleted?

xezon:
Deletion is also an option, but I did not want to go that far.

Stubbjax:
> Deletion is also an option, but I did not want to go that far.

How is it far? It's clearly a busted duplicate. It'd be cleaning up the data and a net positive.

xezon:
For example, if another Mod fixed the FarmerTruck like this change did, and then someone created a new Map for that Mod, then porting that Mod map to Patch104p will be more graceful, because FarmerTruck is present in both game builds.

Stubbjax:
> For example, if another Mod fixed the FarmerTruck like this change did, and then someone created a new Map for that Mod, then porting that Mod map to Patch104p will be more graceful, because FarmerTruck is present in both game builds.

This does not make a whole lot of sense. Dedicated mod maps are not expected to work with 1.04 because they often contain other things from the mod.

xezon:
Alternative change is #1634.

PR #1632, opened 2023-02-05
Fix: Civilian FarmerChickenTruck now has correct display name

xezon:
Civilian FarmerChickenTruck now has correct display name.


Stubbjax (approved):

xezon:
`FarmerTruck` object was removed previously.

PR #1634, opened 2023-02-05
Fix: Remove non-functional FarmerTruck

xezon:
* Closes #1631

This change removes non-functional FarmerTruck.


Stubbjax (approved):
image

This makes me happy.

PR #1638, opened 2023-02-05
Change: Civilian Humvee1 now uses the same transition damage effects as regular Humvees

xezon:
Humvee1 uses legacy transition damage effects. They are over the top. The arm pillar is not even shown in the video below, because pillar explosions are broken in Original. All Humvees now use the same transition damage effects.

## Original

https://user-images.githubusercontent.com/4720891/216846336-5976c6b8-9a7c-402c-a7a6-f8fc51cde6af.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -5294,10 +5294,10 @@ Object Humvee1
   ;  CreationList = OCL_EjectPilotOnGround
   ;End
 
+  ; Patch104p @tweak xezon 05/02/2023 Use same damage transition effects as regular America Humvee.
   Behavior = TransitionDamageFX ModuleTag_12
     ReallyDamagedParticleSystem1 = Bone:Smoke RandomBone:Yes PSys:SmokeSmallContinuous01
-    ReallyDamagedParticleSystem2 = Bone:Smoke RandomBone:Yes PSys:ArmExplosionSmall01
-    ReallyDamagedFXList1 = Loc: X:0 Y:0 Z:0 FXList:FX_HumveeDamageTransition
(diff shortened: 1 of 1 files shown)


MTKing4:
If it's CINE why does it matter? Maybe it was purposefully made that way for some cutscenes in the campaign, in that case we definitely shouldn't change that

The Other one is fine to change though

commy2:
Don't change visuals of CINE units (even if this one is only in CCG).

xezon:
What is the reasoning for not changing the death effect of CINE unit?

commy2:
They're used in cutscenes and are modified to work specifically for them.

xezon:
I understand. But this is not a functional change to CINE unit. Just a visual change. This means the CINE unit will behave the same as before, but would now show damage effects that are identical of those from regular Humvees. This adds consistency between cutscene and gameplay effects.

Other than that, the CINE damage effects have already changed with change #809. In Original game, `ArmExplosionSmall01` never showed. In Patch it shows. This makes the death effect of Humvee even more comical.

I don't feel strongly about this change for CINE unit. I can revert that one if so desired.

commy2:
> Just a visual change. 

CINE units are used in cutscenes. The visuals ARE their function.

Battlefieldkille:
Doesnt Radar Van also have similar explosion effect?

xezon:
It does. GLA Radar Van is the only unit using Humvee transition damage effect.

xezon:
> CINE units are used in cutscenes. The visuals ARE their function.

I disagree with this reasoning. However, it reads you are firm on this opinion, so I have reverted the fix of CINE_Humvee.

ExiLe-ZH:
How can we even judge without having watched the actual CINE?

In general I would agree with commie though, consistency is of no importantance, it's about what looks best in what setting and it's quite usual that cinematics use bigger and more spectular explosions, while for normal gameplay that would be too obstructive/distractive.

PR #1639, opened 2023-02-05
Fix boat damage models and effects

xezon:
**Merge with Rebase**

This change fixes boat damage models and effects.

Adds models for damaged states of
* BoatFishingTrowler
* BoatTugboat

Adds transition damage effects for
* BoatFishingTrowler
* BoatTugboat
* BoatAsianFishing

The other boats do not have models for damaged states.

## Patched

https://user-images.githubusercontent.com/4720891/216848957-8722986e-5f88-43a8-9c83-6b015a4ab274.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -12882,6 +12882,14 @@ Object BoatFishingTrowler
     ConditionState = NONE
       Model = CVTROWLER
     End
+
+    ; Patch104p @fix xezon 05/02/2023 Add models for damaged states.
+    ConditionState = REALLYDAMAGED
+      Model = CVTROWLER_D
+    End
+    ConditionState = RUBBLE
+      Model = CVTROWLER_D
+    End
   End
 
   ; ***DESIGN parameters ***
@@ -13230,6 +13238,14 @@ Object BoatTugboat
   Draw = W3DModelDraw ModuleTag_01
     ConditionState = NONE
       Model = CVTUGBOAT
+    End
+
+    ; Patch104p @fix xezon 05/02/2023 Add models for damaged states.
+    ConditionState = REALLYDAMAGED
+      Model = CVTUGBOAT_D
+    End
+    ConditionState = RUBBLE
+      Model = CVTUGBOAT_D
     End
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -12924,7 +12924,11 @@ Object BoatFishingTrowler
     InitialHealth   = 100.0
   End
 
-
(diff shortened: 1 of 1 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
```diff
@@ -12992,6 +12996,14 @@ Object BoatFishingTrowler
     ConditionState = NONE
       Model = CVTROWLER
     End
+
+    ; Patch104p @fix xezon 05/02/2023 Add models for damaged states.
+    ConditionState = REALLYDAMAGED
+      Model = CVTROWLER_d
+    End
+    ConditionState = RUBBLE
+      Model = CVTROWLER_d
```
Stubbjax:
```suggestion
      Model = CVTROWLER_d1
```

Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
```diff
@@ -13341,6 +13357,14 @@ Object BoatTugboat
     ConditionState = NONE
       Model = CVTUGBOAT
     End
+
+    ; Patch104p @fix xezon 05/02/2023 Add models for damaged states.
+    ConditionState = REALLYDAMAGED
+      Model = CVTUGBOAT_D
+    End
+    ConditionState = RUBBLE
+      Model = CVTUGBOAT_D
```
Stubbjax, line 13256:
```suggestion
      Model = CVTUGBOAT_D1
```

Stubbjax (commented):
Might be good to use the d1 models for rubble state, even if just for a frame.

Why lowercase `_d` for the fishing trowler and uppercase `_D` for the tug boat?

Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
xezon: _D1 for Rubble needs to be separate investigation. Imo the default RUBBLE should always be _D, because proper vehicles spawn their _D1 as separate Hulk and the transition phase to Hulk should remain in _D model. _D1 as RUBBLE should only be used if it serves a clear purpose.

xezon: As for lowercase _d I need to recheck. Likely because the actual W3D model on Disk is written like this.

xezon: Models are called CVtrowler, CVtrowler_d.

xezon: Fixed to upper case _D suffix.

xezon:
Looks good to me. `BoatAsianFishing` explosions are a bit odd because the boat looks all ancient with wood and shit, but perhaps Asian fishermen always have some high explosive liquor on board.

Stubbjax:
This should not have been merged because it had not been approved.

There is still an issue with the destruction transitions where the boats become cars.

image

xezon:
Yes that is the debris of most vehicles. I do not think there are much debris kinds to choose from.

PR #1643, opened 2023-02-06
Revert: Replace Microwave Haze effect from Microwave and GPS Scrambler to avoid rendering issues on some hardware configurations 2 (#1612)

xezon:
Fix #1612 had mistake.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -395,7 +395,7 @@ Object GPSScrambler_InvisibleMarker
 
     DefaultConditionState
       Model = None
-      ParticleSysBone = None GPSMicrowaveScrambler
+      ParticleSysBone = None GPSMicrowaveScambler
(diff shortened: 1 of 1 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
```diff
@@ -395,7 +395,7 @@ Object GPSScrambler_InvisibleMarker
 
     DefaultConditionState
       Model = None
-      ParticleSysBone = None GPSMicrowaveScrambler
+      ParticleSysBone = None GPSMicrowaveScambler
```
Stubbjax, line 398: Maybe could do with a comment to avoid future confusion?
```suggestion
      ParticleSysBone = None GPSMicrowaveScambler ; bone name is misspelled
```

xezon: It is an effect name. I think comment is not necessary for now.

Stubbjax (approved):

PR #1645, opened 2023-02-08
Fix: Set proper animation modes and flags for China Nuke Cannon

xezon:
This change sets proper animation modes and flags for China Nuke Cannon.

The original setup does work, but logically makes no sense. The original setups of Nuke Cannon differ between Regular China and Nuke General.

With this change the model condition states get proper setup:
```
MOVING:    AnimationMode = MANUAL
           Flags         = START_FRAME_FIRST
UNPACKING: AnimationMode = ONCE
           Flags         = START_FRAME_FIRST
PACKING:   AnimationMode = ONCE_BACKWARDS
           Flags         = START_FRAME_LAST
DEPLOYED:  AnimationMode = MANUAL
           Flags         = START_FRAME_LAST
```

I looked up animation code in Thyme, and MANUAL flag is synonymous for paused state. in MOVING and DEPLOYED states it now pauses animation at first and last frames respectively, and in the UNPACKING and PACKING states it will play animation ONCE from first frame or last frame respectively.

Thyme code does not yet have Deployment logic implement, but I suspect animations work with slightly inaccurate setups because it is setup in a way where it can borrow Animation mode and flags for UNPACKING and PACKING states from MOVING AND DEPLOYED states, if UNPACKING and PACKING are set to MANUAL, aka pause. So in a way code defends itself from configuration errors.

With new configuration, animation setup is explicit.

## Patched

https://user-images.githubusercontent.com/4720891/218091512-c93798fb-7cb8-42f8-bd7b-3f6acc7edf4d.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
@@ -1667,6 +1667,15 @@ Object CINE_ChinaVehicleNukeLauncher ; Alias CINE_ChinaVehicleNukeCannon
   ;UpgradeCameo5 = NONE
 
   ; Patch104p @bugfix xezon 08/02/2023 Shows NVNukeCn_D model on RUBBLE as well.
+  ; Patch104p @tweak xezon 10/02/2023 Sets proper animation modes and flags on condition states for robustness:
+  ;   MOVING:    AnimationMode = MANUAL
+  ;              Flags         = START_FRAME_FIRST
+  ;   UNPACKING: AnimationMode = ONCE
+  ;              Flags         = START_FRAME_FIRST
+  ;   PACKING:   AnimationMode = ONCE_BACKWARDS
+  ;              Flags         = START_FRAME_LAST
+  ;   DEPLOYED:  AnimationMode = MANUAL
+  ;              Flags         = START_FRAME_LAST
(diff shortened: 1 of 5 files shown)


xezon:
It is not clear to me what the discrepancy of Animation modes between `ChinaVehicleNukeLauncher`, `Infa_ChinaVehicleNukeLauncher` and `Nuke_ChinaVehicleNukeLauncher`, `Tank_ChinaVehicleNukeLauncher` means.

Are there animation issues? Needs testing.

commy2:
Why call it "fix" when you and I don't know what these changes do? Not totally against aligning these values, but if the change does nothing effectively, why bother with it? I think it increases the code difference between base and this project needlessly if it doesn't actually do anything.

xezon:
Needs testing. If this is no bug, then it can be closed. However, I do remember there is animation issue with Nuke Cannon where it creeps forward before shooting and the turret animation clips for a moment. Perhaps this is related to this. I do not have a video at hand.

Generally it is ok to open tentative pulls for suspicions of error. Better that way than just ignoring it.

xezon:
Here is the issue I was mentioning. Appears to happen with both China and Nuke General. So all Nuke Cannons are broken like this.

* #1651

Edit:

* Fixed with #1653

xezon:
I have now reviewed the codes and put proper change here.

xezon:
Looks good. Tested with Nuke Cannon from Regular China, Infa and Nuke.

PR #1646, opened 2023-02-08
Fix vehicle hulk flags of Civilian Militia Tank, CarLimo03, Humvee1

xezon:
**Merge with Rebase**

This change fixes vehicle hulk flags of Civilian Militia Tank, CarLimo03, Humvee1. Hulks can no longer be attacked and collided with.

Collision issue can not be observed in original Game, because the hulks are not spawned correctly. This will change with fixed setups.


Stubbjax:
Should `CarLimo03DeadHull` and `Humvee1DeadHull` be moved to `Hulk.ini`? Should `Humvee1DeadHull` have `Side = China`?

xezon:
Perhaps would be good to have all hulks in same file. The Combat Cycle hulk is also in another file. About Side, have not checked. Still fixing death effects at large.

xezon:
Anyway, that is no concern for this change.

Stubbjax (approved):

PR #1647, opened 2023-02-08
Change: Add suicide car bomb ability to Civilian AircraftCropDuster

xezon:
This change adds suicide car bomb ability to Civilian AircraftCropDuster, ~~Scooter~~.

Note that other carts and planes can also be used by bombers already.

https://user-images.githubusercontent.com/4720891/217649930-f6ae893e-02c0-46f2-8e46-c576927308da.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -2146,6 +2146,12 @@ Object ShiekLimo
     Weapon = PRIMARY None
   End
 
+  ; Note: This vehicle cannot be a suicide bomber because it has voices.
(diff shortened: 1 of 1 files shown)


commy2:
Alternative would be preventing Terrorists from entering these. Just like how it makes no sense for Hijackers to enter Combat Bikes, it makes no sense for Terrorists to hijack Scooters with the driver still around.

xezon:
Bombers can also enter the Ox thingy and the Horse Cart, which have no Human drivers at all. I don't know, I would make it interactive. That is what these Civilian Cars are for. For everything else there are props.

commy2:
They would be better props if Terrorists couldn't enter them.

xezon:
Up to the map designer. For fun maps these cars are ok.

PR #1649, opened 2023-02-08
Fix: ChinaInfantryOfficer poison death effects

xezon:
This change fixes ChinaInfantryOfficer poison death effects.


xezon:
Untested. Just found by coincidence when searching for OCL's.

Stubbjax (approved):

xezon:
This change should be tested once before merge.

xezon:
Poison death does not look proper. Looks broken on various misc infantry units.

xezon:
This change is superseeded by #2074

PR #1653, opened 2023-02-09
Fix: China Nuke Cannon animation jumps while it creeps forward after firing

### LINKED ISSUE #1651: China Nuke Cannon turret jumps animation when targeting
China Nuke Cannon turret jumps animation when targeting.

Tested with Nuke Cannon from Regular China and Nuke General.

https://user-images.githubusercontent.com/4720891/217897904-03ce7725-9078-411b-9987-b7574bddcb00.mp4

xezon:
* Fixes #1651

This change fixes China Nuke Cannon animation jump while vehicle creeps forward after firing.

https://user-images.githubusercontent.com/4720891/218319935-7506c461-6ad3-40b8-8535-562e1d449e33.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaCINEUnit.ini
@@ -1742,6 +1742,7 @@ Object CINE_ChinaVehicleNukeLauncher ; Alias CINE_ChinaVehicleNukeCannon
     AliasConditionState = RUBBLE PACKING BETWEEN_FIRING_SHOTS_A
 
     ;*** DEPLOYED STATE -- ready to fire ***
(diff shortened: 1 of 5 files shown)


commy2:
Does it work with Neutron Shells?

xezon:
Yes.

MTKing4:
Is it a confirmed fix with no side effects? If so it should be merged

Also maybe some details on how it got fixed would be nice

xezon:
You can clone the branch and check in game to see if it works properly. You can see how it was fixed in the Files changed diff.

xezon:
Works.

Stubbjax:
So all this does is make the bug look a bit nicer?

xezon:
Yes. As for fixing the creeping, I do not know yet how to solve it. Possibly code bug.

PR #1613, opened 2023-01-29
Fix faction vehicle/aircraft hulk physics and sink delays

xezon:
**Merge with Rebase.**

* Fixes #1183

~~This change fixes faction vehicle/aircraft hulk physics, sink delays, destruction delays.~~

* ~~ComancheRubbleHull can no longer bounce~~
* ~~Decreases the quick sink rate of USA Comanche hulk~~
* ~~Increases the slow sink rate of USA Chinook, China Helix hulks~~
* ~~Removes random sink delays from hulks of USA Ambulance, China Gattling, ECM~~
* ~~Enables kill on resting for hulks of USA Microwave, Humvee, Comache, China Battlemaster, Overlord, Dragon~~
* ~~Increases destruction delays for hulks of USA Dozer, Avenger, China Dozer, Supply Truck, Inferno Cannon, Overlord, Nuke Cannon, Helix, GLA Combat Bike, Scud Launcher to avoid deletion before object is sunk into terrain~~
* ~~Decrease destruction delays for hulks of USA Dozer, Humvee, Ambulance, Scout Drone, Battle Drone, China Gattling, ECM, GLA Technical(s), Quad Cannon, Scorpion, Toxin Truck, Radar Van, Rocket Buggy~~
* ~~Decrease sink delay for hulks of China Dragon Tank to start sink right after the fire particle vanished~~
* ~~Set sink delays for hulks of most vehicles and aircraft to at least 3000 milliseconds~~

## ~~Result~~

Dragon Tank sinks after 12000 milliseconds. It has a special death particle that lasts very long, so dead hull should stay with it.

Most small vehicles sink after 3000 ms. This way dead hulls can be seen properly after explosion particles for a bit of time, otherwise hulls can barely be seen on various vehicle kills. The new sink delay of 3000 ms is also consistent with all Infantry, which is 3000 ms as well.

Large vehicles sink after 6000 ms, which are Overlord, Nuke Cannon and Helix.

All faction vehicles and aircraft, except GLA Combat Bike, will sink with uniform speed.

No hulk disappears before it sunk into terrain.

Every hulk will disappear shortly after it sunk into terrain.

No hulk sinks with random intervals.

All changes have been tested on generously sloped terrain of around 40 degrees.

## Next Steps

In next pulls, Debris objects need fixing. They also disappear too quickly. On top of that, some vehicles could do with a small flame on the wreck. Some vehicles already have that. And there are bugs with some vehicles, such as no hulks for USA Repair Drone, Prisoner Transporter.

PR #1659, opened 2023-02-09
Fix: Decrease quick sink rate of USA Comanche hulk, increase slow sink rate of USA Chinook, China Helix hulks

xezon:
**Merge with Rebase**

This change decreases the quick sink rate of USA Comanche hulk, increases the slow sink rate of USA Chinook, China Helix hulks. All vehicles, helicopters, planes, except GLA Combat Bike, sink with same speed.

| Object               | Original Sink Rate | Patched Sink Rate |
|----------------------|--------------------|-------------------|
| USA Comanche Hulk    | 4                  | 2                 |
| USA Chinook Hulk     | 1                  | 2                 |
| China Helix Hulk     | 1                  | 2                 |
| Other vehicles hulks | 2                  | -                 |

## Rationale

Consistent sink speeds for most hulks. No unexplainable discrepancies. Helix hulk is big, and given its size it already takes longer to sink into terrain than all other vehicles. Half the sink speed made this even longer. Comanche no longer sinks 4 times slower than Chinook.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -468,10 +468,12 @@ Object ChinookRubbleHull
     MinLifetime    = 20000   ; min lifetime in msec
     MaxLifetime    = 20000   ; max lifetime in msec
   End
+
+  ; Patch104p @fix xezon 29/01/2023 Change Sink Rate from 1 to 2, Destruction Delay from 16000 to 8000, to be consistent with other hulks.
   Behavior = SlowDeathBehavior ModuleTag_05
     SinkDelay        = 1500
-    SinkRate            = 1     ; in Dist/Sec
-    DestructionDelay = 16000
+    SinkRate         = 2     ; in Dist/Sec
+    DestructionDelay = 8000
   End
 
 End
@@ -502,10 +504,12 @@ Object HelixRubbleHull
     MinLifetime    = 20000   ; min lifetime in msec
     MaxLifetime    = 20000   ; max lifetime in msec
   End
+
+  ; Patch104p @fix xezon 29/01/2023 Change Sink Rate from 1 to 2, Destruction Delay from 16000 to 8000, to be consistent with other hulks.
   Behavior = SlowDeathBehavior ModuleTag_05
     SinkDelay        = 1500
-    SinkRate            = 1     ; in Dist/Sec
-    DestructionDelay = 16000
+    SinkRate         = 2     ; in Dist/Sec
+    DestructionDelay = 8000
   End
 
 End
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -425,10 +425,11 @@ Object ComancheRubbleHull
     MaxLifetime    = 3000   ; max lifetime in msec
   End
 
+  ; Patch104p @fix xezon 29/01/2023 Change Sink Rate from 4 to 2, Destruction Delay from 10000 to 9500, to be consistent with other hulks.
   Behavior = SlowDeathBehavior ModuleTag_05
     SinkDelay        = 3000
-    SinkRate         = 4     ; in Dist/Sec
-    DestructionDelay = 10000
+    SinkRate         = 2     ; in Dist/Sec
+    DestructionDelay = 9500
   End
 
 


Stubbjax:
Your rationale explains the result of the change, but not why it was done. Why should all vehicles, helicopters and planes sink with the same speed? How is this beneficial? How is it detrimental?

Additionally, the Helix hulk is not given enough time to properly sink into the ground. It needs at least an extra three seconds.

xezon:
> Your rationale explains the result of the change, but not why it was done. Why should all vehicles, helicopters and planes sink with the same speed? How is this beneficial? How is it detrimental?

Changed.

> Additionally, the Helix hulk is not given enough time to properly sink into the ground. It needs at least an extra three seconds.

You asked me to split changes. This is changes split. Adjusting durations for terrain clipping issues in not part of this change, but part of a follow up change.

Original Helix disappears too early as well.

Stubbjax:
> Changed.

Your updated description still doesn't explain why. Why is "consistent sink speeds for most hulks." better than different sink speeds for most hulks? If you ask me, it makes sense for larger hulks to sink slower than smaller hulks.

xezon:
I disagree with your assessment. If Helix is meant to sink that slow, then there needs to be strong reason for that. Same for Comanche, why should it sink 4 times quicker than Chinook? Reverse the status quo, then reason about this change.

To illustrate: In follow up change, I would like to make Humvee sink twice as fast and Ambulance half as fast. Reason: EA Logic. Does that make sense to you?

Stubbjax:
> I disagree with your assessment.

What assessment are you disagreeing with? I suggested a base-level counter opinion. It seems intuitive that a larger, slower vehicle would take longer to sink than a lighter, faster one. The Titanic sinks slower and takes longer to corrode than a fishing boat.

> If Helix is meant to sink that slow, then there needs to be strong reason for that.

Should there not be an opposing strong reason to make it sink fast? I don't see any written here.

xezon:
I cannot explain this better. This setup makes most sense in the scope of all hulks. You disagree with it. It's fine. We do not have to agree on everything.

Stubbjax:
> I cannot explain this better. This setup makes most sense in the scope of all hulks. You disagree with it. It's fine. We do not have to agree on everything.

I have not disagreed with anything. You still haven't provided any rationale for the change. _"This setup makes most sense in the scope of all hulks."_ - why? If the change cannot hold up against even basic scrutiny, then maybe it is not worth changing.

MTKing4:
I agree with stubjax, this change makes no sense, and the counter argument makes perfect sense, bigger units should sink slower

Also it would pretty silly to streamline all sink delays to be the same, it will look robotic and reduce immersion

xezon:
This conversation just underlies why splitting hulk sink changes was poor idea. All this is addressed in follow up changes, but you expect it to be inside this change, which you have asked to be split away.

Makes no sense.

Stubbjax:
> This conversation just underlies why splitting hulk sink changes was poor idea. All this is addressed in follow up changes, but you expect it to be inside this change, which you have asked to be split away.
> 
> Makes no sense.

How? On the contrary - it shows that streamlining the sink delays might be a bad idea. I think a good takeaway here is that you should not introduce regressions that you intend to address through follow up changes. Nobody knows what changes may or may not exist in the future. A PR should be a complete feature / tweak / fix. It's hard to conceptualise changes if they're always small fractions of the end result.

xezon:
If you approach a change you can do it differently if you like. I like to approach change like this - in a structured way, aka setup deviations should be done last - if at all.

Stubbjax:
> If you approach a change you can do it differently if you like. I like to approach change like this - in a structured way

This comes across as anti-collaborative. Changes should be presented in the clearest way possible - the whole dish, not each individual ingredient, so to speak. Unorthodox methodologies are not conducive to an efficient team environment.

xezon:
No. I simply do not want to move to from tackling problems in unstructured ways just because this is how you would approach the problem while actually not working on this particular problem.

I do acknowledge that there is merit to make big vehicles sink later or slower. But this change is not responsible for it. As has been presented in the former Pull that this Pull was split off, the big vehicles will sink later than small vehicles.

Big vehicles are Overlord, Nuke Cannon and Helix because of their relative plus in size over other units.

The structured approach is to get Sink Speeds, Sink Delays and Sink Destructions in sync with each other and their respective model sizes. At the very end, one can design discrepancies such as longer delays, slower sink speeds and/or sink delay variances.

Stubbjax:
> No. I simply do not want to move to from tackling problems in unstructured ways just because this is how you would approach the problem while actually not working on this particular problem.

Are you unwilling to compromise or adapt even if it is for the greater good?

MTKing4:
So this was merged regardless of our feedback?

xezon:
I have adhered to the following feedback:

* Split this change off of #1613
* Add rationale text
* Explain changes

Changes have been tested and it looks good. If you want anything changed you can open new Pull, however, I am not yet done with fixing vehicle death issues, so I recommend to postpone it until I am done.

commy2:
This change should be reverted.

MTKing4:
yet it's still here

PR #1660, opened 2023-02-09
Change: Remove random sink delays from hulks of USA Ambulance, China Gattling, ECM

xezon:
This change removes random sink delays from hulks of USA Ambulance, China Gattling, ECM.

## About Sink Delays

`SinkDelayVariance` and `DestructionDelayVariance` are designed a bit unfortunate.

What it does:
```
RealSinkDelay = SinkDelay + SinkDelayVariance
RealDestructionDelay = DestructionDelay + DestructionDelayVariance
```

What it should do:
```
RealSinkDelay = SinkDelay + SinkDelayVariance (*)
RealDestructionDelay = DestructionDelay + SinkDelayVariance (*)
```

(*) Apply same random generated value of variance.

## Rationale

Removal of variance makes setup much easier to maintain. If we want sink variances on hulks, it should be added as last step. Only 3 original hulks have sink delay variance. Removal makes no tangible difference to visuals.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/Hulk.ini
@@ -976,12 +976,11 @@ Object DeadChinaGattlingTankHulk
     MaxLifetime = 1600   ; max lifetime in msec
   End
 
+  ; Patch104p @tweak xezon 29/01/2023 Remove Sink Delay Variance of 1500, Destruction Delay Variance of 4000.
   Behavior = SlowDeathBehavior ModuleTag_05
     SinkDelay                 = 1500
-    SinkDelayVariance         = 1500
     SinkRate                  = 2     ; in Dist/Sec
     DestructionDelay          = 12000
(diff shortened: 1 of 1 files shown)


Stubbjax (approved):

commy2:
lol

commy2 (approved):

MTKing4:
So sink delay variance was never functional? 
What are the 3 original hulks with sink variance?

commy2:
>  What are the 3 original hulks with sink variance?

First sentence. Ambulance, Gattling and ECM tank

MTKing4:
Oh, right xD
But was the variation functional?

commy2:
Apparently no. Instead it is always added to the base value.

xezon:
It does work, but it cannot be configured to work well with correctly timed DestructionDelay. Adding a small variance of like 1000 ms to all hulks would be ok imo. But first step needs them all removed for proper baseline.

MTKing4:
Yes a little bit of variance will look good imo

PR #1657, opened 2023-02-09
Fix China Battlemaster and Overlord death effects

xezon:
* Fixes #1605

This change fixes China Battlemaster and Overlord death effects. Muzzle flashes no longer show on death. And turret barrel no longer recoils on death.

## Patched

https://user-images.githubusercontent.com/4720891/217946777-b4551ae6-d576-4fe5-9711-2423cd504961.mp4

Issue #1605, opened 2023-01-26
Some tanks show muzzle flash upon death


xezon:
Some tanks show muzzle flash upon death. Reproducible with Nuke Battlemaster, Nuke Overlord. Not reproducible with Regular Battlemaster, Regular Overlord.

Potentially related to `FireWeaponWhenDeadBehavior` module. When Isotope Stability upgrade is acquired, then issue is no longer reproducible with Nuke Battlemaster, Nuke Overlord.

https://user-images.githubusercontent.com/4720891/214953508-94590e2f-6355-4613-9a8a-cef4571731cf.mp4

xezon:
Possible cause for this is RUBBLE ConditionState missing a HideSubObject for a specific sub state.

commy2:
I don't think the RUBBLE state is needed. You can just hide it on the DEFAULT state. The animation overwrites the explicit hide.

xezon:
I have fix for it.

xezon:
Affects all Battlemasters, Overlords, Demo Scorpion, Demo Marauder. Anything else?

xezon:
I have troubles fixing this on Demo Scorpion. For some reason keeps showing Muzzle flash on death when doing same fix as for Overlords. There are multiple Condition States.

commy2:
HideSubObject of the muzzle flash section.

xezon:
I tried that with Battlemasters first, but it does not work.

xezon:
Ok I think I got it.

PR #1662, opened 2023-02-10
Fix: (Demo) GLA Marauder death explosion no longer triggers effects on turret barrels

xezon:
* Related to #1605 
* Related to #1657
* Related to #1663

With this change the (Demo) GLA Marauder death explosion no longer triggers effects on turret barrels.

Tested with Demo GLA Marauder, with and without Demo Upgrade, with and without Scraps.

Setup is replicated on all faction Marauders, for the case that Mod Maps & Co want to add death weapons to any Marauder.

https://user-images.githubusercontent.com/4720891/218164722-a25f5df9-8192-4e54-ad23-9358b6c4ece1.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -20056,15 +20056,16 @@ Object Chem_GLATankMarauder
       WeaponLaunchBone = PRIMARY BarrelMS
     End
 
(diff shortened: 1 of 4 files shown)


commy2:
How does it look while it dies while firing? You're overwriting recoil bone, so I assume it would snap forwards immediately instead of being correctly animated.

xezon:
> How does it look while it dies while firing? You're overwriting recoil bone, so I assume it would snap forwards immediately instead of being correctly animated.

This question was addressed in #1657

xezon:
Looks good.

PR #1665, opened 2023-02-10
Change: Eject Pilot on death of USA Comanche immediately without delay of 1500 ms

### LINKED ISSUE #175: USA Comanche pilot does not spawn if helicopter crashes in under 1.5 seconds
Comache's pilot eject codes are in the helicopterslowdeath behavior unlike any other the USA vehicles. If comanche crash before 1.5 seconds pilot can't spawn. 

   " MinBladeFlyOffDelay             = 1500            ; in milliseconds"
    "MaxBladeFlyOffDelay             = 1500            ; in milliseconds"
    
Because of these codes pilot can't spawn. He can spawn  between 35 60 milliseconds but he will spawn with parachute albeit on the ground.

But these codes not only for the pilot but also for the effects in the dying animation.

So my suggestion is that, we should delete "OCLEjectPilot  = OCL_EjectPilotViaParachute" from helicopterslowdeath and add 
"EjectPilotDie" behavior for comanche. What we will lose, pilot will spawn right after heli shot down but not a major loss. What devs think about was the spawn of the pilot right after the propellers broke it seems, we don't have that anyway.
    
https://user-images.githubusercontent.com/103841664/217898441-2615e791-1157-47a8-80d6-6b753dffeb90.mp4

### LINKED ISSUE #1652: Comache pilot can't spawn if helicopter blow up in 1.49 seconds
Comache's pilot eject codes are in the helicopterslowdeath behavior unlike any other the USA vehicles. If comanche crash before 1.5 seconds pilot can't spawn. 

https://user-images.githubusercontent.com/103841664/217898441-2615e791-1157-47a8-80d6-6b753dffeb90.mp4

xezon:
* Fixes TheSuperHackers/GeneralsGameCode#175

This change ejects Pilot on death of USA Comanche immediately without delay of 1500 ms.

## Patched

First 3 pilots eject without delay. Last pilot ejects with 1500 ms delay.

https://user-images.githubusercontent.com/4720891/218223494-619e5c15-abe6-482d-b99b-81db97127df0.mp4

## Patched Advantage

Pilot can spawn when Comanche is killed on repair pad. It spawns like all other USA vehicle pilots without delay.

https://user-images.githubusercontent.com/4720891/218223460-f567e039-3c61-448d-91a5-7c9d6957a3d4.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -2302,7 +2302,7 @@ Object AirF_AmericaVehicleComanche
     ; Most things that eject pilots do so immediately upon death,
     ; via use of EjectPilotDie, but Helicopters are a special case...
     ; they need to do so after their blades are ejected.
-    OCLEjectPilot                   = OCL_EjectPilotViaParachute
+    ;OCLEjectPilot                   = OCL_EjectPilotViaParachute
     FXBlade                         = FX_HelicopterBladeExplosion
     OCLBlade                        = OCL_HelicopterBladeExplosion
     FXHitGround                     = FX_HelicopterHitGround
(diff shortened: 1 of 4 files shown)


commy2:
I wonder if ExemptStatus could be used in such a way to preserve the 1500 delay when flying, but spawning the Pilot immediately when grounded instead of not spawning.

xezon:
> I wonder if ExemptStatus could be used in such a way to preserve the 1500 delay when flying, but spawning the Pilot immediately when grounded instead of not spawning.

Such approach would mitigate it a bit, but not guarantee Pilot spawn 100%, because a Helicopter in air could also crash before 1500 ms.

xezon:
Do we want this change or not? I am neutral on it. It would be nice to have consistent Pilot eject timings, but it is also ok to leave as it is originally.

MTKing4:
The original looked nicer, but this change is a gameplay improvement no doubt, so i have to go with that

commy2:
I would favour an engine change where the Pilot is spawned at 1500 msec OR when the helicopter crashes before that.

xezon:
I agree with that.

MTKing4:
That would be the best solution yes

PR #1670, opened 2023-02-11
Change: China Overlord will always reload when idle

### LINKED ISSUE #1669: China Overlord does not reload after single shot
China Overlord does not reload after single shot.

https://user-images.githubusercontent.com/4720891/218266968-5133bf2e-8d37-4754-868b-a420952ac770.mp4

xezon:
* Resolves #1669

With this change the China Overlord will always reload when idle.

This makes the unit a bit better in scenarios where it managed to only fire with one of its barrel on its target and some time passes until the next target is attacked.

This situation is perhaps not as rare as one may assume. There is a `ClipReloadTime` of 2000 ms vs a `DelayBetweenShots` of 333 (300) ms. This means there is a 14% chance that Overlord gets into state with one clip in barrels when attacking and moving Overlord around.

# Rationale

Eliminates situations where Overlord would only fire once on a fresh engagement. This likely is what player would expect always.

## Patched

First attack fires 2 clips normally, second attack fires only 1 clip because of interruption (Stop), and third attack fires 2 clips again because of new auto reload when idle.

https://user-images.githubusercontent.com/4720891/218267994-e0c773cf-7fef-4c08-a572-05c6ddf1ed3c.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -236,6 +236,7 @@ Weapon ScorpionTankGunFXWeapon
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @tweak xezon 11/02/2023 Enables automatic reload when idle.
 Weapon OverlordTankGun
   PrimaryDamage = 80 ;100.0
   PrimaryDamageRadius = 5.0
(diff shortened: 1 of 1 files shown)


commy2 (approved):

MTKing4:
Good Change

Stubbjax (approved):

PR #1672, opened 2023-02-11
Fix: Set correct firing sequence on GLA Marauder with double barrel

### LINKED ISSUE #1671: GLA Marauder with 2 barrels always shoots just once before it can shoot twice
GLA Marauder with 2 barrels always shoots just once before it can shoot twice.

https://user-images.githubusercontent.com/4720891/218270787-8369392f-4b57-4bcf-b55b-af7a4ecf3679.mp4

xezon:
* Fixes #1671

This change sets correct firing sequence on GLA Marauder with double barrel.

It will now always start by firing 2 clips instead of just 1 clip.

This gives the unit an inconsequential buff.


MTKing4:
I'm fine with this change

Stubbjax (approved):

PR #1674, opened 2023-02-11
Change: GLA Marauder with double barrel will always reload when idle

### LINKED ISSUE #1673: GLA Marauder with double barrel does not reload after single shot
GLA Marauder with double barrel does not reload after single shot.

https://user-images.githubusercontent.com/4720891/218271511-2c2762aa-80d9-4efb-a060-95b8fb2fd743.mp4

xezon:
* Merge after #1672
* Resolves #1673

With this change the GLA Marauder will always reload when idle.

This makes the unit a bit better in scenarios where it managed to only fire with one of its barrel on its target and some time passes until the next target is attacked.

This situation is perhaps not as rare as one may assume. There is a `ClipReloadTime` of 800 (750) ms vs a `DelayBetweenShots` of 133 (100) ms. This means there is a 14% chance that Marauder gets into state with one clip in barrels when attacking and moving Marauder around.

# Rationale

Eliminates situations where Marauder would only fire once on a fresh engagement. This likely is what player would expect always.

## Patched

First attack fires only 1 clip because of interruption (Stop), and second attack fires 2 clips again because of new auto reload when idle.

https://user-images.githubusercontent.com/4720891/218272291-7894e528-f8a1-4d54-92b7-104b094231cc.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -73,6 +73,7 @@ End
 ;------------------------------------------------------------------------------
 ; Patch104p @bugfix xezon 11/02/2023 Set DelayBetweenShots from 750 to 100
 ;   and ClipReloadTime from 100 to 750 to correct the firing sequence.
+; Patch104p @tweak xezon 11/02/2023 Enables automatic reload when idle.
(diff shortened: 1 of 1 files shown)


MTKing4:
Good Change

Stubbjax (approved):

PR #1675, opened 2023-02-11
Fix: Add CRATE flag to SalvageCrate

xezon:
This change adds the CRATE flag to SalvageCrate object. All other Crates have this flag too. Appears to make no difference in player facing behavior.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1675_salvage_crate_crate_flag.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1675_salvage_crate_crate_flag.yaml
@@ -1,0 +1,18 @@
+---
+date: 2023-02-11
+
+title: Adds missing CRATE flag to SalvageCrate
+
+changes:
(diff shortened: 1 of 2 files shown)


Stubbjax:
It would be good to know what the `CRATE` flag does so we don't end up with unforeseen consequences.

xezon:
Thyme does not yet have any CRATE logic laid out, so we cannot see exactly what CRATE does, if anything. However, we know that the other Crate objects have this flag, and we know that SalvageCrate has KindOf setup issues, such as SELECTABLE flag, so to me it looks like this flag is just missing. When there is a CRATE KindOf, then a Crate Object more likely than not should have it. I was unable to observe difference in behaviour of SalvageCrate with and without this flag however.

xezon:
```
;gs I added parachutable as a catch all to prevent making new kindofs (like CRATE)
```

Looks like CRATE was originally created to make the object PARACHUTABLE.

xezon:

Issue #648, opened 2022-01-21
Scrap crates can be removed by placing scaffolds


ReLaX82:
Scrap crates can be removed by placing scaffolds on top of them. This works with all factions but is counter-intuitive, because USA and China factions cannot even see the scrap crates. GLA cannot effectively prevent the scaffold placement, but can try to kill the scaffold to at least incur a cost penalty for dropping the scaffold.

Noob and Semi players generally do not care about this exploit. Pro players use this exploit and think it helps balance China in 1.04, because China struggles versus scrapped GLA units.

ReLaX82:
> I know its a controversial and highly critical point, but this one needs to be decided asap before anything else in this regard (with direct/indirect interferences) will be done. It needs to be disussed also with GR stuff as it should then be a global rule in the fairplay rules of GR. Please discuss. My approach is to take the most accepted Pros and form an answer out of them so we have that fixed.

Related reports:
* #592 
* #56 
* #647 
* #641
* #1827

### Proposal 1 APPLIED
Leave scrap crates intact when a scaffold is put over them. Same as we did for money crates.
* #1676

### Proposal 2 REJECTED
~~Allow all factions to see scrap crates. This makes it more intuitive to interact with them.~~

### Proposal 3 REJECTED
~~Allow all factions to collect scrap crates. This way GLA can be prevented from picking them up naturally. When USA and China picks up the scrap with a unit, then it just disappears.~~
* #641
* #1677

### Proposal 4 REJECTED
~~Do not drop scrap crates from China Dozer, USA Dozer and China Truck.~~
* #592

### Proposal 5
Drop different kind of scraps from different units. USA Dozer, China Dozer and China Truck drop money scraps only. Battle units drop regular scraps. Technically possible.

### Proposal 6 REJECTED (by 2)
~~Allow scraps to be destroyed by weaponry. Could incur problems in regular battles with splash damage causing too many scraps to disappear.~~

### Proposal 7 REJECTED (by 2)
~~Add "Clear Scrap" ability to Dozers.~~

### Proposal 8 REJECTED (by 2)
~~Add "Clear Scrap" ability to China Truck.~~

### Proposal 9
Reduce Scrap timeout.

### Proposal 10
Scrap crates can be seen by all players, can be selected by all players, and can be detonated by all players with a click of a button in the command set.

### Proposal 11
Lets scaffold Remove Crates, but scaffolds will not spawn, nor deform the terrain until the Builder is very close to the build site.

[...]

PR #1676, opened 2023-02-11
Fix: GLA Salvage Crate can no longer be removed by scaffolds

xezon:
* Merge after #1675
* Related to #648

With this change the GLA Salvage Crate can no longer be removed by scaffolds. Crates simply stay intact under the scaffold until they are picked up or time out.

## Patched

shot_20230212_000305_1

# Rationale

It took until around year 2017 for top players to figure out the obscure exploit of removing visible and invisible Salvage Crate objects by placing Scaffolds over them. In top level matches this is now exploited accordingly. The reason this is accepted by top players is the weakness of China against (scrapped) GLA. A fully scrapped GLA Technical for example can present a major problem in the early stages of a China build.

Since this Project aims to strengthen China and make all faction match-ups more playable, this exploit ideally will no longer be required to achieve fair chances for China vs GLA in the future.

The big advantage is clean game mechanics, and no exploits sold as features to achieve some sort of balancing. The big disadvantage is that this fundamentally changes how top level GLA matches are played, which almost always revolve around Scrap scaffold exploits.

For many Noob, Semi and Pro players this change is inconsequential, because they do not use this obscure exploit to gain the advantage.

This change also makes the mechanics consistent with the inability to delete money crates and any other pickable crate, which was implemented with #443.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1676_salvage_crate_exploit.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1676_salvage_crate_exploit.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-02-11
+
+title: Fixes GLA Salvage Crate being removable by scaffolds
+
+changes:
+  - fix: The GLA Salvage Crate is no longer deleted when a scaffold is placed on top of it. It remains intact under the scaffold until it is picked up or times out.
+
+labels:
+  - bug
+  - controversial
+  - gla
+  - major
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1676
+
+authors:
+  - xezon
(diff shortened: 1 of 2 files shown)


ExiLe-ZH:
Rage invented it quite a few years earlier, closer to 2012'ish I think. Anyways, for a long time it wasn't a thing at all.

MTKing4:
Hallelujah

Stubbjax (approved):
What does the `CRATE` flag do here?

xezon:

xezon:
iDEarikas
> The thing that hurt me the most was the scrap deny removal. Unbelievably awful. Some so called 'bugs' like scrap deny have become a part of the game well enough that removing them is absurd and provocative. 20 years of excessive uses gives them the right to not be called bugs anymore and thus they should stay. Any so called 'bug' that will change the gameplay then better stay. Because it's not a bug anymore if it's the mainstream thing. It's just there, you got to deal with it.

PR #1677, opened 2023-02-11
Change: USA and China can now steal salvage crate from GLA

### LINKED ISSUE #641: Shall USA and China be able to collect scrap?
Looks wired in the first place, but it could change a bit the scrap up effectiveness of GLA.
USA/China could collect without having a direct benefit (not scrapping up) from it. May puts more skill into the game.
Your comment are welcome.

xezon:
* Merge after #1676
* Closes #641
* Related to #648

With this change USA and China can now steal salvage crates from GLA simply by moving a unit over it. The crate will disappear with a sound. No rewards are given other than removing the crate from the world.

## Rationale

This change mitigates the impact of the removal of the Scrap scaffold exploit with #1676. It replaces the original exploit with a graceful and readable game mechanic that is transparent to all players.

This is a major and visible change, but it may add additional depth to the game, as all players are now encouraged to race to Scrap Crates. Even though USA and China receive no direct rewards from it, preventing GLA from scrapping up to powerful units is a reward on its own.

The usefulness of this reward is underlined by the popularity of the original Scrap scaffold abuse among competitive top level players that this change aims to replace.

## Salvage Crate pickup by GLA

Is as per Original.

https://user-images.githubusercontent.com/4720891/218285658-15ce60d6-02bb-4dcb-a3ec-0a67e3291bc1.mp4

## Salvage Crate pickup by China

https://user-images.githubusercontent.com/4720891/218285670-ba8b9803-20cf-45be-b17a-fb451bfaa626.mp4

## Salvage Crate pickup by USA

https://user-images.githubusercontent.com/4720891/218285679-28224683-e93a-4e9d-b308-faa2e4620360.mp4


Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
```diff
@@ -192,7 +193,10 @@ Object SalvageCrate
   EditorSorting   = MISC_MAN_MADE
 
   ; *** ENGINEERING Parameters ***
-  KindOf = PARACHUTABLE SELECTABLE ;Only Salvage crates are marked Selectable so the Mouse can interact with them and give Salvage voices.  Being dead makes them not actually selectable
+  ; Patch104p @fix xezon 11/02/2023 Adds missing crate flag.
+  ; Patch104p @bugfix xezon 11/02/2023 Adds inert flag to avoid removal by scaffolds.
+  ; Patch104p @bugfix xezon 11/02/2023 Removes selectable flag to avoid crate selecting over terrain.
```
xezon, line 198: SELECTABLE flag somehow caused issues for China and USA. With this flag it was not possible to click on the crate to order the unit to move to it, because the mouse click would select the Crate instead. Why this never was a problem with GLA is unclear.

Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
```diff
@@ -206,6 +210,22 @@ Object SalvageCrate
     MaxMoney = 75
   End
 
+  ; Patch104p @feature xezon 11/02/2023 China can pickup this crate with no reward. Crate just vanishes with a sound and GLA cannot collect reward.
+  Behavior = UnitCrateCollide ModuleTag_UnitCrateCollide_1
+    ForbiddenKindOf = PROJECTILE
+    PickupScience = SCIENCE_CHINA
+    UnitCount = 0
+    UnitName = SalvageUnitDummy
```
xezon, line 218: A unit has to be referenced here, otherwise this module does not work. So I created the SalvageUnitDummy object and set count to 0. So this spawns nothing but works to trigger the pickup. Perhaps we can also just use another existing object for it.

Patch104pZH/GameFilesEdited/Data/INI/SoundEffects.ini
```diff
@@ -4414,11 +4414,11 @@ AudioEvent CrateMoney
 End
 
 AudioEvent CrateFreeUnit
-  Sounds = gcramona
+  Sounds = gsalvage2 ; Patch104p @tweak xezon 11/02/2023 From gcramona.
```
xezon, line 4417: This originally played the money pickup sound for `UnitCrateCollide` event. I changed this to be a shortened version of the original Salvage sound, which fits better I think.

Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
```diff
@@ -206,6 +210,22 @@ Object SalvageCrate
     MaxMoney = 75
   End
 
+  ; Patch104p @feature xezon 11/02/2023 China can pickup this crate with no reward. Crate just vanishes with a sound and GLA cannot collect reward.
```
xezon, line 213: I also tested the following, which does NOT work for some reason:
```
  Behavior = SalvageCrateCollide ModuleTag_SalvageCrateCollide_2
    ForbiddenKindOf = PROJECTILE
    PickupScience = SCIENCE_CHINA
    WeaponChance = 0%
    LevelChance = 0%
    MoneyChance = 100%
    MinMoney = 25
    MaxMoney = 75
  End
  Behavior = SalvageCrateCollide ModuleTag_SalvageCrateCollide_3
    ForbiddenKindOf = PROJECTILE
    PickupScience = SCIENCE_AMERICA
    WeaponChance = 0%
    LevelChance = 0%
    MoneyChance = 100%
    MinMoney = 25
    MaxMoney = 75
  End
```

And I tested the following which DOES work:
```
  Behavior = MoneyCrateCollide ModuleTag_MoneyCrateCollide_1
    ForbiddenKindOf = PROJECTILE
    PickupScience = SCIENCE_CHINA
    MoneyProvided = 100
    ExecuteAnimation = MoneyPickUp
    ExecuteAnimationTime = 4.0
    ExecuteAnimationZRise = 15.0
    ExecuteAnimationFades = Yes
  End
  Behavior = MoneyCrateCollide ModuleTag_MoneyCrateCollide_2
    ForbiddenKindOf = PROJECTILE
    PickupScience = SCIENCE_AMERICA
    MoneyProvided = 100
    ExecuteAnimation = MoneyPickUp
    ExecuteAnimationTime = 4.0
    ExecuteAnimationZRise = 15.0
    ExecuteAnimationFades = Yes
  End
```

So in theory we could also give America and China rewards on crate pickup, but they will not show money text on pickup. It will just play money sound and the dollar sign as with any other Money Crate.

xezon:
Needs to be tested in Multiplayer.

Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
xezon: Actually, `SalvageCrateCollide` does work for USA and China. It does however require `SALVAGER` KindOf on all units.

And that in turn would enable GLA player with China and USA units to create Salvage Crates on kills via

```
CrateData SalvageCrateData
  CreationChance = 1.0
  KilledByType = SALVAGER
  KillerScience = SCIENCE_GLA ; Killed by a salvager unit owned by a GLA player
  CrateObject = SalvageCrate 1.0
End
```

But that side effect could be mitigated, by assigning `ARMOR_SALVAGER` to all GLA Salvager units and make that the requirement for the above `SalvageCrateData`.

So proper money-only Salvage Crates would be possible for China and USA, but would require bigger INI setup.

ExiLe-ZH:
I don't like this option/implementation, it doesn't make much sense how it's visible for the other factions, but just disappears upon collision. I don't like money pickup either, even if it's a small amount.

Target-able/destroy-able/crush-able (with wreck/audio feedback like crushing cars) would be better, but still not sure if ideal.

xezon:
The problem with removing Crates by damage is that Crates would eventually disappear under random hits at where the crate is located in regular battle scenarios, for example by SCUD Launcher missile hit.

I played around with this change a bit and it is very satisfying to drive over the crates with Humvees :-)

ExiLe-ZH:
I think there are a few better options to migitate around the imbalances of (non-deniable) scraps.
1. Trucks/dozers spawn money crates only.
2. Tweak experience levels for Techs, easier scraps but vetting slower.
3. Tweaking experience levels and DPS models of Quads. https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1054

MTKing4:
Few notes here
- with this change i predict a new meta will pop up where they would put at least 1 infantry in all transport units to auto deny scrap, units like vees, techs, chinas might mix more infantry with armies for quick deny, which i think is good for diversity
- In every battle vs A GLA, it's always guaranteed to get at least 1 Scrap since you won't have anymore units around to deny the scrap, so it's not nearly as effective as scrap deny
- this should go hand in hand with removing scrap rewards from dozers and trucks otherwise super techs mauraders will still own
- I don't recommend messing with Units' damage, hp or xp requirements, the first two will change the game much more than this change does, and the latter will have no meaningful impact on the balance, super techs are still very strong even without the vet

xezon:
Too risky.

Issue #1145, opened 2022-09-06
Truck gives off quite some XP


ExiLe-ZH:
```
Object ChinaVehicleSupplyTruck
  ExperienceValue = 50 50 50 50

Object AmericaVehicleChinook
  ExperienceValue = 50 50 50 50

Vees/Quads/Gats:
  ExperienceValue = 50 50 100 150

Object GLAVehicleTechnicalChassisOne
  ExperienceValue = 25 25 50 100
```

1. Truck kill grants 50xp, which is the same amount as many combat units like Vees, Quads, Gats etc. A Technical even gives off 25xp only. The obvious difference here is that it's defenseless and can't easily escape danger. It's vulnerable vs any type of unit.

3. Chinooks give off the same amount, but is obviously a higher tech/value unit. Chinook collection can be interrupted from greater distances, but it can also escape faster and has less enemies than Trucks.

4. Workers on the other hand don't grant any XP at all. They're more vulnerable than Trucks, can't crush infantry, but can escape easier into tunnels.

5. Another thing to keep in mind is that they leave Scraps behind aswell.

6. 25-40xp might be a more balanced number and would buff China a little bit.

ReLaX82:
It is a little dependent on how the scraps willbe handeld I think. If scrapping up keeps nearly the same then I would reduce the XP of trucks for sure.

Stubbjax:
Definitely agree. I used to think even 40 would be controversial but seeing the numbers laid out, 25 seems very reasonable. This is especially true considering the primary element of gameplay that is affected as a result - Technical harassment.

It might also be worth considering altering the (kind of strange) Technical experience requirements in tandem, e.g. [0 50 75 150] → [0 50 100 150], so that each rank requires two truck kills and the effect of Technical Training remains as a consistent 1/3 bonus.

xezon:
I am very skeptical of changing Technical 0 50 75 150 to 0 50 100 150, because this is very important unit in competitive matches and raising its veteran requirement will have significant impact. We would first need to determine if 2 Star Technical indeed should be more rare. I do not know how to do that.

Stubbjax:
[...]

Issue #592, opened 2021-10-22
Should trucks and dozers leave scraps?


ExiLe-ZH:
(Controversial) balance solution to make Techs a little less dangerous when chasing dozers or trucks. Imo it makes no sense either a Tech can upgrade it's gun from units that are not even armed.

This thing alone will probably improve the winrate of China quite a bit since one Super Tech can beat a China single-handedly.

EnigmaZift:
I agree, but a change like this is maybe too drastic?

xezon:
Such change would have direct effect on how GLA vs China is played early and I am not sure if that is for the better.

RisingZH:
They should definitely leave scraps, scraps are a super fun and interesting game mechanic.

The answer is just to reduce the damage of a double scrapped vetted tech. Single scrapped techs are fine and even double scrapped non vetted techs are fine.

I haven't been following the Discord chat really and only checking GitHub; is the idea still to only change unit costs and build times? I would say this is another example of where that won't work.

Madrageeeeee:
I think this has a major effect on gameplay and i wouldnt change it.

this will turn this patch into a similar game experience as 1.06. What happened there is that because very powerful units like vees were nerfed and supertechs were very hard to come by due to no scraps elft behind by trucks was that the gameplay became quite ‘sluggish’ and defense based.

the reason why zero hour can be so quick and turn easily is because of this units. If gla has no chance to do significant damage with supertechs / techs in general then you will get a very defensive game because gla units get shredded out in the open.

seeing the bufs that china has gotten i think this will make china too powerful against gla and it will affect gameplay negatively. 

I think we should at really prevent the game from becoming too static / defensive

Even reducing the damage output of supertechs isnt a good idea imo. With mines / cheaper units and better driving dozers it will already be difficult enough for gla’s especially on large open maps. Also with cheaper units for china the cc keep strat is already more viable

RisingZH:
Interesting post rage. 

I think part of this discussion needs to be should denying scraps with scaffolds still be possible. I'm pretty strongly against scrap deny being allowed even on 1.04 and I believe I'm correct in saying you Rage made it become a thing. 

The only real argument for scrap deny on 1.04 I can get behind is that super techs do just wreck china and they are pretty easy to come by without scrap deny so it does balance the matchup.

I would be interested to hear, do you think the game would still be better with super strong techs and the ability to scrap deny, or more balanced super techs and no scrap deny?


[...]

PR #1680, opened 2023-02-12
Change: Super Technical balance

Stubbjax:
This PR is a response to #1145 and #592 (and somewhat #648) and aims to strike a balance between reducing the strength of GLA's Super Technical[^1] and preserving the original feeling and traditional gameplay of 1.04.

### Changes include:

1. Increased GLA Technical xp requirements from [0 50 75 150] to [0 50 100 150]
2. Reduced China Supply Truck xp yield from 50 to 25

# Rationale

### 1. Technical xp requirements

In 1.04, the Technical's 25 xp requirement to go from Veteran to Elite rank is extremely small, and somewhat skewed in terms of progression (50 → 25 → 125). At the bare minimum, this allows a Technical to achieve Elite rank from killing a couple of infantry, but otherwise doesn't make any difference when it comes to vehicular combat, as most vehicles are worth at least 50 xp. Adding a further 25 xp to the Elite rank requirement is a small price to pay for a substantial reduction in Super Tech potential.

Before and after the change, a Technical still ranks up from Veteran to Elite by killing its prey of choice, the Humvee (which is worth 50 xp).

https://user-images.githubusercontent.com/11547761/218027015-98ddb214-5efe-470a-a570-616d72c7de54.mp4

In 1.04, a Technical only needs to kill two rocket infantry / rangers (worth 20 xp each) in order to progress from Veteran to Elite rank (50 + 20 + 20 = 90). This can be seen in the below footage, where a Technical reaches Elite rank by merely driving to an opponent's oils on Defcon 6, which is a very common occurrence. However, after the change, an additional infantry needs to be killed to reach Elite rank (50 + 20 + 20 + 20 = 110).

https://user-images.githubusercontent.com/11547761/218043401-53eb322b-aa6f-4992-b0d7-7388741b1068.mp4

Another way of looking at it is that a Technical now only needs an additional 10 xp to reach Elite rank. A Technical typically reaches Elite rank via infantry kills in 1.04 by going from 70 xp to 90 xp to cross the 75 xp requirement. The practical difference can thus be considered as an extra 10 xp (90 xp vs 100 xp).

### 2. Supply Truck xp yield

For the Technical xp adjustment to make a difference, the xp yield of Supply Trucks must also be reduced. Reducing the xp yield of Supply Trucks primarily targets the Super Tech issue while not drastically impacting too many other areas of the game. It is reasonable that Supply Trucks yield half the xp of Chinooks, as they cost half as much and are far easier to destroy. Other tanks / vehicles are typically worth 50+ xp for comparison. Combined with the Technical xp requirement increase, a Technical now only reaches Elite rank from killing two Supply Trucks, instead of Heroic as in 1.04. The Technical must destroy an extra tank / vehicle or hunt down another two Supply Trucks to achieve Heroic status - a feat that is far more worthy of the rank.

Not only do Elite Technicals have less firepower than Heroic Technicals, but they also have less health. This affords the opponent more time to deal with the Technical, and is less punishing over the same time frame. Note that the Technical does not acquire the fire rate boost that it typically would after killing the first truck (Veteran to Elite), making the destruction of the next truck a slightly more timely endeavour.

## 1.04:
A Technical can kill two Supply Trucks to reach Heroic rank. This assumes Technical Training is applied, which gives the Technical 50 xp by default. The two Supply Truck kills add +50 xp each to a total of 150 xp. The (high) reward does not align with the (relatively low) effort required.

https://user-images.githubusercontent.com/11547761/218010778-f8e4dda9-24df-41c7-a859-537d823492bb.mp4

## Patch:
A Technical will only reach Elite rank after killing two Supply Trucks. This assumes Technical Training is applied, which gives the Technical 50 xp by default. The two Supply Truck kills add +25 xp each for a total of 100 xp. The (medium-high) reward is closer to the (relatively low) effort required.

https://user-images.githubusercontent.com/11547761/218010868-51cf7060-3f0e-4305-b9a2-3ef1f82ca324.mp4

## Further considerations

If the halved xp yield is deemed too low, it can be increased to 30 or 40 while maintaining the altered Technical xp requirements to achieve the same outcome with reduced external impact. A Technical would still need to destroy two 40 xp Supply Trucks to reach Elite rank (50 + 40 + 40 = 130), though they would be much closer to reaching Heroic rank. Requiring three Supply Trucks to reach Heroic rank may be a reasonable compromise if this change is deemed too controversial or test results are negative.

If Technical xp requirements are more of a concern than Supply Truck xp yields, then another approach would be to change the Technical's Elite xp requirement to 80 xp (only +5 xp from 1.04), and leave Supply Trucks yielding 25 xp. This would result in the Technical still needing to destroy two 25 xp Supply Trucks to reach Elite rank (50 + 25 + 25 = 100), while effectively maintaining the same quick ranking progression via infantry kills.

Other ideas may involve subtly reducing the damage of the Technical's RPG weapon, e.g. from 50 damage to 45 damage to match the cannon weapon's damage (it may allow a Gattling Tank to take an extra hit). A damage increase counterbalanced with a fire rate reduction could also amplify the Technical's role as a hit-and-run type of unit, and make it more susceptible to prolonged engagements.

## Summary

The xp alterations serve as a reasonable balance approach because they are relatively subtle, intuitive, preserve the original feeling of 1.04, and make Super Techs harder to acquire instead of reducing their performance, which many players enjoy. Once acquired, a Super Tech can wreak the same havoc as it does in 1.04. The heightened requirement provides the opponent with a bit more counterplay potential, and heightens the skill required to acquire a Super Tech. The higher requirement ties in better with the devastating capabilities of a Super Tech, and adds more weight to their acquisition.

At the end of the day, an Elite Technical is a bit easier for China to deal with than a Heroic Technical. A Veteran Gattling Tank can effectively go head to head with an Elite Technical. An unranked Gattling Tank can even destroy the Technical if micromanaged correctly (#1199). This, of course, assumes limited micro on the Technical's part.

https://user-images.githubusercontent.com/11547761/218047806-5c4ebd12-ef16-4e25-b89f-304238e1cc55.mp4

https://user-images.githubusercontent.com/11547761/218047828-741b0bcc-33f9-40e8-9c56-127950bc9503.mp4

https://user-images.githubusercontent.com/11547761/218295923-c7e833ea-48e8-4a97-a047-a72c1efa701e.mp4

Also note that altering the xp yield for Supply Trucks will affect other units, and effectively double their requirements in terms of killing Supply Trucks. Humvees, for example, will need to kill four Supply Trucks to acquire Veteran rank, instead of two as in 1.04. However, this could also result in a positive outcome, and help reduce the severity of other issues, such as #640.

At the very least, this branch can serve as a testing ground to observe the impact Super Techs have on China's prospects. I believe this proposal will directly address the issue while incurring minimal costs elsewhere.

[^1]: A Super Technical is a Technical that has achieved god-like status by ranking up to Heroic (third) rank and collecting two salvage crates to acquire the RPG weapon. The RPG weapon deals 50 `EXPLOSION` damage, which is a damage type that is not resisted by many units in the game.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1680_supply_truck_xp_reward.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1680_supply_truck_xp_reward.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-02-12
+
+title: Decreases XP reward of China Supply Truck from 50 to 25
+
+changes:
+  - tweak: The China Supply Truck now gives 25 XP instead of 50 XP on kill. This takes a bit of pressure off of China when losing Supply Trucks to the opposition by not promoting enemy units as much.
+
+labels:
+  - buff
+  - china
+  - controversial
+  - design
+  - major
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1680
+
+authors:
+  - Stubbjax
--- a/Patch104pZH/Design/Changes/v1.0/1680_technical_xp_requirement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1680_technical_xp_requirement.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-02-12
+
+title: Increases XP requirement of GLA Technical from 50 75 150 to 50 100 150
+
+changes:
+  - tweak: The GLA Technical now requires 100 XP instead of just 75 XP to reach elite rank. This makes it a bit more challenging for GLA to acquire strong Technicals.
+
+labels:
+  - controversial
+  - design
+  - gla
+  - major
+  - nerf
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1680
+
+authors:
+  - Stubbjax
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -15082,7 +15082,8 @@ Object Boss_VehicleSupplyTruck
     Mass = 5.0
   End
 
-  ExperienceValue    = 50 50 50 50 ;Experience point value at each level
+  ; Patch104p @balance Stubbjax 12/02/2023 Decreases XP reward from 50. (#1680)
+  ExperienceValue    = 25 25 25 25 ;Experience point value at each level
 
   Behavior = SlowDeathBehavior ModuleTag_05
     DeathTypes = ALL -CRUSHED -SPLATTED
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -19557,8 +19557,9 @@ Object Chem_GLAVehicleTechnicalChassisOne
     Object = Chem_GLAArmsDealer
   End
 
+  ; Patch104p @balance Stubbjax 12/02/2023 Increases required XP from 0 50 75 150. (#1680)
   ExperienceValue         = 25 25 50 100    ;Experience point value at each level
-  ExperienceRequired      = 0 50 75 150  ;Experience points needed to gain each level
+  ExperienceRequired      = 0 50 100 150  ;Experience points needed to gain each level
   IsTrainable             = Yes             ;Can gain experience
   CrusherLevel            = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
   CrushableLevel          = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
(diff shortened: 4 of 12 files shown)


ExiLe-ZH:
Actually got an alternative idea for this case, what about changing the Veteran requirement to 25XP instead? 

So [0 25 75 150].

It has the exact same effect as the suggested change, but also gives unvetted Techs more potential. This will make the General Promotion less of a no-brainer and makes alternatives like Marauders more attractive.

xezon:
> So [0 25 75 150]

I do not like this. Makes the General Promotion less valuable.

ExiLe-ZH:
> > So [0 25 75 150]
> 
> 
> 
> I do not like this. Makes the General Promotion less valuable.

That's the intention. The genpoint still remains completely meta no doubt though. The extra firepower for ALL Techs straight from the arms dealer is a huge deal.

xezon:
Great presentation. I like this proposal. It looks simple and intuitive. GLA Super Technical becomes just a bit more difficult to acquire, most notably against China Trucks.

xezon:
XP reward changes to CINE units are reverted. Documentation added. Commit titles fixed.

PR #1682, opened 2023-02-12
Pathfinder balance

Stubbjax:
Welcome to the Pathfinder balance branch! This PR is a response to TheSuperHackers/GeneralsGamePatch#690 and can hopefully serve as a testing ground for various balance adjustments involving the fearsome Pathfinder.

The Pathfinder is one of the most overpowered units in the game due to its ability to immediately eliminate infantry at a great distance with very little recourse for opponents. This is a unit that severely tips the balance once it becomes available, and effectively renders all infantry-based gameplay obsolete from that point forward.

The primary objective in addressing this issue is to reduce the Pathfinder's effectiveness in the most impactful _and_ subtle way possible while maintaining traditional mechanics and gameplay. Unfortunately, it is practically impossible to completely balance the Pathfinder without significantly altering the feel of the game and upsetting a lot of players, so a compromise is naturally required. Changes that result in infantry taking multiple shots to take down for example, while perhaps an effective balance option, would be incredibly visible and thus highly susceptible to controversy. These types of changes are thus generally best avoided.

Keep in mind that not all of the following changes have to be agreed upon, and even a single nerf is still better than 1.04's massively overpowered state. Pathfinders will likely always remain overpowered to some degree, but the extent to which they are overpowered can certainly be reduced in an elegant manner.

The suggested changes below intend to spread the impact across a broad range of attributes - most of which I'd confidently expect to go unnoticed by the majority of players. They are roughly listed in order of impact over visibility.

### Changes include:

1. Double the delay between shots from 2s to 4s
2. Double experience requirements from [0 50 100 200] to [0 100 200 400]
3. Increase build time from 10s to 12s
4. Reduce attack range from 300 to 275
5. Increase unit cost from $600 to $800
6. Reduce move speed from 30 to 25 (ColonelBurtonGroundLocomotor → BasicHumanLocomotorPlus25)
7. Increase stealth delay from 0s to 1s
8. Increase experience yields from [40 40 60 80] to [60 60 80 100]

### 1. Shot delay
Increasing the delay between shots from 2s to 4s may seem like a big change, but it is an incredibly effective way of reducing DPS without the accompanying visibility cost of an equivalent damage reduction. A 2s increase is really as high as it can go before it starts to become conspicuous, and I'd argue most players would not notice any difference. The realism card can also be played here, as it makes sense that snipers (especially of the meticulous and conscientious USA) take time and care to aim and perfectly execute their shots. The higher delay between shots also contrasts nicely with the much faster Jarmen Kell, who wastes no time mercilessly firing off no-scoped shots like a cold-blooded killer at the cost of shorter range and revealing himself when doing so.

The greater delay between shots also affords opponents a larger window to react after witnessing the first shot, providing a greater degree of counterplay and allowing them to rescue or retreat ~more~ some of their soldiers.

A fire rate reduction is preferable to an equivalent damage reduction due to the fact that players are accustomed to interpreting a single shot as a kill. If damage were to be altered instead, Pathfinders would potentially fire multiple shots per target, which would significantly reduce feedback clarity and certainly change the feeling of the game in comparison to 1.04.

Pathfinder fire rate in 1.04:
```
60 frames at rank 0
50 frames at rank 1 (120% fire rate = 60 / 1.2 = 50)
43 frames at rank 2 (140% fire rate = 60 / 1.4 = 42.6)
38 frames at rank 3 (160% fire rate = 60 / 1.6 = 37.5)
```
Pathfinder fire rate in patch:
```
120 frames at rank 0
100 frames at rank 1 (120% fire rate = 120 / 1.2 = 100)
 86 frames at rank 2 (140% fire rate = 120 / 1.4 = 85.7)
 75 frames at rank 3 (160% fire rate = 120 / 1.6 = 75)
```
Increasing the delay between shots by a factor of two is a very reasonable adjustment for a long-range unit that typically kills all of its targets in one shot. With such a long range, Pathfinders can be likened to or considered artillery units, where a slower fire rate is a very common and reasonable tradeoff. A reduced fire rate would more closely align with the design principles of other long-range units.

| Unit | Fire Rate |
| - | - |
| Pathfinder (1.04) | 2s |
| Pathfinder (this) | 4s |
| Inferno Cannon | 4s |
| Rocket Buggy | 6s |
| Tomahawk | 7s |
| Nuke Cannon | 10s |
| SCUD Launcher | 10s |

Below is some footage demonstrating how effortlessly three Pathfinders can take out an entire squad of Tunnel Defenders.

https://user-images.githubusercontent.com/11547761/197380205-09aacc4e-1f78-45f2-8537-c75918a6a8a9.mp4

After such violent and graphic footage, you must be thinking - wow, Pathfinders are really strong! But guess what? The above footage was not taken on 1.04, but on this branch with the respective changes applied! The Tunnel Defenders would be eliminated in less than half the time in 1.04, and each Pathfinder would be Elite rank and 1 - 2 kills from Heroic. The increased shot delay appears to strike a perfect balance between reducing the unit's effectiveness and maintaining the same feeling as 1.04.

### 2. Experience requirements
On average, Pathfinders rank up faster than every other single-target unit in the game, and can reach Veteran rank after 4 seconds, Elite after 7 seconds, and Heroic after 15 seconds - assuming the targets are unranked missile infantry (the most common encounter). It takes as little as 7 seconds to reach Heroic rank with Advanced Training.

Doubling the experience requirements, in combination with the reduced fire rate, results in a 4x longer rank up duration in comparison to 1.04. With the respective change, a Pathfinder has to take out 20 / 10 (~62s / ~31s) missile infantry to reach Heroic rank, rather than 10 / 5 (~15s / ~7s) as in 1.04. This slower rate of progression feels more natural and a lot closer to the time investment required of other units to achieve similar ranks - especially considering the lack of risk Pathfinders typically face against their targets. As Pathfinders become more valuable over time, their loss also becomes much more impactful. This leads to a higher potential for / degree of rubberbanding. The requirements are compared in the table below. Time is displayed in {minutes:seconds.frames}. Each row is a single shot.

image

This change technically serves as an additional fire rate reduction, without actually reducing the fire rate directly, by delaying access to faster fire rates via ranking. This reduces visibility and allows the game to feel the same while still reducing DPS further. The slower fire rate and ranking speed result in a much greater time investment from the user, and affords opponents more time to react and adapt to the situation before incurring heavy losses.

### 3. Build time
A slight build time increase from 10s to 12s will make the unit a bit harder to spam, and gives opponents slightly more time to prepare. An increase of two seconds is reasonable as it is an even number and relatively close to the original value. Increasing the time any further would likely incur too much of a visibility cost. Note that the build times do add up, and more Pathfinders are already required than previously to achieve the same effect.

### 4. Attack range
A slight attack range reduction gives opponents a bit more flexibility, and requires Pathfinders to get a bit closer to danger. Pathfinders are often combined with the Search & Destroy battle plan, and it feels reasonable to select a value that would put the maximum range at a nice midpoint between the two original ranges (275 → 330 vs 300 → 360). A value of 275 also ties in with the ranges of some base defences, such as the Firebase and EMP Patriot, which allows opponents to more accurately protect their infantry within or behind their defences. It is hard to reduce the range much further without affecting the realism aspect of the unit and potentially becoming too noticeable.

The below image illustrates the 25 range difference (300 vs 275), which is about the length of a Humvee.

image

### 5. Price
A $200 increase from $600 to $800 will make the unit a bit harder to spam, and will potentially require players to make compromises elsewhere. A player might not be able to afford a few extra Missile Defenders or another Humvee, which could be the distinguishing factor in an important encounter. A price of $800 is reasonable as it is not too different from 1.04 and has a degree of consistency due to being shared with other infantry units like the Angry Mob and Saboteur. It also helps that 8 is visually similar to 6. A $200 / 33% increase is probably as big of a change as you'd want to make when considering the unit's prevalence and specialisation while minimising visibility / controversy.

### 6. Movement
It does not seem fair that infantry are more or less screwed if they attempt to outrun Pathfinders. Applying a slower movement speed affords opponents a greater opportunity to retreat and slightly increases the time it takes to load up Humvees. The only infantry that can 'outrun' (match) Pathfinders in 1.04 are Saboteurs, Hijackers, Workers (with shoes) and heroes. After the change; Rangers, Red Guard, Rebels, Workers (without shoes), Terrorists and Hackers* join the club. Also note that the `BasicHumanLocomotorPlus25` locomotor has a slower turn rate of 360, compared with the original 500. (Not only is the slower rotation or 'aiming' more realistic, but a value of 360 fits right in with internet culture.)

### 7. Stealth delay
A one second stealth delay allows Pathfinders to be briefly seen after they stop moving, which provides opponents with enhanced readability and a greater opportunity to recall their position. This heightens the risk of movement, and ties in more effectively with the other conditional stealth behaviours in the game, which all feature some form of delay. It also allows opponents to more effectively see when a Pathfinder is evacuated from a vehicle, and more care will have to be taken when evacuating.

### 8. Experience yields
Increasing the experience yields is a subtle way to reward opponents for killing such a powerful and specialised unit. Though Pathfinders are often contained in some form of transport (where xp is not distributed if the occupant dies with the transport; see TheSuperHackers/GeneralsGameCode#55), it could add some much-needed risk to leaving Pathfinders out in the open, or make players think twice about evacuating a critically damaged Humvee.

### Strategy Center prerequisite
Another idea would be to introduce the Strategy Center as a prerequisite. This would provide opponents with some additional counterplay options / strategies, where they can destroy an opponent's Strategy Center to halt Pathfinder production. It also makes sense that such a powerful, late-game unit is locked behind the tech building. It would also increase the risk of selling one's Command Center and protecting one's dozers, as it might prevent the user from being able to acquire a Strategy Center and Pathfinders by extension. The only caveat with such a change - and the reason I have not included it in the initial list of changes - would be its high visibility and susceptibility to controversy, as it would likely be noticed by the majority of players. Regardless, it could certainly be considered reasonable enough for most players to accept.

### Hero resistance
It might also be an idea to introduce alternate 'hero' armour types for Colonel Burton, Black Lotus and Jarmen Kell, with the only difference being a reduced SNIPER damage multiplier. This would allow the respective heroes to take a few extra shots from Pathfinders, which could be an acceptable change as heroes can already sustain a hit from a lesser ranked Pathfinder in 1.04.

### Fire vulnerability
As Pathfinders wear gillie suits, it would add a bit of humour and realism if they made good kindling. Sometimes a Dragon's firewall ability or a MiG force-firing on the ground is all China has at its disposal to deal with an immediate Pathfinder threat in their base.

## Summary

I firmly believe the above combination of changes strikes a very reasonable balance between reducing the Pathfinder's effectiveness and maintaining the original feeling of 1.04, with no major discernible behavioural or mechanical differences. The doubled weapon reload time and experience requirements may seem fairly substantial, but when considering the fact that a _single_ Pathfinder typically ends up killing dozens of infantry in 1.04, the adjustments are more than reasonable. In an ideal world, Pathfinders would likely be nerfed even further / differently.

Anyway, feel free to test the proposed changes and see how the game feels!

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4590,7 +4590,7 @@ Object AirF_AmericaInfantryPathfinder
     AlwaysHeal = 0.25 ; if we get below 25%, find healing right away.
   End
 
-  Locomotor = SET_NORMAL ColonelBurtonGroundLocomotor
+  Locomotor = SET_NORMAL BasicHumanLocomotorPlus25
 
   Behavior = ExperienceScalarUpgrade ModuleTag_05
     TriggeredBy = Upgrade_AmericaAdvancedTraining
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4607,7 +4607,7 @@ Object AirF_AmericaInfantryPathfinder
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
   Behavior = StealthUpdate ModuleTag_09
-    StealthDelay                = 0           ; msec
+    StealthDelay                = 1000        ; msec
     StealthForbiddenConditions  = MOVING ; stays stealthy while attacking
     FriendlyOpacityMin          = 30.0%
     FriendlyOpacityMax          = 80.0%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4541,8 +4541,8 @@ Object AirF_AmericaInfantryPathfinder
   BuildCost = 800
   BuildTime = 12.0          ;in seconds
 
-  ExperienceValue = 40 40 60 80    ;Experience point value at each level
-  ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
+  ExperienceValue = 60 60 80 100    ;Experience point value at each level
+  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel      = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet      = AmericaInfantryPathfinderCommandSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4538,8 +4538,8 @@ Object AirF_AmericaInfantryPathfinder
     Object = AirF_AmericaBarracks
     Science = SCIENCE_Pathfinder
   End
-  BuildCost = 600
-  BuildTime = 10.0          ;in seconds
+  BuildCost = 800
+  BuildTime = 12.0          ;in seconds
 
   ExperienceValue = 40 40 60 80    ;Experience point value at each level
   ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1788,8 +1788,8 @@ Object AmericaInfantryPathfinder
     Object = AmericaBarracks
     Science = SCIENCE_Pathfinder
   End
-  BuildCost = 600
-  BuildTime = 10.0          ;in seconds
+  BuildCost = 800
+  BuildTime = 12.0          ;in seconds
 
   ExperienceValue = 40 40 60 80    ;Experience point value at each level
   ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1791,8 +1791,8 @@ Object AmericaInfantryPathfinder
   BuildCost = 800
   BuildTime = 12.0          ;in seconds
 
-  ExperienceValue = 40 40 60 80    ;Experience point value at each level
-  ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
+  ExperienceValue = 60 60 80 100    ;Experience point value at each level
+  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel      = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet      = AmericaInfantryPathfinderCommandSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1856,7 +1856,7 @@ Object AmericaInfantryPathfinder
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
   Behavior = StealthUpdate ModuleTag_09
-    StealthDelay                = 0           ; msec
+    StealthDelay                = 1000        ; msec
     StealthForbiddenConditions  = MOVING ; stays stealthy while attacking
     FriendlyOpacityMin          = 30.0%
     FriendlyOpacityMax          = 80.0%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1839,7 +1839,7 @@ Object AmericaInfantryPathfinder
     AlwaysHeal = 0.25 ; if we get below 25%, find healing right away.
   End
 
-  Locomotor = SET_NORMAL ColonelBurtonGroundLocomotor
+  Locomotor = SET_NORMAL BasicHumanLocomotorPlus25
 
   Behavior = ExperienceScalarUpgrade ModuleTag_05
     TriggeredBy = Upgrade_AmericaAdvancedTraining
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17888,7 +17888,7 @@ Object Boss_InfantryPathfinder
     AlwaysHeal = 0.25 ; if we get below 25%, find healing right away.
   End
 
-  Locomotor = SET_NORMAL ColonelBurtonGroundLocomotor
+  Locomotor = SET_NORMAL BasicHumanLocomotorPlus25
 
   Behavior = ExperienceScalarUpgrade ModuleTag_05
     TriggeredBy = Upgrade_AmericaAdvancedTraining
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17905,7 +17905,7 @@ Object Boss_InfantryPathfinder
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
   Behavior = StealthUpdate ModuleTag_09
-    StealthDelay                = 0           ; msec
+    StealthDelay                = 1000        ; msec
     StealthForbiddenConditions  = MOVING ; stays stealthy while attacking
     FriendlyOpacityMin          = 30.0%
     FriendlyOpacityMax          = 80.0%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17836,8 +17836,8 @@ Object Boss_InfantryPathfinder
     Object = Boss_Barracks
     Science = SCIENCE_Pathfinder
   End
-  BuildCost = 600
-  BuildTime = 10.0          ;in seconds
+  BuildCost = 800
+  BuildTime = 12.0          ;in seconds
 
   ExperienceValue = 40 40 60 80    ;Experience point value at each level
   ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -17839,8 +17839,8 @@ Object Boss_InfantryPathfinder
   BuildCost = 800
   BuildTime = 12.0          ;in seconds
 
-  ExperienceValue = 40 40 60 80    ;Experience point value at each level
-  ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
+  ExperienceValue = 60 60 80 100    ;Experience point value at each level
+  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel      = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet      = AmericaInfantryPathfinderCommandSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -4078,8 +4078,8 @@ Object Lazr_AmericaInfantryPathfinder
     Object = Lazr_AmericaBarracks
     Science = SCIENCE_Pathfinder
   End
-  BuildCost = 600
-  BuildTime = 10.0          ;in seconds
+  BuildCost = 800
+  BuildTime = 12.0          ;in seconds
 
   ExperienceValue = 40 40 60 80    ;Experience point value at each level
   ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -4081,8 +4081,8 @@ Object Lazr_AmericaInfantryPathfinder
   BuildCost = 800
   BuildTime = 12.0          ;in seconds
 
-  ExperienceValue = 40 40 60 80    ;Experience point value at each level
-  ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
+  ExperienceValue = 60 60 80 100    ;Experience point value at each level
+  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel      = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet      = AmericaInfantryPathfinderCommandSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -4130,7 +4130,7 @@ Object Lazr_AmericaInfantryPathfinder
     AlwaysHeal = 0.25 ; if we get below 25%, find healing right away.
   End
 
-  Locomotor = SET_NORMAL ColonelBurtonGroundLocomotor
+  Locomotor = SET_NORMAL BasicHumanLocomotorPlus25
 
   Behavior = ExperienceScalarUpgrade ModuleTag_05
     TriggeredBy = Upgrade_AmericaAdvancedTraining
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/LaserGeneral.ini
@@ -4147,7 +4147,7 @@ Object Lazr_AmericaInfantryPathfinder
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
   Behavior = StealthUpdate ModuleTag_09
-    StealthDelay                = 0           ; msec
+    StealthDelay                = 1000        ; msec
     StealthForbiddenConditions  = MOVING ; stays stealthy while attacking
     FriendlyOpacityMin          = 30.0%
     FriendlyOpacityMax          = 80.0%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -4631,7 +4631,7 @@ Object SupW_AmericaInfantryPathfinder
     CanDetectWhileContained   = No ;Contained means being in a transport or tunnel network.
   End
   Behavior = StealthUpdate ModuleTag_09
-    StealthDelay                = 0           ; msec
+    StealthDelay                = 1000        ; msec
     StealthForbiddenConditions  = MOVING ; stays stealthy while attacking
     FriendlyOpacityMin          = 30.0%
     FriendlyOpacityMax          = 80.0%
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -4614,7 +4614,7 @@ Object SupW_AmericaInfantryPathfinder
     AlwaysHeal = 0.25 ; if we get below 25%, find healing right away.
   End
 
-  Locomotor = SET_NORMAL ColonelBurtonGroundLocomotor
+  Locomotor = SET_NORMAL BasicHumanLocomotorPlus25
 
   Behavior = ExperienceScalarUpgrade ModuleTag_05
     TriggeredBy = Upgrade_AmericaAdvancedTraining
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -4562,8 +4562,8 @@ Object SupW_AmericaInfantryPathfinder
     Object = SupW_AmericaBarracks
     Science = SCIENCE_Pathfinder
   End
-  BuildCost = 600
-  BuildTime = 10.0          ;in seconds
+  BuildCost = 800
+  BuildTime = 12.0          ;in seconds
 
   ExperienceValue = 40 40 60 80    ;Experience point value at each level
   ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SuperWeaponGeneral.ini
@@ -4565,8 +4565,8 @@ Object SupW_AmericaInfantryPathfinder
   BuildCost = 800
   BuildTime = 12.0          ;in seconds
 
-  ExperienceValue = 40 40 60 80    ;Experience point value at each level
-  ExperienceRequired = 0 50 100 200  ;Experience points needed to gain each level
+  ExperienceValue = 60 60 80 100    ;Experience point value at each level
+  ExperienceRequired = 0 100 200 400  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
   CrushableLevel      = 0  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
   CommandSet      = AmericaInfantryPathfinderCommandSet
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -911,7 +911,7 @@ End
 Weapon USAPathfinderSniperRifle
   PrimaryDamage = 100.0
   PrimaryDamageRadius = 0.0
-  AttackRange = 300.0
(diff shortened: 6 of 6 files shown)


ExiLe-ZH:
Would also look into damage values, would there be a value where a hero can survive 1 shot but normal infantry can't? (well maybe vetted inf).

Edit: I see you mentioned armour type, that's the only way to achieve this?

ExiLe-ZH:
as for build time, here an overview of all other infantry:

|Unit|Build Time|
|:----|:----|
|Ranger|5|
|Missile Defender|5|
|Pathfinder|10|
|Colonel Burton|20|
|Red Guard [China]|10|
|Red Guard [Tank, Nuke]|12|
|Minigunner|10|
|Tank Hunter [China, Inf]|5|
|Tank Hunter [Tank, Nuke]|7|
|(Super) Lotus|20|
|Hacker|20|
|Rebel|10|
|RPG|10|
|Terrorist|10|
|Angry Mob|30|
|Hijacker [GLA]|20|
|Hijacker [Stealth]|30|
|Saboteur|30|
|Jarmen Kell|40|

commy2:
You could consider giving him a pre attack delay in place of this increased reload time. That way, he has to take aim before killing a unit. Vs mass units, the effect is essentially the same.

https://github.com/TheSuperHackers/GeneralsGamePatch/issues/690#issuecomment-1186443240

penfriendz:
Can we remove pathfinder stealth detection? I don't see a need for it. It makes pathfinders way too good against heroes and USA already has ample stealth detection without a brainless automatic hero killer. It's also a pretty low visibility change imo.

ExiLe-ZH:
> Can we remove pathfinder stealth detection? I don't see a need for it. It makes pathfinders way too good against heroes and USA already has ample stealth detection without a brainless automatic hero killer. It's also a pretty low visibility change imo.

This is a key ability indeed and I feel we have to look at those kinda abilities first before details like ROF.

All are mentioned here: https://github.com/TheSuperHackers/GeneralsGamePatch/issues/690

Stubbjax:
> Would also look into damage values, would there be a value where a hero can survive 1 shot but normal infantry can't? (well maybe vetted inf).

The `USAPathfinderSniperRifle` weapon deals 100 damage, which is modified to 200 damage by all human-based armours (not sure why the weapon doesn't just deal 200 damage in the first place with a standardised 100% modifier, but that's EA for ya). We could probably change the modifier to 180% to spare only heroes. This would have the side-effect of allowing vetted Rangers to survive hits from vetted Pathfinders.

A major part of the problem is that Pathfinders are often Heroic rank, and so deal ×1.3 damage for a total of 260 damage. Only a Heroic (×1.5 hp) hero can survive this in 1.04.

```
Weapon USAPathfinderSniperRifle
  PrimaryDamage = 100.0
End
```
```
Armor HumanArmor
  Armor = SNIPER 200%
End
```

For reference:

| Unit | Health |
|-|-|
| Colonel Burton | 200 |
| Black Lotus | 200 |
| Jarmen Kell | 200 |
| Ranger | 180 |
| Pathfinder | 120 |
| Red Guard | 120 |
| Minigunner | 120 |
| Rebel | 120 |
| Terrorist | 120 |
| Saboteur | 120 |
| Missile Defender | 100 |
| Tank Hunter | 100 |
| Tunnel Defender | 100 |
| Hacker | 100 |
| Hijacker | 100 |
| Worker | 100 |

> You could consider giving him a pre attack delay in place of this increased reload time. That way, he has to take aim before killing a unit. Vs mass units, the effect is essentially the same.
> 
> [#690 (comment)](https://github.com/TheSuperHackers/GeneralsGamePatch/issues/690#issuecomment-1186443240)

I did consider this, but strongly dislike the idea of reducing unit responsiveness as I (and I'd confidently say every player on the planet) find it to be one of the most frustrating things to play with. I can easily imagine clicking to attack and briefly wondering why and being annoyed by my unit not responding. Remember the Sentry Drone.

Maybe if there was some kind of 'aiming' or 'raising rifle' animation then it could be acceptable, but it wouldn't be a total replacement for the fire rate reduction.

> > Can we remove pathfinder stealth detection? I don't see a need for it.
> 
> This is a key ability indeed and I feel we have to look at those kinda abilities first before details like ROF.

I believe ROF is the single most important attribute behind the Pathfinder's power, and the most 'bang for buck' nerf we can apply. Removing the stealth detection may be a viable option, but it would also involve removing an ability / mechanic, which could be quite visible and upsetting to players who rely on it or use it often (e.g. dropping Pathfinders into Infantry General Hackers). It would also do very little to quell the massacre that always occurs out on the battlefield when a group of unsuspecting infantry enter a Pathfinder's range, which is the biggest problem. I do agree that it is a superfluous ability though, and USA already has ample stealth detection. Minor secondary abilities like this can certainly be looked at after the biggest problems are dealt with.

I suggest you give the current changes a _shot_. You might be surprised how much like 1.04 it feels.

xezon:
> dislike the idea of reducing unit responsiveness

I agree with this.

The alternative to this would be Proposal 11

> Make Pathfinder no longer use a hitscan weapon and use a dummy projectile instead. Dummy projectile doesn't track targets, but the weapon has a secondary damage radius with lower damage. Therefore, if Pathfinder hits stationary target, it gets hit with full damage and dies and if Pathfinder's target is moving, it'll only be hit by secondary damage, and require multiple hits to die.

commy2:
> Maybe if there was some kind of 'aiming' or 'raising rifle' animation then it could be acceptable, but it wouldn't be a total replacement for the fire rate reduction.

The diff I linked contains an "aiming" animation, which makes the unit act quite differently than the 1.04 Sentry Drone.

Ref https://github.com/commy2/zerohour/commit/48b06f0d096ba37a7fd48ce210a534524c37929f

Also, a Sniper aiming before taking the shot is expected behaviour I would argue. See also the Sniper from Company of Heroes and probably other RTS games.

MTKing4:
I like the changes, i think only after all the 8 changes combined we will see an impactful nerf on pathfinders, but i suspect one can't be sure without some game tests

Stubbjax:
> The diff I linked contains an "aiming" animation, which makes the unit act quite differently than the 1.04 Sentry Drone.
> 
> Ref [commy2/zerohour@48b06f0](https://github.com/commy2/zerohour/commit/48b06f0d096ba37a7fd48ce210a534524c37929f)
> 
> Also, a Sniper aiming before taking the shot is expected behaviour I would argue. See also the Sniper from Company of Heroes and probably other RTS games.

I tested this and it doesn't feel good. The Pathfinder just stands there for a while before firing. It would be better if it had an animation over the duration of the predelay, in a deploy / undeploy kind of way (perhaps closer to C&C3's Sniper Team). Regardless, I feel like simply increasing the delay between shots more or less accomplishes the same thing but in a less annoying manner.

https://user-images.githubusercontent.com/11547761/218695898-e2d45443-d8b3-4b7f-91b8-f52958768c70.mp4

---
> I like the changes, i think only after all the 8 changes combined we will see an impactful nerf on pathfinders, but i suspect one can't be sure without some game tests

Great! I recommend giving 1v3 Infantry General Hard Armies as USA a shot.

MTKing4:
More like an actual game Vs a player, say USA Vs Infantry to see if it still sucks to play late game vs A USA

commy2:
> I tested this and it doesn't feel good. The Pathfinder just stands there for a while before firing. 

Looks right to me. Rifle is raised, obviously different from standing around idle where the rifle is half down.

There may be a raising rifle slowly/deploy animation too. Would have to check on my machine on the W/E.

ExiLe-ZH:
Had some ideas which probably are too mod'ish but interesting to think about.

Like imagine two shooting modes:

1. Shooting from normal standing position; it's quick, but reveals itself (like Jarmen), misses moving targets (or less damage) and with slower ROF.

2. The second shooting position is prone, it can hit moving targets (full damage), remains stealth while shooting and shoots faster (all exactly like it already is in 1.04). The downside however is the mobility and deployment time, same like a Nuke Cannon.

xezon:
Diverting designs should be discussed in TheSuperHackers/GeneralsGamePatch#690 to not pollute this particular Pull.

xezon:
Nice proposals. My thoughts. I think this Pull needs splitting for each individual change, so they can be discussed in isolation.

### 1. Double the delay between shots from 2s to 4s

Rationale for this reads good. However it is not yet clear to me if halving is the ideal reduction. Perhaps 2s to 3s, 50% slower, would already be sufficient?

### 2. Double experience requirements from [0 50 100 200] to [0 100 200 400]

This looks like very good idea. Looks similar or identical to the Jarmen Kell experience requirements.

### 3. Increase build time from 10s to 12s

Increased build time has the unfortunate consequence of reducing the chance for reactive counter-play. Meaning a Pathfinder might produce a second too late after Lotus has captured a nearby Command Center. With many of the other nerfs, such as increased build cost, it perhaps would no longer be necessary to increase the build time.

### 4. Reduce attack range from 300 to 275 (S&D 360 to 330)

The Search And Destroy range looks like a very important consideration here. Original 360 range is really huge. And in garrison (+33%) that turns to 550 if I am not mistaken (300 * (1.2 + 1.33)). So range in Humvee is 550? I think we should double check.

What is the attack range of Jarmen Kell? 300? Looks like Jarmen Kell is reliably outgunned if so. Range reduction looks plausible.

### 5. Increase unit cost from $600 to $800

600 certainly is cheap for this unit. 800 looks okay, though I am not sure about best price here. Would need to see prices of other infantry.

### 6. Reduce move speed from 30 to 25 (ColonelBurtonGroundLocomotor → BasicHumanLocomotorPlus25)

I like this very much. The Pathfinder does not need to walk this fast.

### 7. Increase stealth delay from 0s to 1s

I like this too. Could we get some intel on how long other stealth units take to stealth when idle? Hijacker, Outpost, Sentry Drone etc.

### 8. Increase experience yields from [40 40 60 80] to [60 60 80 100]

Could do. I would like to see other infantry XP yields for better clarity.

### 9. Strategy Center prerequisite

I do not like this at this time. It would eliminate certain match scenarios, such as USA no longer able to turn around a desperate situation against Infantry General with just a few buildings and a barracks left to produce a deadly Pathfinder to kill the approaching troops.

### 10. Hero resistance

I like if Hero's could survive one shot reliably. This could give Lotus and Burton the chance to successfully attack a building or Jarmen Kell to counter attack.

### 11. Fire vulnerability

Interesting idea. Would buff Infantry General mostly though. And Infantry Generals is most vulnerable against Pathfinders.

ExiLe-ZH:
I think the best way to go about Pathfinder balancing is a 2 step approach.

Step 1: Nerf the things where pathfinders have an advantage over Jarmen Kell. Examples: detecting stealth, not revealed while shooting.

Step 2: Further nerfing, because we don't want Jarmen copies (only lacking the vehicle snipe) that are build cheaper and faster. Neither do we want to bump the price and build time to actual hero values. They are not hero units and shouldn't perform like one either.

commy2:
> So range in Humvee is 550?

Buildings use GarrisonContain, which grants GARRISONED to passengers, but generally vehicles (TransportContain) do not. Exception is the Helix (HelixContain also grants GARRISONED).

Iirc, not even Bunker Overlord does, because while the Bunker object receives GARRISONED from OverlordContain, the bunker inside uses TransportContain again.

xezon:
Would you consider it a bug that Helix gives extended range?

ExiLe-ZH:
> I think the best way to go about Pathfinder balancing is a 2 step approach.
> 
> Step 1: Nerf the things where pathfinders have an advantage over Jarmen Kell. Examples: detecting stealth, not revealed while shooting.
> 
> Step 2: Further nerfing, because we don't want Jarmen copies (only lacking the vehicle snipe) that are build cheaper and faster. Neither do we want to bump the price and build time to actual hero values. They are not hero units and shouldn't perform like one either.

### To give more detailed feedback on this particular pull, I think most of these will fit step 1:

2) Experience requirements, there's no reason why a pathfinder should promote faster than Jarmen Kell, especially with USA having Advanced Training it's highly imbalanced.

3) Build time, 10 seconds is definitely too fast compared to 20 seconds of heroes. 12-15 seconds sounds appropriate.

4) Attack range, I think it would be fair if S&D equalized the Pathfinder's range with Jarmen Kell, what is Jarmen's range exactly?

5) Increase the cost, $600 is almost 3x as cheap as hero price, which definitely is way too low. $800 sounds more appropriate.

8. Experience yields, I agree this needs some tweaking, but the currents suggestions are too high. Original base value of 40 is good because it should be a little less than heroes, not more. Higher veteran levels are definitely too low in 1.04. I think the optimal values would be 40-40-80-120.

### These should be highly considered as further nerfs for step 2:

1. Firerate, this makes sense to me, in real life sniper units usually use bolt-action sniper rifles, they tend to be more accurate than semi-automatics. The current firerate is way too fast to reflect bolt-action usage.
Ofcourse it's not all about realism, I'm sure Jarmen doesn't use a semi-automatic rifle either, but he can keep his unearthly skills because he's a hero. 

6) Movement speed, I think a no-brainer, it should be like normal infantry, not heroes.

7) Stealth delay, how does it compare to heroes and other stealth units (GPS'ed)?

### Missing nerfs for step 1 (my own suggestions):

* Pathfinder being able to detect heroes is unfair, feature should be removed.
* Pathfinder should be revealed while shooting, it's too OP and unfair vs Jarmen Kell.

xezon:
I think we can start splitting this change into smaller chunks and then conclude them separately. This way it will be easier to finalize.

ExiLe-ZH:
> I think we can start splitting this change into smaller chunks and then conclude them separately. This way it will be easier to finalize.

I agree, this is kinda hard to discuss, it's too much.

Do you generally agree with my suggestion of the 2 step approach though? Or should things be balanced assymetrically? Like different pros and cons for both units (Jarmen and Pathfinders) like currently, this makes things harder if you ask me though.

xezon:
I do not understand what the goal with the 2 steps are. I just see all nerfs as their own entity.

> Pathfinder being able to detect heroes is unfair, feature should be removed.

This could cause troubles in Mission maps where Pathfinder is used to detect demo traps and such things.

> Pathfinder should be revealed while shooting, it's too OP and unfair with Jarmen Kell.

Is there any other unit in the game that stays stealthed while shooting?

ExiLe-ZH:
> I do not understand what the goal with the 2 steps are. I just see all nerfs as their own entity.
> 

I think the goal would be to make Pathfinder perform like good special units, but not like actual heroes. The only advantage vs heroes should be the price, build time and unlimited quantity. 

Like currently you barely can even utilize Jarmen anymore vs a USA with Pathfinders. His survival chances are just too low without putting him on a Bike or in a Bus, which both have their disadvantages aswell.

That's why I think it makes sense to first take away unfair advantages. Then as a second step apply appropriate nerfs.

> > Pathfinder being able to detect heroes is unfair, feature should be removed.
> 
> This could cause troubles in Mission maps where Pathfinder is used to detect demo traps and such things.

Something to keep in mind indeed, is it ok to edit missions and add Sentry Drones?

> 
> > Pathfinder should be revealed while shooting, it's too OP and unfair with Jarmen Kell.
> 
> Is there any other unit in the game that stays stealthed while shooting?

I don't think so.

commy2:
> Would you consider it a bug that Helix gives extended range?

Definitely, although it is required to make Helix viable at all, so it is better to keep it (and remove GARRISONED range bonus from Minigunner AA weapon :)

xezon:
Excal:

> Inf vs USA:
Inf is much better in the early game and it takes quite a bit of work for, and is hard for, USA to get level 3 as infantry gives off low XP. When USA hits level 3, they are aiming to get a “death blob” of at least 2 avengers + vee pathfinder combo. Not easy to do.

> By this stage of the game, inf usually has more of the map and has answers in the form of the propaganda tech (ecm etc) + multiple airfield of migs.

> I personally like how the advantage can swing to USA but it for sure isn’t an autoloss.

> I actually think at the highest level USA may overall have the edge, but at lower level it is inf for sure as it is a lot easier to a move an infantry blob than it is to micro multiple units as USA. Since the patch is noob friendly you may not want to nerf the pathfinder too hard.

MTKing4:
Commy2

> When you're testing multiplayer, test Tracers of Pathfinders. Are they visible for enemy Pathfinder if they're hidden? Same for Jarmen firing from a building.

ExiLe-ZH:
What Excal says is exactly my experience when playing vs good players, early game is hard to play vs Inf and it's by no means an auto-win when getting Pathies. It does turn it around a little but Inf usually has more eco/map control and most likely Migs in the mid game. This means you're on a timer to move out/do damage for which you'll need a good army composition with atleast like 3 Avengers, which is only possible if you've microed and preserved your units very well. When getting a solid composition it's still no auto-win because good Inf players will use ECM's/Lotus to disable your Avengers and time deadly Migstrikes. Usually you'll only have one chance to move out and finish the game when you're behind in eco/map control.

Yes Pathies are OP, however I wouldn't take the nerfs too far because USA is gonna struggle more vs Inf. So I still stand by my original point to first take away the advantages over Jarmen Kell and look at price/xp levels.

ExiLe-ZH:
One thing we discussed before is to increase minigunner xp, this gives usa opportunity to get pathies sooner to balance things out. 

Overall Inf is in the top 3 I believe? So I think it's justified, especially considering things like doortimes and truck xp makes inf actually stronger than it already is.

MTKing4:
I don't think over nerfing is a good idea either, but we don't have a basis as how much is an ok nerf which is why we need to test this

Also important to note that it isn't only OP vs inf only
For example other china's depend on manspam to break positions too sometimes consisting of troop crawlers and tank hunters spam, pathfinders eliminate that composition entirely, a composition that is fundamental for ChinaV for example

Float1ngFree:
Completely missed the main reason - armor types. That's where the answers are at. Default is 200% for SNIPER damage type, meaning it one-shots anyone.

A proper fix, first and foremost, is to reduce the default to base 100% for all human armor types. Then it will make it simple to decide on a damage relatively to the hp of different infantry. Or you account for x2 damage instead, which IMO is silly cuz sniper can attach only infantry. Sniper should one shot a non vetted infantry and 2 shots for Hero units.

275 range Nerf and stealth delay are good choices as well, but secondary after all.

xezon:
I suspect that armor change was no option because it would be a highly visible change compared to the original. In the original we assume 1 shot equals 1 kill. Taking 2 shots to kill would double the time to kill. But this can also be achieved in other ways, such as increasing reload time, and slowing promotion rate, that are less visible and give finer control.

ExiLe-ZH:
I agree that a pathfinder shouldn't 1 shot a hero. Is there a difference of HP with normal infantry and heroes? If so, I think the damage threshold should at a level where a hero survives one shot at low hp (and vetted inf), but where normal unvetted infantry dies.

xezon:
With "Hero" do you mean Vet 3 or Jarmen/Burton/Lotus?

Stubbjax has considered Hero aka Jarmen/Burton/Lotus in his opening post and it can be implemented by adding new armor type(s) to the hero units.

FiReArM-ZH:
Something like 1.06?

penfriendz:
Would it be possible to do something like this?
No vet: 1 shot
1 vet: 2 shot, damaged state after first shot
2 vet: 2 shot, not damaged state after first shot
3 vet: 3 shot?
Heroes could (for example) have +1 shot at all levels.

xezon:
User HenkOnderwater has tested original Pathfinder damage and kills.

> The Pathfinder used in the test was at vet 0, so no additional damage (from a vet 3 Pathfinder, for example) is included.

> The percentage indicates roughly how much HP the unit has left after taking a shot from a Pathfinder.

> An 'x' simply means the unit was killed by one shot.

pathfinder_kill_stats

ExiLe-ZH:
This would be my balance suggestion:

Screenshot 2024-01-29 at 12 43 49

I also like stubjax suggestion of reducing the firerate, but I would reduce this to 3 instead of 4 seconds.

xezon:
> This would be my balance suggestion

Keep in mind this has big implications. A 5% health left over after 1 shot on pristine enemy unit means that the effectiveness of the Pathfinder is halved, because it will require 2 shots to kill. On top of that, it will be a very visible difference to the original, where 1 shot was expected to kill. I recommend to not go this route in favour of more subtle tweaks.

Also: It would be a disproportional buff for Infantry General, because all its infantry starts at Vet 1.

penfriendz:
> > This would be my balance suggestion
> 
> Keep in mind this has big implications. A 5% health left over after 1 shot on pristine enemy unit means that the effectiveness of the Pathfinder is halved, because it will require 2 shots to kill. On top of that, it will be a very visible difference to the original, where 1 shot was expected to kill. I recommend to not go this route in favour of more subtle tweaks.

I disagree and don't think it goes far enough. His STK table is either a 1 or a 2, so in that sense while I think it's a step in the right direction 3 vet infantry and heroes (especially 3 vet heroes) can take even more shots.

While a damage change is a very visible change, other tweaks (depending) could substantially change the flavour of the game instead. Pick your poison, but something has to be done.

xezon:
> This would be my balance suggestion

The big design advantage of this change would be that the veterancy of the Pathfinder would get more meaning. Originally it does not matter "much" whether the Pathfinder is Vet 1, 2 or 3. It will have faster firing rate, more armor and self heal, but the additional damage is mostly meaningless. By applying Exile's progression, the additional damage from veterancy would suddenly matter.

ExiLe-ZH:
It's a pretty big change, but gives the game a little more depth, a Pathfinder now has to become vet 1 to kill a vet 1 (Inf) Tank Hunter with one shot like before. The basic idea is that the Pathfinder's experience has to match the enemy's experience to one shot that unit.

It should also take a very experienced Pathfinder (vet 2) to outperform Jarmen? Or just 2 of them. Currently Pathfinders are superior in almost every way it really hurts Jarmen's utilization and effectiveness once they're out, this is just silly design imo.

I feel the Pathfinder is still plenty strong with this change, the biggest change is versus heroes and Inf as a faction.

MTKing4:
> This would be my balance suggestion:
> 
> Screenshot 2024-01-29 at 12 43 49
> I also like stubjax suggestion of reducing the firerate, but I would reduce this to 3 instead of 4 seconds.

Big Change for sure, would have to test it to minmax the values, but my initial thoughts are positive, seems like a good way to balance out inf vs USAs

xezon:
If we go ahead with Pathfinder nerfs, then Infantry General will need a really good look at, because we have already applied many changes that favoured Infantry General and it is only getting stronger currently.

ExiLe-ZH:
> If we go ahead with Pathfinder nerfs, then Infantry General will need a really good look at, because we have already applied many changes that favoured Infantry General and it is only getting stronger currently.

Correct,

In terms of balance you see that it's a pretty balanced matchup at expert level, however at lower levels Inf is highly positive because it's less unforgiving to make mistakes as Inf.

Two things that might help USA:

1. Increase MiniGunner XP award https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1150
This helps Pathfinders rank quicker and USA as a faction.
2. Buffing Vanilla Raptors by either decreasing reload speed and/or decreasing the price. Raptors have the ability to one shot Outposts which can really put USA well ahead.

penfriendz:
2 is a dangerous change as it changes the early game before pathfinders are out. Obviously should be taken into consideration with other changes

Float1ngFree:
The thing about Inf Gen is that majority of the time it is packed in outposts/crawlers and when those get destroyed, the infantry escapes damaged. So even if Pathfinder doesn't one-shot vet2 Minigunner on full hp, he will one-shot damaged survivor.

Also, keep in mind the +20% Bombardment damage bonus - with a proper S&D nerf it could become viable option, especially in these "one or two shots to kill" cases.

Over all, it makes sense for regular Vet 2/3 unit to survive non-vetted sniper shot, the only inconvenience is that the **base HP spread of infantry is quite ridiculous**, which I've tweaked in the ~~patch~~ mod I'm working on, therefore it was easier to deal with.

Burton/Kell/Lotus should take 2 shots regardless, especially given that Pathfinders are Stealthed while not moving.

Float1ngFree:
But at least `StealthDelay = 1000` and `AttackRange = 275` are a must-do tweaks. Both stats were op and needed adjustments.

PR #1685, opened 2023-02-12
Helix experience requirements

Stubbjax:
This change aims to slightly reduce the Helix's ranking requirements so it is a bit easier to rank up.

### Changes include:

1. Reduced Helix experience requirements from [0 100 200 400] to [0 75 150 300]

## Rationale

It is extremely rare that a Helix happens to destroy anything, as its passengers often do the job. This significantly limits its ranking prospects (also exacerbated by TheSuperHackers/GeneralsGameCode#87) and makes even a Veteran Helix a rare sight. Taking out many targets or a big target with such a weak gun is certainly a rank-worthy achievement, and it seems reasonable to lower the bar by 25% to facilitate this notion.

## Further considerations

It should be noted that the Gattling Cannon attachment or Napalm bomb may provide experience to their creators as a result of future modifications.

## Summary

Traditional gameplay is preserved and the change is unlikely to be noticed by even the most skilled players. I suspect this change will have very little impact, and the majority of Helixes will remain unranked. But the increased prospects are still enticing and certainly better than in 1.04.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -15685,7 +15685,7 @@ Object Boss_VehicleHelix
     Object = Boss_Airfield
   End
   ExperienceValue     = 50 100 150 200 ;Experience point value at each level
-  ExperienceRequired  = 0 100 200 400  ;Experience points needed to gain each level
+  ExperienceRequired  = 0 75 150 300  ;Experience points needed to gain each level
   IsTrainable         = Yes
   CommandSet          = ChinaVehicleHelixCommandSet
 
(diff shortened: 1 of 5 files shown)


penfriendz:
You can make the gun less weak and that would help it rank up.

xezon:
Perhaps the better change here is to fix TheSuperHackers/GeneralsGameCode#87, TheSuperHackers/GeneralsGamePatch#1687, then see if promotion is still as bad? A successful Napalm Bomb attack on a War Factory would promote the Helix to 2 Star.

ExiLe-ZH:
> You can make the gun less weak and that would help it rank up.

Don't think it needs any buffs, mass lix (DPS) already pretty strong. 

Curious though how it compares to Humvee MG?

MTKing4:
> Perhaps the better change here is to fix TheSuperHackers/GeneralsGameCode#87, TheSuperHackers/GeneralsGamePatch#1687, then see if promotion is still as bad? A successful Napalm Bomb attack on a War Factory would promote the Helix to 2 Star.

Agree with this, there's no urgent need to reduce the xp requirements right now, fixing the two issues above may be enough, can close this imo

PR #1689, opened 2023-02-12
Change: Salvage Crates now sink for about 5 seconds into terrain before deletion

xezon:
This change makes Salvage Crates sink for about 5 seconds into terrain before deletion.

### TODO

- [ ] Fix decal ring update with Thyme

## Rationale

Original Salvage Crate gives no clue on about when it will disappear. Racing for a crate can become a matter of luck in certain situations. The randomized deletion of Salvage Crate at some time between 30000 ms and 35000 ms exacerbates this problem.

With Sink mechanic added to Salvage Create, the deletion can now be anticipated for at least 5 seconds before the crate is deleted.

## Video

https://user-images.githubusercontent.com/4720891/218304962-2bdb22f9-857c-438f-848d-5e243b4fcefb.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Crate.ini
@@ -206,9 +206,19 @@ Object SalvageCrate
     MaxMoney = 75
   End
 
-  Behavior = DeletionUpdate ModuleTag_03 ; Not LifetimeUpdate, since I can't die.  This will DestroyObject me.
-    MinLifetime = 30000
-    MaxLifetime = 35000
+  ; Patch104p @tweak xezon 12/02/2023 Change module from DeletionUpdate.
+  ; Patch104p @tweak xezon 12/02/2023 Decrease min lifetime from 30000 by 5150, max lifetime from 35000 by 5150 to accomodate slow death destruction delay.
+  Behavior = LifetimeUpdate ModuleTag_03
+    MinLifetime = 24850
+    MaxLifetime = 29850
+  End
+
+  ; Patch104p @tweak xezon 12/02/2023 Adds slow death to sink crate into ground. It can still be picked up while it sinks.
+  Behavior = SlowDeathBehavior ModuleTag_Death01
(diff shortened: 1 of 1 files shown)


Stubbjax:
This is a great idea. The crate becomes a bit hard to see towards the end of its life, though. Is there a way to get the green ring to persist until a few frames before despawn?

https://user-images.githubusercontent.com/11547761/218306937-fec6569d-3da9-40cb-8b82-d7291af33640.mp4

xezon:
Hmm first question is why do I not see this green ring in my local build?

Stubbjax:
> Hmm first question is why do I not see this green ring in my local build?

The ring will not show for preexisting crates placed in the World Builder.

xezon:
Hmm yeah, we first need to find where this ring is created.

ExiLe-ZH:
good idea this is

xezon:
I suspect salvage decal is hardcoded with `SalvageCrateCollide` module. Changing Geometry radii of `SalvageCrate` will modify decal. This can only be adjusted in Thyme.

shot_20230212_123935_1

xezon:
Alternatively perhaps we can add a particle around it somehow, and leave the decal as is. Though I do not think the visibility is that bad as is.

MTKing4:
I do like the idea, but the sinking isn't looking the best at the last few seconds, maybe slow down the sinking speed and make the crate disappear a bit after the green ring is gone? Meaning we don't have to see all the sinking animation to get the idea we're running out of time, and it will also be less visible of a change

xezon:
We could make it sink just half way, but then it will be less intuitive as to when exactly it disappears. With it sinking fully into terrain it is quite clear when.

penfriendz:
If it's possible I'd like green ring visible=here, invisible=gone

xezon:
Then we either need to find different method to sink it or wait for Thyme to address decal ring.

MTKing4:
It just looks weird picking up sunken crates, they would need a shovel for that, and a shovel we do not have.

Yes i agree about the green ring idea

PR #1691, opened 2023-02-12
Change: Decrease spin up time of China Gattling Tank Air Gun by 6 frames

xezon:
* Successor to #1199
* Closes #1567

This change decreases barrel spin up time of China Gattling Tank Air Gun by 6 frames. An alternative to #1567.

| Weapon                           | ContinuousFireOne | ContinuousFireTwo | ContinuousFireCoast |
|----------------------------------|-------------------|-------------------|---------------------|
| Original CCG GattlingTankGun     | 3                 | 6                 | 2000                |
| Original CCG GattlingTankGunAir  | 3                 | 6                 | 2000                |
| Original ZH GattlingTankGun      | 2                 | 6                 | 1000                |
| Original ZH GattlingTankGunAir   | 3                 | 6                 | 2000                |
| Patched GattlingTankGun (1)      | 2                 | 6                 | 2000                |
| **Patched GattlingTankGunAir (this)** | 2            | 6                 | 2000                |

### Gattling Tank Air Gun firing sequence in frames

Exclamation mark (!) represents a shot fired in timeline.

```
Original 12          12          6     6     3  3  3  3  3  3
         !...........!...........!.....!.....!..!..!..!..!..!

Patched  12          6     6     6     3  3  3  3  3  3
         !...........!.....!.....!.....!..!..!..!..!..!
```

## Rationale

Original `GattlingTankGun` and `GattlingTankGunAir` have inconsistent spin up times. With this change they are consistent.

This gives Gattling Tanks a minuscule better chance at damaging and killing air units, such as USA Raptor, Aurora and Comanche. This in turn makes China intangibly better against USA Airforce General and air force centrist strategies. Has no impact against GLA.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -1149,6 +1149,7 @@ End
 
 ;------------------------------------------------------------------------------
 ; Patch104p @tweak xezon 02/02/2023 Change DamageType from SMALL_ARMS.
+; Patch104p @tweak xezon 12/02/2023 Change ContinuousFireOne from 3.
 ;------------------------------------------------------------------------------
 Weapon GattlingTankGunAir
   PrimaryDamage         = 12.0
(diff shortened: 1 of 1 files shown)


ExiLe-ZH:
Hmmm I find there to be no need for this change, Gats are already good vs air units. Will also make Migs and Lixes less effective.

xezon:
Need to choose. This or #1567 :-)

ExiLe-ZH:
Because of consistency? 

Range is already different aswell, they don't need to be exactly the same.

Although if it's combined with the bigger buff of ContinuousFireCoast to 2000 that outweights that change, it could be considered.

MTKing4:
I don't see the reasoning behind either changes, the consistency argument is quite weak, we shouldn't mess with balance for the sake of consistency

Also the point that's been raied about reduced lixes/migs durability is vaild, we don't want that

xezon:
No consensus.

Issue #1147, opened 2022-09-06
Combat Cycle Experience Values & Requirements are insanely high


ExiLe-ZH:
```
Object GLAVehicleCombatBike
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level

Object GLAVehicleCombatBikeRocket
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level

Object GLAVehicleCombatBikeTerrorist
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level

Object Chem_GLAVehicleCombatBike
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level

Object Demo_GLAVehicleCombatBike
  ExperienceValue       = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired    = 0 200 400 800  ;Experience points needed to gain each level
```

Now compare that to Techs:

```
Object GLAVehicleTechnicalChassisOne
  ExperienceValue         = 25 25 50 100    ;Experience point value at each level
  ExperienceRequired      = 0 50 75 150  ;Experience points needed to gain each level
```

Or Quads/Gats/Vees (all the same):

```
  ExperienceValue = 50 50 100 150    ;Experience point value at each level
  ExperienceRequired = 0 100 150 300  ;Experience points needed to gain each level
```

It's obvious that Rebel/RPG bikes don't have the same killing potential or survivability as the units named above.

DemoBikes should probably stay the same though as they're trading units.

commy2:
I would keep all bike XP consistent including Demogen Bikes. The riders can be changed and these values would stick per bike, not per rider.

xezon:
Looks to me that Combat Bike XP should be closer to Infantry than Vehicles, because realistically we expect the damage to be what the Infantry does.

ExiLe-ZH:
@commy2 
> I would keep all bike XP consistent including Demogen Bikes. The riders can be changed and these values would stick per bike, not per rider.

RPG and regular Terrorist Bikes can't be produced from the factory, why can it be tweaked separately in INI?

[...]

Issue #954, opened 2022-08-22
Units take different amounts of explosion damage from different angles


xezon:
Units take different amounts of explosion damage from different angles.

Stubbjax:
> Seems like unit position and pitch also plays a part. Happens with quads too.

https://user-images.githubusercontent.com/4720891/185853386-10362a17-5862-464e-8dcc-5f5b0cba0877.mp4

xezon:
I tested this in current Patch main branch and it is not reproducible. Also tested vs Scorpion, Quad, Buggy, Technical, Tractor and the issue did not happen. On both straight and elevated terrain.

https://user-images.githubusercontent.com/4720891/218325907-41b15c1e-fa7b-4aa5-8df8-48af17d0316b.mp4

Stubbjax:
Here's a replay on 1.04: [Bug.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/10718615/Bug.zip)

And some more footage:

https://user-images.githubusercontent.com/11547761/218369696-e699d85c-2e88-4e31-a739-ebee97f7b585.mp4

https://user-images.githubusercontent.com/11547761/218369704-8b1513cc-61ec-4475-9ff6-a8a667d55a1b.mp4

xezon:
Are you able to reproduce it in current Patch main branch?

Stubbjax:
> Are you able to reproduce it in current Patch main branch?

Sure. Here is a replay of it occurring as of commit [3474c97f4842fb256ea6de5651c845e2fba6f85d](https://github.com/TheSuperHackers/GeneralsGamePatch/commit/3474c97f4842fb256ea6de5651c845e2fba6f85d): [Bug.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/10719913/Bug.zip)

xezon:
In what game mode do you test? Network? I tested in Skirmish on Defcon 6 with GLA and it was not reproducible. Wondering what I did wrong.

Stubbjax:
I do all testing via MULTIPLAYER → NETWORK (LAN). I don't believe Skirmish generates replays?

xezon:
It does. Perhaps this is Network specific issue.

Stubbjax:
Can you repro it in LAN?

xezon:
[...]

PR #1694, opened 2023-02-12
Change: Increase fire rate of GLA RPG Trooper on Combat Bike by 88%

Stubbjax:
This change halves the delay between shots for the Combat Cycle's RPG weapon from 1000ms to 500ms. This is to provide some tangible benefit to going to the trouble of replacing a Combat Cycle's rider with a Tunnel Defender, which is a pointless endeavour in pretty much every scenario in 1.04.

There will likely never be a worthwhile use for this unit due to its vulnerability to its preferred targets, vehicles and tanks, in combination with the effort involved in its acquisition and the relatively poor output for the price / time investment. A Technical + Tunnel Defender is cheaper and far more flexible. The goal of this change isn't to make the unit viable, but to make it slightly less of a joke. The fire rate reduction was chosen for its simplicity. See #1147 for other proposals.

Below is a comparison between one Tunnel Defender, two Tunnel Defenders, and the RPG Combat Cycle after the change.

https://user-images.githubusercontent.com/11547761/218325613-b0be6255-25e1-4df9-aa9d-906e31ebc817.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1694_biker_rocket_weapon.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1694_biker_rocket_weapon.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-02-12
+
+title: Increases fire rate of GLA RPG Trooper on Combat Bike by 88%
+
+changes:
+  - tweak: The GLA RPG Trooper on the Combat Bike now shoots a rocket every 533 ms instead of 1000 ms.
+
+labels:
+  - buff
+  - controversial
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1694
+
+authors:
+  - Stubbjax
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2151,7 +2151,7 @@ Weapon TunnelDefenderBikerRocketWeapon
   VeterancyProjectileExhaust  = HEROIC HeroicMissileExhaust
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
   ScatterRadius               = 0      ; This weapon will scatter somewhere within a circle of this radius, instead of hitting someone directly
-  DelayBetweenShots           = 1000  ; time between shots, msec
+  DelayBetweenShots           = 500  ; time between shots, msec
   ClipSize                    = 0             ; how many shots in a Clip (0 == infinite)
   ClipReloadTime              = 0    ; how long to reload a Clip, msec
   AutoReloadsClip             = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -7802,7 +7802,7 @@ Weapon Chem_TunnelDefenderBikerRocketWeapon
   VeterancyProjectileExhaust  = HEROIC HeroicMissileExhaust
   RadiusDamageAffects         = ALLIES ENEMIES NEUTRALS
   ScatterRadius               = 0      ; This weapon will scatter somewhere within a circle of this radius, instead of hitting someone directly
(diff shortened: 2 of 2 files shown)


xezon:
What was the reason to halve the fire rate as opposed to making fire rate and damage of both weapons identical?

Stubbjax:
> What was the reason to halve the fire rate as opposed to making fire rate and damage of both weapons identical?

I'm not sure I follow. Both weapons are effectively identical in 1.04.
```
Weapon TunnelDefenderRocketWeapon
  PrimaryDamage               = 40.0
  DelayBetweenShots           = 1000
End
```
```
Weapon TunnelDefenderBikerRocketWeapon
  PrimaryDamage               = 40.0
  DelayBetweenShots           = 1000
End
```

xezon:
Ok that is odd. The Bike Rocket shoots much more missiles than the Foot Trooper, and yet they take about the same time to kill the Bus. Why is that?

Stubbjax:
> Ok that is odd. The Bike Rocket shoots much more missiles than the Foot Trooper, and yet they take about the same time to kill the Bus. Why is that?

Perhaps I should redo the video to avoid confusion. I noted the armour bug of #954 occurred. It seems to actually be fairly prevalent.

xezon:
Oh right. Perhaps should test against a structure instead.

Stubbjax:
https://user-images.githubusercontent.com/11547761/218325613-b0be6255-25e1-4df9-aa9d-906e31ebc817.mp4

xezon:
Ok that is clear. The reason the Bike Rocket takes longer to kill the structure is because the fire rate is raised to 533.33 ms with the hardcoded weapon tick of 66.66 ms.

I am ok with this change, though I do wonder, are there any other feasible potential improvement for Rocket Bike other than this one?

ExiLe-ZH:
Nice find, love this is gets some attention.

commy2:
Missing Chem_TunnelDefenderBikerRocketWeapon, Chem_TunnelDefenderBikerRocketWeaponGamma and DEMO_TunnelDefenderBikerRocketWeapon.

---

1.04 Bike Rebel also has increased range over Rebel. Maybe the same should apply to Bike?

Stubbjax:
> I am ok with this change, though I do wonder, are there any other feasible potential improvement for Rocket Bike other than this one?

> 1.04 Bike Rebel also has increased range over Rebel. Maybe the same should apply to Bike?

I thought about this too, but it felt slightly off due to the rockets having the same 1250ms fuel lifetime, which would increase the likelihood of them missing fleeing targets when fired from a longer distance. It may be fine though. We could try adding +25 range for a total of 200 if everyone is in agreement.

This would be the difference:

image

xezon:
As for fuel times, we could create a copy of Tunnel Defender Weapon and Tunnel Defender Missile and adjust this.

Would Rocket Bike with attack range 200 then have same attack range as Rebel Bike?

Stubbjax:
> As for fuel times, we could create a copy of Tunnel Defender Weapon and Tunnel Defender Missile and adjust this.
> 
> Would Rocket Bike with attack range 200 then have same attack range as Rebel Bike?

Here are the attack ranges in 1.04:

| Infantry | Range | Bike Range |
| - | - | - |
| Rebel | 100 | 150 |
| Tunnel Defender | 175 | 175 |
| Jarmen Kell | 225 | 225 |

xezon:
The original Attack Ranges make sense. 150 being the common vehicle Attack Range.

MTKing4:
I like the change, the range increase is also fine, if it doesn't cause any accuracy issues

xezon:
Rebel: 150
RPG Trooper: 300
Bike with Rebel: 500

Pure Bike cost is 350. With RPG Trooper is 650. Real cost is 800.

From cost perspective, fire rate benefit is justified. RPG Trooper on Bike is more mobile, which could be good against fast and light targets such as Humvee, Sentry Drone, Technical, Comanche.

Rocket Men on Bike could be a suitable counter against Super Technical, because Bikes are faster.

If this change can create more strategic depth, then that would be good. It does not hurt to try this.

xezon (approved):

xezon:

xezon:
RPG on Bike with its cliff jump ability can also help with Dozer hunt.

PR #1699, opened 2023-02-12
Change: GLA Scorpion Rocket will always reload when idle

### LINKED ISSUE #1698: GLA Scorpion with double rocket does not reload after single shot
GLA Scorpion with double rocket does not reload after single shot.

https://user-images.githubusercontent.com/4720891/218334285-5448f966-af4d-4bf2-b953-b91c1742e260.mp4

xezon:
* Resolves #1698

With this change the GLA Scorpion Rocket will always reload when idle.

This makes the unit a bit better in scenarios where it managed to only fire with one of its rockets on its target and some time passes until the next target is attacked.

This situation is rare. There is a `ClipReloadTime` of 15000 ms vs a `DelayBetweenShots` of 200 ms. This means there is a measly 1.3% chance that Scorpion Tank gets into state with one rocket in racks when attacking and moving Scorpion Tank around.

# Rationale

Eliminates situations where Scorpion Tank would only fire one rocket on a fresh engagement. This likely is what player would expect always.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -197,6 +197,7 @@ Weapon ScorpionTankGunPlusOne
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @tweak xezon 11/02/2023 Enables automatic reload when idle.
 Weapon ScorpionMissileWeaponPlusTwo
   PrimaryDamage = 100.0
   PrimaryDamageRadius = 5.0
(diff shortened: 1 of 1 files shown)


commy2 (approved):

Stubbjax:
> This situation is rare. There is a `ClipReloadTime` of 15000 ms vs a `DelayBetweenShots` of 200 ms. This means there is a measly 1.3% chance that Scorpion Tank gets into state with one rocket in racks when attacking and moving Scorpion Tank around.

I don't think 1.3% is quite accurate. When the Scorpion is moving, it is far less likely to fire both of its rockets, which is quite a common occurrence when there's a lot of kiting and movement going on.

https://user-images.githubusercontent.com/11547761/218371478-8c5f6b76-07dd-40f9-a82d-dc871617e197.mp4

Stubbjax (approved):

PR #1704, opened 2023-02-13
Change: China Tank Battlemaster with Autoloader will always reload when idle

### LINKED ISSUE #1702: China Tank Battlemaster with Autoloader does not auto reload before entire clip is emptied
China Tank Battlemaster with Autoloader does not auto reload before entire clip is emptied. Clip size is 3.

https://user-images.githubusercontent.com/4720891/218558233-5fcee6c8-cefe-4a95-8049-7e2a80d6cd25.mp4

xezon:
* Resolves TheSuperHackers/GeneralsGamePatch#1702

With this change the China Tank Battlemaster with Autoloader will always reload when idle.

This makes the unit significantly better in scenarios where it managed to only fire 1 or 2 times and some time passes until the next target is attacked.

This situation is common. There is a `ClipReloadTime` of 2000 ms vs a `DelayBetweenShots` of 533 (500) ms with 3 clips. This means there is a 44% chance that Tank Battlemaster with Autoloader gets into state with just 1 or 2 loaded clips in regular battle engagements.

# Rationale

Eliminates situations where Tank Battlemaster with Autoloader would only fire 1 or 2 shells on a fresh engagement. This likely is what player would expect always.

## Original

https://user-images.githubusercontent.com/4720891/218558233-5fcee6c8-cefe-4a95-8049-7e2a80d6cd25.mp4

## Patched

https://user-images.githubusercontent.com/4720891/218562135-13c0d00c-cead-47f3-8faf-c7164ba996a1.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1704_battlemaster_autoloader_idle_reload.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1704_battlemaster_autoloader_idle_reload.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-02-11
+
+title: Adds idle auto reload after 2100 ms to China Tank Battlemaster with Autoloader upgrade
+
+changes:
+  - tweak: The China Tank Battlemaster with Autoloader upgrade will now always reload when idle. This makes the unit much better in scenarios where it managed to only fire 1 or 2 of its shells and at least 2 seconds pass until the next attack.
+
+labels:
+  - buff
+  - china
+  - controversial
+  - design
+  - major
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1704
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -6641,6 +6641,7 @@ End
 
 
 ;------------------------------------------------------------------------------
+; Patch104p @tweak xezon 11/02/2023 Enables automatic reload when idle. (#1704)
 Weapon Tank_BattleMasterTankGun
   PrimaryDamage           = 40.0
   PrimaryDamageRadius     = 5.0
@@ -6661,6 +6662,8 @@ Weapon Tank_BattleMasterTankGun
   DelayBetweenShots       = 500               ; time between shots, msec
   ClipSize                = 3                    ; how many shots in a Clip (0 == infinite)
   ClipReloadTime          = 2000              ; how long to reload a Clip, msec
+  AutoReloadsClip         = Yes
+  AutoReloadWhenIdle      = 2100
   WeaponBonus             = PLAYER_UPGRADE DAMAGE 125% ; UraniumShells
 
   ; note, these only apply to units that aren't the explicit target


xezon:
I am skeptical of this change. Makes this unit much stronger.

commy2:
I don't think it makes much of a difference, but would be neat to see a reloaded and a random group with same veterancy duke it out.

penfriendz:
I'd fix it then tweak if it's op. Consistent expected behaviour is good.

Stubbjax:
I wouldn't say this change makes it _much_ stronger. The Autoloader upgrade is rather misleading in that it only increases damage per second by ~30%, when it should be expected to be closer to 200% due to the additional two shells. Nationalism can be considered marginally better in 1.04 (due to TheSuperHackers/GeneralsGamePatch#810 / TheSuperHackers/GeneralsGameCode#173) at +25% DPS for $500 less.

`BattleMasterTankGun` deals 1×60 damage per 60 frames and `Tank_BattleMasterTankGun` deals 3×40 (120) damage per 92 (16 + 16 + 60) frames. Note that the 500ms delay between each shot in the clip translates to 533.33ms and thus 16 frames.

| Upgrade | Damage | Fire Rate | DPS |
| - | - | - | - |
| None | 60 | 2,000ms | 30 |
| Nationalism | 60 | 1,600ms | 37.5 |
| Autoloader | 120 | 3,067ms | 39.13 |
| Both | 120 | 2,400ms | 50 |

It's also important to note that Tank Battlemasters start at Veteran rank, and often Elite rank. This reduces the window for a partial clip to less than 24 frames (2×400ms) and 20 frames (2×333ms), respectively. The remaining shell(s) may also still be fired before the clip's reload duration. Movement may also significantly extend these windows.

So if an Autoloader Battlemaster only fires one shot, it is technically worse than without the upgrade. If it fire two shots, it is only marginally better. I don't see anything overpowered about this.

Stubbjax (approved):

MTKing4:
Looks good

xezon:
Would someone be able to make test map with classic Tank Gun vs patched Tank Gun in big battle to see difference, if there is any?

kABUSE1:
Map to test battlemaster strength attached.

Let me know what you think or if adaptions are required.

Play this map vs. tank general and just let it run.

Auto-loader will be purchased automatically.

After 60 sec quads keep spawning in. Amount of kills is being tracked and can be seen in bottom right corner.

[battlemaster-strength-tester.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/11222718/battlemaster-strength-tester.zip)

xezon:
I ran test 10 times on test map vs Tank General, Medium Army.

## Before this change

Traitors Killed: 63
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59
Traitors Killed: 59

## After this change

Traitors Killed: 46
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49
Traitors Killed: 49

## Conclusion

???

Judging by these numbers it implies that the Battlemaster now performs worse than before. How is this possible? It is supposed to perform better.

xezon:
I ran test anew vs Tank General, Medium Army. But this time I did not Restart map with button, and instead went back to lobby and started new game. I think only this will reset the random seed to new value. Result numbers ordered.

## Before this change:

35, 59, 60, 76, 78, 97, 103, 117, 118, 253

Average: 99.6

## After this change:

49, 68, 79, 96, 113, 121, 126, 157, 202, 251

Average: 126.2

## Conclusion

There is too much variation in this test. 35 to 253 is just too much discrepancy. We would need to collect thousands of samples to get a meaningful average. The lower the sample count, the more randomness it will contain.

xezon:
It looks like we cannot evaluate impact of this change in a meaningful way. So we would have to yolo it and see how it turns out.

xezon:
Documentation added.

If China Tank becomes too strong, we have many thing to dial back.

PR #1710, opened 2023-02-14
Change: Enable kill on resting and set consistent lifetime for hulks

xezon:
**Merge with Rebase**

This change enables `KillWhenRestingOnGround` on all vehicle hulks where it is missing. Additionally, all ground vehicles and helicopters now have consistent life times. User facing this has no tangible effect, however it does affect when exactly hulk object is pushed into dead state.

What `KillWhenRestingOnGround` does in code:

```c++
UpdateSleepTime PhysicsBehavior::Update()
{
    ...
    if (data->m_killWhenRestingOnGround && !above && Is_Low_Velocity(m_vel)) {
        if ((!obj->Is_KindOf(KINDOF_DRONE) || obj->Is_Effectively_Dead()
                || obj->Get_Disabled_State(DISABLED_TYPE_DISABLED_UNMANNED))) {
            obj->Kill(DAMAGE_UNRESISTABLE, DEATH_NORMAL);
        }
    }
    ...
}
```

It kills an object when resting as the name implies.

With a life time such as
```
  Behavior = LifetimeUpdate ModuleTag_04
    MinLifetime = 1500   ; min lifetime in msec
    MaxLifetime = 1600   ; max lifetime in msec
  End
```
it means that vehicle is set dead either after 1500 to 1600 ms or when it is resting on ground.

This setup is now consistent on all hulks. Helicopter hulks have longer life times because they slowly fall to ground and their death is supposed to begin after the crash landing.

Vehicles: 1500 to 1600
Helicopters: 20000 to 20000

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/CivilianUnit.ini
@@ -2275,10 +2275,11 @@ Object CarLimo03DeadHull
     InitialHealth   = 1.0
   End
 
-
+  ; Patch104p @tweak xezon 29/01/2023 Enable kill when resting.
   Behavior = PhysicsBehavior ModuleTag_03
     Mass = 50
     AllowBouncing = Yes
+    KillWhenRestingOnGround = Yes
   End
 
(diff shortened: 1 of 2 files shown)


xezon:
Looks good.

PR #1711, opened 2023-02-14
Fix: Increase destruction delays for hulks of USA Dozer, Avenger, China Dozer, Supply Truck, Inferno Cannon, Overlord, Nuke Cannon, Helix, GLA Combat Bike, Scud Launcher to avoid deletion before object is sunk into terrain

xezon:
This change increase destruction delays of various hulks to avoid deletion before object is sunk into terrain.

Affects

* USA Dozer
* USA Avenger
* China Dozer
* China Supply Truck
* China Inferno Cannon
* China Overlord
* China Nuke Cannon
* China Helix
* GLA Combat Bike
* GLA Scud Launcher

Note: The destruction delay of China Nuke Cannon was optimized for its Hulk prior change #1618. With the D1 model tweaked, it no longer takes as much height.

Note 2: The destruction delays are not optimized for sideway wrecks on slope. For that the long models would require much longer destruction delays. Ideally in future we will have hulks aligned to terrain properly, which will avoid such destruction before complete sink.

## Original Deletion Delay Sample

https://user-images.githubusercontent.com/4720891/218875893-3fa368ad-dd83-43dc-b8bd-4023cbfd5f39.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GLAMiscUnit.ini
@@ -148,10 +148,12 @@ Object GLACombatBikeToppledHulk
     MaxLifetime = 0   ; max lifetime in msec
   End
 
+  ; Patch104p @tweak xezon 29/01/2023 Remove obsolete Sink Delay of 1.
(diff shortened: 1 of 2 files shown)


xezon:
Increased China Helix destruction delay by another second. Looks good.

PR #1713, opened 2023-02-15
Fix: Super Weapon General now uses the correct Command Center model during construction and deconstruction

Stubbjax:
Super Weapon General now uses the correct Command Center model during construction / sold states. This is only noticeable when construction nears 100% or for the first ~44 frames when sold (the faction logo is now visible during these times).

Before:
image

After:
image


xezon:
Patch104p comments missings.

xezon (approved):
Looks good.

PR #1714, opened 2023-02-15
Fix: All Scud Launcher variants now correctly show their salvage upgrades during their rubble state

Stubbjax:
All Scud Launcher variants now correctly show their salvage upgrades during their rubble state. In 1.04, only the vGLA Scud Launcher shows its salvage upgrades during its rubble state.

Below is some 1.04 footage from before this change. Note that the salvage upgrade disappears from the top Scud Launcher (`Demo_GLAVehicleScudLauncher`) upon destruction:

https://user-images.githubusercontent.com/11547761/219004446-16e1234c-556f-4167-90d5-85d1a046e3e2.mp4


xezon:
Nice find. Patch104p comments are missing.

xezon (approved):
Comments added. Looks good.

PR #1715, opened 2023-02-15
Fix: Condition state formatting

Stubbjax:
Fixed various condition state formatting. Equals signs were missing in various places in the Angry Mob and Combat Cycle draw modules.


xezon (approved):
Looks good.

PR #1717, opened 2023-02-15
Change: Remove superfluous upgrade modules

Stubbjax:
Removed superfluous upgrade modules from various units:

- Black Napalm upgrade from the Nuke Helix
- AP Bullets upgrade from the Toxin Rebel
- Rocket Buggy Ammo upgrade from the Combat Cycle

This is just a minor cleanup and has no effect in-game.


xezon:
Did you remove the ones that were originally ignored? I am asking because it should remove only the real obsolete ones, so that Mod Maps referencing the correct modules would still reference the correct modules.

xezon (approved):
Comments added. Looks ok.

PR #1719, opened 2023-02-15
Fix duplicate module name in BoatAsianFishing

xezon:
* Introduced by #1639


commy2:
You can check by starting the worldbuilder btw. It throws a fit when modules have name conflicts.

xezon:
Yeah that is how I noticed this :D

PR #1723, opened 2023-02-16
Change: China Nuke Cannon can no longer be crushed by Overlord

xezon:
With this change the China Nuke Cannon can no longer be crushed by Overlord.

## Original

https://user-images.githubusercontent.com/4720891/219507841-4719b7a0-b410-45c3-9384-cfe866b87786.mp4

# Rationale

The Nuke Cannon is as large as the Overlord. In deployed state it towers higher than all other vehicles. Yet the Overlord can crush it as if it was a small vehicle.

Naturally one could expect that the Nuke Cannon is not crushable. This change delivers. It makes the Nuke Cannon a bit more unique and distinct from the Inferno Cannon and other tanks. It also rewards the high cost and inaccessibility of this unit.

For most battle situations this has no tangible consequences, because the Nuke Cannon will likely be shot dead before the enemy Overlord gets the chance to crush it.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -1748,8 +1748,10 @@ Object ChinaVehicleNukeLauncher ; Alias ChinaVehicleNukeCannon
   ExperienceValue = 50 100 200 400    ;Experience point value at each level
   ExperienceRequired = 0 400 600 1000  ;Experience points needed to gain each level
   IsTrainable = Yes             ;Can gain experience
+
+  ; Patch104p @tweak xezon 16/02/2023 Nuke Cannon is no longer crushable by Overlord.
   CrusherLevel           = 2  ;What can I crush?: 1 = infantry, 2 = trees, 3 = vehicles
-  CrushableLevel         = 2  ;What am I?:        0 = for infantry, 1 = for trees, 2 = general vehicles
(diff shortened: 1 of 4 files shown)


commy2 (approved):
Agreed
lol at wreck

penfriendz:
The crush wrecks are so ugly

MTKing4:
Not a good change, Scud Luncher can be crushed in comparison, all long range weapons do, there's no reason to make a specific rule for nuke cannon, plus it doesn't matter if the unit is higher, overlord has large treads that can climb any tank so it's fine

Stubbjax:
This is an interesting idea and makes intuitive sense, though it's a rather trivial buff and a better case for faction diversity.

My only minor concern with this change is the breaking of the 'Overlords can crush everything' rule that players are already familiar with. Players who expect Overlords to crush everything may experience some form of confusion / breakdown when they attempt to crush a Nuke Cannon but the Overlord just drives around it.

commy2:
The nuke cannon death explosion should still happen when driven over, no?

xezon:
I am confused now. Do we want this change or not? I don't feel strongly about this change because this is a 1 in a million match situation, I was just surprised to see that Nuke Cannon can be crushed when I expected that it is too big for that.

MTKing4:
I wouldn't change it personally

commy2:
Keep game simple. Overlord = can crush everything

The only issues here is the wreck model and probably how crushing deletes the explosion.

xezon:
> Keep game simple. Overlord = can crush everything

This is not entirely true. It cannot crush some of the civilian cars. But it can crush all regular faction vehicles.

Issue #1721, opened 2023-02-16
China Listening Outpost model issues


xezon:
* NVLOutpost_D1.W3D uses _D texture

PR #1727, opened 2023-02-18
Change: China Listening Outpost now spawns whole body hulk on death

xezon:
With this change the China Listening Outpost now spawns whole body hulk on death.

## Original

https://user-images.githubusercontent.com/4720891/219853149-76088931-c4b3-468d-8202-c6d85b689f5d.mp4

## Patched

https://user-images.githubusercontent.com/4720891/219853162-2c0b89ec-9738-4522-93f4-1c30f67f3273.mp4


xezon:
I am not sure about this. The original design has the advantage that the popped passengers are better visible after the final explosion.

MTKing4:
I do think the original is better, second explosion is for the back part being destroyed, the patched only showing tries flying with second explosion

xezon:
If we close this, then we can also close #1721

PR #1728, opened 2023-02-18
Fix USA Ambulance debris

xezon:
**Merge with Rebase**

This change fixes USA Ambulance debris setup.

* Tires and doors now spawn in correct locations
* Third tire now spawns because the hulk object is missing 3 tires

### TODO

- [x] Use Humvee doors for debris
- [x] Remove doors from _D1 texture

## Patched

https://user-images.githubusercontent.com/4720891/219858087-72160e6a-c700-4ea0-adb4-222c17df0f47.mp4


xezon:
The doors still not look right. I think we can spawn the Humvee door debris instead.

xezon:
I have completed second pass. Swapped odd Ambulance door debris with Hummer door debris. Removed passenger doors in Ambulance rubble textures. Optimized debris positions. Looks good now, apart from the physics.

PR #1729, opened 2023-02-18
Fix: Set better spawn positions of USA Humvee debris

xezon:
This change sets better spawn positions of USA Humvee debris.

* The tires now spawn at the right height
* The doors now spawn at better heights and no longer inside the chassis


xezon:
Superseeded by #1735

PR #1730, opened 2023-02-18
Fix: Add missing rubble model state to USA Humvee

xezon:
This change adds missing rubble model state to USA Humvee. Originally the undamaged Humvee would show for one frame after death.


xezon:
Superseeded by #1735

PR #1731, opened 2023-02-18
Fix USA Sentry Drone death model setup

xezon:
**Merge with Rebase**

This change fixes USA Sentry Drone death model setup.

* Adds proper rubble states
* Sets up better hulk object fit for the debris and upgrade variants
* Sets correct debris positions and counts

## Original

https://user-images.githubusercontent.com/4720891/219867081-53901572-7c47-42d8-9337-4f2da61c644b.mp4

## Patched

https://user-images.githubusercontent.com/4720891/219867094-18421b7f-6d16-4889-856d-39e58825350f.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1731_sentry_drone_death.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1731_sentry_drone_death.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-02-18
+
+title: Fixes death setup of USA Sentry Drone
+
+changes:
+  - fix: The USA Sentry Drone no longer shows its wreck model before the final death explosion.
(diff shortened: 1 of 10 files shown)


xezon:
I optimized debris positions a bit further after I found how to extract coordinates from Blender.

There is bug related to Sentry debris, but it does not block submission of this fix.

* #1733

xezon:
The light debris flies up, but the plate debris does not. It is barely visible after the smoke settled. I think both pieces need to fly up.

xezon:
I have improved the debris physics. And streamlined the hulk sink delay to not start before the explosion effect vanish. Documentation is left to be added.

xezon:
Documentation added.

xezon:
Good.

PR #1732, opened 2023-02-18
Fix USA Avenger death states and debris

xezon:
**Merge with Rebase**

This change fixes USA Avenger death states and debris.

* Sets proper model for rubble state
* Adds proper debris

The turret will reset on rubble state. I currently see no way to fix it. It appears the `OverlordContain` module simply stops update in Rubble, therefore will disappear, even if laser turret also has a rubble state added.

Debris physics will need some tweaking. That is something for follow up change(s).

### TODO

- [x] Setup rubble state as if code bug did not exist
- [x] Remove destruction delay to hide rubble state
- [ ] Close the hole at the bottom of the turret debris TheSuperHackers/GeneralsGamePatch#1734

## Original

https://user-images.githubusercontent.com/4720891/219878728-4c9bc2b9-3927-45bf-b7eb-332af0f20bb7.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -7138,6 +7138,18 @@ Object AirF_AmericaTankAvenger
     Mass = 50.0
   End
 
+  ; Patch104p @tweak xezon 21/02/2023 Replaces original slow death behaviours with sudden death to
+  ;   a) be consistent with Humvee death
+  ;   b) mask the executable issue that allows for no rubble state of the slaved laser turret
+  Behavior = SlowDeathBehavior ModuleTag_05
+    DeathTypes = ALL -CRUSHED -SPLATTED
+    ProbabilityModifier = 100
+    DestructionDelay = 0
+    DestructionDelayVariance = 0
+    FX  = FINAL    FX_AmericaVehicleAvengerDeathExplosion ; Patch104p @tweak from FX_GenericTankDeathExplosion
+    OCL = FINAL    OCL_AvengerTankDeathEffect
+  End
+
   ; Turret fly off death
   Behavior = SlowDeathBehavior ModuleTag_05
     DeathTypes = ALL -CRUSHED -SPLATTED
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -7031,8 +7031,10 @@ Object AirF_AmericaTankAvenger
       HideSubObject       = TURRET01
     End
 
+    ; Patch104p @bugfix xezon 18/02/2023 Shows currect model before final death.
     ConditionState        = RUBBLE
-      Model               = AVAVNGER_D1
+      Model               = AVAVNGER_D
+      ShowSubObject       = TURRET01  ;Show static turret
     End
 
     ;When a bombtruck disguises as an avenger, show the turret!
@@ -7042,7 +7044,7 @@ Object AirF_AmericaTankAvenger
     End
     ConditionState        = REALLYDAMAGED DISGUISED
       Model               = AVAVNGER_D
-      ShowSubObject       = TURRET01  ;Hide controlled turret
+      ShowSubObject       = TURRET01  ;Show static turret
     End
 
     TrackMarks = EXTireTrack.tga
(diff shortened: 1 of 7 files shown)


Stubbjax:
The patched version looks jankier than the original due to the immediate resetting of the turret and the anti-air laser briefly originating from a different location and for a while after death. It is not so clear that there is a problem with the original, whereas the patched version has a fairly conspicuous defect. However, the before and after footage is not demonstrated under the same conditions, which makes things harder to judge.

The added debris is a welcome change and definite improvement.

xezon:
The original has same issue. Rubble simply has no turret movement. Executable bug. I will remove the videos as they appear to add confusion, eventhough the change text did attempt to explain it.

xezon:
Wonder if Overlord has same issue with Gattling Turret.

Stubbjax:
> I will ~~remove the~~ **record better** videos as they appear to add confusion

Fixed it for you.

There is no confusion on my end. The patch video was extremely helpful in highlighting the defect introduced by this change. (The additional issues I noted were supplemental and a result of the inconsistent video comparison.)

The main problem with this change is that the Avenger's turret resets its orientation when transitioning to the same damaged model in the rubble state, which looks worse than the turret 'resetting' when transitioning to the hulk model.

image

While hulk transitions are rarely perfect, the `_D` → `_D1` transition (and `_D1` → `_D1` in the 1.04 Avenger's case) looks better and is easily preferable to `_D` → reset turret → `_D1`. To make things crystal clear, below is a comparison under identical conditions in 1.04 using the Inferno Cannon and Overlord. One resets the turret upon transitioning to the rubble state and the other does not.

https://user-images.githubusercontent.com/11547761/220236256-88e431f5-933c-4879-85ae-81695530df58.mp4

---

https://user-images.githubusercontent.com/11547761/220236264-8917c0e9-4931-4600-94a1-4b7b8471b6fa.mp4

---

https://user-images.githubusercontent.com/11547761/220255284-23fca573-a78d-4d54-8917-8b4bda2168ef.mp4

---

https://user-images.githubusercontent.com/11547761/220255299-0268ddc9-40e1-404d-b402-7ed98b694645.mp4

---

> The original has same issue. Rubble simply has no turret movement. 

As should be evident from the second and fourth videos, the turret resetting during the transition between the `REALLYDAMAGED` and `RUBBLE` states looks terrible and is extremely obvious. Conversely, the final transition from rubble to hulk (`_D` → `_D1`), while not perfect, looks fine. This is because the entire model changes / deforms, is combined with (obscured by) an effect / explosion / screen shake, and is the standard behaviour for all hulk transitions.

As shown, the issues with the Inferno Cannon and Overlord's hulk transitions are much less apparent, despite their turrets 'resetting' and the barrel debris spawning from the wrong location (or not at all). They are relatively difficult defects to spot unless intentionally sought out.

No vehicle in the game resets the orientation of its turret when transitioning to the `RUBBLE` state in this way. Introducing such an inconsistency further highlights the defect as it is not an established and expected behaviour. (Perhaps EA couldn't fix the respective issues with the `OverlordContain` module and used the `_D1` model for the Avenger's `RUBBLE` state as a workaround for this exact reason.) Even if this issue was apparent elsewhere, it would undoubtedly be just as unacceptable to add such a regression.

So I will reiterate: It is not so clear that there is a problem with the original, whereas the patched version has a fairly conspicuous defect.

xezon:
It is an issue with code, not an issue with this change. The `OverlordContain` turret must not reset to identity on rubble. The expected flow is

REALLYDAMAGED _D
RUBBLE _D
HULK _D1

There is no defect with this change. Just needs a fix for `OverlordContain` in rubble state.

The alternative is to remove delayed explosion on vehicles with `OverlordContain` module to simply explode it in one go.

Stubbjax:
> It is an issue with code, not an issue with this change. The `OverlordContain` turret must not reset to identity on rubble. The expected flow is
> 
> REALLYDAMAGED _D RUBBLE _D HULK _D1

I am aware of the expected flow.

> There is no defect with this change. Just needs a fix for `OverlordContain` in rubble state.

Before this change: Turret does not awkwardly reset → looks fine.
After this change: Turret does awkwardly reset → looks bad.

Ergo, this change introduces a visible defect.

> The alternative is to remove delayed explosion on vehicles with `OverlordContain` module to simply explode it in one go.

This is a much better idea.

xezon:
Ok. Then I will setup rubble state as if code bug did not exist and then disable destruction delay. And then when we have code fix, we can reinstate destruction delay. Would that work for you?

xezon:
Fixes applied.

* Avenger rubble states are now setup to accommodate potential code fix in the future
* Avenger debris now has same physics behavior as Humvee debris
* Avenger now has sudden death behavior like Humvee does to mask OverlordContain rubble issue (#1747)

There is still one problem with the big turret debris piece. Can either fix in this change or a new one.

Stubbjax:
The changes seem fine from a brief examination, but I unfortunately cannot compare them with the previous iteration because you force-pushed the changes and wiped the history. (This needs to stop.)

It would be good to get some updated footage of the changes so that everyone can compare the visual differences.

xezon:
I did add fixup! commits and did not squash them yet, so you should be able to see the diffs in the separate commits. There are 6 commits in this pull.

xezon:
I had one more idea for the turret death: Perhaps rubble state can be fixed by adding SlowDeathBehaviour to laser turret object.

Edit: Does not work.

PR #1735, opened 2023-02-18
Fix USA Humvee death model states and debris positions

xezon:
This change fixes USA Humvee death model states and debris positions.

Originally the undamaged Humvee would show for one frame after death.

* Debris tires now spawn at more accurate spot
* Doors now spawn at better heights and no longer inside the chassis

### TODO

- [ ] Remove Slow Death delay? Needs testing.


Stubbjax:
I would suggest reducing the Humvee's `DestructionDelay` from 1 to 0 so that the rubble does not persist for a frame after the `OCL_InitialHumveeDebris` object is spawned.

```
Behavior = SlowDeathBehavior ModuleTag_11
  DestructionDelay = 0
End
```

xezon:
Can do, we just need to check that the death effects work okay when asked to spawn at same time.

PR #1736, opened 2023-02-19
Fix USA Crusader death effects

xezon:
**Merge with Rebase**

This change fixes USA Crusader death effects.

* Removes all Mammoth Tank debris from Crusader death
* Always spawns Crusader Turret debris on death

Note: Crusader Tank has 2 Slow Death behaviors. One spawns a fire before final death and the other a generic explosion.

### TODO

- [ ] Add new USA vehicle debris pieces

## Original

https://user-images.githubusercontent.com/4720891/219944908-83b12d08-dd73-4764-99fa-c328d0696377.mp4

## Patched

https://user-images.githubusercontent.com/4720891/219944926-2d96f40b-7f3e-4c8b-b9cb-2620aaf7fc77.mp4

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaCINEUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaCINEUnit.ini
@@ -2224,7 +2224,7 @@ Object CINE_AmericaTankCrusader
     DestructionDelay  = 2000
     DestructionDelayVariance  = 300
     FX  = INITIAL  FX_CrusaderCatchFire
-    OCL = FINAL    OCL_GenericTankDeathEffect
+    OCL = FINAL    OCL_CrusaderTurret ; Patch104p @tweak xezon 18/02/2023 From OCL_GenericTankDeathEffect
     FX  = FINAL    FX_GenericTankDeathExplosion
   End
 
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaVehicle.ini
@@ -1546,7 +1546,7 @@ Object AmericaTankCrusader
     DestructionDelay  = 2000
     DestructionDelayVariance  = 300
     FX  = INITIAL  FX_CrusaderCatchFire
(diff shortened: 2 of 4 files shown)


Stubbjax:
> - Always spawns Crusader Turret debris on death

This is an objective improvement and a good change.

> - Removes all Mammoth Tank debris from Crusader death

This is a matter of personal taste and is problematic because 1) it substantially changes the look and feel and 2) it diminishes the spectacle.

xezon:
> This is a matter of personal taste and is problematic because 1) it substantially changes the look and feel and 2) it diminishes the spectacle.

Is not problematic. It spawns 12 less debris pieces that looked different than host vehicle. Less stuff is better for performance. If we want to add some more debris pieces, then we should make them with Crusader texture, not Mammoth texture.

But this is not necessary right now. Maybe more can be added on second pass. Note that China Battlemaster also just spawns 2 debris pieces and this fine.

Stubbjax:
> Is not problematic. It spawns 12 less debris pieces that looked different than host vehicle. Less stuff is better for performance. If we want to add some more debris pieces, then we should make them with Crusader texture, not Mammoth texture.

It is problematic. Visual fidelity is decreased. To many players, the effect would look worse as a result. The ideal approach is to of course replace the debris with something more Crusader-specific, though that will be far less noticeable than the debris not appearing at all.

xezon:
I tracked it as task here
* #1738

commy2:
Doesn't the Crusader have two death effects that are played at random? This makes both of them the same.

xezon:
They are different as in one of them spawns an initial fire with longer destruction delay, and the other spawns an initial explosion with shorter destruction delay.

commy2:
Only one spawns the turret in 1.04 iirc.

xezon:
Yes. One just spawned 12 Mammoth debris, and the other 12 Mammoth debris + turret + barrel.

commy2:
Yes, and after this change, both spawn with the turret flying around. I know the variants aren't masterpieces of animation, but why the regression?

commy2:
I propose:

- keep the two different death variations, but make the one without turret spawn a turret, but make that turret not fly as high into the air, but just drop to the ground.

---

Maybe fixing the missing turret randomly should be separate from the upcoming PR that adds faction specific debris. Same for Laser tank change.

PR #1737, opened 2023-02-19
Fix USA Laser Tank death states and effects

xezon:
This change fixes USA Laser Tank death states and effects.

* Removes all Mammoth Tank debris from Laser Tank death
* Shows correct damaged Laser Tank model in Rubble state
* Adds new fitting Laser Tank debris

Note: Laser Tank has 2 Slow Death behaviors. One spawns a fire before final death and the other a generic explosion.

Note 2: Debris physics behavior currently matches Crusader debris.

### TODO

- [ ] Add new USA vehicle debris pieces

## Original

https://user-images.githubusercontent.com/4720891/219945254-9b8e65f4-1709-4211-8de2-d1e5f9e0b458.mp4

## Patched

https://user-images.githubusercontent.com/4720891/219953919-8196e159-6b4c-4ef2-bdf2-0c8c28b3442a.mp4


Stubbjax:
Add back the missing Mammoth debris and it's a yes from me.

xezon:
Mammoth Tank debris will not be re-added to USA Tanks, because its
* yellow color does not fit grey USA textures
* pieces are too big and do not match anything on the USA tanks
* pieces count of 14 is way too high and causes performance degradation on mass kill events

If we want more debris, then we will need to have new USA debris that matches the grey look of USA vehicles. Around 6 debris pieces max would be appropriate.

Stubbjax:
> Mammoth Tank debris will not be re-added to USA Tanks

All changes are negotiable.

> If we want more debris, then we will need to have new USA debris that matches the grey look of USA vehicles. Around 6 debris pieces max would be appropriate.

Absolutely. It would be best to change this PR to a draft until we get something closer to the original effect.

PR #1740, opened 2023-02-19
Fix: avsentry models

### LINKED ISSUE #1160: USA Sentry Drone texture has black stripes
USA Sentry Drone texture has black stripes.

shot_20220908_130519_5

xezon:
**Merge with Rebase**

* Fixes #1160

This change fixes avsentry models.

### avsentry, avsentry_d

* Fixes UV mapping of several meshes with house color
* Fixes UV mapping at the bottom of the vehicle
* Fixes TURRETFX01 bone position
* Fixes muzzle flash texture positions
* Fixes muzzle flash effect position
* Fixes muzzle smoke effect position
* Fixes muzzle shells effect position
* Fixes missing alpha test setting on radar dish
* Fixes z clipping at the bottom rear of the chassis
* Fixes slightly misplaced house color plate at the rear end

### avsentry_d

* Fixes house color meshes clipping into chassis
* Fixes radar dish using avsentry instead of avsentry_d texture
* Fixes wrong base rotation of radar dish
* Fixes slightly misplaced house color plate at the rear end

### avsentry_d2, avsentry_d4

* Fixes incorrect size of debris

---

TURRETFX01 Bone has been moved into proper position.

avsentry_d1, avsentry_d2, avsentry_d3 have not been touched.

### TODO

- [x] Fix headlamps culling
- [x] Fix Z clipping at bottom of chassis
- [x] Fix muzzle flash position
- [x] Fix muzzle fx position
- [x] Fix muzzle smoke position
- [x] Fix gun tracer position
- [x] Fix gun shell position
- [x] Remove pre-exported w3d models when Mod Builder allows so
- [x] Remove Basic_Sphere.007 mesh?

## Patched (with texture fixes)

shot_20230219_194140_3


xezon:
Various effects in `WeaponFX_SentryDroneGunFire` need fixing.

xezon:
Headlamps culling is now fixed. All gun fire effects are fixed.

xezon:
~~I have now also found the Blender file compression option. That reduced avsentry.blend file size from 1500 kb to 170 kb.~~

Edit: Compression is not meant to be used for git versioning.

Stubbjax:
A before comparison / showcase of each of the fixes would be nice.

Why do we need the Blender file?

xezon:
Blender files are meant to be kept for the same reasons we preserve PSD source textures, or STR language files, or C++ source code. Going forward the goal is to only add blend files, not w3d exports. Note that w3d file can have lossy compression too.

Stubbjax:
> Blender files are meant to be kept for the same reasons we preserve PSD source textures, or STR language files, or C++ source code. Going forward the goal is to only add blend files, not w3d exports. Note that w3d file can have lossy compression too.

I don't think those same reasons apply here. Is it not basically the same as saving a PSD file with a single rastered layer instead of a PNG? What additional information is stored in a Blender file that we would need?

It's generally preferable to minimise external build dependencies where possible.

xezon:
That discussion is beyond the scope of this pull request. It should be discussed in the Discord instead, where there already are relevant notes in the tool development section.

To give hints as to why W3D is not optimal as source file, here are 5 reasons:

1. W3D files can have compression. Ideally we avoid re-importing and re-exporting compressed files many times, because it is unclear how detrimental it can be.

2. W3D files have different export formats and settings, such as Hierarchical, Animated Hierarchical, Mesh only, Compression, No Compression, etc, which need to be carefully picked on manual W3D export. When automated, then W3D format and settings just needs to be setup once and will remain correctly set on exports according to our model setup.

3. W3D files can contain chunks of complex models, for example Actor Animation W3D file can be separate from Actor Mesh W3D. Or Vehicle Debris W3D can be separate from Vehicle W3D. Now one can manually manage separate asset files and transfer from one file to another, OR create one uber blend file to have it all sit together, and export separated W3D files in automated export.

4. Perhaps Blender plugin or W3D format itself will get new fixes or new features in the future. If W3D Export was not automated, then this means existing W3D models will not benefit from changes and would have to be manually re-imported and re-exported. With automated source model export, there is no such problem.

5. Upon W3D model import, perhaps something very specific is not functional and may need manual fixing until the W3D Importer is fixed. Exporting and re-importing W3D may lose this fix again. In that case having the source asset is essential.

xezon:
There appears to be an alien sphere in the blend file. Investigate
```
Basic_Sphere.007 is MESH
  basis.pos: [0.000, 0.000, 0.000]
  local.pos: [0.000, 0.000, 0.000]
  world.pos: [0.000, 0.000, 0.000]
  basis.ang: [0.000, -0.000, 0.000]
  local.ang: [0.000, -0.000, 0.000]
  world.ang: [0.000, -0.000, 0.000]
  MESH Basic_Sphere.007
    vertices: 62
    polygons: 72
```

MTKing4:
It seems You have a Mesh Linked to a Ghost Object Called `Basic_Sphere.007` and `Basic_Sphere` in both `AVSentry.blend` and `AVSentry_d.blend` respectively that is not assigned to an object nor has a material added to it therefore not showing in your viewport

I assigned it to an object in the Picture below and this is how it looks like

image

It is Unlikely that it serves any purpose since it doesn't even belong to any object so i went ahead and removed it for you.

I'll add the files as a change suggestion and we can test if it has any effect on the game

MTKing4 (dismissed):
Remove Basic Sphere

MTKing4:
I'm not sure i did this right 😄

xezon:
I have already noted this in the TODO of first post.

MTKing4:
ah ok, well i already committed it, you can review if everything is correct

xezon:
* Spheres removed from sentry blend files
* sentry w3d files removed
* sentry blend files moved to correct folder

xezon:
Thy only issue left on the models is the z clipping at the bottom of the chassis. But I have not figured out how to fix it and it is not typically visible in game, so probably needs no attention.

xezon:
If #2211 is merged before this, then HEADLIGHT needs to be renamed to MANUALLIGHT in both models. And the avsentry w3d files need to be removed.

commy2:
Permanent muzzle flash if upgrade happens while moving:

shot_20230806_220040_10


- Turret resets into middle position for a second before returning to actual state.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/3969e038-bc52-4e89-baee-0d774a8c27ba

- black stripe still visible, unlike how OP looks

commy2:
shot_20230806_221211_7

- Muzzleflash is way way too big.

commy2:
If this were Renx, I think you'd have to tick this option on the muzzle flash:

Unbenannt

No idea where to find that in Blender.

xezon:
Muzzle flash size is as per original.

As for the other 2 things, does this not happen with original?

commy2:
> As for the other 2 things, does this not happen with original?

No, can't repro on main.

MTKing4:
> If this were Renx, I think you'd have to tick this option on the muzzle flash:
> 
> Unbenannt
> 
> No idea where to find that in Blender.

What does this hide button do? Hide it in a certain timeframe? Or just hide it altogether?

If it's the latter you can just select the object and press H
Or press the Eye button from the scene collection List

commy2:
Hides the section by default. It will still be unhidden by engine when needed, e.g, by ShowSubObject or by hard-coded name like MuzzleFX or Headlight when appropriate.

MTKing4:
Ok i will look into this when i get home

commy2:
Simplified W3D ini by using SubObjectsUpgrade instead of many redundant WEAPONSET_PLAYER_UPGRADE states.
xezon, you can just use REALLYDAMAGED RUBBLE combined. The game picks the closest match and both are mutual, so that's fine. EA uses that trick everywhere.

PR #1745, opened 2023-02-21
Change: Begin sink China Dragon Tank hulk right after the fire effect vanished

xezon:
This change begins to sink China Dragon Tank hulk right after the rubble fire effect vanished. It now sinks after 12000 instead of 14000 milliseconds. This way it does not stay in the world longer than necessary.


commy2:
Why?

xezon:
> Why?

Looks better. And there is no need to have the Dragon Tank wreck linger too much longer around than every other vehicle wreck.

PR #1746, opened 2023-02-21
Change: Set consistent sink delays for faction vehicles

### LINKED ISSUE #1183: USA Humvee has significantly longer sink time than other vehicles

xezon:
**Merge with Rebase**

* Fixes #1183

This change increases or decreases sink delays of

* all regular faction vehicles to 3000 ms
* all large faction vehicles to 6000 ms

Large vehicles are:

* China Overlord
* China Nuke Cannon
* China Helix

## Rationale

All faction things now sink after at least 3000 ms. Infantry already sinks after 3000 ms. This allows the observer to see the hulks for just a bit longer after the death effects disappeared.

Large vehicles are set to sink slower because the Original Overlord also stayed longer and they are big (and strong) vehicles, so their hulks can be presented for a bit longer to highlight the significance of the large vehicle kill for just a bit longer than regular smaller vehicles.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1746_wreck_sink_delays.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1746_wreck_sink_delays.yaml
@@ -1,0 +1,62 @@
+---
+date: 2023-01-29
+
+title: Sets consistent sink delays for faction vehicle wrecks
+
+changes:
+  - tweak: Sets sink delays for wrecks of regular vehicles and aircraft to 3000 ms.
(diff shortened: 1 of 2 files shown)


xezon:
Documentation added. I need to think about whether or not 3000 ms delay is optimal choice. I think I should finish wreck particles fixes first before concluding this. Some vehicles have no wreck smoke or fire.

xezon:
3000 ms looks good. It starts sinking about the time when the Crusader fire disappears.

PR #1750, opened 2023-02-24
Change: Add 3D shadows to big bombs

Stubbjax:
This change adds shadows to all carpet bombs (all weapon objects using the `EXCarptBmb` model). This makes it much easier to determine the height and trajectory of the bomb.

Values were copied from other aircraft so that the shadow aligns with them.

---

Where is the bomb going to land?

image

It's much clearer now!

image

_I can see many shadows appearing in the distance._

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1750_bomb_shadows.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1750_bomb_shadows.yaml
@@ -1,0 +1,22 @@
+---
+date: 2023-02-24
+
+title: Adds 3D shadows to bomb objects
+
+changes:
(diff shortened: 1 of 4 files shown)


MTKing4:
Very Nice

Is this the New Bomb Model we made? it looks like it

xezon:
Perhaps add it to all other bombs as well then? MOAB & Co.

xezon:
Documentation added. 3D shadows added to objects using models EXAMine_A and AVBomber_B. Not added to object using PMMoab, because the shadow somehow does not show for it. I could not figure out why.

xezon (approved):
Looks good.

PR #1752, opened 2023-02-24
Change: Removes the destruction delay variance from the USA Alpha Aurora's bomb

Stubbjax:
This change removes the 100ms destruction delay variance from the Aurora Alpha's bomb weapon. It's not ideal that important gameplay elements (particularly arbitrary ones like this one) are randomised in such a way, and it's best to minimise them where possible.

After this change, the Aurora Alpha bomb's fuel-air explosion will always detonate exactly 1000ms (30 frames) after igniting, instead of a random time between 1000ms (30 frames) and 1100ms (33 frames).

Although this is technically a buff, it should be noted that this partially cancels out the increased detonation delay of 3 to 5 frames introduced by #292. (In 1.04, the fuel ignites 2 to 4 frames _before_ impact, whereas the fuel ignites one frame _after_ impact as of #292.)

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1752_alpha_aurora_bomb_random_delay.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1752_alpha_aurora_bomb_random_delay.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-02-24
+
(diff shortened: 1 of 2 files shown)


xezon (approved):
Looks good.

Issue #655, opened 2022-02-18
USA Aurora cannot be destroyed when it locks on target


xezon:
USA Aurora cannot be destroyed when it locks on target. Aurora is banned in semi player games because of it. There is no counter to it, other than attacking the airfield in the enemy base. Maybe Aurora should not have this feature.

### Proposal 1
Aurora is not invincible, but stealth while attacking. This way stealth detection + anti air can kill it. Similar to Stealth Fighter. Nerf for base attacks, but buff for army attacks in the open, because the opponent will not see attacks coming.

### Proposal 2
Aurora is not invincible, but tanks a lot more damage while attacking. This way it can still be shot down if there is enough firepower present. Nerf for base + army attacks.

### Proposal 3
Aurora cost is changed.

### Proposal 4
Increase reload time. Then other players have more time to kill it while in the airfield and have more time to recover in between Aurora attacks.

### Proposal 5
Remove SF GP unlock, and give it to the Aurora instead

### Proposal 6
Increase the turn radius of the Aurora's slow locomotor so it has a higher chance of getting shot down after it attacks

### Proposal 7
Lessen its invincibility miss distance so AA weapons re-target faster after the Aurora comes out of supersonic mode and weapons with large AoE (nuke mig missiles) have a chance of killing them even in supersonic mode

### Proposal 8
Alpha Aurora only: Increase delay of their FAB explosion to make it less reliable in hitting moving targets

### Proposal 9
Decrease Aurora HP or increase the damage it takes while in low fuel

### Proposal 10
Aurora is not invincible for Avengers and Laser Turrets (Light Speed > Aurora Speed)

MTKing4:
Proposal 2.

ExiLe-ZH:
Don’t see reason to remove this feature tbf, it’s the only reason to go for them (and the Alpha because of it’s damage ofcourse). Other than that non-alphas are hardly ever economically viable. They can’t destroy more than they’re worth in 1 run and there’s no guarantee they survive when killing high value targets, like SW’s. 99% of the time it’s a 10k vs 5k trade (+airfield that takes awful lot of space). Sure they can be used to kill defenses at frontlines but then it’s still a risky overinvestment. 

I forsee that with either proposal noobs will still ban them for being OP (while they never were). 

Noobs will always play with weird rules anyway and you can’t balance around that, like the No-Rush rule is exactly the same as the No-Whatever rule.

[...]

PR #1753, opened 2023-02-24
Change: Decrease USA Alpha Aurora bomb drop speed to match the regular Aurora bomb drop speed

Stubbjax:
This change slightly reduces the Aurora Alpha's bomb speed to match the vanilla Aurora. From maximum range, this increases the time to impact by roughly three frames. Based on several comments in #655, it seems reasonable to slightly nerf the Aurora Alpha in a subtle way that maintains the same feeling of 1.04 in terms of damage output, but with a higher skill requirement to use and greater potential for counterplay from attentive opponents. And what better way to do it than streamlining some counterintuitive differences?

## Before:
Bombs strangely travel at different speeds, which feels somewhat counterintuitive. It is evident that even though the bombs are dropped on the same frame, the Alpha Aurora's bomb travels slightly faster to the ground and impacts four frames earlier.

https://user-images.githubusercontent.com/11547761/221228395-8f99eb2a-0c36-4eb3-b8b0-695ea2e15948.mp4

## After:
Both bombs now have identical speed and impact the ground at roughly the same time (only one frame apart due to the slightly different travel distances). Note that the detonation from the vanilla Aurora bomb coincidentally destroys the Aurora Alpha bomb and stops it from detonating, which is a phenomenon that occurs with most if not all projectile weapons in 1.04.

https://user-images.githubusercontent.com/11547761/221228417-9aed8c4e-d475-4f6d-a301-1d768d983b92.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1753_alpha_aurora_bomb_speed.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1753_alpha_aurora_bomb_speed.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-02-24
+
+title: Fixes Alpha Aurora bomb movement discrepancies
+
+changes:
+  - tweak: The Alpha Aurora bomb now travels as quickly as the original regular Aurora Bomb. Effectively it hits the ground around 5 frames slower than before.
+
+labels:
+  - controversial
+  - design
+  - nerf
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1753
+
+authors:
+  - Stubbjax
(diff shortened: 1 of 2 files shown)


xezon:
According to video, Alpha Bomb hits 5 frames sooner than the other Bomb.

xezon:
With this change I observe Alpha Aurora bomb drop time of 37 frames vs original 26 frames. Plus fuel bomb triggers 2 frames later than before. So all together new Alpha Aurora bomb is 13 frames slower - almost half a second.

xezon:
New measured sample frames

### Original

Alpha Aurora Bomb hit 30
Alpha Aurora Fuel trigger 26
Alpha Aurora Fuel explode 58
Aurora Bomb hit 34

### Patched

Alpha Aurora Bomb hit 35
Alpha Aurora Fuel trigger 35
Alpha Aurora Fuel explode 66
Aurora Bomb hit 36


## Observations

1. There appears to be some variance in Aurora Bomb flight duration. I have measured 33 to 37 frames. Not sure where this rather large discrepancy comes from.

2. The Alpha Aurora bomb stops for around 2 frames before impact. Something is not right there.

xezon:
Impact issue appears to come from `GeometryMajorRadius = 12.0` vs `GeometryMajorRadius = 2.0`

xezon (approved):
Looks proper.

PR #1755, opened 2023-02-25
Change: Assist weapons of USA Patriot will always reload when idle

xezon:
* Related to #1580
* Related to #1178

With this change the assist weapons of USA Patriot will always reload when idle.

This makes the weapon a bit better in scenarios where it managed to only fire with a few of its rockets on its target and some time passes until the next target is attacked.

This situation is common. There is a `ClipReloadTime` of 2000 ms (from regular weapon to assist to) vs a `DelayBetweenShots` of 266 (250) ms. This means there is a 35% chance that assisted USA Patriot gets into state with just some rocket in its battery after attacking targets.

# Rationale

Eliminates situations where USA Patriot would only fire some rockets on a fresh engagement. This likely is what player would expect always.

This change also makes the assisted weapon behavior consistent with the patched non-assisted weapons. See
* #1178

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1178_patriot_battery_auto_reload.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1178_patriot_battery_auto_reload.yaml
@@ -1,13 +1,14 @@
 ---
 date: 2022-09-10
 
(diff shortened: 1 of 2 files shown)


xezon:
Looks good.

PR #1759, opened 2023-03-05
Fix: Remove move voices from USA Spectre Gunship

xezon:
This change removes the move voices from USA Spectre Gunship. They can be triggered after the shooting, but the Gunship will not move where ordered contrary to what the voices imply.

## Original

https://user-images.githubusercontent.com/4720891/222984064-1d4392b2-a0fd-4227-8c58-df70dc1f7857.mp4


MTKing4:
This Voice can be used when unlocking the unit as a playable unit using INI for fun mod maps, so I'd rather have it there in the game files

xezon:
Then mod map INI can add this sound.

xezon:
Looks good.

PR #1774, opened 2023-03-18
Fix: Listening Outpost Damaged Smoke Effect Sticks After Repair (2)

xezon:
* follow up for #261

Adds missing comments and corrects the smoke effect removal on
* Infa_ChinaVehicleTroopCrawler
* Infa_ChinaVehicleListeningOutpost


commy2 (approved):
No idea how this mix-up happened.

PR #1413, opened 2022-10-30
Change: Enable and fix audio for USA Battle Drone repairs

xezon:
This change enables and fixes audio for USA Battle Drone repairs.

The sound files already existed, but had some humming and quite "crispy" sparks. I reduced the humming and made the sparks less "crispy", so they do not jump as much into the ears. Thus they sound subtle in 3d world.

https://user-images.githubusercontent.com/4720891/198886343-1d8d3a52-3493-4942-820a-b63943fe9c8c.mp4

https://user-images.githubusercontent.com/4720891/198886354-791f6877-456e-4d72-8292-da2a6ad13b1f.mp4

PR #1785, opened 2023-03-26
Change: Set lowest audio priority on USA Battle Drone RepairSparks

xezon:
* Follow up for #1413

Sets lowest audio limit for Battle Drone repair sparks. This way it does not mask more important sounds.


xezon:
Works.

PR #1787, opened 2023-03-26
Change: Add buttons for Guard Mode and Attack Move to USA Airforce Combat Chinook

xezon:
This change adds buttons for guard and attack move to USA Airforce Combat Chinook.

It would also be possible to add the button for air guard, but the Helix does not have it - perhaps because of lack of space - so it was not added to Combat Chinook either.

Tested and worked.

shot_20230326_121232_1


xezon:
Any comments for this?

MTKing4:
It is good enough

commy2:
If LO and TC got a guard button, so should CC. 👍

xezon:

PR #1801, opened 2023-04-06
Change: Reduce audio range and volume of Remote Demo Charge

xezon:
* Fixes #1793

This change reduces the audio range and volume of the Remote Demo Charge to make its presence less obvious.

| Sound                      | MinRange (*1)   | MaxRange (*1)   | MinVolume | MaxVolume |
|----------------------------|-----------------|-----------------|-----------|-----------|
| Original BombTickRemote    | 175             | 800             | 40        | 65        | 
| **Patched BombTickRemote** | 100 (-66%)      | 250 (-90%)      | 20        | 40        |

(*1) = Difference in area coverage in percent.

## Patched

https://user-images.githubusercontent.com/4720891/230464175-dde52016-584d-493e-91eb-4e77aac99231.mp4

# Rationale

Placing the original Remote Demo Charge with USA Colonel Burton or GLA Demo Jarmen Kell makes the presence of both the bomb and the instigator very obvious. At an audible range of 800 it can be heard over a very large area (see sample in #1793).

The massive reduction in range and volume makes the presence much less obvious and allows for more sneaky plays with Burton and Jarmen. The defender needs more attention over his base to hear the bomb(s).

GLA is generally less vulnerable to demo charges due the structure rebuild from hole and the common presence of stealth detecting Tunnel Network in key areas. China is able prevent demo charge placements by acquiring structure mines. Placing too many demo charges in the China base could be risky if the Hero unit steps on mines. All pending demo charges will be lost then. Both China and USA need stealth detection units to prevent enemy Hero units from causing damage.

Issue #1793, opened 2023-04-03
Demo charge ticking sound is loud and makes hidden bomb presence obvious


xezon:
Demo charge ticking sound is loud and makes hidden bomb presence obvious.

Explore whether or not sound volume and radius needs reduction.

commy2:
This is talking about Burton / Demo Jarmen / USA04 Agent, right?

xezon:
Yes.

xezon:
kABUSE:
> If you were a Mr. Burton highly trained super stealth secret agent rambo dude trying to take out a target, would you add a beeping sound to your remote bombs? It makes no sense that they beep and it makes them entirely useless. It would be a lot better if the remote bombs did not beep. The timed bombs can still beep, as they have other advantages (aka not disappearing when burton is killed). Maybe add just an initial beep. The remotely triggered bombs are not overpowered at all. Players exclusively use timed bombs, as they have the advantage of still going off if Burton dies, 
so giving a reason to actually use remote bombs could be interesting.

MTKing:
> It is visibility feature, Burton is overpowered enough having to deploy bombs unannounced i think the tick sound is justified to at least give the opponent a chance to respond by selling, it make sense gameplay wise. Now Imagine a tickless bomb burton will plant all over your base and destroy it all at once without you even realizing. China struggles a lot vs Burton, with little stealth detection, no announcement on placements, and now no ticking sound will give the player very little chance to detect it, as opposed to how lotus is detected when capturing for example. I think the beeping sound is fine for remote bombs. I think of it like a RF signal beeper, not a time ticker.

Exile:
> I like the idea of removing the sound from the remote charge, beeping makes sense as a timer. Timed charge is indeed mostly preferred because there is less risk. The initial sound + short reveal does not make it overpowered I think. There are only few cases now where the remote is preferred, for instance when planting the charge on a truck from a mined supply.

xezon:
> Burton is a pretty good Hero unit already. Removing the bomb sound would be a significant buff. Lotus also makes noises when attacking buildings. I am open to exploring reduction of beep sound, so it is not as obvious and requires a bit more player care to detect. So for example sound can be quieter, and less audible radius. If my memory does not fail me, original bomb ticking sound is quite loud and has generous radius, so it is almost always spotted.

xezon:
The audible range is definitely too much. It is also quite loud, especially because the beep sits on a mostly unchallenged sound frequency.

[...]

PR #1802, opened 2023-04-06
Change: Reduce audio range of Timed Demo Charge

xezon:
* Relates to #1793
* Relates to #1801

This change reduces the audio range of the Timed Demo Change to make its presence less obvious.

| Sound                      | MinRange        | MaxRange (*1)   | MinVolume | MaxVolume |
|----------------------------|-----------------|-----------------|-----------|-----------|
| Original BombTickTimed     | 175             | 800             | 40        | 65        | 
| **Patched BombTickTimed**  | 175             | 500 (-61%)      | 40        | 65        |

(*1) = Difference in area coverage in percent.

## Patched

https://user-images.githubusercontent.com/4720891/230466133-e5b86405-1b0c-4552-8210-9740feaaaf51.mp4

# Rationale

Placing the original Timed Demo Charge with USA Colonel Burton or GLA Demo Jarmen Kell or China Tank Hunter makes the presence of both the bomb and the instigator very obvious. At an audible range of 800 it can be heard over a very large area (see sample in #1793).

The reduction in range makes the presence much less obvious and allows for more sneaky plays with Burton, Jarmen and Tank Hunters. The defender needs more attention over his base to hear the bomb(s).

GLA is generally less vulnerable to demo charges due the structure rebuild from hole and the common presence of stealth detecting Tunnel Network in key areas. China is able prevent demo charge placements by acquiring structure mines. Placing too many demo charges in the China base could be risky if the Hero unit steps on mines. All pending demo charges will be lost then. Both China and USA need stealth detection units to prevent enemy Hero units from causing damage.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1802_timed_demo_charge_audio.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1802_timed_demo_charge_audio.yaml
@@ -1,0 +1,23 @@
+---
+date: 2023-04-06
+
+title: Reduces audio range of Timed Demo Charge
+
+changes:
+  - tweak: Reduces the Timed Demo Charge's maximum audio range from 800 to 500. This makes the presence of a demo charge and its instigator much less obvious as it has its audible area reduced by 61%.
+
+labels:
+  - audio
+  - china
+  - controversial
+  - design
+  - gla
+  - minor
+  - usa
+  - v1.0
+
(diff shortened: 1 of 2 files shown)


xezon:
Any feedbacks for this?

MTKing4:
I think it's a safe change

xezon:

PR #1804, opened 2023-04-07
Fix: Inaccurate tooltip text of GLA Booby Trap upgrade

### LINKED ISSUE #1794: GLA Booby Trap upgrade tooltip text has inaccurate information
GLA Booby Trap upgrade tooltip text has inaccurate information.

It says: "Rebels can place booby traps on buildings and neutral units"

Booby traps cannot be placed on neutral units (vehicles).

xezon:
* Fixes #1794

This change fixes the inaccurate tooltip text of the GLA Booby Trap upgrade for all languages except Arabic. It no longer claims it can be placed on neutral units.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1804_booby_trap_upgrade_text.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1804_booby_trap_upgrade_text.yaml
@@ -1,0 +1,19 @@
+---
+date: 2023-04-07
+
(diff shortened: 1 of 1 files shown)


ExiLe-ZH:
Should you be able to place it on neutral units though? Sounds like a fun feature but kinda useless aswell.

commy2:
^ This would've to be implemented via engine changes.

PR #1808, opened 2023-04-07
Change: Remove Rally Point button from all command sets

xezon:
This change has been reverted with #1851

~~This change removes the Rally Point button from all command sets. It will not be missed.~~

## Rationale

~~The button is obsolete. Rally Point functionality is always enabled as long as structure is selected.~~

## Original

shot_20230407_141859_1

## Patched

shot_20230407_142002_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1808_remove_rally_point_button.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1808_remove_rally_point_button.yaml
@@ -1,0 +1,19 @@
+---
+date: 2023-04-07
+
+title: Removes obsolete Rally Point button from all command sets
+
+changes:
+  - tweak: Removes the obsolete Rally Point button from all command sets. The functionality of the Rally Point feature is unchanged.
+
+labels:
+  - design
+  - gui
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1808
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -20,6 +20,7 @@
 ;------------------------------------------------------------------------------
 
 ; Patch104p @bugfix commy2 03/09/2021 Align evacuate buttons on same slot.
+; Patch104p @tweak xezon 07/04/2023 Removes Command_SetRallyPoint from all command sets because it has no use.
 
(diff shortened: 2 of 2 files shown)


commy2:
I remember someone objecting to this change when I made it in my project. I don't remember why though. As far as I can tell, this button is completely pointless.

MTKing4:
I agree that it is pointless, but I'm not a fan of removing things that could help newcomers to the game if any, especially if we're not in dire need of that extra slot currently

xezon:
In what way does it help newbies though? Clicking this button really changes nothing. The only this this button does it tell that this structure has Rally Points.

MTKing4:
Yeah in case there's someone confused and doesn't know anything about RTS and Wondering what does the flag does, he can read the command button to understand

Basic RTS tutorial in Essence.

commy2:
>Set Rally Point

Tooltip:
>Set the rally point for this building

Idk if that ever helped anyone.

MTKing4:
Yeah I'm not too fussed about removing it either, just stating what the purpose of it was wether it helped anyone in the past or not.

commy2:
If ingame info about rallypoints is truly needed (I don't think so personally), then it should be placed better than on a button that does absolutely nothing. I mean, this thing is more confusing right now than clearing up anything, because the button does literally nothing at all (as far as I can tell). Not even change cursor or anything.

MTKing4:
Placed better like where?

commy2:
Anywhere really. Anywhere that is not a useless button, that can be clicked, but nothing happens for some reason.

xezon:
Merge this?

xezon:
By coincidence I noticed that the button was not removed from Superweapon Command Center. Fixed. Documentation added. Looks good.

kABUSE1:
There is a situational difference between having the command button and not having it:

https://user-images.githubusercontent.com/47530719/232312579-8a6616ce-4166-4287-bdd2-401db0ddf733.mp4

If you have the command button you can place the rally point into a group of units via the hotkey without accidentally selecting any of them.

I only noticed after looking for a difference, I Have never done this in an actual match.

xezon:
Ohh. So it does make a difference. Well then we can revert this change again :D

kABUSE1:
So let's do the opposite, the boss general airfield is lacking this command button by default lmao

kABUSE1:
Please also take this into consideration:
https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1985

(Nevermind, current issue is already closed. I am still in the progress of understading how git works lol)

xezon:
This change was merged and reverted.

PR #1800, opened 2023-04-06
Fix: Improve Daisy Cutter explosion effect and sound sequence, remove destruction delay variance

xezon:
**Merge with Rebase**

* Relates to #1752

This change fixes up the entire Daisy Cutter explosion effect and sound sequence.

* Removes the destruction delay variance of 100 ms
* Removes the 250 ms delay baked into the explosion audio effect.
* Adds smooth fade out to the explosion audio effect.
* Syncs the ignition particles better with the ignition sound.
* Syncs the explosion effects better with the explosion sound.

The sound is used for Fuel Bomb, MOAB, Alpha Aurora, Bunker Buster.

## Patched, before this change

https://user-images.githubusercontent.com/4720891/230896420-459f9ba2-04fa-40ef-ad81-d3a74961ba36.mp4

## Patched, after this change

https://user-images.githubusercontent.com/4720891/230896448-a0059246-8add-47cc-878f-bd7ced5bfdf8.mp4

PR #1810, opened 2023-04-07
Fix: Flawed audio quality in Explosion FX sounds

xezon:
This change fixes clicking and fade out in all `gexp` sounds, except gexpdaia.wav which is already covered by #1800. The explosion sounds now fade out gently without hard cuts.

All wav files are used as is uncompressed for the time being. The original respective sampling rates of 22050 or 44100 are preserved.


xezon:
Sounds good to me. Documentation needs to be updated.

xezon:
Documentation updated.

PR #1812, opened 2023-04-07
Fix: Flawed audio quality in various FX sounds

xezon:
This change fixes noises, clicking and fade out in all g* sounds, except gexp* which are already covered by #1800 and #1810. The sound effects now fade out gently without hard cuts.

All wav files are used as is uncompressed for the time being. The original respective sampling rates of 22050 or 44100 are preserved.


xezon:
Second review pass completed. Documentation added. Sounds good.

Issue #1809, opened 2023-04-07
Tank speed balancing


ExiLe-ZH:
Unit | Speed / Acceleration | Upgraded | 
-- | -- | -- 
Overlord / Emperor | 20/15 | 30/30
(Tank) Battle Master | 25/1000 | 35/1000
(Laser) Crusader / Paladin | 30/1000 | - 
Scorpion / Marauder | 40/1000 | - 
Quad Cannon | 40/1000 | - 
Gatling Tank | 40/1000 | - 
Dragon Tank | 40/1000 | - 
Nuke Overlord | 40/30 | - 
Nuke Battle Master | 45/1000 | - 

Battle Masters and especially Overlords (Emperors) are painfully slow, a huge part of why they suck, Battle Masters often are often easily outflanked by Dragon Tanks, while they're supposed to be the counter unit. Overlords are extremely easy targets for Terror Techs.

40/1000 seems to be the standard most units and BM's/Overlord seem to be the outliers. Would a small buff like 27.5 be justified?

xezon:
Speed increase would help China a lot. Speed of 27.5 would break the speed step size of 5 which all units adhere to if I am not mistaken. It should be 30 then.

Also, it is odd that upgraded Overlord is 30 only while Battlemaster is 35. Would it not be better if they upgraded to the same speed?

commy2:
Vast majority of Chinese units move at 40. Would be neat if at least with Nuclear Tanks, BM reached 40 speed, although it might be op, idk.

MTKing4:
I doubt 40 speed would make it op, it's already underused, and that upgrade is quite an investment and very lategame, i think we can get away with increased speed for both BMs and Ovies

xezon:
Battle Master base:30 upgraded:40
Overlord base:30 upgraded:40

?

MTKing4:
Base for both ~25~ 30 , but yeah Upgraded to 40

Nuke BM will still be faster at 45 speed
~And Nuke Ovie will still be faster to respond/rotate with faster Acceleration at 30 
compared to normal ovie at 15~

Maybe 40 is too much for ovie idk, 35 is good as well

ExiLe-ZH:
The only thing why I still doubted the 30 speed, is the combination of changes we made to China, like Stealth might even struggle a bit vs Tank BM spam. Remember the faster building dozer for faster truck rush and less Elite BM XP.

[...]

PR #1813, opened 2023-04-08
Change: Increase China Overlord speed by 25%, upgraded speed by 16%

### LINKED ISSUE #649: Upgraded Vanilla Overlords should be as fast as Nuke Overlords?
With the Nuclear Tanks upgrade Vanilla Overlords are still slower than Nukes Overlords, should they be just as fast?

xezon:
**Merge with Rebase**

* Closes #649
* Relates to #1809

This change increases the regular China Overlord speed by 25% and upgraded speed by 16%. The Emperor Tank speed is unchanged.

| Object                            | Speed | Dmg. Speed | Accel. | Turn | Upgr. Speed | Upgr. Dmg. Speed | Upgr. Accel. | Upgr. Turn |
|-----------------------------------|-------|------------|--------|------|-------------|------------------|--------------|------------|
| Original China Nuke Overlord      | 40    | 40         | 30     | 80   | -           | -                | -            | -          |
| Original China Emperor            | 20    | 20         | 15     | 60   | 30          | 30               | 30           | 60         |
| Original China Overlord           | 20    | 20         | 15     | 60   | 30          | 30               | 30           | 60         |
| **Patched China Overlord (this)** | 25    | 25         | 20     | 70   | 35          | 35               | 30           | 70         |

# Rationale

Regular China is the one of the weakest factions in the game. Overlords are rarely used, because they are expensive and slow. Mass Humvees and Buggies can kill them quickly. Jarmen Kell snipes them. A slight increase in movement speed can increase the utility of the Overlord Tank. With the increased speeds, before upgrade it still is 17% slower than regular tanks, 37.5% slower than GLA tanks and 37.5% slower than Nuke Overlord. After Nuclear Tanks upgrade it is just 12.5% slower than Nuke Overlord. The Nuclear Tanks upgrade costs 2000 and requires the Nuclear Silo, so it is rather inaccessible and late game only.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1813_overlord_speed.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1813_overlord_speed.yaml
@@ -1,0 +1,27 @@
+---
+date: 2023-04-08
+
+title: Increases movement speed of China Overlord by up to 25%
+
+changes:
+  - tweak: |
+      Increases regular China Overlord
+        - speed from 20 to 25
+        - acceleration from 15 to 20
+        - turn rate from 60 to 70
+        - upgraded speed from 30 to 35
+        - upgraded turn rate from 60 to 70
+
+labels:
+  - buff
+  - china
+  - controversial
+  - design
+  - major
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1813
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
@@ -4559,14 +4559,40 @@ End
 ;------------------------------------------------------------------------------
 Locomotor Tank_OverlordLocomotor
   Surfaces            = GROUND
-  Speed               = 25   ; in dist/sec
-  SpeedDamaged        = 25   ; in dist/sec
-  TurnRate            = 60   ;25   ; in degrees/sec
-  TurnRateDamaged     = 60   ;20   ; in degrees/sec
+  Speed               = 20   ; in dist/sec
+  SpeedDamaged        = 20   ; in dist/sec
+  TurnRate            = 60   ; in degrees/sec
+  TurnRateDamaged     = 60   ; in degrees/sec
   Acceleration        = 15   ; in dist/(sec^2)
   AccelerationDamaged = 15   ; in dist/(sec^2)
   Braking             = 50   ; in dist/(sec^2)
   MinTurnSpeed        = 0    ; in dist/sec
+  ZAxisBehavior       = NO_Z_MOTIVE_FORCE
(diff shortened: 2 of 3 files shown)


xezon:
Could also increase the Turn Rate from 60 to 70, so it sits between former 60 and Nuke's 80.

ExiLe-ZH:
Exactly how I had it in mind for the Vanilla Overlord, but any reason why the Emperor should be slower?

xezon:
I am not clear what to do with Emperor. We already gave it the Subliminal Buff and I consider the Emperor the stronger Tank, because it can get Gattling + free Propaganda Tower. So perhaps it should not be so fast.

ExiLe-ZH:
As for balance reasons I share the same concerns, however logic wise it's odd how the supposedly better Emperor has a downgraded engine so to say. Perhaps increasing it's cost can somewhat make up it.

xezon:
Visually the Emperor is bulkier than the Overlord, also can carry 2 attachments with Propaganda Tower + Gattling Tower. So from that angle the lower speed is plausible.

Plus China Tank appears to be more competitive than regular China in Original Game, so it does not need speed buff as much.

xezon:
TurnRate increased from 60 to 70.

ExiLe-ZH:
> Visually the Emperor is bulkier than the Overlord, also can carry 2 attachments with Propaganda Tower + Gattling Tower. So from that angle the lower speed is plausible.
> 
> 
> 
> Plus China Tank appears to be more competitive than regular China in Original Game, so it does not need speed buff as much.

Double attachment is a good point, but only valid when it's actually installed.

It does look more bulky, but does it actually have any additional armor? Compared to a vet 1 Vanilla PropLord?

Btw question, how does the free prop compare to the regular upgradable prop?

Anyways yes, Tank is a little more competitive overall, however I can safely state that Emperors very marginally contribute to the higher winrate, if at all? I rarely see those made, let alone change games, in 1v1 atleast. I can only remember one Rage vs Size WS game really.

xezon:
Armor of Overlord and Emperor is identical. TankArmor + 1100 Health. Propaganda Tower is the same. This change does not aim to make the Emperor quicker.

ExiLe-ZH:
While doing this change, we should highly consider doing it for the Emperor aswell in a separate PR.

xezon:
That is what the issue report is for. You do not need to discuss Emperor in this Pull. This Pull does not change Emperor stats.

MTKing4:
I do think that it's fine to do the same buff to the Emperor, it's rather identical to Overlord Minus the Vet1 and the free prop which only improve it by a margin

xezon:
You should post it in the linked issue report. This pull does not intent to change the Emperor, because the Emperor is a different unit.

xezon:
Anything to add to this (besides the request to make Emperor drive faster) ?

MTKing4:
None that i can think of

ExiLe-ZH:
Needed buff, nothing to add.

Issue #819, opened 2022-08-05
China Battlemaster is rarely used


xezon:
China Battlemaster is rarely used. It drives slow.

Change that is good for Battlemaster already:
* #412
* #1815

## Original Movement Stats

| Object                              | Speed |
|-------------------------------------|-------|
| China Battlemaster                  | 25    |
| China Battlemaster Upgraded         | 35    |
| Tank China Battlemaster             | 25    |
| Tank China Battlemaster Upgraded    | 35    |
| Nuke China Battlemaster             | 45    |
| China Overlord                      | 20    |
| China Overlord Upgraded             | 30    |
| Tank China Emperor                  | 20    |
| Tank China Emperor Upgraded         | 30    |
| Nuke China Overlord                 | 40    |
| China Dragon Tank                   | 30    |
| China Gattling                      | 40    |
| China Outpost                       | 40    |
| China ECM                           | 40    |
| USA Crusader                        | 30    |
| USA Humvee                          | 60    |

## Original Weapon Stats

DPS based on damage since they all have similar RoF and all
DPU = depleted uranium shells
AL = autoloaders
WGU = weapons grade uranium shells (cut upgrade)

| Object                               | DPS                     |
|--------------------------------------|-------------------------|
| China Battlemaster                   | 60                      |
| China Battlemaster + DPU             | 75                      |
| Tank China Battlemaster              | 60                      |
| Tank China Battlemaster + DPU        | 75                      |
| Tank China Battlemaster + AL         | 120 (40x3)              |
| Tank China Battlemaster + AL + DPU   | 150 (50x3)              |
| Nuke China Battlemaster              | 60 + Radiation          |
| China Overlord                       | 160 ( 80x2)             |
| China Overlord + DPU                 | 200 (100x2)             |
| Tank China Emperor                   | 160 ( 80x2)             |
| Nuke China Overlord                  | 160 ( 80x2) + Radiation |

### Proposal 1

Increase movement speed. Done with
* #1815

### Proposal 2

Increase movement speed with Nuclear Tanks upgrade. Done with
* #1815

### Proposal 3

Add radiation puddle with Uranium Shells upgrade.

### Proposal 4

Reduce cost

xezon:
Looks safe to me to raise movement speed to at least 29 to match China Tank Battlemaster.

commy2:
(fixed an error in OP - Tank Emperor w/o upgrade)

commy2:
[...]

PR #1815, opened 2023-04-08
Change: Increase China Battlemaster speed by 20%, upgraded speed by 14%

xezon:
* Relates to #819
* Relates to #1809

This change increases China Battlemaster speed by 20%, upgraded speed by 14%.

| Object                                     | Speed | Dmg Speed | Upgr. Speed | Upgr. Dmg Speed |
|--------------------------------------------|-------|-----------|-------------|-----------------|
| Original China Nuke Battlemaster           | 45    | 37        | -           | -               |
| Original China Battlemaster                | 25    | 25        | 35          | 32              |
| **Patched China Battlemaster (this)**      | 30    | 25        | 40          | 35              |
| Original China Tank Battlemaster           | 25    | 25        | 35          | 32              |
| **Patched China Tank Battlemaster (this)** | 30    | 25        | 40          | 35              |

# Rationale

The original regular China Battlemaster is considered to be one of the worst units in the game. It drives very slow. The original China Tank Battlemaster is performing considerably better with its free Vet1 or Vet2 promotion, but is also this slow. It drives slower than the Dragon Tank (30). This change increases movement speed of the regular Battlemaster and Tank Battlemaster to elevate it to the same speeds as USA Crusader, Paladin (30). It will still be slower than GLA tanks however (40). After the Nuclear Tanks upgrade, which can be acquired from the Nuke Silo for 2000 cash, it will drive as quick as GLA tanks. This way the Battlemaster is more competitive.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1815_battlemaster_speed.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1815_battlemaster_speed.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-04-08
+
+title: Increases movement speed of China Battlemaster by up to 20%
+
+changes:
+  - tweak: Increases China Battlemaster speed from 25 to 30, upgraded speed 35 to 40. This affects regular China General and Tank General. Movement speed is now competitive with tank speeds of USA and GLA.
+
+labels:
+  - buff
+  - china
+  - controversial
+  - design
+  - major
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1815
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
@@ -694,8 +694,8 @@ End
 ;------------------------------------------------------------------------------
 Locomotor BattleMasterLocomotor
   Surfaces            = GROUND
-  Speed               = 25   ; in dist/sec
-  SpeedDamaged        = 25   ; in dist/sec
+  Speed               = 30   ; in dist/sec ; Patch104p @tweak from 25 (#1815)
+  SpeedDamaged        = 25
   TurnRate            = 180  ;90   ; in degrees/sec
   TurnRateDamaged     = 180  ;60   ; in degrees/sec
   Acceleration        = 1000 ;240  ; in dist/(sec^2)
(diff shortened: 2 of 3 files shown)


MTKing4:
There's still a discrepancy of damage speed, VChina and TankGen tanks have identical Speed and Damaged Speed but only before the upgrade, is that intentional?

xezon:
That is what the original does yes.

MTKing4:
It feels wrong that a damaged unupgraded unit moves fine but starts moving slower after being upgraded

xezon:
I think it is fine. This change preserves that behavior.

xezon:
Should the Tank Battlemaster speed change be removed here? Just make regular China Battlemaster drive faster?

MTKing4:
I say keep it, and future test will dictate if it's too strong to remove, personally I don't see it that way, bms are slow and have many counters, i say the strongest thing about tank gen is their dragon tanks, not bms

ExiLe-ZH:
Agree with change, like MTKing said, can revert if it proves to be too strong.

xezon:
Looking at the speeds of other tanks, I would suggest to bring back damaged speed of unupgraded Battlemaster back to original 25. This way it is consistent with USA Tanks. And the GLA Tanks also have significant speed reduction when damaged, especially the Quad Cannon.

Thoughts?

| Object                                     | Speed | Dmg Speed | Upgr. Speed | Upgr. Dmg Speed |
|--------------------------------------------|-------|-----------|-------------|-----------------|
| Original GLA Scorpion                      | 40    | 30        | -           | -               |
| Original GLA Marauder                      | 40    | 30        | -           | -               |
| Original GLA Quad Cannon                   | 40    | 25        | -           | -               |
| Original China Gattling Tank               | 40    | 40        | -           | -               |
| Original China Dragon Tank                 | 30    | 25        | -           | -               |
| Original USA Crusader Tank                 | 30    | 25        | -           | -               |
| Original USA Paladin Tank                  | 30    | 25        | -           | -               |

xezon:
Updated proposal:

| Object                                     | Speed | Dmg Speed | Upgr. Speed | Upgr. Dmg Speed |
|--------------------------------------------|-------|-----------|-------------|-----------------|
|   Patched China Battlemaster               | 30    | 25        | 40          | 35              |
|   Patched China Tank Battlemaster          | 30    | 25        | 40          | 35              |

ExiLe-ZH:
Yes this sounds fair

MTKing4:
I already suggested this Above so yes i agree

xezon:
Ok. Settings updated. Unupgraded Battlemasters now drive as fast as Crusader, Paladin, Dragon. Looks fair. Documentation added.

PR #1816, opened 2023-04-08
Fix: Add Isotope Stability upgrade icon to China Nuke Overlord Tank

xezon:
This change adds the missing Isotope Stability upgrade icon to the China Nuke Overlord Tank. And it moves the position of the Isotope Stability icon on the Nuke Battlemaster from 4 to 2, so the icons share the same position on both units.

## Patched

shot_20230408_134344_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1816_isotope_stability_cameo.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1816_isotope_stability_cameo.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-04-08
+
+title: Fixes Isotope Stability upgrade icon placements
+
+changes:
+  - fix: Adds the missing Isotope Stability upgrade icon to the China Nuke Overlord tank.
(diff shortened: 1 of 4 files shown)


commy2:
Why all those unnecessary unrelated changes? Those would break custom maps for no reason. You can distribute these upgrade via map, and the weapons are configured to get bonuses.

Ref: https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesOriginalZH/Data/INI/Weapon.ini#L7691
https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesOriginalZH/Data/INI/Weapon.ini#L7775

xezon:
What would break?

xezon:
Changes to Nuke_Upgrade_ChinaWGUraniumShells WeaponBonusUpgrade reverted.

xezon:

PR #1820, opened 2023-04-09
Fix: USA Avenger Target Designator audio loop gaps

xezon:
This change fixes the gaps in the USA Avenger Target Designator fire sound loop.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -492,6 +492,7 @@ Weapon AvengerAirLaserTwo
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @bugfix xezon 09/04/2023 Sets FireSoundLoopTime from 120 to avoid gaps in loops. (#1820)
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
```diff
@@ -506,7 +507,7 @@ Weapon AvengerTargetDesignator
   LaserBoneName       = TurretFX03
   FireFX              = WeaponFX_AvengerTargetDesignator
   FireSound           = AvengerPaintWeaponLoop
-  FireSoundLoopTime = 120                ; loop the firing sound until there's this much delay between shots
+  FireSoundLoopTime   = 200 ; loop the firing sound until there's this much delay between shots
```
xezon, line 510: Should be set to 220 or so.

xezon: 200 appears to be sufficient. Sounds good. Other loops have this a bit higher than respective value of `DelayBetweenShots`, but here 200 likely is ok because it is a multiple of weapon tick 66.66.

xezon:

PR #1821, opened 2023-04-09
Fix: USA Microwave weapon audio loop interval and frequency irregularities

xezon:
This change fixes unpleasant audio loop sounds of Microwave beam.

I was unable to fix the sound files, so I opted to disable 3 of 4 sounds to get rid of the interval and frequency irregularities and maintain a consistent beam sound.


xezon:
Documentation added. Sounds ok.

PR #1822, opened 2023-04-09
Fix: Flawed audio quality in vehicle sounds (2)

xezon:
This change fixes clicking, fade in, fade out in all vehicle sounds.


xezon:
Completed 3 more review passes. Documentation added. Sounds good.

PR #1828, opened 2023-04-10
Fix: USA Alpha Aurora bomb freezing for 1 or 2 frames near ground before hit

xezon:
* Follow up for #1753

This change fixes the USA Alpha Aurora bomb idling 2 frames near ground before hit. It is caused by the large geometry size. Geometry size is now decreased to match that of the regular Aurora bomb.

Additionally the design parameters have been streamlined with the regular Aurora bomb to further decrease setup discrepancies.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1828_alpha_aurora_bomb_impact.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1828_alpha_aurora_bomb_impact.yaml
@@ -1,0 +1,19 @@
+---
+date: 2023-04-10
+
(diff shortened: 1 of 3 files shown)


xezon:
The reduction of size introduces an issue where the bomb model is now slightly clipping through the top of the plane model. This issue is also present with the regular Aurora. It is less visible on the brighter Aurora model however. Perhaps the weapon bone (WeaponA) needs to be moved down a bit on all Aurora models.

xezon:

xezon:
This change does not fix the issue.

PR #1829, opened 2023-04-10
Remove Helix bomb shadows

xezon:
* Follow up for #1750

After submit I realized I forgot to test Helix bomb and realized shadow is useless on these. Removing these again.


MTKing4:
Why is it useless?

xezon:
Because typically the helix shadow will be above it. Or a structure or unit underneath. So in common situations the shadow can not be seen and is therefore a wasted resource.

MTKing4:
But there's a use case for it after the helix moves away, i think it's fine to have plus how much resource is it going to take anyway? It's only used very little and most of the times with 1 or 2 helixs max, and it only lasts a few seconds, surely that won't be a problem whatever

PR #1830, opened 2023-04-10
Fix: Optimize drop location of USA Daisy Cutter bomb

xezon:
This change optimizes the drop location of the USA Daisy Cutter bomb. It no longer falls as much behind the intended target when flying in vertically. And drops a bit before the target when flying in straight. ~~The distance matches the one of the MOAB.~~ The drop position is optimized for fix TheAssemblyArmada/Thyme#752.


xezon:
I figured the fix TheAssemblyArmada/Thyme#752 will fix the Special Power position discrepancies, so I will optimize the new drop position for this fix. With that, the optimal drop distance is 180 in this case.

xezon:
~~The MOAB distance of 160 is already perfect. Needs no change.~~

Edit: I am mistaken. Small optimization with #1831 makes it more accurate.

shot_20230410_180249_3

xezon:
Airforce Carpet Bomber also needs no change. Very balanced.

shot_20230410_180150_2

PR #1832, opened 2023-04-10
Fix: Optimize drop location of China Cluster Mines

xezon:
This change optimizes the drop location of the China Cluster Mines. It now falls at the very center of the target location. The drop location in practice makes no difference as to where the mines spawn, so this effectively is a small buff for China, because the plane eventually draws less enemy fire before the bomb drop. To compensate for the decreased bomb drop time, the explosion height has been roughly halved from 60 to 35, which also looks better in practice. The drop position is optimized for fix https://github.com/TheAssemblyArmada/Thyme/pull/752.

## Patched

Moment of explosion (with TargetHeight 30 instead of 35).

shot_20230410_190015_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -1391,7 +1391,7 @@ Object ClusterMinesBomb
 
 
   Behavior = HeightDieUpdate ModuleTag_06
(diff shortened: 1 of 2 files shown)


xezon:
TargetHeight updated from previously 30 to 35 to help it explode it even more in center. Now looks very accurate right above the flag.

commy2:
Have you tested this with plane moving in diagonally vs straight? That changes the speed of plane and at least for carpet bombs, where they land drastically.

PR #1836, opened 2023-04-11
Fix: Remove SELECTABLE flag from SalvageCrate

xezon:
This change was reverted by #1997

~~This change removes the SELECTABLE flag from SalvageCrate. Oddly the GLA faction cannot select the crate anyway, but the USA and China factions can if the SalvageCrate would show for them. Being able to select the crate means it is no longer possible to properly move a unit to it, because the terrain underneath cannot be clicked.~~


xezon:

PR #1859, opened 2023-04-19
Change: Optimize button placements of Boss Command Center

xezon:
This change optimizes button placements of Boss Command Center.

* Boss and China now share same button position for Carpet Bomber
* Boss and China now share same button position for Cluster Mines
* Boss and China now share same button position for Artillery Barrage
* Boss and China now share same button position for EMP Pulse
* Spectre and Sneak Attack buttons were moved

No image because unclear how to test.

### TODO

- [x] Move China Radar back to position 11

## Patched Boss Command Center Buttons

shot_20230422_140220_1

## Patched China Command Center Buttons (Reference Only)

shot_20230422_140307_2

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1859_boss_commandcenter_buttons.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1859_boss_commandcenter_buttons.yaml
@@ -1,0 +1,25 @@
+---
+date: 2023-04-19
+
+title: Optimizes button placements of Boss Command Center
+
+changes:
+  - tweak: |
+      Optimizes button placements of Boss Command Center
+        - Boss and China now share same button position for Carpet Bomber
+        - Boss and China now share same button position for Cluster Mines
+        - Boss and China now share same button position for Artillery Barrage
+        - Boss and China now share same button position for EMP Pulse
+        - Spectre and Sneak Attack buttons are moved
+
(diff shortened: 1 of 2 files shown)


commy2:
Command_UpgradeChinaRadar should be moved from 9->11 (top right corner, tucked in with Rallypoint and Landmines). That is where it is in CCG, but they forgot to move it in ZH along with all the other right aligned buttons when expanding the control bar from 12 to 14 buttons, probably because it is a one of button unlike Stop, Sell, Guard etc. 

It is similar to how they forgot to move Air Guard on (King) Raptors, but remembered for MiGs.

xezon:
So we ok with moving China Radar button by one position?

kABUSE1:
I believe above 99% of players wouldn't know from memory in what slot the radar upgrade even is and the <1% who do know use the hotkey to purchase it lol

xezon:
Ok then I will create another change for moving Radar position first.

xezon:
Radar upgrade moved back to original position. Documentation added.

xezon:
Comparison images added to opening post. Looks very good.

commy2:
I think those "@tweak from X" meassges here are excessive. You can always look it up where they were, or look up git history.

xezon:
Looks good.

PR #1861, opened 2023-04-21
Change: Move Radar upgrade button above Mines upgrade button on China Command Center

xezon:
This change moves the Radar upgrade button one field to the right, above the Mines upgrade button on the China Command Center. It is now consistent with the placement on the Boss Command Center.

## Patched

shot_20230421_210806_1


xezon:
Documentation added.

commy2 (approved):

Nevermore2790:
Apparently, the current iteration, is causing double Radar upgrades for those with the 14 Command Bars on all Chinese Command Centers.

PR #1862, opened 2023-04-22
Change: Add Starting Money options for 30000 and 40000

xezon:
This change adds Starting Money options for 30000 and 40000, because the original gap between 20000 and 50000 is quite wide.

shot_20230422_132735_1


xezon:

PR #1863, opened 2023-04-22
Fix: Boss Dozer button image and text

xezon:
This change fixes Boss Dozer button image and text. It no longer shows the China Dozer picture and China specific tooltip text.

## Patched Boss Dozer GUI

shot_20230422_144509_1


xezon:

PR #1864, opened 2023-04-22
Fix: Set unique default house color for Boss General

xezon:
This change sets a more appropriate default color for Boss General. Was green, is now gold, matching the red/yellow China GUI theme of Boss General better.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1864_boss_color.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1864_boss_color.yaml
@@ -1,0 +1,18 @@
+---
+date: 2023-04-22
+
+title: Sets unique default house color for Boss General
+
+changes:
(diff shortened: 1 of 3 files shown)


commy2:
Where does this color appear?

xezon:
I suspect this is default color in World Builder faction. Not sure.

MTKing4:
Maybe give it a distinctive 4th color? So that Chinese boss units can be distinguished from actual Chinese units like overlords

I think Gold is a good choice for this

xezon:
We can go with Yellow yes. Orange would also work.

MTKing4:
I chose Yellow because it's a primary color, same as Red, blue and green

xezon:
Boss default house color is now gold. Documentation added.

xezon:
Documentation fixed. Looks good.

PR #1865, opened 2023-04-22
Change: Optimize button placements of Boss War Factory

xezon:
This change optimizes button placements of Boss War Factory. The units are now located at more familiar positions, with all China units taking China positions and GLA and USA buttons being close to where they are with the respective original faction.

## Original

shot_20230422_173224_2

## Patched

shot_20230422_180807_1


xezon:

PR #1061, opened 2022-09-01
Fix: Improve English sound quality by removing hum, noise, hiss and crackling

xezon:
**Merge with Rebase**

* Fixes #1058

This change fixes humming, noises, hissing and crackling for all Unit and EVA sounds. It uses 72 MB uncompressed and 58 MB 7-zip compressed. Due the large size requirement it is therefore unfit for Patch Core and will be optional content only.

The sounds are bundled into the following archives:
* OptionalAudioArabic
* OptionalAudioBrazilian
* OptionalAudioChinese
* OptionalAudioEnglish
* OptionalAudioFrench (uses just a few of the sounds)
* OptionalAudioItalian
* OptionalAudioKorean
* OptionalAudioPolish
* OptionalAudioSpanish

# Affected units

## USA

* Ranger
* Laser Ranger
* Missile Defender
* Pathfinder
* Burton
* Dozer
* Humvee
* Ambulance
* Crusader
* Microwave
* Combat Chinook
* Comanche
* Raptor
* Stealth Fighter
* Aurora
* Spectre Gunship

## China

* Red Guard
* Minigunner
* China Lotus
* Hacker
* Dozer
* ECM
* Troop Crawler
* Listening Outpost
* Inferno Cannon
* Nuke Cannon
* Overlord
* Helix

## GLA

* Worker
* GLA Rebel
* GLA Tunnel Defender
* Hijacker
* Jarmen
* Saboteur
* Terror
* Car Bomb
* Cycle
* Toxin Tractor
* Technical
* Quad
* Scorpion
* Radar Van
* Rocket Buggy
* Bomb Truck
* SCUD
* Marauder
* POW

PR #1867, opened 2023-04-23
Fix: Flawed audio quality in gps scrambler sound

xezon:
* Follow up for #1061

This change removes the noise at the end of the gps scrambler audio effect.


xezon:

PR #1871, opened 2023-04-23
Fix: Improve USA Alpha Aurora bomb explosion effects

xezon:
* Closes #1751

This is an alternative implementation for fixing and improving Alpha Aurora bomb effects. The main issues were already raised in #1751. This change addresses additional problems and avoids the down sides of #1751.

* The initial Aurora bomb hit now shows the hit particle effects and sound. Is consistent with regular Aurora bomb hit effects.
* The initial Aurora bomb hit no longer draws a scorch, because it is covered by the particles at the time of the impact and would be overdrawn soon after by the bigger scorch for the gas explosion. Total explosion scorch count is 1 (optimal).
* The ignite sound effect is preserved, but is made a bit quieter than the one of the Daisy Cutter ignition, because the sparks are smaller and there is now the other explosion sound triggered at the same time.
* The first ignition sparks are now visible earlier and spread further apart.
* The second ignition sparks are now less in numbers and spread further apart.
* The second ignition sparks now properly fade out.
* The second ignition sparks now have random size.
* The lingering green gas clouds are removed.
* The final gas explosion is now just one explosion instance instead of 2 to 4.
* The final gas explosion is now synced better with the actual damage and sound event.
* The final gas explosion now has much less random size variation, so it never looks too small.
* The original Alpha bomb shock wave effect is now a smaller copy of the Daisy Cutter shock wave.

## Patched (Before)

https://user-images.githubusercontent.com/4720891/233851401-7742e212-5f8c-4e8c-87a5-cea6485700a9.mp4

## Patched (This)

https://user-images.githubusercontent.com/4720891/233851416-60f58c28-89f3-40da-8b48-148e0821dc04.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1871_alpha_aurora_bomb_effects.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1871_alpha_aurora_bomb_effects.yaml
@@ -1,0 +1,35 @@
+---
+date: 2023-04-23
+
+title: Fixes and improves explosion effects of USA Alpha Aurora bomb
+
+changes:
+  - tweak: Fixes and improves the USA Alpha Aurora bomb explosion effects.
+
+subchanges:
+  - tweak: The initial Aurora bomb hit now shows the hit particle effects and sound. Is consistent with regular Aurora bomb hit effects.
+  - tweak: The initial Aurora bomb hit no longer draws a scorch, because it is covered by the particles at the time of the impact and would be overdrawn soon after by the bigger scorch for the gas explosion. Total explosion scorch count is 1 (optimal).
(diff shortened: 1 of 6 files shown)


xezon:
I see a problem. AirF_AuroraBombScatterIgnite2 is used in FXList Neutron_WeaponFX_NukeCannon. So that needs a copy.

~~And for ignite sound, perhaps we can just use the original daisy cutter sound without aurora bomb explosion sound. That way we would remove one sound from playback, aka less audio demand.~~

xezon:
Added size variation for second ignition sparks like we did for Fuel Bomb and Neutron sparks. And reduced the previously added spreading of ignition sparks a bit. Documentation added. Looks good.

PR #1872, opened 2023-04-24
Fix: Improve China Neutron Shell explosion effects

xezon:
* Relates to #1871

This change fixes and improves the China Neutron Shell explosion effects. It no longer shows the explosion effect of the Alpha Aurora and instead just shows the Neutron specific effects and the scattered flares in matching colors and sizes.

* Removes the Alpha Aurora fuel bomb explosion and smoke
* Leaves the scattered flares with matching colors and sizes
* Moves all Neutron particle effects closer to the ground where the shell actually hits
* Removes one of the dust rings because there are 2 of them moving at different speeds
* Halves the duration of the remaining dust ring effect
* Increases speed and reduces lifetime of the neutron rings
* Gives the outer neutron ring a fixed size to better indicate the weapon effect radius
* Increases the size growth of the centered neutron flare ball to look a bit more impressive
* Removes the unnecessary terrain scorch

## Original particles

https://user-images.githubusercontent.com/4720891/234041830-c5ccad90-5126-457f-9f53-35ebea934c56.mp4

## Patched particles

https://user-images.githubusercontent.com/4720891/234048733-6699f991-f62e-42fb-99ae-345642b3ad66.mp4


xezon:

ExiLe-ZH:
Looks much better than original! 

It fits the type of weaponry and it looks unique now, not like a weird silent Alpha Aurora bomb copy paste.

Maybe we can do something about the sound aswell? The breaking glass sounds so.. cheesy?

MTKing4:
The "breaking Glass" sound is actually for the unman action, you can also hear it with the jarmen snipe, it's not strict for nuke cannon

ExiLe-ZH:
Yea true, is it extra loud when multiple units get 'sniped' by a neutron?

xezon:
I suspect it is this

```
AudioEvent VehicleSnipeVictim
  Sounds      = gvehsnia
  Priority    = high
  VolumeShift = -15
  Volume      = 70
  Type        = world shrouded everyone
End
```

No limit.

PR #1884, opened 2023-04-29
Change: Move Radar upgrade button of China Command Center in optional bundle only

xezon:
* Follow up for #1861

This change moves the Radar upgrade button position for optional bundle only. Radar button for both Boss and China will take original China Radar button position. This way legacy players are not confused by new Radar button position.


xezon:
Documentation added and updated.

xezon:
Tested and worked.

PR #1887, opened 2023-04-30
Fix: Add Stop command button to all worker units

xezon:
* Split off of #1579

This change adds the Stop command button to all worker units (except Boss). This way button can be pressed to stop worker units, including in group selection with other units.

To put the Stop button at its expected position, the Clear Mines, GLA Fake Structures and USA Airfield buttons had to be moved.

This change applies to the optional Patch bundle only.

### Advantages

* Stop button added
* Clear Mines button now at same Position as Guard Mode button
* USA Airfield button is closer to other structures, and in same row as China Airfield

### Disadvantages

* Clear Mines, GLA Fake Structures, USA Airfield buttons moved
* GLA Fake Structures button is closer to Arms Dealer button

## Patched, Before

shot_20230121_131853_1

shot_20230121_131859_3

shot_20230121_131902_4

## Patched, After (This)

shot_20230121_133026_2

shot_20230121_133030_3

shot_20230121_133037_4

All regular worker units selected.

shot_20230121_132855_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1887_dozer_worker_stop_button.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1887_dozer_worker_stop_button.yaml
@@ -1,0 +1,23 @@
+---
+date: 2023-04-30
+
+title: Adds missing Stop command button to all worker units
+
+changes:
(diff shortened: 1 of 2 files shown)


xezon:

PR #1888, opened 2023-04-30
Change: Optimize button placements of China Dozer

xezon:
* Split off of #1579
* Follow up for #1887

This change optimizes button placements of China Dozer. Swaps the positions of Airfield, Propaganda Center and Internet Center to achieve consistency with button placements of USA Dozer and GLA Worker. This applies to the optional Patch bundle only.

Advantages:

Disadvantages:
* Buttons moved around

## Patched, Before

shot_20230121_133030_3

## Patched, After (This)

shot_20230128_095927_2


xezon:

PR #1889, opened 2023-04-30
Change: Optimize button placements of USA Dozer

xezon:
* Split off of #1579
* Follow up for #1887

This change optimizes button placements of USA Dozer. Swaps the positions of Airfield, Strategy Center, Supply Drop Zone, Particle Cannon, Command Center to achieve consistency with button placements of China Dozer and GLA Worker. This applies to the optional Patch bundle only.

Advantages:

Disadvantages:
* Buttons moved around

## Patched, Before

shot_20230121_133037_4

## Patched, After (This)

shot_20230430_111722_1


xezon:

PR #1890, opened 2023-04-30
Change: Optimize Combat Drop button placement of USA Chinook

xezon:
This change optimizes the Combat Drop  button placement of the USA Chinook. It is moved one position to the right, according to the draft shown in the linked discussion. Optional content only.

* https://github.com/TheSuperHackers/GeneralsGamePatch/discussions/1543#discussioncomment-4685730

## Patched, Optional
shot_20230430_123718_1


MTKing4:
This doesn't make sense and it just looks weird now, the rationale in the linked discussion is weak, it looked better when it was below the evacuate button, and it's not strange to have an ability near the slots, look at the helix, also it's more of an alternative type of evacuate than an ability, so it makes sense to be near the slots

xezon:
I do not think this change is necessary either.

PR #1891, opened 2023-04-30
Fix: Keep disabled Overlord attachment upgrade buttons in Command Set

xezon:
* Relates to #1542

This change simplifies China Overlord and Emperor attachment Command Set. Buttons will no longer disappear after researching the Gattling Cannon or Propaganda Tower upgrades. They will still disappear when researching Bunker upgrade however, because that claims the upgrade button positions with passenger buttons.

With this change the upgrade buttons can be used on multi Overlord selection to build an upgrade any time, as long as one of the selected Overlords does not own the Bunker upgrade.

This behavior is consistent with USA vehicle drone upgrade buttons and Helix upgrade buttons (as of #1542).

shot_20230430_163340_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1891_overlord_attachment_upgrade_buttons.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1891_overlord_attachment_upgrade_buttons.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-04-30
+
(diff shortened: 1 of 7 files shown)


MTKing4:
Maybe it would have been better to put them where there are no slots just like with helix, so that they all can multi upgraded

xezon:

PR #1893, opened 2023-04-30
Change: Add Countermeasures armor upgrade bonus to Boss Raptor

xezon:
This change adds the Countermeasures armor upgrade bonus to the Boss Raptor. The Boss Aurora and Spectre already have it.


xezon:

PR #1894, opened 2023-05-01
Change: Move China Overlord evacuation button to common position

xezon:
* Follow up for #165

This change moves the China Overlord evacuation button to the common evacuation position. This allows to evacuate the Overlord with other transport units in a group selection.

shot_20230501_115208_2


xezon:

PR #1895, opened 2023-05-01
Change: Fixes order of China Overlord and Helix attachment cameo icons

xezon:
This change fixes the order of the China Overlord and Helix attachment cameo icons.

| Object                  | Cameo left           | Cameo mid        | Cameo right          |
|-------------------------|----------------------|------------------|----------------------|
| Original Emperor        | Subliminal Messaging | Propaganda Tower | Gattling Cannon      |
| Original Overlord       | Propaganda Tower     | Gattling Cannon  | Battle Bunker        |
| Original Helix          | Battle Bunker        | Propaganda Tower | Gattling Cannon      |
| Original Infantry Helix |                      |                  | Battle Bunker        |
| **Patched Emperor (this)**  | Gattling Cannon      | Propaganda Tower | Subliminal Messaging |
| **Patched Overlord (this)** | Gattling Cannon      | Propaganda Tower | Battle Bunker        |
| **Patched Helix (this)**    | Gattling Cannon      | Propaganda Tower | Battle Bunker        |

## Patched

shot_20230501_120159_1

shot_20230501_120205_2

PR #1896, opened 2023-05-01
Change: Improve China Overlord upgrade command button and cameo icon placements

xezon:
* Closes #1544
* Closes #1895
* Follow up for #1891

This change improves China Overlord upgrade command button and cameo icon placements.

### Advantages

* The Overlord upgrade cameo icon positions and order now matches the Helix, Emperor and the patched upgrade button order.
* The Overlord upgrade command button positions and order now matches Helix button positions and order.
* The Overlord will keep the disabled upgrade buttons after acquiring the Bunker upgrade.
* The Overlord and Emperor can now build the Gattling Cannon upgrade in a group selection.
* Helix buttons and cameo icons are unchanged

### Disadvantages

* The Overlord upgrade button placement and order has changed.
* The Overlord upgrade button placement looks a bit oddly placed unless the Bunker upgrade is built.

### Upgrade buttons

| Object                  | Upgrade left         | Upgrade mid      | Upgrade right        |
|-------------------------|----------------------|------------------|----------------------|
| Original Emperor        |                      | Gattling Cannon  |                      |
| Original Overlord       | Battle Bunker        | Gattling Cannon  | Propaganda Tower     |
| Original Helix          | Battle Bunker        | Propaganda Tower | Gattling Cannon      |
| Original Infantry Helix |                      |                  | Battle Bunker        |
| **Patched Overlord (this)** | Battle Bunker        | Propaganda Tower | Gattling Cannon      |
| **Patched Emperor (this)**  |                      |                  | Gattling Cannon      |

### Upgrade cameo icons

| Object                  | Cameo left           | Cameo mid        | Cameo right          |
|-------------------------|----------------------|------------------|----------------------|
| Original Emperor        | Subliminal Messaging | Propaganda Tower | Gattling Cannon      |
| Original Overlord       | Propaganda Tower     | Gattling Cannon  | Battle Bunker        |
| Original Helix          | Battle Bunker        | Propaganda Tower | Gattling Cannon      |
| Original Infantry Helix |                      |                  | Battle Bunker        |
| **Patched Overlord (this)** | Battle Bunker        | Propaganda Tower | Gattling Cannon      |

## Patched

shot_20230501_155848_4

shot_20230501_155826_3

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1891_overlord_attachment_upgrade_buttons.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1891_overlord_attachment_upgrade_buttons.yaml
@@ -1,10 +1,10 @@
 ---
 date: 2023-04-30
 
-title: Fixes issue where China Overlord attachment upgrade buttons disappear after research
+title: Fixes upgrade buttons of China Overlord and Emperor disappearing after research
 
 changes:
-  - fix: China Overlord upgrade buttons will no longer disappear after researching the Gattling Cannon or Propaganda Tower upgrades. They will still disappear when researching Bunker upgrade however, because that claims the upgrade button positions with passenger buttons. The upgrade buttons can now be used on multi Overlord selection to build an upgrade any time, as long as one of the selected Overlords does not own the Bunker upgrade.
+  - fix: The China Overlord and Emperor upgrade buttons will no longer disappear after researching the Gattling Cannon, Propaganda Tower and Battle Bunker upgrade. The upgrade buttons can now be used in a group selection even if some selected Overlords already own upgrades.
 
 labels:
   - bug
@@ -15,6 +15,7 @@ labels:
 
 links:
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1891
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1896
 
 authors:
   - xezon
--- a/Patch104pZH/Design/Changes/v1.0/1896_overlord_emperor_gattling_cannon_button.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1896_overlord_emperor_gattling_cannon_button.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-05-01
+
+title: Improves Gattling Cannon upgrade button of China Overlord and Emperor to be usable in group selections
+
+changes:
+  - tweak: The Gattling Cannon upgrade of the China Overlord and Emperor can now be acquired in group selections.
+
+labels:
+  - china
+  - enhancement
+  - gui
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1896
+
+authors:
+  - xezon
--- a/Patch104pZH/Design/Changes/v1.0/1896_overlord_upgrade_buttons_and_icons.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1896_overlord_upgrade_buttons_and_icons.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-05-01
+
+title: Improves upgrade command button and upgrade cameo icon placements of China Overlord and Emperor
+
+changes:
+  - tweak: Matches the position and order of the China Overlord and Emperor upgrade command button with those of the China Helix.
+  - fix: Matches the order of the China Overlord and Emperor upgrade cameo icons with the order of the command buttons.
+
+labels:
+  - china
+  - gui
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1896
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -1259,7 +1259,6 @@ CommandButton Command_UpgradeChinaOverlordGattlingCannon
   UnitSpecificSound       = OverlordTankVoiceModeGattling
 End
 
-;patch104p-optional-begin
 ; Patch104p @tweak xezon 01/05/2023 Adds a new button variant for the Gattling Cannon upgrade that works for both Overlord and Emperor.
 ;   This allows the button to be used in Overlord and Emperor group selections.
 CommandButton Multi_Command_UpgradeChinaOverlordGattlingCannon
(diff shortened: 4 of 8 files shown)


xezon:
I think this is good solution as it does not require touching Helix upgrade icons and cameo icons to streamline it.

MTKing4:
This increases functionality, so it shouldn't be optional

xezon:
If we change core buttons, then we confuse legacy players. So this must be optional.

MTKing4:
Missing out on a functionality has bigger impact than a small confusion, besides it won't be confusing at all, we didn't change the buttons order, it would be very clear, i don't think anyone would disagree if they knew why we did it

We've agreed in the past that only subjective and art/non gameplay improvements should be optional, it's not ok that optional players will have an advantage over core ones with changes like this

xezon:
Is not clear to me what you are on about.

If player likes legacy buttons, then he uses core patch buttons.

If player likes optimized buttons, then he uses optional patch buttons.

There is no downside, no discrimination. Players have the freedom to choose.

MTKing4:
It's as simple as this, if Players choose core patch, they will miss out on being able to Multi purchase Upgrades when having Bunkerlord Among the group Selection

This is no different than the streamlining of evacuate buttons

xezon:
> If Players choose core patch, they will miss out on being able to Multi purchase Upgrades when having Bunkerlord Among the group Selection.

Yes. It is their choice. What is the problem? There is a change log. All changes are transparent.

MTKing4:
It's not much of a choice then, players are then forced to use optional just for increased functionality.

Question, is Evacuate streamlining optional?

xezon:
~~Not yet, but I am considering making Evacuate button changes mostly optional, so that legacy players do not complain about changed button positions.~~ The main goal of core button layout is to avoid too much discomfort when transitioning from legacy game to new game, which is a very reasonable goal, because adoption comfort is very important.

MTKing4:
I don't like that route, there will be a big discrepancy between core and optional and then it core players either switch to optional bundle or be at a huge disadvantage, i think that would upset the players far more than a few button order changes as they'll feel they have no real choice there.

I don't know about the others but I'm fine with sacrificing trivial muscle Memory for increased functionality, what I'm not fine with is changing them with no functionality gain, like the new Dozer buildings order, and that can be optional, but this can not

xezon:
I checked evacuation button changes and I think there needs to be nothing optional, apart from Helix button change.

MTKing4:
Yes and nor should this

xezon:
Exile
> I think it is fine

kABUSE
> It makes absolutely no sense to make something that directly affects (and improves) gameplay optional and I have a hard time to follow the reasoning. I don't think the case "OMG the overlord keeps his upgrade command buttons visible even after you purchased an upgrade, I UNINSTALL THE PATCH NOW" exists. The order in which the command buttons are displayed is even more negligible. I personally could not for the love of god tell in which order the command buttons are currently displayed. I really have no idea why somebody would consider making this optional.

kABUSE1:
The most problematic part about this are custom scripted maps.

Sometimes scripters remove command buttons from units like dozers and put a different command button into the same slot position (by fixed numeric index). If now you move buttons around, this might mess up with the intended logic of the maps by putting custom command buttons above existing command buttons.

It's all good if some maps need to be re-worked for the patch. But it's impossible to accustom for the fact that people will have different command set versions by individual preference.

xezon:
Perhaps can just set all buttons for a given command set in a custom map?

Or reference a new command set in the object?

kABUSE1:
> 
Speaking entirely on scripting level:
The "fix" would be to remove all command buttons that are potentially in a different (custom) order, then assign different command buttons in these slots in fixed order.

You cannot assign command-sets on script-level, you can only modify default command-sets per type. With scripts you have exactly two options:
1. Remove specific command button from unit-type specific default command set, regardless of which slot it is in
2. Assign specific command button in slot X of unit type's default command set (numeric slot 01 - 18)

xezon:
Essentially this would already be a problem, because users can already play with custom Command Sets, because game does not CRC checks it and game clients are compatible with different Command Sets.

So the robust way would be to remove all Command Buttons in Script, and then assign a clean set of Command Buttons always to ensure the Command Set looks as expected. Is that feasible?

kABUSE1:
Of course. This would be the way I do it. But the average mapper would not have this in mind when working on command buttons.

My point being, I would not make matters worse. Making this optional would make matters worse.

Right now custom command button orders in ZH 1.04 are probably insanely rare. We are talking below 1% rare. But if you offer this as an optional feature it would become a common problem.

xezon:
Judging by comments and interactions

- 3 persons want this in Core
- 2 persons are fine with either Core or Optional
- 1 persons against Core

xezon:
We can also add Bunker and Propaganda Buttons to Emperor. These will be disabled, but can then be used in group selection with other Overlords:

shot_20230523_192810_1

MTKing4:
I don't know about that last part, it could cause confusion seeing a bunker upgrade on an emperor giving the implication that it has one installed, the way i see it is emperor and overlord are two different units so they don't  really need to be multi-upgraded.

That doesn't mean I'm necessarily against this change.

xezon:
Yes it is not ideal.

kABUSE1:
> Yes it is not ideal.

You can move the disabled bunker and propa buttons to slot 17 and 18 for emporor overlords. This way they are never visible for human players.

xezon:
> You can move the disabled bunker and propa buttons to slot 17 and 18 for emporor overlords. This way they are never visible for human players.

Buttons must take same slot to work in group selection.

kABUSE1:
> > You can move the disabled bunker and propa buttons to slot 17 and 18 for emporor overlords. This way they are never visible for human players.
> 
> Buttons must take same slot to work in group selection.

In that case both variants are not ideal.

I'd say the use-case of a player owning emperors and overlords at the same time AND having the need to equip overlords with something other than gattlings is so rare that it's acceptable that he has to do some manual group selection.

In such a case I would expect the player to only want gat upgrade, as his emperors already have propaganda, which should suffice for the mixture. And seriously, at that point, who would bother with bunker upgrades :D

commy2:
Also, Emperor has a dummy upgrade button for Speaker Tower (as it spawns with one), so it makes no sense to both have the upgrade highlighted, but the button disabled.

xezon:
Changes promoted to core. Opening text updated. Please confirm satisfaction.

MTKing4:
Satisfied.

xezon:
Documentation added and updated. Looks good.

PR #1897, opened 2023-05-01
Change: Swap Drop Bomb and Evacuate buttons on China Helix command set

xezon:
This change swaps the Drop Bomb and Evacuate buttons on China Helix command set. This way the Evacuate button shares the common position and it can be evacuated together with other vehicles. Optional only.

### Advantages

* Helix Evacuation button now works in group selection with all other vehicles, such as the USA Chinook.
* Helix Drop Napalm Bomb button can still be used with Regular, Infantry and Tank Helix in group selection.
* Button layout still looks very close to the original, because just 2 buttons swapped.
* Optional only, core Helix button layout is unchanged

### Disadvantages

* Buttons are moved in optional bundle.
* Napalm bombs are visually further apart which looks odd on first sight. This can be improved by changing the visuals of the buttons.

## Patched Regular Helix

shot_20230501_172249_2

## Patched Infantry Helix

shot_20230501_172245_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1897_helix_evacuation_button_placement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1897_helix_evacuation_button_placement.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-05-01
+
+title: Swaps Drop Bomb and Evacuate buttons on China Helix command set
+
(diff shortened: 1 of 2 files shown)


xezon:
Documentation added. Looks ok.

PR #1899, opened 2023-05-03
Change: Add new cursor for ground force attack

xezon:
This change adds a new cursor for ground force attack. This way there is a clear distinction between force attack object and force attack ground, which are conceptually different, because a ground attack does not follow an object.

Original attack cursor:

sccattack

Patched ground force attack cursor:

SCCAttack3


ExiLe-ZH:
Good little improvement I think, why not?

xezon:
The only problem I have that this perhaps is too little of a distinction for color blind people. Shades of Red and Orange may look very similar. I tried to decorate the arrows differently on top of the color change, but did not come up with a good variant.

MTKing4:
I do think a design change is better than a color change, i mean this color can be difficult to tell apart even for normal sighted people at certain angles and different monitor types

xezon:
Options:
* Make arrows black border thicker
* Spawn a small growing ring from the center when the arrows meet
* Draw 4 long vertical lines from the center when the arrows meet

ExiLe-ZH:
perhaps rotate cursor 45 degrees?

xezon:
> perhaps rotate cursor 45 degrees?

Then it would look more similar to the green move cursor, which is vertical like that. I do not think that fits.

ExiLe-ZH:
Or just make the orange slightly brighter for better contrast with the red one.

MTKing4:
> Options:
> 
> * Make arrows black border thicker
> * Spawn a small growing ring from the center when the arrows meet
> * Draw 4 long vertical lines from the center when the arrows meet

Second Option Sounds nice, could be just a static one, no need to regrow, think it would look better that way

4 Vertical lines? How's that supposed to look?

xezon:
> 4 Vertical lines? How's that supposed to look?

Here is a potential candidate. Basically the yellow impact thingy in the middle expands further away from the center, so it looks more distinct from the original, but is overall still the same cursor style.

SCCAttack7 SCCAttack7

Reference cursors:

SCCAttack1 sccattack

SCCAttack3 SCCAttack3

xezon:
I tested SCCAttack7 in game and it looked nice. Looks sufficiently different I think. Change updated.

MTKing4:
I was confused by the fact you said vertical, but i guess you meant diagonal :P

Looks good

xezon:
Fix TheSuperHackers/GeneralsModBuilder#30 before merge.

xezon:
TheSuperHackers/GeneralsModBuilder#30 has been fixed. Cursor install and uninstall works properly. Documentation added. Looks good.

PR #1900, opened 2023-05-03
Fix: Set audio limit for Vehicle Snipe

xezon:
* Follow up for #1872

This change sets an audio limit of 2 for Vehicle Snipe. This way it can no longer spam many sounds on Neutron Shell hits.


xezon:

PR #1902, opened 2023-05-04
Fix: Prevent shells of China Artillery Barrage from killing each other

xezon:
* Relates to TheSuperHackers/GeneralsGameCode#210

This change prevent shells of the China Artillery Barrage from killing each other by increasing their health to 400. This fix potentially avoids scenarios where 1 or 2 Oil Derricks survive a Level 3 Artillery Barrage at random.

400 health was chosen because with that the Artillery deals reliably max damage on a single point attack in a test scenario. See TheSuperHackers/GeneralsGameCode#210 for more details.

(**health** refers to Artillery shell health)

Untitled-1

400 health is still low enough to be destructible by Super Weapons, such as GLA SCUD Storm with 500 damage per missile or China Nuke Missile with 700 damage per blast.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -5467,9 +5467,11 @@ Object ChinaArtilleryBarrageShell
 
   ; *** ENGINEERING Parameters ***
   KindOf            = PRELOAD CAN_CAST_REFLECTIONS PROJECTILE ;SMALL_MISSILE
+
+  ; Patch104p @bugfix xezon 04/05/2023 Change health from 100 to prevent shells eventually killing each other on impact at random.
   Body = ActiveBody ModuleTag_02
-    MaxHealth       = 100.0
-    InitialHealth   = 100.0
+    MaxHealth       = 400.0
+    InitialHealth   = 400.0
   End
 
 ; ---- begin Projectile death behaviors


ExiLe-ZH:
Should it be destructible at all though? Under most circumstances it's by friendly fire.

xezon:
Carpet Bomb deals 300 damage. We could make Artillery shells practically indestructible if so desired.

MTKing4:
We could compare it to A10 Strike, are A10 Missiles Destructible?

ExiLe-ZH:
> We could compare it to A10 Strike, are A10 Missiles Destructible?

Avengers can stop A10 missiles

MTKing4:
that is true, in that case, a 400hp seems a solid solution

ExiLe-ZH:
A10 rockets are small though, that's why they're counterable? Whereas shells are normally never counterable.

Like it only creates inconsistency when combining powers (with allies), which can lead to frustration if the target building doesn't die while it should have. That's the only scenario where it can happen really, you're not gonna Nuke yourself to counter an artillery strike lol.

xezon:
Another thing we could add on top is applying the shells damage on destruction of said shells. This way if it is killed, then it will still deal its damage at the height it was killed at. Currently it just evaporates.

xezon:
Jundiyy
> I still would like Artillery Shells to have health and be destructible. A player can and does destroy his own Shells, it's something he has to calculate.

commy2:
Instead of doctoring with shell health, one could add NOT_SIMILAR flag to RadiusDamageAffects token of ArtilleryBarrageDamageWeapon. A10 missiles use this flag for example and do not seem to suffer from this issue. This would keep shell destruction, but not shell destruction from other Artillery Barrage shells.

ExiLe-ZH:
> Jundiyy
> 
> > I still would like Artillery Shells to have health and be destructible. A player can and does destroy his own Shells, it's something he has to calculate.

I feel we've this discussion many times with many different subjects. It comes down to involving the player to either be rewarded for paying attention (and time stuff well) or to punish him for not doing so. Generally I think we agreed the prior is better.

Ofcourse this also depends on logic. For instance it should be fairly obvious a Carpet Bomber can be shot out of the sky with a Nuclear Missile, there's clear visual feedback for this happening. Is this the case for Artillery Strikes aswell? Or will it only cause confusion when an enemy structure survives?

PR #1905, opened 2023-05-05
Fix: Prevent shells of China Artillery Barrage from killing each other

xezon:
* Relates to TheSuperHackers/GeneralsGameCode#210
* Closes TheSuperHackers/GeneralsGamePatch#1902

This change prevents shells of the China Artillery Barrage from killing each other by excluding damaging others of their kind. Therefore random damage discrepancies on target structures are reduced. This fix potentially avoids scenarios where 1 or 2 Oil Derricks survive a Level 3 Artillery Barrage at random.


xezon:
Tested and worked.

shot_20230505_191711_3

commy2 (approved):

xezon:

Issue #176, opened 2021-09-04
Investigate disabled audio lines in Voice.ini


xezon:
There appears to be a number of unused speech files in Voice.ini. Investigate whether or not some of these could be resurrected.

Stubbjax:
Here's my list of changes that don't affect online compatability:

AudioEvent WorkerVoiceUpgradeShoes
iworu1a, iworu1b were unused
iworu1a = "I like my new shoes"
iworu1b = "These shoes fit nicely"

AudioEvent ToxinTractorVoiceUpgradeGamma
vtoxu2a, vtoxu2c were unused
vtoxu2a = "Look, General, the anthrax gamma upgrade is here"
vtoxu2c = "The new anthrax gamma is here, General"

AudioEvent RangerVoiceUpgradeChemSuits
iranu2a, iranu2c were unused
iranu2a = "These chemical suits are going to help"
iranu2c = "Upgrading to chemical suits"

AudioEvent RPGTrooperVoiceUpgradeCamoNetting
irpgu1a, irpgu1b were unused
irpgu1a = "We can hide our defenses with camo-netting now"
irpgu1b = "Our defense will be hidden from the enemy"

AudioEvent RangerVoiceUpgradeAdvancedTraining
iranu4b was unused
iranu4b = "We've completed the advanced training"

AudioEvent DozerUSAVoiceMove
vdoumod was unused
vdoumod.wav - "Like the neighbourhood"

AudioEvent DozerChinaVoiceClearMine
vdocclc was misspelled as vdocdlc
vdocclc.wav - "Our people will be safer"

AudioEvent DragonTankVoiceModeFireStorm
vdramfa, vdramfb, vdramfc were unused
vdramfa.wav - "Firestorm ready"
vdramfb.wav - "Ready for firestorm attack"
vdramfc.wav - "Where should I unleash the Dragon?"

AudioEvent GattlingTankVoiceSelect
vgatsec was unused
vgatsec.wav - "Let's get into battle"

AudioEvent GattlingTankVoiceMove
vgatmod was unused
vgatmod.wav - "Looking for some action"

AudioEvent GattlingTankVoiceUpgrade
vgatcrc was unused
vgatcrc.wav - "This might save us some bullets"

AudioEvent HumveeVoiceSelect
vhumsec, vhumsed, vhumsee were unused
vhumsec.wav - "Four wheelin"
vhumsed.wav - "Hey, little off-roading?"
[...]

PR #1906, opened 2023-05-05
Add scripts and info about unused sounds and speech

xezon:
* Relates to #176


xezon:
Scripts and info updated. Now also for Speech files.

xezon:
All research files added. Audio reviews are up-to-date if no mistakes were made.

PR #1908, opened 2023-05-07
Change: Improve voice setup of China Black Lotus

xezon:
* Relates to #176

This change adds several unused sounds to Lotus and Super Lotus.

### BlackLotusVoiceCaptureComplete

* iblacob "I'm in. We control the building now"
* iblacoc "Building capture complete"
* iblacod "We have their building. What's next?"

### BlackLotusVoiceHackVehicle

* iblaatc "I'll make this quick"
* iblaatd "I got it covered"

### BlackLotusVoiceDisableComplete

* iblac2c "That's it. Their vehicle is down"
* iblac2d "The vehicle is down. What's next?"

### BlackLotusVoiceHackCash

* iblah3b "They won't even notice"

### BlackLotusVoiceCashComplete

* iblah3c "Credit transfer ready"

### BlackLotusVoiceAttack

* Enabled audio event (unused)

### SuperBlackLotusVoiceCreate

* iblssea "Super Lotus here. Give me an update"

### SuperBlackLotusVoiceSelect

* iblsseb "Super Lotus"

### SuperBlackLotusVoiceMove

* iblssec "I'm quick"

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1908_unused_black_lotus_voices.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1908_unused_black_lotus_voices.yaml
@@ -1,0 +1,29 @@
+---
+date: 2023-05-05
+
+title: Adds unused voice variation(s) to China Black Lotus
+
+changes:
+  - tweak: Adds unused voice "I'm in. We control the building now" to BlackLotusVoiceCaptureComplete.
+  - tweak: Adds unused voice "Building capture complete" to BlackLotusVoiceCaptureComplete.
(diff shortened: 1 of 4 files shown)


xezon:
Demote to optional?

commy2:
I like the Super Lotus one. It's a different unit with a different portrait.

On the other hand, we probably need some guidelines to decide which to make mandatory and which to make optional.

xezon:
Just write which ones you do not like to hear by default and I will check to demote them.

MTKing4:
Different creation sound for a Varied unit is fine, it helps distinguishing it from the Vanilla one, all this can be on core

xezon:

PR #1909, opened 2023-05-07
Change: Improve voice setup of China Nuke Battlemaster

xezon:
* Relates to #176

This change adds new sounds to China Nuke Battlemaster.

### NukeBattleMasterTankVoiceSelect

* vbansed "Nuklear Battlemaster tank reporting"

### NukeBattleMasterTankVoiceCreate

* vbansec "Nuklear Battlemaster here"


xezon:
Demote to optional?

MTKing4:
Hmmm i think this one can use a new creation sound to distinguish it from normal battlemasters, so it might be good to add it to core, only issue this creation sound is too long, any other options?

Think the selection sound here is even more fitting for creation, so maybe switch them? I'm not sure, probably need to listen to both to judge

xezon:
> only issue this creation sound is too long

It is not longer than other regular ones:

```
Battlemaster:      "Chinas Tank Division"
Nuke Battlemaster: "Nuklear Battlemaster tank reporting"
Troop Crawler:     "Let's get these men to the battle"
Outpost:           "I will tell you everything I hear"
Gattling:          "I have many bullets to spare"
```

What I would actually prefer is
"Nuklear Battlemaster reporting"
"Nuklear Battlemaster here"

We can assemble these sounds I think.

xezon:
I have now added new assembled sounds.

"Nuklear Battlemaster tank reporting"
"Nuklear Battlemaster here"

I think this is better.

MTKing4:
> > only issue this creation sound is too long
> 
> It is not longer than other regular ones:
> 
> ```
> Battlemaster:      "Battlemaster tank reporting"
> Nuke Battlemaster: "Nuklear Battlemaster tank"
> Troop Crawler:     "Let's get these men to the battle"
> Outpost:           "I will tell you everything I hear"
> Gattling:          "I have many bullets to spare"
> ```
> 
> What I would actually prefer is "Nuklear Battlemaster reporting" "Nuklear Battlemaster here"
> 
> We can assemble these sounds I think.

True, also I'm pretty sure the Creation for battlemaster is "China's Tank Division" which might be wordy but it's lighter on the tongue, that's what i meant by long, and that's why i needed to hear it otherwise I can't tell

What do you mean by assembled? You created new voices with AI?

Can you attach them before and after in the opening post please?

xezon:
Yes it is "Chinas Tank Division". It is not possible to upload sounds to github. Assembled means it has been put together with other sounds.

Here are voice duration comparisons

| Voice (object)                        | Duration (s)   |
|---------------------------------------|----------------|
| Battlemaster Create                   | 1.08           |
| "Nuklear Battlemaster here"           | 1.23           |
| Supply Truck Create                   | 1.28           |
| Overlord Create                       | 1.29           |
| Dragon Create                         | 1.41           |
| Gattling Tank Create                  | 1.52           |
| Dozer Create                          | 1.71           |
| Troop Crawler Create                  | 1.83           |
| "Nuklear Battlemaster tank reporting" | 1.93           |
| Nuke Cannon Create                    | 1.94           |

xezon:
I suggest we make "Nuklear Battlemaster here" the creation sound and "Nuklear Battlemaster tank reporting" one of the select sounds. Good?

xezon:
English sounds swapped. Creation now uses the short line "Nuklear Battlemaster here"

MTKing4:
Will check that later, my pc is broken atm

xezon:

PR #1911, opened 2023-05-07
Enable Dragon Tank and Toxin Tractor secondary attack button sounds, but do not use them

xezon:
This change enable the Dragon Tank and Toxin Tractor secondary attack button sounds, but does not use them. User facing this has no effect, but is technically a better setup - consistent with other disabled setups of this kind.


xezon:
Requires no extra documentation.

xezon:
Good.

PR #1914, opened 2023-05-07
Add comments for unused sound files

xezon:
* Relates to #176

This change just adds comments for unused sound files.


xezon:
Comments look very good.

PR #1917, opened 2023-05-07
Move new upgrade audio event INI definitions near the original ones

xezon:
* Follow up for #1026

This change moves the new upgrade audio event INI definitions near the original ones. No change in functionality.


xezon:

PR #1923, opened 2023-05-08
Change: Adds sound variation for selecting the GLA Barracks

xezon:
* Relates to #176

This change adds a sound variation for selecting the GLA Barracks. The sound is similar to the original one. For optional bundle.


xezon:
We can drop this change, because it breaks the design of one distinct sound per building select.

PR #1925, opened 2023-05-08
Change: Improve voice setup of USA Paladin Tank

xezon:
**Merge with Rebase**

* Relates to #176

This change adds 1 former unused sound to the USA Paladin Tank and swaps other sounds.

### PaladinTankVoiceSelect

* ~~vpalsed "Polished and Ready"~~ **(Optional)**
* vpalseg "Preserving freedom" **(Optional)**
* vpalatf "Protecting our people"

### PaladinTankVoiceCreate

* ~~vpalseg "Preserving freedom"~~ **(Optional)**
* vpalsed "Polished and Ready" **(Optional)**

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1925_improve_paladin_voices.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1925_improve_paladin_voices.yaml
@@ -1,0 +1,23 @@
+---
+date: 2023-05-05
+
+title: Improves voice setup of USA Paladin
+
+changes:
+  - tweak: Moves voice "Preserving freedom" from PaladinTankVoiceCreate to PaladinTankVoiceSelect.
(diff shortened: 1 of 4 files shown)


MTKing4:
You're changing the voice upon creation and adding it to selection? I don't like that, it would confuse players, I'd rather have the same setup as it is now

ExiLe-ZH:
I mean you get used to it fast enough, think 'polished and ready' makes sense when rolling out of the factory. But why remove it from select?

xezon:
Creation voice change demoted to optional bundle pack.

xezon:
"Polished and Ready" really only makes sense as creation sound. Having it for selection is odd because the Paladin could be badly damaged at that time.

xezon:

PR #1927, opened 2023-05-08
Fix: Sets correct decay and attack sounds for AmbientCrowdAngryArabs1, AvalancheTextureLoop

xezon:
* Relates to #176

This change sets the correct decay and attack sounds for AmbientCrowdAngryArabs1 and AvalancheTextureLoop.


xezon:

PR #1928, opened 2023-05-08
Fix: Set less intense unique set of fire sounds for GenericFireMediumLoop

xezon:
* Relates to #176

This change adds the unused fire sounds to GenericFireMediumLoop.

Originally GenericFireLargeLoop and GenericFireMediumLoop use the same set of intense fire sounds, while there is a second set of less intense fire sounds. GenericFireMediumLoop now uses it. New fade in and fade out sounds were created for it accordingly.


xezon:

PR #1929, opened 2023-05-08
Fix: Set unique set of insect sounds for Amb_DesertVillageNightInsect2

xezon:
* Relates to #176

This change adds the unused night insect sounds to Amb_DesertVillageNightInsect2.

Originally Amb_DesertVillageNightInsect and Amb_DesertVillageNightInsect2 use the same set of sounds, while there is a second set of night insect sounds.


commy2:
literal bugs

xezon:

PR #1931, opened 2023-05-08
Change: Add unique machine gun firing sounds to USA Battle Drone

xezon:
* Relates to #176

This change enables the unique gun firing sounds for the USA Battle Drone. This way the Battle Drone gun no longer sounds like the guns of the Helix and Comanche. The sounds already exist in the game but were unused.

### Original

https://user-images.githubusercontent.com/4720891/236870858-137787e2-cec6-4ab5-9638-12fd6288663e.mp4

### Patched

https://user-images.githubusercontent.com/4720891/236870930-f830f1dc-a1da-4d8e-b5ad-fbd0a05cb65e.mp4

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1931_battle_drone_gun_audio.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1931_battle_drone_gun_audio.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-05-05
+
+title: Adds unique machine gun firing sounds to USA Battle Drone
+
+changes:
+  - feature: Adds unique machine gun firing sounds to the USA Battle Drone. Originally it uses the machine gun sounds of the Comanche helicopter.
+
+labels:
+  - audio
+  - enhancement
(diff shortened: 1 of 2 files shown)


MTKing4:
Sounds a bit too loud giving the impression it's stronger than it should be, maybe toning it down would be a good idea

ExiLe-ZH:
Agree with MTKing, the sound is cool but sounds more powerful than Commie MG, that's not quite right.

xezon:
I do not think the volume of the drone is too high. The Comanche gun is just relatively quiet.

Here is Battle Drone gun volume compared vs Ranger gun and Humvee gun, both of which are louder than new drone gun.

battle_drone_volume

xezon:
Comanche Gun -15 db
Ranger -10 db
Drone -12 db
Humvee -8 db

My rough observation.

xezon:
I did experiment with repurposing the original double shot audio files and creating single shot versions with them, but the added variation did not sound better.

PR #1932, opened 2023-05-08
Change: Improve voice setup of USA Stealth Fighter

xezon:
* Relates to #176

This change adds 1 new sound to the USA Stealth Fighter. The sound has been assembled for English, French, German, (Russian).

### StealthFighterVoiceCreate

* ~~vstesea "Ready for covert mission"~~
* vsteseg "Stealth Fighter, ready for take-off"


MTKing4:
Please stop changing Creation Voicelines, people have gotten used to original ones and this change not only is unnecessary but will cause confusion

You can just add the new sounds to the select, don't mess with the creation

xezon:
Demoted to optional bundle pack.

MTKing4:
Just to be clear, I would still like the new Voicelines added to selection, and the creation change is the only one that should be optional

xezon:
"Stealth Fighter, ready for take-off" is no fit voice line for selection.

MTKing4:
Hmm fair enough

xezon:

PR #1933, opened 2023-05-08
Fix: Remove invalid sound from SupplyTruckVoiceSupply

xezon:
* Relates to #176

This change removes an invalid sound from the SupplyTruckVoiceSupply.


xezon:

PR #1939, opened 2023-05-09
Change: Increase volume of ComancheWeaponMachineGun, HelixWeaponMachineGun

xezon:
This change increases the volume of the ComancheWeaponMachineGun from 70 to 90, HelixWeaponMachineGun from 80 to 90. This way it is no longer much quieter than guns of Ranger, Humvee and Co.

## Original

Humvee Gun: ~ -8 db
Ranger Gun: ~ -10 db
Comanche Gun: ~ -15 db
Helix Gun: ~ -12 db

## Patched

Comanche Gun: ~ -10 db
Helix Gun: ~ -10 db

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1939_comanche_gun_volume.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1939_comanche_gun_volume.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-05-09
+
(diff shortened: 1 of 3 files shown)


MTKing4:
What about the helix gun?

xezon:
Helix Gun fixed.

xezon:
Helix gun is still quieter than the Gattling Cannon. So that is good.

PR #1945, opened 2023-05-11
Fix: exhorde textures

xezon:
This change fixes exhorde textures

### exhorde, exhorde_up, exhordeb, exhordeb_up, exhordec_up

* Rotates stars to align with the direction of the units
* Removes minor artifacts from alpha channel

## Original

Decal rotation is not correct. 

shot_20230511_161911_1

## Patched

Red arrow always points forward.

shot_20230513_111431_1


xezon:

PR #1956, opened 2023-05-16
Fix: Add missing text variants for Nuclear Battlemaster and Nuclear Overlord

xezon:
This change adds dedicated text for Nuclear Battlemaster and Nuclear Overlord. This is consistent with other specialized units, such as Attack Outpost and Assault Troop Crawler.

shot_20230516_163240_1


xezon:

PR #1958, opened 2023-05-17
Change: Improve latin text of volume options

xezon:
This change improves latin text of volume options. The motivation is to help clarify that the Sound FX Volume also affects unit voices. There is limited space for the text in English, so we cannot put all that much text. Non-English languages have a bit more room, because they use different Font.

shot_20230517_175354_1

shot_20230517_203133_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1958_audio_volume_options_text.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1958_audio_volume_options_text.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-05-17
+
+title: Improves latin text of volume options
+
+changes:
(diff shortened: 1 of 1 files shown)


bobtista:
What is the speech volume controlling? Is it like the voice acting eg "Tanks, build more tanks!" from single player campaigns? 
According to the manual: `SOUND FX VOLUME Adjust the volume of sound effects in the game.
VOICE VOLUME Adjust the volume of voices in the game.`

xezon:
The bottom audio slider is for Campaign, Challenge, Attack Notification voices (2D sounds).
The middle audio slider is for Sound Effects, Unit Voices (3D sounds).

xezon:
Documentation added. Is not perfect. But hopefully a bit better than original.

PR #1962, opened 2023-05-18
Change: Remove obsolete voice line from USA King Raptor

xezon:
This change removes an obsolete voice line from King Raptor.

### KingRaptorVoiceSelect

* Removes vrapsea "Raptor here", because vraksed says "King Raptor here"


MTKing4:
Makes sense. I approve

xezon:

PR #1964, opened 2023-05-19
Fix: English audio restoration (2)

xezon:
* Follow up for #1061

This change fixes humming, hissing, crackling, fade in, fade out for all Unit sounds that were not already covered by #1061. The fixed audio issues are not as severe as they were in #1061 and therefore the affected files are stored in separate folders, but merged into the same bundle pack. It uses 30 MB uncompressed. Due the large size requirement it is therefore unfit for Patch Core and will be optional content only.

The sounds are bundled into the following archives:
* OptionalAudioArabic
* OptionalAudioBrazilian
* OptionalAudioChinese
* OptionalAudioEnglish
* OptionalAudioFrench (uses just a few of the sounds)
* OptionalAudioItalian
* OptionalAudioKorean
* OptionalAudioPolish
* OptionalAudioSpanish


xezon:
2 more review passes completed.

xezon:
3 more review passes completed.

xezon:
Finished another review pass. Added and updated documentation. I think this is enough now. I could do 100 more review passes and still find errors in these audio files.

I call it done for now.

PR #1968, opened 2023-05-20
Remove audio shadowing of gexpdaia, gsparksa, gsparksb

xezon:
* Follow up for #1413
* Follow up for #1800

This change removes audio shadowing of gexpdaia, gsparksa, gsparksb.


xezon:
Documentation added. Works.

PR #1920, opened 2023-05-07
Change: Add various unused sounds

xezon:
* Relates to #176

This change adds several unused sounds to various things.

### ~~F15MoveLoop~~

* ~~Attack vb52lo1a~~

### DozerUSAMoveStart

* Sounds vtrostae

### ~~B52AmbientLoop~~

* ~~Attack vb52lo1a~~

### ~~CINE_B52AmbientLoop~~

* ~~Attack vb52lo1a~~

### BuildingDie

* Sounds gbuidieh

### CowAmbience

* Sounds addfco1d
* Sounds addfco1e

### Amb_UrbanCityAmbientLoop2

* Sounds aurbci2c

### Amb_DesertMarketWallaLoop2

* Attack amarke1f
* Decay amarke1g

### Amb_DesertMarketWallaLoop3

* Attack amarke1f
* Decay amarke1g

### CivilianPanicMale

* Sounds gpanmald
* Sounds gpanmale

### ExplosionPatriotEMP

* Sounds gexpempc

PR #1970, opened 2023-05-20
Fix: Add fade in sound to B52 engine loop

xezon:
* Follow up for #1920

This changes adds a fade in sound to the B52 engine loop. This way sound fade is gentle when Bomber flies into camera view. Adding a custom sound for it, because the original vb52lo1a sound has a too long fade in which makes it sound as if the engine is powering on.


xezon:

PR #1971, opened 2023-05-20
Restore original fire sound on tree objects

xezon:
* Follow up for #1928

This change restores original fire sound on tree objects, because the previously referenced medium fire sounds were changed.


xezon:
Documentation added.

PR #1918, opened 2023-05-07
Change: Add various unused voices to faction units

xezon:
* Relates to #176

This change adds several unused voices to various faction units.

### RangerVoiceSelect

* iransed "Will lead the way"

### ComancheVoiceSelect

* vcomsec "Need little ground support?"

### ChinookVoiceFalling

* vchifab "Ahhhhhhhhhhhhhhh"

### HumveeVoiceSelect

* vhumsee "Alright let's have some fun"

### DozerUSAVoiceMove

* vdoumod "Like the neighbourhood"

### ~~RedGuardVoiceCaptureComplete~~

* ~~iredcob "Building captured"~~

### HackerVoiceHackComplete

* ihaccob "Hack complete"
* ihaccoc "Hehe. I have disabled their building"
* ihaccod "Enemy building is disabled"

### HackerVoiceSurrender

* ihacsuc "I surrender. Just don't hurt my computer"

### GattlingTankVoiceSelect

* vgatsec "Let's get into battle"

### GattlingTankVoiceMove

* vgatmod "Looking for some action?"

Note: The original sound is spoken with oddly large gaps between the words, which perhaps is the reason this sound was not chosen. I have reduced the gaps in the audio file and put it in the core bundle.

### NukeCannonVoiceSelect

* vnuksec "They will never forget"

### POWTruckChinaVoiceSelect

* vpocsea "Are there more pickups?"
* vpocsec "I love meeting new people"

### POWTruckChinaVoiceMove

* vpocmob "Looking for some company"
* vpocmod "More candidates?"

### ~~RebelVoiceCaptureComplete~~

* ~~irebcoe "An enemy building has been captured"~~

### TerroristVoiceSelect

* itersed "Our following is strong"

### TerroristVoiceMove

* itermod "I move to a sacred place"

### TerroristVoiceEnter

* itermod "I move to a sacred place"

### ToxinTractorVoiceAttack

* vtoxath "Ahhh hehehe"

### QuadCannonVoiceSelect

* vqussec "I put them all to rest" (Quad Sniper)

### QuadCannonVoiceAttack

* vqusatb "Put them out of their misery" (Quad Sniper)
* vqusatd "Mow them down" (Quad Sniper)
* vqusate "Turn them into mud" (Quad Sniper)

### WorkerVoiceSurrender

* iworsue "Here. Take me. I don't care"

### ScudLauncherVoiceModeAnthrax

* vscumab "Anthrax missile ready"

### ScudLauncherVoiceModeHiEx

[...]

PR #1977, opened 2023-05-23
Change: Add unused voice(s) to GLA Worker (2)

xezon:
* Relates to #176
* Follow up for #1918

This change adds unused voice(s) to GLA Worker.

### WorkerVoiceMoveUpgraded

* iworu1b "These shoes fit nicely"


xezon:
Tonality of voice does not fit the other move voices. Discarding.

PR #1994, opened 2023-06-07
Fix: Gattlings start spinning pre attack (1)

### LINKED ISSUE #1991: Gatling Gun doing a Gatling spin animation when locking on target
### Issue Observable in Original and Patch

it seems to happen when the unit locks on a target, so it's best reproduced by a manual attack click on an enemy unit, otherwise an attack move won't trigger it until the target is in the units vision range.

## **Affected Units**
Gatling Tank
Overlord Gatling Cannon
Helix Gatling Cannon

## Replay
[Gat Spin](https://www.mediafire.com/file/jkii2p362ii9kpc/OMAK_gmail_replay_spinning_gatling.rep/file)

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/e72d5b99-d0e5-43bb-8c41-b5dd6b429e62

xezon:
* Fixes #1991

This change fixes the Gattling spin issue on China Gattling Tank. It works, but the implementation is complicated. The fix cannot be applied as is to Overlord Gattling Cannons, because they do not move.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -1888,6 +1888,7 @@ Object ChinaTankGattling
 
   ; Patch104p @bugfix commy2 08/10/2022 Fix muzzle flash when aiming at grounded targets.
   ; Patch104p @tweak xezon 09/10/2022 Rescale Gattling AnimationSpeedFactorRange's from 0.5, 0.8, 1.5, 3.0 to 0.60, 0.60, 1.20, 1.80 to match firing rate properly.
+  ; Patch104p @bugfix xezon 05/06/2023 Adds AliasConditionStates for MOVING to prevent the turret from spinning when attacking and moving without firing.
 
   Draw                      = W3DTankDraw ModuleTag_01
     OkToChangeModelColor    = Yes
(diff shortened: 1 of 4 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
```diff
@@ -1902,8 +1903,16 @@ Object ChinaTankGattling
...
+    AliasConditionState = USING_WEAPON_B ATTACKING MOVING
 
-    ConditionState        = REALLYDAMAGED RUBBLE
+    ConditionState        = REALLYDAMAGED
+      Model               = NVGattTank_D
+    End
+    AliasConditionState = REALLYDAMAGED ATTACKING MOVING
+    AliasConditionState = USING_WEAPON_B REALLYDAMAGED ATTACKING MOVING
+
+    ConditionState        = RUBBLE
```
commy2, line 1915: What's this for?

xezon: It is split off of `ConditionState = REALLYDAMAGED RUBBLE` above so that the Condition States are clean.

commy2: I would not introduce unnecessary states to keep code and diff short and simple.

xezon: RUBBLE is a valid state for units. Hulks are not always spawned right away.

commy2: It works fine as alias, as there is no difference between RD and Rubble.

xezon: Yes you are right it can be an Alias.

commy2:
Missing Boss General. Otherwise lgtm.

PR #1999, opened 2023-06-09
Change: Shorten English Particle Cannon timer text

xezon:
This change shortens "Particle Uplink Cannon" to "Particle Cannon" for the English Superweapon timer text.

There are 2 reasons for this:
1. The Particle Cannon is referred to as "Particle Cannon" is every other text
2. "Particle Cannon" is shorter than "Particle Uplink Cannon" and therefore takes less precious space on the screen

## Original

shot_20230609_194212_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1999_particle_cannon_text.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1999_particle_cannon_text.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-06-09
+
+title: Shortens English Particle Cannon timer text
+
+changes:
+  - tweak: "Particle Uplink Cannon" now reads "Particle Cannon" on the English Superweapon timer.
+
+labels:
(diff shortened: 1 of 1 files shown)


ExiLe-ZH:
Would be better and less bloaty.

MTKing4:
could we fix the "MISSING" text as well from unused Superweapons?

xezon:

commy2:
Good change.

> could we fix the "MISSING" text as well from unused Superweapons?

Yes, I have a PR planned for this. It also fixes the missile rotation. "Laser Cannon" text already is fixed. 1.04, it also says "Missing: blah"

MTKing4:
Oh nice, just noticed, ICBM is also fixed, so only Cruise Missile left

commy2:
ICBM text works in vanilla.

MTKing4:
Oh nvm then

PR #2000, opened 2023-06-10
Fix: Add missing Attack Move button to China ECM Tank

xezon:
This change adds the AttackMove button to the China ECM Tank. Attack Move works as expected.

## Patched

ECM Tank is attack moving on its own.

shot_20230610_072020_3


commy2 (approved):

xezon:
Documentation added.

PR #2001, opened 2023-06-10
Change: Add Guard button to GLA Radar Van

xezon:
This change adds the Guard button to the GLA Radar Van.

Guard does not do anything special for the Radar Van. It just moves to the target location and idles there. However, adding the Guard button allows it to be used in group selections, which can be useful if the player wants to send a group of units together with a Radar Van somewhere to guard.

The Attack Move button is not required to be added to the Radar Van, because Attack Move already works in group selections regardless of the absence of the button.

## Original

GLA units group selection. Guard button is missing.

shot_20230610_070759_1

## Patched

GLA units group selection. Guard button is present.

shot_20230610_073631_3

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2000_ecm_attackmove_button.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2000_ecm_attackmove_button.yaml
@@ -4,7 +4,7 @@ date: 2023-06-10
 title: Adds missing Attack Move button to China ECM Tank
 
 changes:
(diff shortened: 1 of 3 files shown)


commy2 (approved):

xezon:
This functionality is useful when attacking enemy base with a group of GLA units and using Guard Mode to have them kill the structures on their own.

xezon:
Documentation added.

PR #2005, opened 2023-06-10
Revert: TechRepairPad is deleted when destroyed (#568)

xezon:
This change partially reverts #568.

After #568, the destroyed Repair Pad looked almost identical to the really damaged Repair Pad, because the Drop Zone object does not have a dedicated rubble object.

The original setup is now reinstated to provide a notable visual difference between the really damaged and destroyed Repair Pad.

## Repair Pad after 568

Really damaged vs Destroyed Repair Pad.

shot_20230610_151218_2

## Original Repair Pad, after this change

Really damaged vs Destroyed Repair Pad.

shot_20230610_151549_1


xezon:

commy2:
Oops.

PR #2011, opened 2023-06-14
Fix: Remove obsolete structure damage particle effects

xezon:
This change removes the obsolete structure damage particle effects. Some effects are never visible and others rarely, depending on the geometry of the structure. Despite, the damage particles always spawn at the very center of a structure, which makes no sense when the damage impact is somewhere else.

## Original

shot_20230614_174508_4

shot_20230614_174722_6


xezon:

Issue #802, opened 2022-07-31
Toxin GLA Demo Trap is not practical in regular match


ExiLe-ZH:
Costs $600, builds very slow, ~~automatic detonation reacts slow,~~ very low damage output.

## Demo Trap Specs

Demo Trap | price | build time (seconds) | primary damage | 30 s anthrax beta damage | Prerequisite
-- | -- | -- | -- | -- | --
vGLA | 400 | 10 | 600 | - | Arms Dealer 
Tox | 600 | 16 | 250 | 140 | Arms Dealer 
Demo | 200 | 6 | 700 |  - | Supply Stash 
Stealth | 400 | 10 | 600 | - | Arms Dealer 

## Contamination Times

| Object                   | Contamination Time (s) |
|--------------------------|------------------------|
| Toxin Shells (Scorpion)  | 10                     |
| Toxin Tractor Sprinkler  | 180                    |
| SCUD Storm               | 30                     |
| SCUD Launcher Anthrax    | 30                     |
| Bio Bomb Truck           | 30                     |
| Demo Trap                | 30                     |
| Toxin Bomb               | 60                     |

## Puddle Damages

Demo Trap uses MediumPoisonFieldWeaponUpgraded and Chem_MediumPoisonFieldWeaponGamma.

| Weapon                                          | Damage | DPS  |
|-------------------------------------------------|:-------|:-----|
| SmallPoisonFieldWeapon (Toxin Shells)           | 2.0    | 3.75 |
| SmallPoisonFieldWeaponUpgraded (Toxin Shells)   | 2.5    | 4.69 |
| Chem_SmallPoisonFieldWeaponGamma (Toxin Shells) | 2.5    | 4.69 |
| MediumPoisonFieldWeapon                         | 2.0    | 3.75 |
| MediumPoisonFieldWeaponUpgraded                 | 2.5    | 4.69 |
| Chem_MediumPoisonFieldWeaponGamma               | 2.5    | 4.69 |
| LargePoisonFieldWeapon (Anthrax Bomb)           | 15     | 28.1 |
| LargePoisonFieldWeaponUpgraded (Anthrax Bomb)   | 25     | 46.9 |
| Chem_LargePoisonFieldWeaponGamma (Anthrax Bomb) | 25     | 46.9 |
| AnthraxBombPoisonFieldWeapon                    | 40     | 75.0 |
| AnthraxBetaBombPoisonFieldWeapon                | 40     | 75.0 |
[...]

PR #2015, opened 2023-06-18
Change: Decrease build time of GLA Toxin Demo Trap from 16 to 8 seconds

xezon:
* Relates to #802

This change decreases the build time of the GLA Toxin Demo Trap from 16 to 8 seconds. 

| Object                        | Build Cost | Build Time | Explosion Damage | 30s Poison Damage |
|-------------------------------|------------|------------|------------------|-------------------|
| Original GLA Trap             | 400        | 10         | 600              |                   |
| Original GLA Demo Trap        | 200        | 6          | 700              |                   |
| Original GLA Stealth Trap     | 400        | 10         | 600              |                   |
| Original GLA Toxin Trap       | 600        | 16         | 250              | 140               |
| Patched GLA Toxin Trap (this) | 600        | 8          | 250              | 140               |

# Rationale

The original GLA Toxin Trap provides poor value. Its damage output is vastly inferior to that of other GLA Traps. It is more expensive. And ultimately the long build time of 16 seconds (without power) makes it unattainable for practical applications. If the opposition can see a GLA Worker swinging his hammer in the air for 16 seconds, then it is much less likely for the Toxin Trap to succeed in a surprise.

8 seconds over 10 seconds build time compensates a bit for the fact that the Toxin Trap costs more and deals less damage than the other Traps.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2015_toxin_trap_build_time.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2015_toxin_trap_build_time.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-06-18
+
+title: Decreases build time of GLA Toxin Demo Trap from 16 to 8 seconds
+
+changes:
+  - fix: Decreases build time of GLA Toxin Demo Trap from 16 to 8 seconds. This makes it more practical to build before the enemy approaches it.
+
+labels:
+  - buff
+  - controversial
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2015
+
+authors:
+  - xezon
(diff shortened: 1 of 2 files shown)


ExiLe-ZH:
Tox trap is such annoying case for me, like I don't understand why it's so expensive and slow building, it would've made much more sense if it was the same as other GLA's, but better vs infantry and worse vs vehicles. They should've just removed it from Tox all together, I prefer that over the current setup.

Ofcourse we're not going to remove it and neither should we buff it too much because Toxgen doesn't need buffs. Equalizing like I described above is probably too much of a buff already.

So I guess 8 seconds is fine, players with brains are not going to spam $600 traps. 

Though what if we nerf explosive damage even a bit more (200), buff the poison damage and maybe it's radius aswell? Then equalize build time and price? Would that be safe? Then it's really specialized as an anti-infantry trap, but still affordable.

xezon:
I suggest discussing damage buffs in #802. This change is only concerned about lowering the build time.

To me the 8 seconds build time looks very safe given the current damage potential. The Toxin Trap still sucks given its price, but at least it can be used for what it is intended with the reduced build time. So it is no longer useless.

xezon:
MTKing
> Looks good to me

Jundiyy
> Fine by me

xezon:

PR #2020, opened 2023-06-19
Change: Prioritize larger Toxin and Radiation puddles over smaller ones

xezon:
wip


xezon:
This approach does not work. Puddles eventually will just kill each other and there will be no survivor. I think this could only work if there was a way to guarantee that no 2 puddles spawn at the exact same logic frame.

PR #2014, opened 2023-06-17
Fix: Decrease small poison field radius from 12 to 7.5

xezon:
* Fixes #794

This change decreases small poison field radius from 12 to 7.5 to match small Anthrax Beta and Anthrax Gamma fields radius. The small poison field is used by Scorpion Tank toxin shells.

The particle effect size was adjusted to closer match the damage radius.

| Weapon                           | Original Damage Radius | Patched Damage Radius |
|----------------------------------|------------------------|-----------------------|
| SmallPoisonFieldWeapon           | 12                     | 7.5                   |
| SmallPoisonFieldWeaponUpgraded   | 7.5                    |                       |
| Chem_SmallPoisonFieldWeaponGamma | 7.5                    |                       |
| GC_Chem_GLARebelGunGamma         | 10                     |                       |
| Chem_ToxinTruckGunGamma          | 10                     |                       |
| SmallRadiationFieldWeapon        | 15                     |                       |
| Nuke_SmallRadiationFieldWeapon   | 12                     |                       |

# Rationale

The damage radius of Toxin Shells no longer becomes worse after acquiring the Anthrax Beta upgrade.

PR #2022, opened 2023-06-20
Fix: Decrease small poison field object radius from 6 to 4

xezon:
* Follow up for #2014

This change decrease the small poison field object radius from 6 to 4. This is consistent with the other small poison puddle objects. It only affects the size of the area it can be deleted from by other puddles.


xezon:

Issue #705, opened 2022-07-17
USA Paradrop Generals Power is very rarely used


xezon:
USA Paradrop Generals Power is very rarely used. Players typically prefer the A10 and Spectre Gunship Promotions. Check if it can be better and more attractive.

## Stats
```
SpecialPower SuperweaponParadropAmerica
  ReloadTime = 240000

ObjectCreationList SUPERWEAPON_Paradrop1
  Payload = AmericaInfantryRanger 5

ObjectCreationList SUPERWEAPON_Paradrop2
  Payload = AmericaInfantryRanger 10

ObjectCreationList SUPERWEAPON_Paradrop3
  Payload = AmericaInfantryRanger 20
```

## Proposal 1
Reduce cooldown time.

## Proposal 2
Add a few Missile Defenders into the Ranger group (2 or 3, suggested by Dominator)

## Proposal 3
Reduce to 1 Special Power only

## Proposal 4
Make Soldiers in Parachute controllable - like in Contra Mod.

commy2:
Missile Defenders do have a parachute animation. As have Pathfinders and Col. Burton (as have Red Guards, Tank Hunters, Hackers and Lotus).

xezon:
Nice. Let's add 2 Pathfinders and 6 Burtons in the Paradrop.

USA-Bruce:
It's best just to make the cooldown non-existent, It should be four minutes now.
Try 1 minute, It already gets 20% more infantry than rebel ambush, 16 vs 20.
If 1 minute is too high increase to 90 seconds

ExiLe-ZH:
> It's best just to make the cooldown non-existent, It should be four minutes now.
> 
> Try 1 minute, It already gets 20% more infantry than rebel ambush, 16 vs 20.
> 
> If 1 minute is too high increase to 90 seconds 

Thing is you can make lvl 1 more attractive, but 2 & 3 will never be chosen ever.

Proposal: remove lvl 1 and 3

ReLaX82:
Thought I answered already, put some MDs in it, maybe Pathis on level 3?

Maybe decrease timer as well.

Will add strats if used more frequently, which is good.

commy2:
Pathfinders are interesting, because that could be an alternative to "unlocking" them without having to spend a genpoint on them.

MTKing4:
Proposal 2 and 4 for me

commy2:
Stage 1: 3 Ranger 2 MD
[...]

PR #2026, opened 2023-06-21
Change: Replace 40% of Rangers with Missile Defenders in USA Paradrops

xezon:
* Relates to #705

This change replaces 40% of Rangers with Missile Defenders in USA Paradrops.

The latin localization has been updated accordingly.

## Original

Paradrop Level 1: 5 Ranger
Paradrop Level 2: 10 Ranger
Paradrop Level 3: 20 Ranger

## Patched

Paradrop Level 1: 3 Rangers + 2 Missile Defenders
Paradrop Level 2: 6 Rangers + 4 Missile Defenders
Paradrop Level 3: 12 Rangers + 8 Missile Defenders

shot_20230621_180828_2

shot_20230621_180842_3

shot_20230621_180851_4

# Rationale

The original USA Paradrop is seldom used, because the Spectre Gunship and A10 powers are typically better value than a group of Rangers. The newly added Missile Defenders increase the usefulness of the Paradrop, because it can now also engage vehicles, helicopters and planes in both offense and defense.

The maximum monetary value of the spawned infantry increases by 600, from 4500 to 5100.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2026_usa_paradrop_payload.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2026_usa_paradrop_payload.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-06-21
+
+title: Replaces 40% of Rangers with Missile Defenders in USA Paradrops
+
+changes:
+  - tweak: Replaces 40% of Rangers with Missile Defenders in USA Paradrops. Rank 1 drops 3 Rangers and 2 Missile Defenders, Rank 2 drops 6 Rangers and 4 Missile Defenders and Rank 3 drops 12 Rangers and 8 Missile Defenders.
+
+labels:
+  - buff
+  - controversial
+  - design
+  - major
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2026
+
+authors:
+  - xezon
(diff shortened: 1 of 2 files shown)


commy2 (approved):
code: good
translation: no idea

xezon:
Jundiyy
> Like it but not sure. 2 MDs for first drop seems fine but the 8 seem like quite a lot.

Exile
> I feel the opposite way, 2 MD's with level 1 feels like a bigger buff than 8 with level 3 to me, because it's a huge sacrifice to spent so many points on paradrop and not on meta powers. Spending only 1 point is easier to consider. Overall I think the changes are fine, because something is needed to make it an attractive alternative to meta powers.

xezon:
MTKing
> I like the current setup

xezon:

xezon:
Zarz:
> Very strong upgrade. I do not think it is a good idea. Because in late game, if you will drop these guys, then almost no chance to kill them if opponent has low money and some units. 8 Missile Defenders on Level 3 looks too hard.

ExiLe-ZH:
40% looks a little too OP for me, the MD's can do a lot of damage in different stages of the game, like killing expensive landed Aircraft or hunting Dozers.

My proposal is:
- Keep level 1 like original, no MD's.
- 2 MD's with level 2.
- 4 MD's with level 3. 

I realize this is still powerful, however every promotion spent on the Paradrop, can't be spent on the A10, which is a huge sacrifice. Therefore level 2 and 3 definitely need a considerable buff.

xezon:
Would be good if this was tested.

PR #2028, opened 2023-06-22
Fix Supply Lines bonus value on USA Supply Drop Zone crates

xezon:
* Follow up for #1390

This change fixes a mistake introduced by #1390. The Supply Lines would add just 5% bonus instead of 10% bonus for Supply Drop Zones crates. This is now fixed.


xezon:
Documentation added.

commy2 (approved):

PR #2030, opened 2023-06-23
Null clip sizes of all stream weapons

xezon:
This change nulls the clip sizes of all stream weapons. Some stream weapons already have this setup. Game play wise this makes no difference.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2494,6 +2494,7 @@ Weapon HijackVehicle
 End
 
 ;------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


xezon:
This change requires no further documentation.

commy2:
As far as I can tell, this changes how Ambo Detox ability works, as the button fires 1 full clip.

xezon:
I expect that it fired until the puddle is deleted. The clip reload was instant.

commy2:
ClipReload is instant, but it fires 30 shots * DelayBetweenShots=40

xezon:
Ok. It is not really clear to me if this could make a difference somehow. Perhaps there is a Script Action or an AI behavior which says "fire one clip of your weapon" and if we changed this perhaps behavior would change unexpectedly. Right now firing 1 clip would mean all (most) these stream weapons shoot for 2 seconds.

So we can close this.

PR #2031, opened 2023-06-23
Remove unnecessary definitions from DummyWeaponProjectile

xezon:
* Follow up for #279

This change removes unnecessary physics and ai update definitions from DummyWeaponProjectile.


commy2 (approved):

xezon:
Documentation updated.

PR #2033, opened 2023-06-23
Fix: Minimize dummy weapon damage

xezon:
This change minimizes dummy weapon damage. This way it will be very unlikely to ever kill a unit.


commy2:
This doesn't do anything. The projectiles are deleted immediately and never hit a target.

PR #2035, opened 2023-06-24
Optimization: Decrease performance cost of toxin and hazard cleanup stream splash effects by 20%

xezon:

This change decreases the cost of toxin and hazard cleanup stream hit splash effects by 40%. The overall performance improvement of the stream weapon is significantly less than that, because there are other expensive components.

Affects

* GLA Toxin Truck
* GLA Toxin Rebel
* GLA Toxin Tunnel
* USA Ambulance

## Before

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/687f2c7d-3ad8-49bf-b013-ba3026091118

## After (Outdated)

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/989d51d8-9c6d-4d0d-89dc-ab451473acfc

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2035_toxin_stream_splash_performance.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2035_toxin_stream_splash_performance.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-06-25
+
+title: Decreases performance cost of toxin and hazard cleanup stream splash effects by 20%
+
+changes:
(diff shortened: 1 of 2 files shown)


MTKing4:
Looking good

xezon:
The 40% cut was a bit too much. I brought it to a 20% cut, which looks better and is also sufficient because this effect is not so expensive to begin with.

Documentation added.

PR #2036, opened 2023-06-24
Optimization: Decrease performance cost of toxin stream puddles by 70%

xezon:

This change decreases the cost of toxin stream puddles by 70%. The overall performance improvement of the stream weapon is significantly less than that, because there are other expensive components.

Affects

* GLA Toxin Truck
* GLA Toxin Rebel
* GLA Toxin Tunnel

The performance increase is achieved by reducing the puddle lifetime from 100 to 30 frames. Design wise the shorter puddle lifetime also makes more sense, because the visual puddle itself deals no damage and the slow fade out of 100 frames would indicate that the location is still hazardous when it actually is not.

The cleanup puddle of the USA Ambulance already has a lifetime of 30 frames.

## Before

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/687f2c7d-3ad8-49bf-b013-ba3026091118

## After

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/4e5aaac7-440b-4c88-afaf-6cc14dcd2da2

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2036_toxin_stream_puddle_performance.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2036_toxin_stream_puddle_performance.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-06-24
+
+title: Decreases performance cost of toxin stream puddles by 70%
+
+changes:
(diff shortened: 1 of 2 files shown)


MTKing4:
Very nice

xezon:

PR #2037, opened 2023-06-24
Restore original China Overlord upgrade command button order in core bundle pack

xezon:
* Follow up for #1896

This change restores the original China Overlord upgrade command button order (before #1896) in the core bundle pack. Reason being, the changed order can conflict with the memory of legacy players. Therefore it is better to keep the corrected order in the optional bundle pack only.


xezon:
Documentation updated.

PR #2038, opened 2023-06-24
Fix: excleanupstream texture

xezon:
This change fixes the excleanupstream texture color for the USA Ambulance cleanup stream. It now matches the particle colors better.

## Original

shot_20230624_171120_1

## Patched

shot_20230624_171255_1


commy2:
Idk, I like the contast of the original more.

MTKing4:
There's a merit to the original colors, it makes the stream distinct from its particles, right now they're just one entity

xezon:
Perhaps there is another way to fix it, by shifting particle colors from cyan to purple.

xezon:
Obsolete by #2041

commy2:
This IS https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2038

PR #2041, opened 2023-06-25
Fix: Improve color blending of hazard cleanup stream particles

xezon:
* Closes #2038

Alternative implementation for #2038. Instead of changing the cyan stream texture, the surrounding particles are adjusted to blend from cyan to purple.

## Original

shot_20230624_171120_1

## ~~Patched (1)~~

shot_20230625_111316_2

## ~~Patched (2)~~

shot_20230628_155033_2

## Patched (This)

shot_20230702_103520_3


MTKing4:
Too much cyan

You know what i think the original colors are just fine

xezon:
I do not like the original transition. The stream color does not match the particle colors.

xezon:
Cyan amount reduced.

shot_20230625_113335_2

MTKing4:
Still think the original looks better

Cyan particles are changing too much from the original color

xezon:
Ok. Not the feedback I expected, but ok. This is subjective stuff. I like the edit very much, it also is better than #2038. It smoothly transitions from cyan stream to purple particles. Originally the stream would be cyan and the particles just a different color, which looks odd without any transition between them.

MTKing4:
How about we just make the stream a lighter shade of purple? That way they'll blend in and still be distinguishable without the need to adjust the particles

MTKing4:
Something like this #c06de3

xezon:
I think it is fine as is. I suggest you take a look at it in game in motion. Perhaps it will not look as bad as you think it does from the image.

ExiLe-ZH:
tbf like the original better too

xezon:
Ok. I don't quite understand your graphical tastes, but I will try demote it to Optional content then.

MTKing4:
Cyan color is just not right, cleaning spray is meant to be purple, you making it more cyan is not a good idea, why not just make the stream pinkish like i suggested and be done with it?

I think you should keep in mind that art is subjective and currently the only opinion that's being considered for the optional content is yours, i think you need more than one person to decide that branch, because what if everyone disagrees with your art decisions? I certainly haven't seen anyone agreeing with you internally yet, and that should be indicative enough.

You wouldn't have many Optional users that way, and me personally I'd like to consider using the optional bundle when it comes out, on the condition that has well made art decisions and is appealing for the basic users

xezon:
> currently the only opinion that's being considered for the optional content is yours

This is not true. In #2038

commy2:
> Idk, I like the contast of the original more.

MTKing:
> There's a merit to the original colors, it makes the stream distinct from its particles, right now they're just one entity

These comments indicate that cyan and purple was preferred. So I tried to make them work by blending particles from cyan stream color to purple particle color, preserving both original colors and therefore adhering to the given feedback.

MTKing:
> Too much cyan.

So I reduced cyan amount in puddle accordingly.

xezon:
The only issue I take with current implementation is that the stream splash color still does not properly blend. I am not sure why that is, but I can take a nother look.

MTKing4:
Yes but this was all meant for core, but when the final result wasn't agreed upon you decided that it should be added to the optional bundle, which no one else have provided their opinion on for that branch yet, which has been the case with past optional pulls as well, consisting of one opinion 

The reason why the original color is still preferred is because pretty much no one has ever noticed the cyan color, most just assume it has always been purple, so we didn't really prefer it because it had cyan in it

The other reason is that we've gotten used to being able to distinguish the stream from the particles

So in the end what we want is 2 things

1. Distinguishable stream
2. Purple colored theme

And in my opinion, to do that without having to mess with particles or change colors drastically, you can just put the stream as light purple, so that it can be distinguished from its particles while still maintaining the purple theme with an organic color blend, aka leave the particles unchanged, can you try that?

xezon:
Yes, when something subjective cannot be agreed upon then I will try push it to Optional bundle. The particle blending is objectively better than no blending, because originally the colors do not match. Why should the stream have different color than the particles, when the particles are meant to be part of the entire stream visuals? The colors should fit together when these things are meant to belong together.

There is one more thing I can try, despite fixing the splash thingy.

MTKing4:
Particle blending is still subjective though, blending colors is an art change regardless, which has to be agreed upon by the majority before greenlighting

That being said i understand your point, that's why i suggested to give the stream a lighter shade of purple therefore you will have your color blend and we get to keep the purple theme, win win

xezon:
The particles are actually dark blue. What makes them so purple is the ADDITIVE shader of the particles.

I have now blended the cyan stream texture with the blue color half way to reduce cyan strength, then blended from this new stream color to the blue particle color.

And also fixed the color/shader mismatch of the splash particle.

shot_20230628_155033_2

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/a3dff995-5e4a-47ef-bd04-d80d7ac868ea

ExiLe-ZH:
> Ok. I don't quite understand your graphical tastes, but I will try demote it to Optional content then.

Can't explain really, just a taste :p

MTKing4:
> shot_20230628_155033_2
> 

Ok that is a lot better already, nice work :)

xezon:
> Ok that is a lot better already, nice work :)

Ok great because I started to lose my mind over here.

So we move this to Optional or no?

MTKing4:
> Ok great because I started to lose my mind over here.

Haha no this is just right, you did a good job :)

This can be on core

xezon:
Splash changes reverted for now. Need revisit together with other splashes. Documentation added. Looks good.

PR #2042, opened 2023-06-26
Change: Fix, improve and optimize spray particles and puddles of GLA Toxin Tractor

xezon:
**Merge with Rebase**

This change fixes, improves and optimizes the Sprayer particles and puddles of the GLA Toxin Tractor.

1. Removes non-functional projectile detonation effect from Sprayer weapons
2. Simplifies and improves the Toxin Tractor spray puddle particle effects to streamline visuals and cut cost by 62.5%
3. Decreases cost of the Toxin Tractor spray particles by 50%
4. Increases visibility and matches visuals of Toxin Tractor spray particles
5. Fixes the incorrect SizeRateDamping setup of the ToxinSpray particle system
6. Increases gravity of Toxin Tractor spray particles by 20%

## Before

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/a5138fa8-1679-4c58-9861-e2f037f95b78

## After

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/d8f6ed49-e3fd-4114-858f-96b27ac61d42


Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
```diff
@@ -18145,14 +18146,14 @@ Object Chem_GLAVehicleToxinTruck
...
-      ParticleSysBone = none AnthraxGammaPuddleContinuous
-    END
+    ;ConditionState = USING_WEAPON_B
+    ;  Model = UVToxinTrk
+    ;  ParticleSysBone = none AnthraxPuddleContinuous
+    ;END
+    ;ConditionState = USING_WEAPON_B WEAPONSET_PLAYER_UPGRADE
+    ;  Model = UVToxinTrk
+    ;  ParticleSysBone = none AnthraxGammaPuddleContinuous
+    ;END
```
commy2: I would just delete these blocks entirely.

commy2:
I think this setup is preferable, but the name "nozzle" doesn't fit anymore.

xezon:
Can we just add an Alias name in comment?

xezon:
Or we add new effect copy and leave the old one as it was.

commy2:
New one.

Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
xezon: Ok. Done.

xezon:
Done. Tested and worked.

MTKing4:
Sprinkling particles are definitely less visible now, or clouds are too bright and visible now that the particles can't be seen anymore?

xezon:
Yes, because the puddle underneath is more dense now. What we can try is throw the spray particles a bit further away at random to cover a larger area.

MTKing4:
Why does it need to be so dense though?

xezon:
It is the original effect of the medium toxin puddle.

MTKing4:
Then what have you changed from original that made puddles look different?

xezon:
This:
> Simplifies and improves the Toxin Tractor spray puddle particle effects to streamline visuals and cut cost by 62.5%

This means by using this puddle instead of the previous ones the performance was improved and the puddle visuals are now consistent.

MTKing4:
Which ones was it using originally? Small puddles?

xezon:
Before it was using 2 other puddle particle systems. See code page.

MTKing4:
Don't know what I'm reading

Anyway, is there no way around it? Reduced puddle opacity or using small puddles instead? Is that possible?

xezon:
I would like to keep the puddle as is if possible and tweak the sprinkler instead.

MTKing4:
Hmmm it would be pretty hard to make the sprinkles visible over such High contrasted puddle, but I'm curious what will your solution look like

xezon:
We can do this.

## Image

shot_20230703_181114_1

## Video

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/11df9e00-859e-40c8-994b-b79a4ba2156b

xezon:
Spray particles are quite vibrant now. Very visible. Consistent look of particles. Perhaps needs to revisit colors in the future. But for now I think it looks ok.

Documentation added.

MTKing4:
Yeah this looks better, maybe contrast is a bit too high, but for now it's fine

PR #2023, opened 2023-06-20
Change: Prioritize larger Toxin and Radiation hazard fields over smaller ones

xezon:
* Relates to #55

This change prioritizes larger Toxin and Radiation fields/puddles over smaller ones. Larger fields will require more hazard damage to clean up.

This change fixes the exploitative hazard field behavior where players would typically fire a single shot on any hazard field to remove it in an instant. It also fixes the bug where the death field of the Toxin Tractor would remove its Sprinkler contamination field.

## Original

| Field                                  | Health      | Hazard Damage |
|----------------------------------------|-------------|---------------|
| Small Poison and Radiation fields      | 100 or 120  | 200           |
| Medium Poison and Radiation fields     | 100 or 120  | 200           |
| Large Poison and Radiation fields      | 100 or 120  | 200           |
| Anthrax Bomb fields                    | 100 or 120  | 200           |
| Nuke Missile field                     | 150         | 200           |

## Patched

| Field                                  | Health      | Hazard Damage |
|----------------------------------------|-------------|---------------|
| Small Poison and Radiation field       | 100         | 100           |
| Medium Poison and Radiation field      | 1000        | 1000          |
| Large Poison and Radiation field       | 5000        | 5000          |
| Anthrax Bomb fields                    | 10000       | 10000         |
| Nuke Missile field                     | 10000       | 10000         |

# Consequences

Originally any field could be cleaned up with any other field. Now larger fields will require more damage from smaller fields to clean up. Fields of equal size clean up each other with one shot.

The Anthrax Bombs and Nuke Missiles now have the largest and therefore most persistent fields. They will take 100 small, 10 medium or 2 large field hazard damages to clean up.

Smaller fields can stack on top of larger fields now, as long as the smaller fields do not manage to clean up the bigger ones. But larger fields will always clean up smaller ones if they spawn afterwards and with their core inside the smaller field's object radius.

In practice it will take more than 5 Scorpions with Toxin Shells and no Junk Repair to clean the Anthrax Bomb field and more than 5 Nuke Battlemaster tanks to clean the Nuke Missile field before the tanks die on the respective fields. The more tanks the quicker the cleanup can be completed.

A single USA Ambulance will take 200 frames, or 6.66 seconds to cleanup the Anthrax Bomb and Nuke Missile. The more Ambulances the quicker the cleanup can be completed.

## 6 Scorpions with Toxin Shells vs Anthrax Bomb toxins

[...]

PR #2045, opened 2023-06-26
Change: Increase field damage of GLA Anthrax Gamma Bomb by 20%

xezon:
* Follow up for #2012

This change increases the field damage of the GLA Anthrax Gamma Bomb by 20%. The field damage of the Anthrax Beta Bomb is unchanged, which means the upgrade damage progression from the blue to the purple toxin field is now consistent with all other toxin field weapons.

| Weapon                            | Original Damage (DPS) | Patched Damage (DPS) |
|-----------------------------------|:----------------------|:---------------------|
| SmallPoisonFieldWeapon            | 2.0 (3.75)            |                      |
| SmallPoisonFieldWeaponUpgraded    | 2.5 (4.69)            |                      |
| Chem_SmallPoisonFieldWeaponGamma  | 2.5 (4.69)            | 3.0 (5.63)           |
| MediumPoisonFieldWeapon           | 2.0 (3.75)            |                      |
| MediumPoisonFieldWeaponUpgraded   | 2.5 (4.69)            |                      |
| Chem_MediumPoisonFieldWeaponGamma | 2.5 (4.69)            | 3.0 (5.63)           |
| LargePoisonFieldWeapon            | 15 (28.1)             |                      |
| LargePoisonFieldWeaponUpgraded    | 25 (46.9)             |                      |
| Chem_LargePoisonFieldWeaponGamma  | 25 (46.9)             | 30 (56.3)            |
| AnthraxBombPoisonFieldWeapon      | 40 (75.0)             |                      |
| AnthraxBetaBombPoisonFieldWeapon  | 40 (75.0)             |                      |
| AnthraxGammaBombPoisonFieldWeapon | 40 (75.0)             | 48 (90)              |

# Rationale

The Anthrax Gamma upgrade has more utility now and the damage progression of the field weapon becomes consistent with that of other toxin fields.

See

* #2012 

Although this change is a notable buff for GLA Toxin General, it has received a number of major nerfs too:

* https://github.com/TheSuperHackers/GeneralsGamePatch/pull/699
* https://github.com/TheSuperHackers/GeneralsGamePatch/pull/882
* https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2002

This buff does not affect the other GLA Generals, because they natively have no access to the Anthrax Gamma upgrade.

PR #2044, opened 2023-06-26
Change: Decrease field damage of GLA Toxin Bomb by 20%

xezon:
This change decreases the field damage of the GLA Toxin Bomb by 20%. The field damage of the Anthrax Beta Bomb is unchanged, which means the upgrade damage progression from the green to the blue toxin field is now consistent with all other toxin field weapons.

| Weapon                            | Original Damage (DPS) | Patched Damage (DPS) |
|-----------------------------------|:----------------------|:---------------------|
| SmallPoisonFieldWeapon            | 2.0 (3.75)            |                      |
| SmallPoisonFieldWeaponUpgraded    | 2.5 (4.69)            |                      |
| Chem_SmallPoisonFieldWeaponGamma  | 2.5 (4.69)            | 3.0 (5.63)           |
| MediumPoisonFieldWeapon           | 2.0 (3.75)            |                      |
| MediumPoisonFieldWeaponUpgraded   | 2.5 (4.69)            |                      |
| Chem_MediumPoisonFieldWeaponGamma | 2.5 (4.69)            | 3.0 (5.63)           |
| LargePoisonFieldWeapon            | 15 (28.1)             |                      |
| LargePoisonFieldWeaponUpgraded    | 25 (46.9)             |                      |
| Chem_LargePoisonFieldWeaponGamma  | 25 (46.9)             | 30 (56.3)            |
| AnthraxBombPoisonFieldWeapon      | 40 (75.0)             | 32 (60)              |
| AnthraxBetaBombPoisonFieldWeapon  | 40 (75.0)             |                      |
| AnthraxGammaBombPoisonFieldWeapon | 40 (75.0)             | 48 (90)              |

# Rationale

The Anthrax Beta upgrade has more utility now and the damage progression of the field weapon becomes consistent with that of other toxin fields. Although this change is a notable nerf, with #2023 there is a major buff that makes the Anthrax Bombs much better against GLA and USA. Overall, China benefits most from these changes.

This is a nerf for all GLA Generals, except Toxin General. Most notably it is a nerf for GLA Demolition General, because it natively does not have access to the Anthrax Beta upgrade. However, the Demolition General has many other perks, such as Demolition upgrade and stronger Scud Storms, which makes up for the weaker toxin puddles.

As for GLA General and GLA Stealth General, the nerf manifests before Anthrax Beta upgrade only. This gives more reward to the Anthrax Beta upgrade, which is originally seldom acquired with Stealth General.

PR #2049, opened 2023-06-28
Change: Add inherent hazard field cleanup abilities to GLA poison and China flame weapons

xezon:
**Merge with Rebase**

* Relates to #55
* Follow up for #2023
* Closes #1109

This change adds inherent hazard field cleanup abilities to all GLA poison and China flame weapons. This affects

* GLA Toxin Rebel gun
* GLA Toxin Tunnel gun
* GLA Toxin Tractor gun
* GLA Toxin Tractor sprayer
* China Dragon Tank flame thrower
* China Dragon Tank flame thrower + fire wall
* China Mig fire field
* China Mig fire storm
* China Helix fire storm
* China Inferno Cannon fire field
* China Inferno Cannon fire storm

## Patched hazard field cleanup setup

| Hazard Field Cleanup weapon          | Health | Hazard Dmg | Duration ms | Hazard Dmg in 1000 ms |
|--------------------------------------|--------|------------|-------------|-----------------------|
| Small Poison and Radiation field     | 100    | 100        | -           | 100
| Medium Poison and Radiation field    | 1000   | 1000       | -           | 1000
| Large Poison and Radiation field     | 5000   | 5000       | -           | 5000
| Anthrax Bomb fields                  | 10000  | 10000      | -           | 10000
| Nuke Missile field                   | 10000  | 10000      | -           | 10000
| USA Ambulance                        | -      | 100        | 66          | 1500
| Toxin Tractor stream                 | -      | 25         | 66          | 375
| Toxin Tractor sprayer                | -      | 75         | 200         | 375
| Toxin Rebel stream                   | -      | 10         | 66          | 150
| Toxin Tunnel stream                  | -      | 10         | 66          | 150
| Dragon Tank flame thrower            | -      | 25         | 66          | 375
| Dragon Tank fire wall (*1)           | -      | 5 to 20    | 66          | 75 to 300
| China Mig fire field                 | -      | 500        | 2500        | 200
| China Mig fire storm (*2)            | -      | 3000       | 6000        | 500
| China Helix fire storm               | -      | 3000       | 6000        | 500
| China Inferno Cannon fire field      | -      | 500        | 2500        | 200
| China Inferno Cannon fire storm (*2) | -      | 3000       | 6000        | 500

(*1) The Fire Wall hazard damage will stack on top of flame thrower damage, because the Dragon Tank fire wall ability has 2 weapons.

(*2) The Fire Storm hazard damage stacks on top of fire field damage. So 3 shell hits with 500 hazard damage each + fire storm with 3000 hazard damage.

## Toxin stream vs Anthrax Bomb field (Anthrax Beta damage)

2 Toxin Tractors can cleanup the Anthrax Bomb effortlessly. Toxin Tractor is immune to Poison and Radiation.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/2f6cee58-d9dc-477d-beef-3e83a940741e

## Flame thrower vs Anthrax Bomb field (Anthrax Beta damage)

2 Dragon Tanks can cleanup the Anthrax Bomb when well positioned. They do take poison damage.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/6d2b5d12-5ee0-4c71-b5f2-d4136d79bb7e

## Fire wall vs Anthrax Bomb field (Anthrax Beta damage)

2 Dragon Tanks can cleanup the Anthrax Bomb when well positioned. They do take poison damage.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/b5631ab6-4025-420d-9424-07d38b1cb098

## Fire storm vs Anthrax Bomb field (Anthrax Beta damage)

2 Fire Storms with 4 Inferno Cannons clean up the Anthrax Bomb.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/dd1d01de-92ee-4555-b21c-d277fb2f6b6c

# Rationale

The new hazard field cleanup setup gives all China and GLA factions the ability to cleanup hazards reliably with various weapons. Originally only radiation and poison fields could do cleanups and it was counter-intuitive that some poison weapons, such as Toxin Tractor streams, did not cleanup hazard fields, while Toxin Shells from Scorpion Tanks did.

China greatly profits from hazard field cleanups with flame weapons due their availability. Originally only Nuke Battle Masters, Nuke Cannons, Nuke missiles and destroyed Nuke Reactors and destroyed Nuke Silos could cleanup hazard fields.

USA only has the Ambulance for hazard cleanups, however it is the superior cleaner, because it cleans the fastest, is immune to toxin and radiation and is easily available at any stage of a match.

This change makes it much more intuitive to perform hazard field cleanups with the various factions, while at the same time the hazard fields pose a greater threat due their durability.

Overall China benefits most from the hazard cleanup design changes, because although it will be challenging to cleanup, it no longer is unpractical to do so.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
@@ -1089,6 +1089,80 @@ Object GC_Chem_ToxinStreamProjectile ; Alias GC_Chem_ToxinTruckStreamProjectile
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @feature xezon 06/07/2023 Stream projectile for Toxin General, a copy of GC_Chem_ToxinStreamProjectile (#2049)
+Object GC_Chem_ToxinRebelStreamProjectile
+
+  ; Explanation - Particles can't do damage, so this is a fast shooting low
+  ; damage invisible missile launcher with a toxin trail for exhaust
+
+  ; *** ART Parameters ***
+  Draw = W3DModelDraw ModuleTag_01
+    DefaultConditionState
+      Model = NONE
+    End
+  End
+  ; Must have a draw module to be allowed to draw detonation FX
+
+  ; ***DESIGN parameters ***
+  EditorSorting     = SYSTEM
+  VisionRange       = 0.0
+  ArmorSet
+    Conditions      = None
+    Armor           = ProjectileArmor
+    DamageFX        = None
+  End
+
+  ; *** ENGINEERING Parameters ***
+  KindOf            = PROJECTILE
+  Body              = ActiveBody ModuleTag_02
+    MaxHealth       = 100.0
+    InitialHealth   = 100.0
+  End
+
+  Behavior = DestroyDie ModuleTag_03
+    ;nothing
+  End
+
+  Behavior = PhysicsBehavior ModuleTag_04
+    Mass                = 1
+    AllowCollideForce   = No  ; toxins collide, but never apply forces when they do so
+  End
+
+  Behavior = MissileAIUpdate ModuleTag_05
+    TryToFollowTarget               = No
+    FuelLifetime                    = 1000
+    InitialVelocity                 = 120                ; in dist/sec
+    IgnitionDelay                   = 0
+    DistanceToTravelBeforeTurning   = 2
+    GarrisonHitKillRequiredKindOf   = INFANTRY
+    GarrisonHitKillForbiddenKindOf  = NONE
+    GarrisonHitKillCount            = 2
+    GarrisonHitKillFX               = Chem_FX_ToxinStreamGammaGarrisonBuildingHit
+    DetonateCallsKill               = Yes
+  End
+
+  Locomotor = SET_NORMAL ToxinTruckStreamLocomotor
+
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinRebelGunHazardFieldWeapon
+    StartsActive = Yes
+  End
+
+  Behavior = BoneFXUpdate ModuleTag_06
+    PristineParticleSystem1  = bone:NULL OnlyOnce:Yes  0  0 PSys:GC_Chem_ToxinTrail01 ; Covers gap at beginning of line
+  End
+
+  Behavior = BoneFXDamage ModuleTag_07
+    ;nothing
+  End
+
+  Geometry = Sphere
+  GeometryIsSmall = Yes
+  GeometryMajorRadius = 2.0
+
+End
+
+;------------------------------------------------------------------------------
 Object GC_Chem_ToxinProjectileStream
 
   ; *** ART Parameters ***
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/GC_Chem_GLAUnits.ini
@@ -1015,7 +1015,7 @@ End
 
 
 ;------------------------------------------------------------------------------
-Object GC_Chem_ToxinStreamProjectile
+Object GC_Chem_ToxinStreamProjectile ; Alias GC_Chem_ToxinTruckStreamProjectile
 
 ; Explanation - Particles can't do damage, so this is a fast shooting low
 ; damage invisible missile launcher with a toxin trail for exhaust
@@ -1063,9 +1063,16 @@ Object GC_Chem_ToxinStreamProjectile
     GarrisonHitKillForbiddenKindOf  = NONE
     GarrisonHitKillCount            = 2
     GarrisonHitKillFX               = Chem_FX_ToxinStreamGammaGarrisonBuildingHit   ; Patch104p @bugfix commy2 28/08/2021 Fix Gamma Anthrax effect to use pink instead of green color.
+    DetonateCallsKill               = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
 
   Locomotor = SET_NORMAL ToxinTruckStreamLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinTruckGunHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   Behavior = BoneFXUpdate ModuleTag_06
     PristineParticleSystem1  = bone:NULL OnlyOnce:Yes  0  0 PSys:GC_Chem_ToxinTrail01 ; Covers gap at beginning of line
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -1537,6 +1537,11 @@ Object FirestormSmall
     MaxLifetime = 6000
   End
 
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = FireStormHazardFieldWeapon
+  End
+
   Geometry            = CYLINDER
   GeometryMajorRadius = 1.0
   GeometryMinorRadius = 1.0
@@ -1595,6 +1600,11 @@ Object BlackNapalmFirestormSmall
     MaxLifetime = 6000
   End
 
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = FireStormHazardFieldWeapon
+  End
+
   Geometry            = CYLINDER
   GeometryMajorRadius = 1.0
   GeometryMinorRadius = 1.0
@@ -1632,6 +1642,11 @@ Object FireWallSegment
 
   Behavior = FireWeaponUpdate ModuleTag_05
     Weapon = FireWallSegmentWeapon
+  End
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = DragonTankFireWallHazardFieldWeapon
   End
 
   Behavior = DeletionUpdate ModuleTag_06
@@ -1679,6 +1694,11 @@ Object FireWallSegmentUpgraded
     Weapon = FireWallSegmentUpgradedWeapon
   End
 
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = DragonTankFireWallHazardFieldWeapon
+  End
+
   Behavior = DeletionUpdate ModuleTag_06
     MinLifetime = 4000
     MaxLifetime = 4000
@@ -1693,7 +1713,7 @@ Object FireWallSegmentUpgraded
 End
 
 ;------------------------------------------------------------------------------
-Object FireFieldSmall
+Object FireFieldSmall ; Used by Inferno Tank, Mig Jet
 
   ; *** ART Parameters ***
   Draw = W3DModelDraw ModuleTag_01
@@ -1717,6 +1737,11 @@ Object FireFieldSmall
     Weapon = SmallFireFieldWeapon
   End
 
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = SmallFireFieldHazardFieldWeapon
+  End
+
   Behavior = DeletionUpdate ModuleTag_04
     MinLifetime = 2500
     MaxLifetime = 2500
@@ -1731,7 +1756,7 @@ Object FireFieldSmall
 End
 
 ;------------------------------------------------------------------------------
-Object FireFieldUpgradedSmall
+Object FireFieldUpgradedSmall ; Used by Inferno Tank, Mig Jet
 
   ; *** ART Parameters ***
   Draw = W3DModelDraw ModuleTag_01
@@ -1753,6 +1778,11 @@ Object FireFieldUpgradedSmall
   End
   Behavior = FireWeaponUpdate ModuleTag_03
     Weapon = SmallFireFieldWeaponUpgraded
+  End
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponUpdate ModuleTag_HazardFieldWeapon
+    Weapon = SmallFireFieldHazardFieldWeapon
   End
 
   Behavior = DeletionUpdate ModuleTag_04
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -4113,9 +4113,15 @@ Object DragonTankFlameProjectile
     GarrisonHitKillCount = 2
     GarrisonHitKillFX = FX_DragonFlameGarrisonBuildingHit
     DistanceToTargetForLock = 0  ; If it gets within 100 of the target, it kills the target.
-
+    DetonateCallsKill = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
   Locomotor = SET_NORMAL DragonTankFlameLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = DragonTankFlameHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   ; Patch104p @refactor xezon 24/06/2023 Removes obsolete particle effects. (#2040)
   Behavior = BoneFXUpdate ModuleTag_06
@@ -4208,9 +4214,15 @@ Object DragonTankFlameProjectileUpgraded
     GarrisonHitKillCount = 2
     GarrisonHitKillFX = FX_DragonFlameGarrisonBuildingHit
     DistanceToTargetForLock = 0  ; If it gets within 100 of the target, it kills the target.
-
+    DetonateCallsKill = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
   Locomotor = SET_NORMAL DragonTankFlameLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = DragonTankFlameHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   ; Patch104p @refactor xezon 24/06/2023 Removes obsolete particle effects. (#2040)
   Behavior = BoneFXUpdate ModuleTag_06
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -3707,9 +3707,16 @@ Object ToxinTruckStreamProjectile
     GarrisonHitKillForbiddenKindOf  = NONE
     GarrisonHitKillCount            = 2
     GarrisonHitKillFX               = FX_ToxinStreamGarrisonBuildingHit
+    DetonateCallsKill               = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
 
   Locomotor = SET_NORMAL ToxinTruckStreamLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinTruckGunHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   Behavior = BoneFXUpdate ModuleTag_06
     PristineParticleSystem1  = bone:NULL OnlyOnce:Yes  0  0 PSys:ToxinTrail01 ; Covers gap at beginning of line
@@ -3774,8 +3781,15 @@ Object ToxinTruckStreamProjectileUpgraded
     GarrisonHitKillForbiddenKindOf  = NONE
     GarrisonHitKillCount            = 2
     GarrisonHitKillFX               = FX_ToxinStreamUpgradedGarrisonBuildingHit
+    DetonateCallsKill               = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
   Locomotor = SET_NORMAL ToxinTruckStreamLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinTruckGunHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   Behavior = BoneFXUpdate ModuleTag_06
     PristineParticleSystem1  = bone:NULL OnlyOnce:Yes  0  0 PSys:AnthraxTrail01 ; Covers gap at beginning of line
@@ -3878,10 +3892,16 @@ Object ToxinTruckSprayProjectile
     IgnitionDelay = 0
     DistanceToTravelBeforeTurning = 0
     DistanceToTargetForLock = 0  ; If it gets within 100 of the target, it kills the target.
-
+    DetonateCallsKill = Yes ; Patch104p @tweak xezon 27/06/2023 Adds DetonateCallsKill to fire weapon when dead.
   End
 
   Locomotor = SET_NORMAL ToxinTruckDribbleLocomotor
+
+  ; Patch104p @feature xezon 27/06/2023 Adds new hazard cleanup field weapon to this weapon object.
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinTruckSprayerHazardFieldWeapon
+    StartsActive = Yes
+  End
 
   Geometry = Sphere
   GeometryIsSmall = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/WeaponObjects.ini
@@ -3806,6 +3806,79 @@ Object ToxinTruckStreamProjectileUpgraded
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @feature xezon 06/07/2023 Stream projectile for Toxin General, a copy of ToxinTruckStreamProjectileUpgraded (#2049)
+Object ToxinRebelStreamProjectileUpgraded
+
+  ; Explanation - Particles can't do damage, so this is a fast shooting low
+  ; damage invisible missile launcher with a toxin trail for exhaust
+
+  ; *** ART Parameters ***
+  Draw = W3DModelDraw ModuleTag_01
+    DefaultConditionState
+      Model = NONE
+    End
+  End
+  ; Must have a draw module to be allowed to draw detonation FX
+
+  ; ***DESIGN parameters ***
+  EditorSorting     = SYSTEM
+  VisionRange       = 0.0
+  ArmorSet
+    Conditions      = None
+    Armor           = ProjectileArmor
+    DamageFX        = None
+  End
+
+  ; *** ENGINEERING Parameters ***
+  KindOf            = PROJECTILE
+  Body              = ActiveBody ModuleTag_02
+    MaxHealth       = 100.0
+    InitialHealth   = 100.0
+  End
+
+  Behavior = DestroyDie ModuleTag_03
+    ;nothing
+  End
+
+  Behavior = PhysicsBehavior ModuleTag_04
+    Mass = 1
+    AllowCollideForce = No  ; flames collide, but never apply forces when they do so
+  End
+
+  Behavior = MissileAIUpdate ModuleTag_05
+    TryToFollowTarget               = No
+    FuelLifetime                    = 1000
+    InitialVelocity                 = 120                ; in dist/sec
+    IgnitionDelay                   = 0
+    DistanceToTravelBeforeTurning   = 2
+    GarrisonHitKillRequiredKindOf   = INFANTRY
+    GarrisonHitKillForbiddenKindOf  = NONE
+    GarrisonHitKillCount            = 2
+    GarrisonHitKillFX               = FX_ToxinStreamUpgradedGarrisonBuildingHit
+    DetonateCallsKill               = Yes
+  End
+  Locomotor = SET_NORMAL ToxinTruckStreamLocomotor
+
+  Behavior = FireWeaponWhenDeadBehavior ModuleTag_HazardFieldWeapon
+    DeathWeapon = ToxinRebelGunHazardFieldWeapon
+    StartsActive = Yes
+  End
+
+  Behavior = BoneFXUpdate ModuleTag_06
+    PristineParticleSystem1  = bone:NULL OnlyOnce:Yes  0  0 PSys:AnthraxTrail01 ; Covers gap at beginning of line
+  End
+
+  Behavior = BoneFXDamage ModuleTag_07
+    ;nothing
+  End
+
+  Geometry = Sphere
+  GeometryIsSmall = Yes
+  GeometryMajorRadius = 2.0
+
+End
+
+;------------------------------------------------------------------------------
 Object ToxinTruckProjectileStream
 
   ; *** ART Parameters ***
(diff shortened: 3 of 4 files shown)


MTKing4:
Great work!

Looks very nice indeed, i like the setup

MTKing4:
One thing to keep in mind though, giving toxin tunnels the ability to clear will also give toxin general a bigger advantage at clearing than other factions, kinda feels like we're back at square one here, i fear toxin general will still clear puddles too easily

xezon:
Yes Toxin General will do well if there is a Toxin Tunnel nearby. It could also use the Toxin Rebel Ambush to clean toxins :-)

We can scale hazard damage setup freely as we like, so no problem.

MTKing4:
Yeah exactly, too many hazard cleaners, and tox tunnels are always spammed, should be toned down for toxgen in particular

commy2:
Why would pouring more goo on a poison cloud clear it up? This makes no sense at all.

xezon:
> Why would pouring more goo on a poison cloud clear it up? This makes no sense at all.

Well it is in line with the design of the Ambulance. It also sprays some stuff on the toxins and then the hazard is cleaned. Perhaps there are some anti-contaminants inside the toxins that kill stale toxins? Or the new toxins are so potent that they kill the old ones.

I am not sure if game design needs to be super realistic in this case when the machanic itself should be the deciding factor. I would argue that if Toxin weapons should not remove other toxins, then neither should the toxin puddles. And vice versa.

Gameplay wise current setup is fair, because all factions can remove all hazards equally without discrimination.

* Ambulance can remove Radiation and Toxin.
* Flame can remove Radiation and Toxin.
* Radiation can remove Radiation and Toxin.
* Toxin can remove Radiation and Toxin.

Is it realistic that Toxins remove Radiations? Likely not. Is it even realistic that Radiations disappear after 30 seconds? I don't think so. If this game was realistic then Radiation should stay until the end of the match.

If we start discriminating, like only Flames can kill Toxins and only radiation can kill toxins then perhaps that would be more realistic, but would that also be better for gameplay?

xezon:
Follow up change for this would be to add a text to the affected units that their fire/toxin weapons can clean hazard fields. This way design would be communicated with player.

Another possibility is to explore hazard field auto attack with select units, like the Ambulance does. This way it would also be communicated in an non textual way.

xezon:
Realistically there are 3 routes to the problem. Here I try to compare them.

# Route 1: Keep the status quo

All hazard fields can be removed by placing any other hazard field on top. Anthrax Bomb can be removed with one Toxin Shell from the Scorpion Tank.

### Advantages

- Design is unchanged
- Legacy players are not confused because they know hazard field mechanics
- Interactive hazard fields add depth to game mechanics, because the player gets more options to control the outcome of events

### Disadvantages

- New players are confused because mechanics feel like an exploit
- In exploitative scenarios the Anthrax Bomb fields and other hazard fields are useless against vehicles
- It can create odd issues such as the Toxin Tractor deleting its own hazard field on death
- It is unrealistic that hazard fields can remove each other. In nature puddles would just stack

# Route 2: Redesign hazard cleanup health, damages and weapons (This)

This change (and #2023) are taking Route 2. It is a compromise between Route 1 and Route 3. The original exploitative feel of the hazard field removal is redesigned to keep the mechanics of the hazard removal, expanding it and making it more difficult to exercise. Therefore there is a balance striked between it being easily possible to remove hazard fields and it being impossible to remove hazard fields (except USA Ambulance).

### Advantages

- Legacy players are not confused because they know hazard field mechanics
- Anthrax Bomb fields and other hazard fields become more useful against vehicles
- It requires preparation, skill and time to cleanup a hazard field
- Interactive hazard fields add depth to game mechanics, because the player gets more options to control the outcome of events
- It avoids odd issues such as the Toxin Tractor no longer deleting its own hazard field on death

### Disadvantages

- Design is changed
- New players are confused because mechanics could feel like an exploit
- It is unrealistic that hazard fields can remove each other. In nature puddles would just stack

# Route 3: Remove exploit by allowing enemy hazard fields to stack

Enemy weapons, except USA Ambulance, can no longer remove hazard fields. Only allied weapons can, to avoid hazard field stacks. Change #2023 is compatible with Route 3, as it would still prevent allies from removing their own large hazard field with a small one by mistake.

### Advantages

- New players are not confused because mechanics feel proper
- It is realistic that enemy hazard fields do not remove each other
- All Toxin and Radiation fields work as advertised and unleash their whole damage potential always
- It avoids odd issues such as the Toxin Tractor no longer deleting its own hazard field on death

### Disadvantages

- Design is changed
- Legacy players are confused because they can no longer remove hazard fields with their own
- All enemy hazard fields become impossible to remove for China and GLA, which could be a major frustration point if they render the base production useless for long periods of time
- Has high potential to fundamentally change the way matches are played out with hazard fields
- Non-interactive hazard fields remove depth from game mechanics, because it means the player has less control over the outcome of an event

xezon:
I updated the Patched hazard field cleanup setup table. It now shows hazard cleanup damage per 1000 ms. USA Ambulance is the clear leader.

MTKing4:
So with this change force firing is the same as spraying for toxin weapons, which means toxin tunnels will be meta for clearing, adding to that rebels + all the other puddle creation weapons, I don't like this, it's too much, i think it's better to just use the sprayer, we don't need inherent weapons to clear toxins, it's too risky.

And if we needed to, For the sake of equality, we can consider making dragon tanks only clear with firewalls and infernos and migs only with firewalls, but maybe that's not needed right now, i kidna think it's equilized with toxin already since they have toxin shells + tractors +  toxin terrorists for toxgen

xezon:
> So with this change force firing is stronger than spraying for toxin weapons, which means toxin tunnels will be meta for clearing, adding to that rebels + all the other puddle creation weapons, I don't like this, it's too much, i think it's better to just use the sprayer, we don't need inherent weapons to clear toxins, it's too risky.

Toxin General will still take longer to clear Anthrax than in Original. Toxin General starts with Toxin Shells, so one shot from Scorpion Tank would have cleared Anthrax Bomb. In Patch, it will take more time, even with Toxin Tunnels around. Therefore I do not agree with risk assessment.

MTKing4:
Yes but it will still clear toxins faster than other generals in patch, perhaps much faster with all the clearing options he has available, which is an unfair advantage.

We haven't tested, it might take much less time to clear it than you're anticipating

xezon:
We can reduce it for Toxin Tunnel and Toxin Rebel.

commy2:
Still think this is a bad, unrealistic and counter-intuitive change. I get it for Flame weapons (burns the chemicals, as in the lore), but why can toxin weapons clear toxin now? If GLA really needs a counter to puddles so badly, why not use https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1113? It's a new ability either way, but at least it makes it explicit.

xezon:
I agree with it being unrealistic.

As for intuition, I think it is more intuitive than original, because

| Weapon   | Removes hazard field in Original | in Patch (this) |
|----------|----------------------------------|-----------------|
| Toxin Tractor stream        | no  | yes |
| Toxin Tractor sprinkler     | no  | yes |
| Toxin Tractor contamination | yes | yes |
| Toxin Tractor death puddle  | yes | yes |
| Toxin Shell                 | yes | yes |

Judging from this angle, the added consistency makes it much easier to understand.

In essence it should work all the way, or not at all. In other words all boxes should be either "yes" or "no", otherwise it is confusing and counter-intuitive.

commy2:
> In essence it should work all the way, or not at all. 

I agree, I just think it should just not work at all.

xezon:
> I agree, I just think it should just not work at all.

That would be Route 3. I listed above the advantages and disadvantages of Route 2 (this) and Route 3. There is merit to both. To me it looks as if Route 3 has more disadvantages.

commy2:
This one

- Has high potential to fundamentally change the way matches are played out with hazard fields

applies to both.

These:

- All enemy hazard fields become impossible to remove for China and GLA
- Non-interactive hazard fields remove depth from game mechanics, because it means the player has less control over the outcome of an event

don't apply if you keep vulnerability to flame weapons and a Toxin Tractor decontam ability.

An explicit Toxin Tractor decontam ability would be balanced between GLA factions for the most part too, unlike giving Toxin Tunnels and half of all Toxgen units that ability for free.

I also think it is easy to explain to experienced players: Puddle deletion bug was removed. To make Anthrax Bomb/Tractor Sprinkler not broken, use a Tractor like an Ambo as GLA, or a Dragon tank with force fire as China.

This also balances the ability to clear puddles between the Chinese factions. Currently, only Nukegen can do it unless you could Nuke Cannons.

commy2:
So basically I'm suggesting Route 4: Remove exploit by allowing enemy hazard fields to stack. Add poison puddle decontamination ability to Toxin Tractor and FLAME vulnerability to poison puddles to counter Anthrax Bombs.

Which is
- explicit
- backed up by the lore
- balances the ability between GLA and Chinese factions (everyone has Toxin Tractor and Dragon Tank, not everyone has Toxin Shells for free or Nuke BMs)

Disadvantage:
- Removes Suicide from Demo Tractor (sad, but doesn't matter balance wise in the slightest)
- Adds a new ability (but this applies to any suggestion, except do-nothing)

xezon:
FLAME vulnerability to poison puddles has 2 disadvantages

1) Damage to hazard fields cannot be tweaked independently of the actual FLAME damage.
2) Flames cannot remove radiation, so China is left behind when it comes to hazard field removal.

Hazard cleanup ability for Toxin Tractor has 1 disadvantage

1) Conflicts with suicide ability of Demo Tractor

commy2:
> Damage to hazard fields cannot be tweaked independently of the actual FLAME damage.

You can tweak the armor value of the puddles. I don't see the issue.

> Flames cannot remove radiation, so China is left behind when it comes to hazard field removal.

I mean, you could make radiation puddles vulnerable to FLAME too, but I don't think being able to clean up radiation puddles is necessary. The only one that one would maybe want to clean up is the one after a Nuclear Silo strike, all the other ones are probably gone before you can move in a vehicle and have a small radius.

MTKing4:
Why didn't we make FLAME remove radiation? Toxin weapons and ambulance can remove it, no?

MTKing4:
I like your idea commy, you're leaning towards having 1 puddle cleaner for each faction, but it's still missing something, china armies can still clean puddles with all of their flame weapons, it's not just the Dragon tank, and even though it might be beneficial for them, it's not a consistent design

If we make it so that only ambo, toxin truck, dragon tank can clear toxin/radiation, I'm all for it, otherwise we go with the current route, which i still don't like due to toxin general being able to clean toxins with pretty much every unit, even demo trap

Imo if we go with the current implementation, we just don't do the streaming part, I don't like it and I don't see the obligation to have it all or nothing, i think it's fine if you can only clear with sprinkler, we really don't need to clean with force firing, it's unintuitive and uncreative and it's bad for balance

xezon:
I was giving FLAME vs Toxin some thought and I am ok with it. It means China cannot remove Radiation anymore, but on the other hand it gives them a large arsenal of weapons to remove Poison, which makes it an interesting concept.

This means

### USA

* Has 1 cleaner unit, USA Ambulance, immune to Radiation and Poison
* Can clean Radiation and Poison

### GLA

* Has 1 cleaner unit, Toxin Tractor, immune to Poison only
* Can clean Radiation and Poison

### China

* Has many cleaner units, all flame weapons, not immune
* Can clean Poison only

Do we want this?

MTKing4:
This doesn't feel right, why can't china clean something they themselves can create? This doesn't make sense and it also not a good idea

This gives Nuke General the advantage as now they're the only china faction that can clear both toxin and radiation

What is the issue with flames being able to clear radiation? Doesn't make sense? And toxin cleaning radiation makes sense?

xezon:
It is what commy2 suggested. Flame weapons kill poison, but not radiation. And enemy radiation would no longer kill radiation. Neither would enemy poison kill poison.

commy2:
>What is the issue with flames being able to clear radiation? Doesn't make sense?

Probably biased, because I'm a material science guy, but the issue is that you can burn chemicals irl, but you can't burn radioactive atoms.  :D

There are two sub-choices here:

- make FLAME clear radiation & poison 
  -> + every faction puddle cleaner is equal
  -> - questionable realism
  -> - China has advantage overall by having 3 cleaners (Dragon, Inferno [except Tank], Mig [except Nuke])
- make FLAME only clear poison
  -> + somewhat balances with the fact that China has multiple cleaners
  -> + believable physics (fire burning toxins is part of the CCG China campaign story)
  -> - China cannot clear radiation

Overall I'm leaning towards "flame only clears poison", because I don't think the ability to clean up radiation is super necessary compared to cleaning up Anthrax Bombs (large radius, 80 dmg/sec, 60 sec lifetime) and Toxin Tractor Sprinkler abilities (med radius, med damage, up to 180 sec lifetime). The Nuke Missile field lasts only 30 seconds and has average damage (33 dmg/sec).

Also, radiation exists only when facing other China, so the disadvantage is actually mutual in a 1v1.

> This gives Nuke General the advantage as now they're the only china faction that can clear both toxin and radiation

Nuke would not have an advantage, as Nuke BMs and Nukelords could no longer clear puddles at all. In fact, it would have a slight disadvantage, because it doesn't have Flame Migs, but I think that disadvantage is so small that you can file it under faction difference. Just get a Dragon when facing GLA to guard your base.

xezon:
Then the last pain point is the Demo Tractor. It would be great if we could somehow keep its suicide ability and still have the decontamination. Perhaps there is some trickery we can do by spawning an OCL with the suicide weapon?

commy2:
> Perhaps there is some trickery we can do by spawning an OCL with the suicide weapon?

Perhaps, but in that case you need to edit every other Demogen unit to use the same trick, because the button action type has to be identical for all units, otherwise they shadow each other out when selecting multiple units.

Could also remove the sprinkler instead, but I think the Sprinkler is ten times more important than Suicide.

xezon:
Removing the sprinkler is not a good alternative, as it would affect all GLA factions and make the Tractor poorer for it.

Removing the Suicide Ability from Tractor would also break the suicide group selection.

commy2:
> Removing the Suicide Ability from Tractor would also break the suicide group selection.

True, but there are other units that do this already like Angry Mob, Bomb Truck, Terrorist and Terror Car. Plus anything captured with Jarmen Kell.

Group selection of Suicide is a feature of the patch for the most part, as positioning is completely random in 1.04. At least you're not taking anything away in that sense. Toxin Tractor does not share the position with Quads or any infantry unit in 1.04. It does share it with Rocket Buggy and Scorpion though.

xezon:
One way to implement suicide is with OBJECT_UPGRADE

```
CommandButton Command_ConstructAmericaVehicleBattleDrone
  Command       = OBJECT_UPGRADE
  Upgrade       = Upgrade_AmericaBattleDrone
  Options       = OK_FOR_MULTI_SELECT NOT_QUEUEABLE
  TextLabel     = CONTROLBAR:ConstructAmericaVehicleBattleDrone
  ButtonImage   = SABattleDrone
  ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
  DescriptLabel           = CONTROLBAR:ToolTipUSABuildBattleDrone
End

Object AmericaVehicleHumvee
  Behavior = ObjectCreationUpgrade ModuleTag_06
    UpgradeObject = OCL_AmericanBattleDrone
    TriggeredBy   = Upgrade_AmericaBattleDrone
    ConflictsWith = Upgrade_AmericaScoutDrone Upgrade_AmericaHellfireDrone
  End

ObjectCreationList OCL_AmericanBattleDrone
  CreateObject
    Offset = X:0 Y:0 Z:12 ; Patch104p @tweak from X:0 Y:0 Z:10
    ObjectNames = AmericaVehicleBattleDrone
    IgnorePrimaryObstacle = Yes
    Disposition = LIKE_EXISTING
    Count = 1
    RequiresLivePlayer = Yes
  End
End
```

For that we could use our `Upgrade_DummyUpgrade`. The downside of this approach is that the button would not work properly when there is another object upgrade in the build queue, which currently would only be the case with Bomb Truck (but it has no demo suicide button).

commy2:
I think that approach would remove damage bonus from veterancy. I am also not sure if the "kill yourself" logic of SUICIDE damage would apply if the weapon is fired by dummy object and not the unit itself. Remember that Suicide does not harm allies, which would include yourself if the dummy does the weapon firing.

Eventually, you could patch Thyme to add a QUATERNARY slot lol

commy2:
Eventually, you could patch Thyme to add a QUATERNARY slot lol

MTKing4:
> Probably biased, because I'm a material science guy, but the issue is that you can burn chemicals irl, but you can't burn radioactive atoms. :D

Well if you want to make this game ultra realistic as well as balanced, and you're already adding a decontaminate ability to toxin tractors, why not add it to dragon tanks as well and be done with it?

That way all factions have 1 cleaner unit that can clean both toxin and radiation.

I really don't like the idea that not only are china cleaners not immune to neither toxin or radiation, but they also can only clean just one type of hazard, which makes for a completely flawed and frustrating design.

While it is true that the impact is only on china mirrors, that doesn't make it any less important, gla also only faces toxins on gla mirrors, and most of the toxin cleaning issue started there actually, and that can happen in china mirrors as well, i think china mirrors balance is just as important as gla mirrors and it should be considered, Nuclear missiles are buffed now, which means there's more chance of this happening, and it will impact results of games.

Not to mention that glas can also get china tech and use radiation weapons

xezon:
> Well if you want to make this game ultra realistic as well as balanced, and you're already adding a decontaminate ability to toxin tractors, why not add it to dragon tanks as well and be done with it?

I do not like this. Seeing hazard cleanup goo from Dragon Tank would be disappointing.

> I really don't like the idea that not only are china cleaners not immune to neither toxin or radiation, but they also can only clean just one type of hazard, which makes for a completely flawed and frustrating design.

It is not ideal indeed. It is as if one faction could not clear mines.

For me the major blocker for commy2's vision is the Demo Tractor. It would lose functionality.

I tried another approach with

```
CommandButton Command_GLAToxinTractorCleanupGround
  Command                 = FIRE_WEAPON
  WeaponSlot              = PRIMARY ; Best if this matches up with an AutoChoose listing that forbids this slot normally
  Options                 = NEED_TARGET_POS OK_FOR_MULTI_SELECT USES_MINE_CLEARING_WEAPONSET
  TextLabel               = CONTROLBAR:DisarmMinesAtPosition
  ButtonImage             = SSClearMines_L
  ButtonBorderType        = ACTION ; Identifier for the User as to what kind of button this is
  DescriptLabel           = CONTROLBAR:ToolTipDisarmMinesAtPosition
  RadiusCursorType        = CLEARMINES
  InvalidCursorName       = GenericInvalid
End

Object Demo_GLAVehicleToxinTruck
  WeaponSet
    Conditions        = None
    Weapon            = PRIMARY     ToxinTruckGun
    Weapon            = SECONDARY   ToxinTruckSprayer
    Weapon            = TERTIARY    TerroristSuicideNotARealWeapon
    AutoChooseSources = TERTIARY    NONE
    AutoChooseSources = SECONDARY   NONE ;Special attack only
  End
  WeaponSet
    Conditions        = MINE_CLEARING_DETAIL
    Weapon            = PRIMARY     AmbulanceCleanHazardWeapon
    Weapon            = SECONDARY   ToxinTruckSprayer
    Weapon            = TERTIARY    TerroristSuicideNotARealWeapon
    AutoChooseSources = TERTIARY    NONE
    AutoChooseSources = SECONDARY   NONE ;Special attack only
    Conditions        = None
  End
```

But somehow it does not work. I suspect it quickly loses the MINE_CLEARING_DETAIL on its own.

commy2:
> While it is true that the impact is ~~mostly~~ Only on china mirrors, that doesn't make it any less important,

It only affects China mirrors, as no other faction ever spawns radiation. What makes radiation decontam less important is the fact that nuclear puddles are much rarer, deal less damage and disappear faster than the worst poison puddles (only Anthrax Bomb and MAYBE the Tractor are relevant).

Not being able to clean up radiation as non Nuke China is a non issue in 1.04. I don't see how it would become one in this patch.

Making the Dragon immune to poison is a nogo, as that would break matches vs Toxin Tunnels. Maybe radiation resistance, but there is imo insufficient reason to make that change, as, again, radiation puddles are mostly irrelevant.

---

xezon, from my understanding, mine clearing needs WorkerAIUpdate, which would enable the Tractor to repair, among other things.

xezon:
@MTKing4 
I have now reduced hazard cleanup damage of Toxin Tunnel and Toxin Rebel by 60%. With this setup it takes 4 Toxin Tunnels 16.66 seconds to cleanup Anthrax Bomb. 2 Toxin Tractors take 13.33 seconds.

The table is updated.

MTKing4:
> It only affects China mirrors

Obviously a typo, i meant only, as i highlighted to you above that it's almost the same issue with glas, the matchup that the anthrax bombs issue originated from was in gla mirrors, either getting cleaned too fast for some, or too slow for others.

> Not being able to clean up radiation as non Nuke China is a non issue in 1.04. I don't see how it would become one in this patch.

It will become an issue soon when people realize they can clean only one puddle and not the other, your material sicence argument shouldn't be the defining factor here, we don't need realism in a game that's so far away from it, we need fairness, and gameplay that makes sense, people have been so used to the fact that units who can remove puddles they can remove all kinds of puddles wether it's radiation or toxin, you are creating a disconnect here with this change, people will be very confused, and probably as frustrated as i feel right now knowing how china found a new thing to suck at
Please keep your material sicence degree out of this :P

Also with this change, China nuke will loose out on an ability that they used to have, they can no longer clean radiation, that reads to me as loss of functionality

You're right about the toxin immunity point, but i wasn't demanding that it should have it I'm just highlighting that dragon tank is the worst cleaner of the 3 factions, having radiation immunity will be nice though, but the most important part that it can't clean radiation, i hate that.

MTKing4:
> I do not like this. Seeing hazard cleanup goo from Dragon Tank would be disappointing.

It doesn't necessarily have to be goo, could be something else like throwing a special kind of flame or Gasoline or something similar, or just make the stream orange looking, that is closer to flame liquids, like motlen stuff

MTKing4:
> @MTKing4 I have now reduced hazard cleanup damage of Toxin Tunnel and Toxin Rebel by 60%. With this setup it takes 4 Toxin Tunnels 16.66 seconds to cleanup Anthrax Bomb. 2 Toxin Tractors take 13.33 seconds.
> 
> The table is updated.

That is better indeed, but at this point it feels like each of us is biased to their own ideas as I'm thinking 1 cleaner per faction is the best design with the inclusion of my suggestion to the dragon tank being able to decontaminate radiation :P

We could use a fourth opinion

xezon:
There are many different ways to tackle this. The presented solution here is one way to do it. The advantage is that it does not take away any of the original ways to remove puddles. It adds new ones and achieves consistency.

This change could also be implemented differently, by letting puddles take damage by poison and flame weapons. That way the damages would be bound to the real weapons and it would scale with all upgraded damages automatically. It is not clear to me if that would be better. It would make the setup easier, but take away the liberty to tweak hazard damages independently.

Edit: On second thought, we cannot make Poison Fields take damage by Poison, because then they would take continuous damage from other Poison Fields, which would be similar to #2020, which we already tested and found not working properly because Poison Fields would eventually remove each other prematurely.

So this change is the proper implementation.

xezon:
Here is high level overview of hazard cleanups, Original vs Route 2 (This) vs Route 4 (commy2, Realism).

## Original

| Weapon                 | damages enemy Radiation Field | damages enemy Poison Field |
|------------------------|-------------------------------|----------------------------|
| Ambulance Hazard Clean | yes                           | yes                        |
| Tractor Hazard Clean   | -                             | -                          |
| Radiation Field        | yes                           | yes                        |
| Poison                 | no                            | no                         |
| Poison Field           | yes                           | yes                        |
| Flame                  | no                            | no                         |

## Route 2 (This)

| Weapon                 | damages enemy Radiation Field | damages enemy Poison Field |
|------------------------|-------------------------------|----------------------------|
| Ambulance Hazard Clean | yes                           | yes                        |
| Tractor Hazard Clean   | -                             | -                          |
| Radiation Field        | yes                           | yes                        |
| Poison                 | yes                           | yes                        |
| Poison Field           | yes                           | yes                        |
| Flame                  | yes                           | yes                        |

## Route 4 (Realism)

| Weapon                 | damages enemy Radiation Field | damages enemy Poison Field |
|------------------------|-------------------------------|----------------------------|
| Ambulance Hazard Clean | yes                           | yes                        |
| Tractor Hazard Clean   | yes                           | yes                        |
| Radiation Field        | no                            | no                         |
| Poison                 | no                            | no                         |
| Poison Field           | no                            | no                         |
| Flame                  | no                            | yes                        |

commy2:
Route 4 has the advantage of not giving Nuke and Tox free cleanup in the form of Nuke BM and Toxin Shells Scorpion (Tox Tunnel).

I wouldn't call it the "realism" approach. It is the "fix the bug that gimps Anthrax Bomb and gives some factions an unfair advantage, and compensate by introducing a dedicated cleanup unit to every faction" approach.

There is no reason why toxin puddles should clean up/delete toxin puddles in the first place.

It's also the least intrusive approach with the fewest changes. Fix bug (one line), add ability, change 1 armor set.

--- 

There is a route 4a (fix bug) and route 4b (fix bug + realistic).

Route 4a (fix bug)
Flame | yes | yes
-- | -- | --

Route 4b (fix bug + realisitc)
Flame | no | yes
-- | -- | --

xezon:
If you can find a nice solution for the Demo Tractor then I am ok with Route 4, otherwise I think Route 2 is ok as is.

commy2:
The best possible solution for the Demo Tractor without engine changes is removing the Suicide ability, which is not a big deal, because nobody ever uses Toxin Tractors to suicide and not all Demogen units have that ability in the first place.

A huge unresolved issue with Route 2 is the inherent advantage it gives to Nuke and especially Tox, and to a lesser degree, the disadvantage it gives to Stealth and Demo against GLA with Toxin Shells in xGLA mirrors.

xezon:
> A huge unresolved issue with Route 2 is the inherent advantage it gives to Nuke and especially Tox.

This was already mitigated with #2023. The small puddles no longer do much damage against big puddles. I think it is no issue. Toxin and Nuke have perks. Them being able to kill Infantry units easier is a much bigger perk in comparison with the hazard cleanup.

commy2:
Mitigated, but not resolved. It is an issue in 1.04 as well and Route 4 fixes it entirely.

xezon:
> The best possible solution for the Demo Tractor without engine changes is removing the Suicide ability, which is not a big deal, because nobody ever uses Toxin Tractors to suicide and not all Demogen units have that ability in the first place.

Is it perhaps possible to do something with WEAPON_RIDER1 WeaponSet? Have a CommandSet toggle, like the Worker to activate Decontamination Mode and Decontamination Weapon abilities?

commy2:
WEAPON_RIDER1 breaks Jarmen sniping. The Toxin Tractor would explode instead of leaving an unmanned vehicle.

xezon:
Perhaps we can use another WeaponSet condition?

commy2:
xezon, it is not feasible to give a unit more than 3 weapons without engine changes. All methods are hacks that wouldn't work with the constraints of a self kill ability or would have huge drawbacks that would have to be spread to all Demogen units, because of command set button mass select compatibility.

It was okay to not give Terrorists and Bomb Trucks Suicide for mass select. I don't get why it is an issue here.

How feasible is it to give a QUARTERNARY slot via Thyme? As long term solution I mean.

xezon:
I would expect it to be possible to expand the amount of unit weapons without too much hassle.

commy2:
That would be a good feature to have regardless of this issue.

xezon:
Ok so go Route 4b, remove Demo Tractor Suicide for now, re-add it later when possible.

Can we also put Ambulance and Tractor hazard cleanup weapon on same slot so it can be used in group selection?

commy2:
I wouldn't do that. They would have to use the same weapon slot, but that would change the Force Fire behaviour of these units, because the PRIMARY slot is used for that. Ambulance uses cleanup for force fire, Toxin Tractor uses its goo gun.

In general, cross faction button compatibility is 100 times less important that inner faction compatibility. Unless you really really like Casino maps I guess.

xezon:
Hmm. It would be nice if Tractor can also use force fire for the Hazard Cleanup. Perhaps put Sprinkler in TERTIARY and add a weapon toggle like for Ranger Machine Gun and Flash Bang?

commy2:
Weapon toggle doesn't work well with units with TERTIARY weapons from memory, but can look into that.

xezon:
I mean put Toxin Nozzle and Hazard Nozzle in PRIMARY and SECONDARY, and then have toggle button in Command Set. Then I think force fire would work for either.

xezon:
There are names for many weapons

primary
secondary
tertiary
quaternary
quinary
senary
septenary
octonary
nonary
denary

This is retarded :D

commy2:
You can just name them WEAPON_4, WEAPON_5 etc.

MTKing4:
I would go with route 4a with flames being able to clean radiation, remove detonation button for now and re-add with thyme

ExiLe-ZH:
## Feedback

I stopped replying to the subject since you guys were invested and dedicated to work on a solution I don't like much. Now it's still better than the original so that's why I accepted and ignored. But seeing commie’s feedback I’d like to speak up as well.

So here’s my opinion: I never liked the entire idea of stack deletion, neither with a health system. I feel none of these units except the Ambulance should be able to clean up toxins, it doesn't make much sense either, Toxin Tractors, Toxin Rebels, Toxin Tunnels etc. they're all weapons of war, not cleaners.

A health system isn’t very transparant for the player either, there’s no visual feedback whatsoever (like HP bars), so how is a new player supposed to discover how this all works? One isn’t likely to try and force fire a unit for like 10 seconds in attempt to clear Anthrax. And how will it be described ingame?

## My proposal

My take to balance everything would be to only keep the Ambulance's cleaning ability, but nerf it so that Anthrax Bombs can’t be cleared at an instant. The result should be that low armored units like Humvees can’t be saved from a direct hit. 

Secondly, nerf the Anthrax Bomb's duration and/or potentially it’s damage.

Now It will both a buff and nerf for GLA but I don't think there are any red flags when it comes to balance. In GLA mirrors Anthrax bombs aren't even that deadly when used on the production, it just denies low HP unit production, other units can be saved into tunnels. A good hit in the open should simply be deadly and the only counter should be a dodge.

For China I do like the idea of flamewall clean ups though, also for balance reasons, but it has to be intuitive, well-described and preferably have clear visual feedback. Like it would be really awesome if the fire spreads over the entire Anthrax field before disappearing, the same sort of effect you have when burning down trees.

MTKing4:
Well for gla there will be visual feedback since we're literally adding a decontamination ability :D

ExiLe-ZH:
> Well for gla there will be visual feedback since we're literally adding a decontamination ability :D

I don't like it. GLA doesn't need it either. GLA mirror late game would be more exciting if the Anthrax Bomb had more potential to actually "swing" games.

xezon:
Stealth General is quite screwed when Anthrax Bomb falls on Production sites. Infantry General too.

Perhaps add a new unit upgrade to Toxin Tractor that will change its weapons to hazard cleanup stream and hazard cleanup sprinkler permanently. So it sort of becomes a different unit upon choice.

ExiLe-ZH:
> Stealth General is quite screwed when Anthrax Bomb falls on Production sites. Infantry General too.

--->

> Secondly, nerf the Anthrax Bomb's duration and/or potentially it’s damage.

Quads are savable anyway, the player just needs to make sure to make his tunnel close to the Arms Dealer.

MTKing4:
> I don't like it. GLA doesn't need it either. GLA mirror late game would be more exciting if the Anthrax Bomb had more potential to actually "swing" games.

Well that would be the most extreme approach yet, a big change that we're trying to avoid, players would expect to be able to clean puddles, removing that ability will have huge impact on balance

Having a counter system is nice in game design, and is necessary in this situation where you have impactful weapons such as the A-Bomb

xezon:
What I like about the hazard cleanup is that it makes the game richer in functionality. Taking it away can make player feel helpless and not in control of the outcome.

xezon:
As for Anthrax Bomb damage, I have put changes here: #2044, #2045

ExiLe-ZH:
> @xezon Perhaps add a new unit upgrade to Toxin Tractor that will change its weapons to hazard cleanup stream and hazard cleanup sprinkler permanently. So it sort of becomes a different unit upon choice.

> @MTKing4 Well that would be the most extreme approach yet, a big change that we're trying to avoid, players would expect to be able to clean puddles, removing that ability will have huge impact on balance

It's how it was supposed to be, now it's a pointless level 5(!) genpower.

> @xezon What I like about the hazard cleanup is that it makes the game richer in functionality. Taking it away can make player feel helpless and not in control of the outcome.

Not every faction needs cleanup, not every single power needs a direct counter either. You can counter the plane already anyway.

MTKing4:
Xezon actually tweaked puddle damages nicely, it takes around 13 seconds to get cleaned, a lot of damage is done by then, it's not pointless

xezon:
Puddle health and cleanup damages can be adjusted to our liking. If cleanup is too fast or slow -> tweak.

commy2:
Puddles could be changed such that they display flame particles upon deletion or while taking FLAME damage.

MTKing4:
Also with that approach you're not removing puddle cleaning entirely, you still have the ambulance, which will make it quite unfair that only one faction can clear puddles, if we take your approach we should remove ambulance cleaning as well, otherwise it doesn't make sense, it's either they all have this ability or no one does

ExiLe-ZH:
> Xezon actually tweaked puddle damages nicely, it takes around 13 seconds to get cleaned, a lot of damage is done by then, it's not pointless

I already left my detailed feedback regarding that. 

I simply prefer the game to be more like it was suppossed to be (more realistic). Plus it's obvious the Anthrax bomb was never intended to be counterable by so many different units/weapons. Again, the plane itself is already counterable.

ExiLe-ZH:
And what about radiation, every faction needs a radiation cleaner unit aswell now?

MTKing4:
> I simply prefer the game to be more like it was suppossed to be (more realistic). Plus it's obvious the Anthrax bomb was never intended to be counterable by so many different units/weapons. Again, the plane itself is already counterable.

It clearly was intended to be countered by the ambulance though

> And what about radiation, every faction needs a radiation cleaner unit aswell now?

USA and GLA and all chinas¹ (exept Tank) can already clean radiation, it's just after removing puddle deletion china will no longer be able clean radiation unless we add it to the dragon tank.

---
¹ ChinaV, Infantry have Nuke Cannons, China nuke have BMs, Overlords, Nuke MiGs and Nuke Cannons, China tank has none.

xezon:
China Tank can clean Radiation and Poison with Nuke Missile and Reactor death. It is not very practical though.

commy2:
China Tank can also suicide Nuclear Tank BMs and Emperors.

MTKing4:
> China Tank can clean Radiation and Poison with Nuke Missile and Reactor death. It is not very practical though.

i did not consider them because those are not actually cleaning units, it's more like a suicide if anytihng, not to mention how long it takes for an overloaded reactor to die, or the fact that you can't kill it or the nuke with units since there is already a cloud. so no.

commy2:
Not all my replies are to be taken seriously, even if the tone is very dry. Sorry :D

MTKing4:
ik, the response wasn't meant for you, but yeah i like to be overcritical :P

PR #2050, opened 2023-06-29
Fix: Incorrect clip size and reload time of China Dragon Tank flame thrower with Black Napalm

xezon:
* Similar to #1588

This change fixes the incorrect clip size and reload time of the China Dragon Tank Black Napalm flame thrower.

It is not clear to me if this is a bug per se, but now it is consistent with the setups of the other stream weapons. For human matches nothing changes.


xezon:
Documentation added and updated. Linked to other related pull. Looks good.

PR #2051, opened 2023-06-29
Add dedicated armor class to USA Ambulance

xezon:
This change adds a dedicated armor class to the USA Ambulance. This way it no longer shares the same armor definition with the Toxin Tractor, but keeps the original settings.


xezon:
Looks good.

PR #2052, opened 2023-06-29
Fix: Remove radiation hazard immunity from GLA Toxin Tractor

xezon:
* Follow up for #2051

This change removes the radiation hazard immunity from the GLA Toxin Tractor. It now has the same resistance as the China Dragon Tank, 50%.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2051_toxin_tractor_radiation_armor.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2051_toxin_tractor_radiation_armor.yaml
@@ -1,0 +1,22 @@
+---
+date: 2023-06-29
+
+title: Changes radiation armor of GLA Toxin Tractor from 0% to 50%
+
+changes:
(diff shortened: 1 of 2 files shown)


MTKing4:
I don't feel strongly about this, realistically it shouldn't be immune to anything, so why this matters? Better to not change it when there are no balance reasons

xezon:
The Toxin Tractor is labeled specifically to be immune to Toxins in code. I think it was an oversight that it is immune to Radiation. No other GLA unit is. The Dragon Tank is immune to Fire, but not immune to Poison and Radiation.

I think the Radiation immunity of the Tractor can be considered a bug, because by design the expectation is that everything in contact with Poison and Radiation will take damage, unless there are specific reasons not.

The Toxin Tractor spills Toxins, so it must be able to stand on toxins. That is reasonable. But it does not own Radiation abilities, and therefore having Radiation immunity is counter-intuitive and wrong.

I think this change is proper, and the gameplay impact is negligible. We can increase armor against Radiation, so impact will be less. 50% is not bad for the Tractor. 25% armor would be pretty good.

xezon:

PR #2054, opened 2023-07-01
Change: Infantry paradrop parachutes can be controlled by player

xezon:
* Relates to #705

With this change all infantry paradrop parachutes can be controlled by the owning player. Tank paradrop parachutes can not be controlled.

## Sample execution 1

In this sample a few of the initial parachutes are selected, then E button is pressed and move command is given several times to select and move all parachutes.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/35807a68-60a0-4d66-b79a-0e05bcdb5667

## Sample execution 2

In this sample a few of the initial parachutes are selected, then E button and X button are pressed several times to select and scatter all parachutes.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/892c62e1-b318-4b0f-9604-0309be7cd441

# Rationale

This change slightly increases the usefulness of the infantry paradrop abilities by expanding its capabilities. It allows the player to optimize the drop location on arrival. This can make a difference in how successful the paradrop becomes, depending on the terrain conditions underneath. It adds more depth to the game and requires a bit of micro and attention to execute.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4815,7 +4815,7 @@ Object AirF_ComancheBlades
 End
 
 ;------------------------------------------------------------------------------
-Object AirF_AmericaParachute
+Object AirF_AmericaParachute ; unused
 
   ; *** ART Parameters ***
   Draw = W3DModelDraw ModuleTag_01
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
@@ -143,6 +143,90 @@ Object AmericaParachute
 End
 
 ;------------------------------------------------------------------------------
+; Patch104p @feature xezon 01/07/2023 Adds new parachute identical to AmericaParachute,
+;   but selectable and controllable for the duration of its lifetime.
+Object AmericaControllableParachute
+
+  ; *** ART Parameters ***
+  Draw = W3DModelDraw ModuleTag_01
+    OkToChangeModelColor = Yes
+    DefaultConditionState
+      Model = PMparacht_SKNc
+      Animation = PMparacht_SKL.PMparacht_FAL
+      AnimationMode = LOOP
+    End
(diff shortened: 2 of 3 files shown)


commy2:
Unsure if this actually helps or is just a change for the sake of changing stuff (i.e. bad)

xezon:
I would imagine there are scenarios where this is useful:

1. Paradrop near a collection of enemy airfields. Scattering infantry increases the area they can attack from and reduces the chance of single strikes taking out the whole group. For example by a Mig Napalm strike.

2. Paradrop on unexpected situation. By the time the Paradrop arrives there could be a change of conditions, for example a Toxin Tunnel could be built, or a Dragon Tank could be placed. In that case it can be beneficial to have a method to not let the paratroopers fall to certain death. Enemy player does not know where Paradrop lands, so this would almost always be coincidence, rather than defensive skill.

ExiLe-ZH:
It's super cool though

xezon:
MTKing
> Looks nice. Considerable buff to parachutes. This will definitely have balance implications on certain matchups, especially for minigunners dropping at enemy airfields, migs will need many more shots to clear all units due to the ability to scatter the units before landing

Jundiyy
> Seems like a safe change. The more important reason is that it saves units from dying when parachuting on to mountains.

Battlefieldkille:
This looks very dope

kABUSE1:
You need to test this against dropping above water, then ordering parachutes to go to beach.

What happens if you press "S", garrison or attack something while they are still airborn?

xezon:
> You need to test this against dropping above water, then ordering parachutes to go to beach.

Dropping above water is not possible. It only allows to drop above terrain. But it is possible to steer soldiers into water.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/94329f90-a240-4705-8d21-de5643fb3b9e

Dropping into mountains is possible.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/62f380ff-4037-460f-b960-2cdaff1c65cc

> garrison or attack something while they are still airborn?

The soldiers are not selectable in air. It is the parachute that is selected - without the soldier.

> What happens if you press "S"

It stops the parachute.

Behavior looks as expected.

There appears to be an issue with the Missile Defender land animation transition. It goes into free fall shortly before death.

PR #2056, opened 2023-07-01
Fix: Optimize drop location and spacing of USA Paradrops

xezon:
This change optimizes the drop location and spacing of the USA Paradrops. The soldiers now fall into a line and no longer charge towards to same position. The spacing is tightened consistently and they fall as close to the drop center as possible.

Optimized for fix https://github.com/TheAssemblyArmada/Thyme/pull/752.

## Original

Paradrop 1 spacing is not tight. Soldier fall too far behind.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/8bb69c38-ede8-4d90-9ea4-e764ca4e9666

Paradrop 3 is poorly executed. Depending on drop angle, planes will return a second time.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/22f0cba7-598f-4922-906e-b7cebcf9b98c

## Patched (+Thyme 752)

Paradrop 1 is nice.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/fbb671fc-6c0b-42af-812b-8858d2735179

Paradrop 3 is nice.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/d1a435a4-054f-4826-abfc-c0b5eed4039f

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
@@ -6182,6 +6182,9 @@ End
 
 ; -----------------------------------------------------------------------------
 ; Patch104p @tweak xezon 21/06/2023 Changes payload from 5 Rangers. (#2026)
+; Patch104p @tweak xezon 01/07/2023 Disables ParachuteDirectly to drop payload nicely spaced in a line.
+; Patch104p @tweak xezon 01/07/2023 Decreases DropDelay from 150 to tighten spacing.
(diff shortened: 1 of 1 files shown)


commy2:
I notice that each plane drops 6 rangers, then 4 mds next to each other. Maybe we can mix it up a bit?

Instead of

    Payload = AmericaInfantryRanger 6
    Payload = AmericaInfantryMissileDefender 4

have

    Payload = AmericaInfantryRanger 3
    Payload = AmericaInfantryMissileDefender 2
    Payload = AmericaInfantryRanger 3
    Payload = AmericaInfantryMissileDefender 2

PR #2057, opened 2023-07-01
Reorder and balance payload of USA Paradrops

xezon:
* Follow up for #2026
* Follow up for #2056

This change reorders and balances the payload of the USA Paradrops. The Ranger and Missile Defender drop delays are now more evenly distributed.

## Before (#2026)

shot_20230701_200340_2

## After (This + #2026)

shot_20230701_195124_1


commy2:
I just suggested something like this. :+1:

commy2:
Made obsolete by https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2272

PR #2058, opened 2023-07-02
Fix: Match visuals of toxin and hazard cleanup stream splash effects with nearby puddle effects

xezon:
* Follow up for #2036
* Follow up for #2041

This change matches the visuals of toxin and hazard cleanup stream splash effects with the puddle effects.

## Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/4301659b-c22c-45c8-b110-9a786e95de5c


xezon:
Documentation is added. Looks good.

PR #2063, opened 2023-07-06
Fix: GLA Toxin Tunnel stream no longer times out

xezon:
With this change the GLA Toxin Tunnel stream no longer times out.

It mostly is a problem with targeting the ground or allied units. When targeting enemy units, the stream will collide and deal damage up to max range. So this fix luckily is no practical buff.

## Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/a9a3a27c-a540-4658-8119-ebd8639acc0e


commy2:
What the heck.

xezon:

PR #2064, opened 2023-07-07
Change: Decrease kill experience reward of veteran USA Burton to match other heroes

### LINKED ISSUE #2025: Hero units give little kill XP
Hero units give little kill XP

```
ChinaInfantryBlackLotus
  ExperienceValue = 50 50 100 150 ;Experience point value at each level

AmericaInfantryColonelBurton
  ExperienceValue = 50 100 100 150  ; Experience point value at each level

GLAInfantryJarmenKell
  ExperienceValue = 50 50 100 150  ;Experience point value at each level
```

## Proposal 1

Raise all Heros to 75 100 150 300.

## Proposal 2

Raise all Heros to 75 100 150 200.

## Proposal 3

Make Burton 50 50 100 150 only.

## Proposal 4

Raise all Heros to 75 75 125 175.

xezon:
* Resolves #2025

This change streamlines the kill XP progression of USA Burton with other heroes and infantry.

### Before

```
ChinaInfantryBlackLotus
  ExperienceValue = 50 50 100 150  ; Experience point value at each level

AmericaInfantryColonelBurton
  ExperienceValue = 50 100 100 150  ; Experience point value at each level

GLAInfantryJarmenKell
  ExperienceValue = 50 50 100 150  ;Experience point value at each level
```

### After

```
ChinaInfantryBlackLotus
  ExperienceValue = 50 50 100 150  ; Experience point value at each level

AmericaInfantryColonelBurton
  ExperienceValue = 50 50 100 150  ; Experience point value at each level

GLAInfantryJarmenKell
  ExperienceValue = 50 50 100 150  ;Experience point value at each level
```

# Rationale

Originally the Vet 1 Burton gives twice as much kill experience as the other Heros for no apparent reason. It is not particularly harder to kill or more dangerous than the Vet 1 Jarmen Kell.

The Burton kill experience progression mismatches with the global unit experience progression, which typically looks like so: X X Y Z.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2064_burton_experience_reward.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2064_burton_experience_reward.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-07-07
+
+title: Decreases kill experience reward of veteran USA Burton to match other heroes
+
+changes:
+  - tweak: The USA Burton now rewards 50 50 100 150 instead of 50 100 100 150 experience on kill.
+
+labels:
+  - controversial
+  - design
+  - minor
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2064
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -3114,7 +3114,8 @@ Object AirF_AmericaInfantryColonelBurton
   BuildCost = 1500
   BuildTime = 20.0          ;in seconds
 
-  ExperienceValue     = 50 100 100 150  ; Experience point value at each level
+  ; Patch104p @tweak xezon 07/07/2023 Changes experience value from 50 100 100 150 to match the other heroes. (#2064)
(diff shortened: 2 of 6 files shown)


commy2 (approved):

ImTimK:
Fine change.

OoE-ExCaL:
Another example of  why I think "streamlining and uniforming values", is a super bad approach:


I agree that kill XP could be raised for the heroes(small value).  I totally disagree with the streamlining.  

Reasoning: These 3 heroes are very much different in gameplay, and vary in terms of how hard they are to kill. Lotus I would argue is the easiest to kill therefore should give away the least xp. Jarmen and Burton are much harder to kill, with Jarmen being the most difficult to kill in my opinion.

xezon:
> Reasoning: These 3 heroes are very much different in gameplay, and vary in terms of how hard they are to kill. Lotus I would argue is the easiest to kill therefore should give away the least xp. Jarmen and Burton are much harder to kill, with Jarmen being the most difficult to kill in my opinion.

Black Lotus veterancy progression has already been improved. She can level up quicker. With self heal she will also be harder to catch like the other Heros. Additionally there is a pending change to increase her Vehicle Hack range by some around 30%.

You write Jarmen is harder to kill than Burton. I agree with this statement and it supports this change. Vet 1 Burton should not give twice the kill reward.

Rationale text is left to be added.

xezon:

PR #2034, opened 2023-06-24
Optimization: Decrease cost of toxin and hazard cleanup stream trail particles by around 50%

xezon:
**Merge with Rebase**

* Relates to TheSuperHackers/GeneralsGameCode#124

This change decreases the cost of toxin and hazard cleanup stream trail particles by around 50%. The overall performance improvement of the stream weapon is significantly less than that, because there are other expensive components.

Affects

* GLA Toxin Truck
* GLA Toxin Rebel
* GLA Toxin Tunnel
* USA Ambulance

## Before

Expensive stream particle effects.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/9dfb0010-68ba-436d-8435-4d924ab0d83c

## After (Outdated)

The original look was preserved as much as possible. At a glance the effects look very similar.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/8f63d393-08b4-4599-bea5-fe3c0a5f693f

PR #2065, opened 2023-07-07
Revert system lifetime changes of stream trail effects

xezon:
* Follow up for #2034

This change reverts system lifetime changes of stream trail effects. It did not look proper on Toxin Tunnel stream, which is longer than the other ones.


xezon:

PR #2067, opened 2023-07-08
Change: Infantry killed by USA Laser Turret and Laser Crusader will burn

### LINKED ISSUE #1848: Infantry killed by lasers does not burn
From survey comments:

> Also it would be cool if infantries fried by lasers play the burning death animation instead of the standard death animations ala Reborn v4: The Rise to Power mod.

xezon:
* Closes #1848

This change makes USA Laser Turret and Laser Crusader weapons burn killed infantry. ~~This change is cosmetic only.~~

The Laser Crusader now also triggers the ~~suicide weapon~~ weak crush explosion (#2081) of Terrorist and Demolition units, which makes it more vulnerable in return, but consistent with how the ~~Laser Turret~~ point defense laser of the USA Paladin Tank behaves.

## Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/4498eacc-0a22-49b1-998b-bd1eecf96b01

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2067_laser_weapons_infantry_burn.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2067_laser_weapons_infantry_burn.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-07-08
+
+title: Enables Laser weapons to burn killed infantry units
+
+changes:
+  - tweak: The USA Laser Turret now burns infantry on kill. It does not change gameplay.
+  - tweak: The USA Laser Crusader now burns infantry on kill. The burned death now triggers the suicide explosion of the GLA Terrorist.
+
+labels:
+  - controversial
+  - design
+  - minor
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2067
+
+authors:
+  - xezon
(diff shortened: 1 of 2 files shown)


commy2:
It's cosmetic against (most) infantry, because in patch, flaming (and poisoned) infantry will no longer be auto attacked by units until they are killed again. https://github.com/TheSuperHackers/GeneralsGamePatch/pull/69/commits/2f3e88707ef93fb792c6e43afcd8481108bb3dc2

It is not cosmetic against Terrorists and Combat Bikes carrying Terrorists.

NORMAL death triggers no/CrushedBikeBomb death weapon (low damage), while BURNED triggers Suicide(Bike)Bomb weapon (high damage).

https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini#L14406-L14419

commy2 (approved):
I think that's fine though.

xezon:
Ah right, I missed that detail. I appended this information in opening text. I think it is ok. Makes the lasers act consistently.

I think as next step we could make the Death Type LASERED. But for that we would first need to add LASERED to infantry burn. LASERED will then trigger the smaller Terrorist detonation.

xezon:

PR #2069, opened 2023-07-08
Change: Remove infantry target scatter from USA Laser Crusader weapon

xezon:
* Follow up for #2068

This change removes the infantry target scatter from the USA Laser Crusader (and USA Laser Paladin). Performance wise nothing changes. The infantry target scatter in practice never missed infantry. Therefore this change ~~is cosmetic only~~ does not change gameplay.

## Before

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/c8220727-7fc1-4ef8-b88b-285d48bf58f3

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -6745,10 +6745,10 @@ Weapon GC_Slth_QuadCannonSnipeGun
 End
 
 ;------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


commy2:
My (objective?) threshold for cosmetic changes was if it can corrupt a replay if viewed with or without the change. I would not call this cosmetic, just a negligible gameplay impact (I haven't tested this).

xezon:
Wording fixed.

MTKing4:
Does it still kill the infantry with the same amount of shots for all infantry types anb mobs?

PR #75, opened 2021-08-27
Fix: GLA defense hole survival

alanblack166:
GLA Defence Holes will no longer let the player survive if he is reduced to nothing but holes spawned from defences.

PR #2072, opened 2023-07-08
Fix: Point Defense Laser of USA Paladin no longer attacks dead infantry

### LINKED ISSUE #2070: USA Paladin's PointDefenseLaserUpdate does not target new infantry after a kill reliably
The USA Paladin's PointDefenseLaserUpdate does not target new infantry units after a kill reliably. It takes several seconds before shooting again for no apparent reason. When not killing a target, then it has no problem switching targets.

```
  Behavior = PointDefenseLaserUpdate ModuleTag_07
    WeaponTemplate        = PaladinPointDefenseLaser
    PrimaryTargetTypes    = BALLISTIC_MISSILE SMALL_MISSILE
    SecondaryTargetTypes  = INFANTRY
    ScanRate              = 500
    ScanRange             = 120.0
    PredictTargetVelocityFactor = 3.0
  End
```

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/3d51ba70-6c95-4ee1-9367-fec5ad5058b4

xezon:
* Fixes #2070
* Follow up for #75

With this change the Point Defense Laser of the USA Paladin no longer attacks dead infantry.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2072_paladin_laser_attack_dead_infantry.txt
+++ b/Patch104pZH/Design/Changes/v1.0/2072_paladin_laser_attack_dead_infantry.txt
@@ -1,0 +1,1 @@
+75_dead_target_bug.yaml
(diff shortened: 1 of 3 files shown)


MTKing4:
Next to fix: USA Paladin no longer T-bags dead infantry

commy2 (approved):
lets hope it doesn't break something else

xezon:
If it does then we will find it. Not worried about it.

xezon:

PR #2073, opened 2023-07-08
Fix: Adds missing damaged condition states to AmericaInfantryOfficer, AmericaInfantryCIAOfficer, ChinaInfantryOfficer

xezon:
This change adds the missing damaged condition states to

* AmericaInfantryOfficer
* AmericaInfantryCIAOfficer
* ChinaInfantryOfficer

This way it no longer shows wrong animation(s) when damaged.

## Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/8fe45dc3-1eec-420e-8866-3046f56b65d3

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
@@ -622,7 +622,9 @@ End
 ; This unit is never human-controllable!
 ;
 Object AmericaInfantryOfficer
+
     ; *** ART Parameters ***
+  ; Patch104p @bugfix xezon 08/07/2023 Adds missing condition states for REALLYDAMAGED.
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaMiscUnit.ini
```diff
@@ -877,6 +892,8 @@ Object AmericaInfantrySecretService
       AnimationMode = ONCE
       TransitionKey = TRANS_Stand
     End
+
+    ; Patch104p @todo Add models and condition states.
```
commy2, line 896: You plan on doing these as well?

xezon: Me, no, but someone else could if he wanted.

PR #2075, opened 2023-07-08
Fix: GLA Terrorist explosion is no longer silent when not suicided

xezon:
With this change the GLA Terrorist explosion is no longer silent when not suicided. Affects burned, exploded, crushed, splatted and lasered death. The smaller crushed, splatted and lasered death explosion gets a quieter sound.

## Before, Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/eabf7c6f-31f5-4228-aea1-f5cbaebdf217

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2075_terrorist_death_explosion_sound_fix.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2075_terrorist_death_explosion_sound_fix.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-07-08
+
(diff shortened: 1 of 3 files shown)


xezon:
In next step we can perhaps get rid of `FX_TerroristExplode` entirely.

MTKing4:
Which sound Did you use?

commy2:
It's the same sound as when exploded, but at volume 55 instead of 90.

commy2 (approved):

xezon:

PR #2076, opened 2023-07-08
Fix: Demo GLA Barracks no longer spawns Terrorist explosion effects on death

xezon:
With this change the Demo GLA Barracks no longer spawns Terrorist explosion effects on death. It would happen when the Barracks was exploded.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -8822,6 +8822,7 @@ End
 
 ;------------------------------------------------------------------------------
 ;GLA Barracks
+; Patch104p @bugfix xezon 08/07/2023 Removes wrong exploded death behaviour that would spawn Terrorist death effects. (#2076)
 Object Demo_GLABarracks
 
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -9565,19 +9566,6 @@ Object Demo_GLABarracks
...
-    SinkRate            = 0.5     ; in Dist/Sec
-    DestructionDelay    = 8000
-    FX                  = INITIAL FX_TerroristExplode
-    FlingForce          = 8
-    FlingForceVariance  = 3
-    FlingPitch          = 60
-    FlingPitchVariance  = 10
-  End
-
   Behavior = CommandSetUpgrade ModuleTag_18
```
commy2, line 9569: Do you have an explanation for these? Should remove them as well while at it.

xezon: Ok. Addressed with #2078

commy2 (approved):

xezon:

PR #2078, opened 2023-07-09
Fix: Remove obsolete Worker command sets from GLA Demo Barracks

xezon:
This change removes the obsolete Worker command sets from the GLA Demo Barracks.


commy2 (approved):
weird

xezon:
No special documentation added because this is not user facing. Looks good.

PR #2080, opened 2023-07-09
Fix: Set correct object strings for Partisan01, Partisan02

xezon:
This change sets the correct object string for Partisan01, Partisan02, because they look male. Only Partisan03 looks female. In game nothing changes, because the OBJECT:MogadishuMaleCivilian and OBJECT:MogadishuFemaleCivilian strings read the same.


commy2 (approved):

xezon:
Looks good.

PR #2085, opened 2023-07-09
Fix: Use correct model for poison deaths of GC Terrorists

xezon:
This changes adds correct model for poison deaths of GC Terrorists. For the GC Chemical Terrorist it makes no difference however, because it takes no poison damage.

shot_20230709_112031_1


commy2 (approved):

commy2:
This also prevents a bug where the barrels of the German censored version turn into green/blue/pink guys after getting poisoned in GLA02.

commy2:
Doc?

xezon:
To be done.

xezon:

PR #2089, opened 2023-07-09
Fix: Removes unnecessary line breaks in object labels of German and French localizations

xezon:
This change removes unnecessary line breaks in German and French object labels. I reviewed all OBJECT labels and these are the only ones I found. 5 in DE, 1 in FR.


xezon:

PR #2092, opened 2023-07-10
Fix: Key conflict of Satellite Hack I & II with STOP (S)

xezon:
* Relates to TheSuperHackers/GeneralsGameCode#136
* Relates to TheSuperHackers/GeneralsGamePatch#2091

This change fixes the key conflict of Satellite Hack I & II with Stop on China Internet Center.

The new button mapping is A instead of S.


commy2 (approved):

xezon:
Wow commy2 vry fst.

xezon:

PR #2093, opened 2023-07-10
Change: Streamline key mappings of China Mines upgrades

xezon:
* Relates to #2095

This change streamlines the key mappings of China Mines upgrades. This makes it a bit easier to learn and operate. Optional bundle only.

## Original

| Language   | Land Mines key | Neutron Mines key |
|------------|----------------|-------------------|
| US         | L              | M                 |
| DE         | L              | M                 |
| FR         | M              | M                 |
| ES         | M              | M                 |
| IT         | M              | M                 |
| KO         | L              | M                 |
| ZH         | L              | M                 |
| BP         | M              | M                 |
| PL         | M              | M                 |

## Patched (Optional)

| Language   | Land Mines key | Neutron Mines key |
|------------|----------------|-------------------|
| US         | M              | M                 |
| DE         | M              | M                 |
| FR         | M              | M                 |
| ES         | M              | M                 |
| IT         | M              | M                 |
| KO         | M              | M                 |
| ZH         | M              | M                 |
| BP         | M              | M                 |
| PL         | M              | M                 |


xezon:
Not clear to me if this is desired. But the L, M key mapping looks silly.

xezon:
The least we can do is put this in Optional bundle, so that new players can have better key mapping here.

MTKing4:
i think it's better to give them separate keys, L for Land mines and M for Neutron

also what happens if you select buildings with different mines?? can't upgrade either?

commy2:
You can only ever select 1 building at once, so no multi select issues on buildings.

MTKing4:
Ah yeah true.

Still, I'm used to the L key for mines and i wouldn't wanna change it

xezon:
I will demote this to Optional bundle.

xezon:
Demoted to Optional bundle.

xezon:
Documentation added. Works.

PR #2094, opened 2023-07-10
Change: Move Stop button of USA Strategy Center to common position

xezon:
This change moves the Stop button of the USA Strategy Center to the common position. Optional, to not confuse legacy players.

## Original

shot_20230710_163050_1

## Patched

shot_20230710_174112_3

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2094_strategy_center_stop_button_placement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2094_strategy_center_stop_button_placement.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-07-10
+
+title: Swaps position of Stop and Supply Lines buttons in command set of USA Strategy Center
+
+changes:
+  - tweak: The Stop button is now positioned at its regular place in the command set of the USA Strategy Center.
+
+labels:
+  - design
+  - gui
+  - minor
+  - optional
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2094
+
+authors:
+  - xezon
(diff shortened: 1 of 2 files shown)


MTKing4:
Objective Improvement, should be on core

commy2:
I don't think anyone will be confused by this. I doubt many people could even tell you where these two buttons are without looking.
Top right is the position in CCG. Devs failed to move it when command bar expanded from 12 to 14 in the transition from CCG to ZH.

xezon:
I know I would be confused because I pressed that Supply Lines button many times. So it is safer in Optional bundle.

MTKing4:
We're programed to look for the green looking portrait and click that, i never go to my strategy center and press the far top right corner of the controlbar without looking, and to that's probably how 90% of people approach it, If you had told me that the Patched picture is from the original game, i wouldn't have known, this is definitely a non issue

xezon:
Not every brain works the same. I actually click on remembered position, rather than the picture.

MTKing4:
Yeah that's why we take more than 1 opinion on this and go through with the most voted choice to be more representative of the majority of players

Besides you will probably play the Optional bundle so you are ok with adjusting to the change

ExiLe-ZH:
Good improvement, also the Supply Lines position, before I used hotkeys I always found myself hesitant to click it, always double checking to be sure I'm not pressing the Sell button. New position gives more confidence.

It should be pretty easy to adjust over time for players who don't use the hotkey.

commy2:
I don't think there are any negative consequences for pressing STOP by mistake either. All it does is stop force-firing the ground (Bombardment only, no one ever uses it). You get a distinct noise too, so unless you have the ingame sounds muted, you will notice the mistake immediately.

MTKing4:
And it's only an upgrade, the chances are you're not gonna rush buying an upgrade the same way you rush with building units

ExiLe-ZH:
Also, changing button positions isn't as controversial as changing hotkeys, because clicking buttons involves direct visual interaction, so much easier to get adjusted to.

This makes it less sensitive to human error, you might lose a second having to relocate it (including the cursor), but pressing wrong hotkeys can easily go by unnoticed because there's less interaction/feedback involved.

xezon:

PR #2100, opened 2023-07-12
Fix: Misspelled and invalid string references in CommandButton.ini

xezon:
This change fixes misspelled and invalid string references in CommandButton.ini. It does not fix bugs, because string id's are case tolerant. User facing nothing changes. Requires no special documentation.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -729,7 +729,7 @@ CommandButton Command_DetonateFakeBuilding
   TextLabel         = CONTROLBAR:DetonateFakeBuilding
   ButtonImage       = SSTerroristCarBomb
   ButtonBorderType  = ACTION
-  DescriptLabel     = CONTROLBAR:TooltipDetonateFakeBuilding
+  DescriptLabel     = CONTROLBAR:TooltipDetonateFakebuilding
 End
 
 CommandButton Command_BecomeRealGLACommandCenter
@@ -783,7 +783,7 @@ CommandButton Command_TerrorCell
   Command                 = SPECIAL_POWER
   SpecialPower            = SuperweaponTerrorCell
   Options                 = NEED_SPECIAL_POWER_SCIENCE NEED_TARGET_POS CONTEXTMODE_COMMAND
-  TextLabel               = CONTROLBAR:TerrorCell
+  TextLabel               = GUI:SuperweaponTerrorCell
   ButtonImage             = SCTempDefaultCommand
   ButtonBorderType        = ACTION ; Identifier for the User as to what kind of button this is
   DescriptLabel           = CONTROLBAR:TempDescription
@@ -1194,7 +1194,7 @@ CommandButton Command_UpgradeAmericaRangerCaptureBuilding
   TextLabel     = CONTROLBAR:UpgradeAmericaRangerCaptureBuilding
   ButtonImage   = SSCaptureBuilding
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
-  DescriptLabel           = CONTROLBAR:TooltipUSAUpgradeRangerCaptureBuilding
+  DescriptLabel           = CONTROLBAR:ToolTipUSAUpgradeRangerCaptureBuilding
 End
 
 CommandButton Command_UpgradeAmericaTOWMissile
@@ -1259,7 +1259,7 @@ CommandButton Command_UpgradeChinaRedguardCaptureBuilding
   TextLabel     = CONTROLBAR:UpgradeChinaRedguardCaptureBuilding
   ButtonImage   = SSCaptureBuilding
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
-  DescriptLabel           = CONTROLBAR:TooltipChinaUpgradeRedguardCaptureBuilding
+  DescriptLabel           = CONTROLBAR:ToolTipChinaUpgradeRedGuardCaptureBuilding
 End
 
 CommandButton Command_UpgradeChinaMines
@@ -1566,7 +1566,6 @@ CommandButton Command_UpgradeGLACamoNetting
   ButtonImage       = SUcamo
   ButtonBorderType  = UPGRADE ; Identifier for the User as to what kind of button this is
   DescriptLabel           = CONTROLBAR:ToolTipGLACamoNetting
-;  UnitSpecificSound = BombTruckVoiceModeHiEx
 End
 
 CommandButton Command_UpgradeGLARadarVanScan
@@ -1713,7 +1712,7 @@ CommandButton Command_UpgradeGLARebelCaptureBuilding
   TextLabel     = CONTROLBAR:UpgradeGLARebelCaptureBuilding
   ButtonImage   = SSCaptureBuilding
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
-  DescriptLabel           = CONTROLBAR:TooltipGLAUpgradeRebelCaptureBuilding
+  DescriptLabel           = CONTROLBAR:ToolTipGLAUpgradeRebelCaptureBuilding
 End
 
 
@@ -4176,16 +4175,6 @@ CommandButton GC_Slth_Command_ConstructGLASupplyStash
   DescriptLabel           = CONTROLBAR:ToolTipGLABuildSupplyStash
 End
 
-CommandButton Command_UpgradeCamoNetting
-  Command       = OBJECT_UPGRADE
-  Upgrade       = Upgrade_GLACamoNetting
-  TextLabel     = CONTROLBAR:UpgradeGLACamoNetting
-  ButtonImage   = SSControlRods
-  ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
-  DescriptLabel           = CONTROLBAR:CONTROLBAR:ToolTipGLACamoNetting
-End
-
-
 CommandButton GC_Slth_Command_ConstructGLABarracks
   Command       = DOZER_CONSTRUCT
   Object        = GC_Slth_GLABarracks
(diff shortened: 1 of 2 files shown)


Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -716,7 +716,7 @@ CommandButton Command_DetonateFakeBuilding
   TextLabel         = CONTROLBAR:DetonateFakeBuilding
   ButtonImage       = SSTerroristCarBomb
   ButtonBorderType  = ACTION
-  DescriptLabel     = CONTROLBAR:TooltipDetonateFakeBuilding
+  DescriptLabel     = CONTROLBAR:TooltipDetonateFakebuilding
```
commy2, line 732: Shouldn't the entry in the stringtable fixed instead? Seems like the original makes more sense.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -770,7 +770,7 @@ CommandButton Command_TerrorCell
   Command                 = SPECIAL_POWER
   SpecialPower            = SuperweaponTerrorCell
   Options                 = NEED_SPECIAL_POWER_SCIENCE NEED_TARGET_POS CONTEXTMODE_COMMAND
-  TextLabel               = CONTROLBAR:TerrorCell
+  TextLabel               = CONTROLBAR:TempDescription
```
commy2: This seems broken.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -4111,13 +4111,13 @@ CommandButton GC_Slth_Command_ConstructGLASupplyStash
   DescriptLabel           = CONTROLBAR:ToolTipGLABuildSupplyStash
 End
 
-CommandButton Command_UpgradeCamoNetting
+CommandButton Command_UpgradeCamoNetting ; unused, duplicate of Command_UpgradeGLACamoNetting
```
commy2: Should just delete duplicates entirely imo.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -5485,15 +5485,14 @@ CommandButton Slth_Command_ConstructGLAVehicleRocketBuggy
...
+CommandButton Slth_Command_BattleBusBuildSlthTrap ; unused
   Command           = FIRE_WEAPON
   WeaponSlot        = PRIMARY
   Options           = OK_FOR_MULTI_SELECT
-  TextLabel         = CONTROLBAR:ConstructGLASlthTrap
+  TextLabel         = CONTROLBAR:TempDescription
   ButtonImage       = SSHideBomb
   ButtonBorderType  = BUILD
-  DescriptLabel     = CONTROLBAR:ToolTipGLABUildSlthTrap
+  DescriptLabel     = CONTROLBAR:TempDescription
```
commy2, line 5549: Broken?

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -7090,7 +7089,7 @@ CommandButton Infa_Command_UpgradeChinaRedguardCaptureBuilding
   TextLabel     = CONTROLBAR:UpgradeChinaRedguardCaptureBuilding
   ButtonImage   = SSCaptureBuilding
   ButtonBorderType        = UPGRADE ; Identifier for the User as to what kind of button this is
-  DescriptLabel           = CONTROLBAR:Infa_TooltipChinaUpgradeRedguardCaptureBuilding
+  DescriptLabel           = CONTROLBAR:Infa_ToolTipChinaUpgradeRedGuardCaptureBuilding
```
commy2, line 7146: Idk, don't most entries use "Tooltip" too? Maybe should be fixed in stringtable instead.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: I did not want to do that in this change. This change just cares about correcting the incorrect string references in CommandButtons.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: `CONTROLBAR:ToolTipGLABUildSlthTrap` does not exist. Replaced with placeholder text.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: `CONTROLBAR:TerrorCell` does not exist. Replaced with placeholder text.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: It is about 50/50 Tooltip and ToolTip. Streamlining that would be another change.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: I opted for just fixing the CommandButton.ini and not touching the stringtable yet.

commy2: You should do it depending on which one is better. This one should be fixed in stringtable imo.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: If the text doesn't exist, it should be added instead.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: I suggest we leave this button unchanged in this PR, and then remove it entirely in another.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: I disagree. This adjustment is more sane for now because there is still pending work in stringtable, for example merging Ukrainian localization into master str file. I would like all strings to be an exact match across all loca files out there.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: No, because `Command_TerrorCell` is unused. See comment above button.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: I think it is fine. Can remove, or leave. Does not matter much. It is documented unused, so no harm.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: Why can't the string be normalized in UA before the merge? "FakeBuilding" is the correct camel case form, because the compound is not a proper noun. This is a regression.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: ??? It's unused outside of custom maps. The fallback text "Missing: CONTROLBAR:TerrorCell" is still decipherable to players, although not great. This makes the button illegible in such maps.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: Ah, ok then. For the sake of keeping merge conflicts down, I would've avoided touching the same lines in two different PRs.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: Because then any future language merges eventually need to account for that too and I may not remember. It is technically more robust to keep string id's uniform for the time being, so that tools operating on stringtables do not introduce hard-to-find bugs.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: Ok I can check if `SuperweaponTerrorCell` refers to anything valid or is just dead code.

commy2: Also, a map may implement an actual entry for CONTROLBAR:TerrorCell in map.str. Which would be broken after this PR.

xezon: Is not in the official ones.

commy2: Don't break custom maps needlessly.

To me this is like the missing text for the Laser Cannon and the Cruise missile. The correct way to go about it is to add these texts to str.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: It should be removed to avoid confusion. It's dead code.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
commy2: I don't see the issue. Just check if stringtable contains the key in arbitrary case. I still maintain that FakeBuilding is correct and this is a regression.

xezon: I agree FakeBuilding looks nicer but Fakebuilding is safer for now. And no this is no regression. Names are an exact match now, which is an objective improvement, because case sensitive compares now work.

commy2: Before: 1 wrong use of camel case
After: 2 wrong uses of camel case

Not an improvement.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: On second look I found there is `GUI:SuperweaponTerrorCell`, so I will put that.

Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
xezon: Removed.

xezon:
Documentation for button removal added. Looks good.

PR #2101, opened 2023-07-12
Fix: Remove hidden Terrorist Bike button from GLA Stealth Arms Dealer

xezon:
This change removes the hidden Terrorist Bike button from GLA Stealth Arms Dealer.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -3053,6 +3053,7 @@ CommandSet Slth_GLATunnelNetworkCommandSet
   14 = Command_Sell
 End
 
+; Patch104p @bugfix xezon 12/07/2023 Removes hidden button 15.
 CommandSet Slth_GLAArmsDealerCommandSet
   2  = Slth_Command_ConstructGLAVehicleTechnical
   3  = Slth_Command_ConstructGLAVehicleRadarVan
@@ -3065,7 +3066,7 @@ CommandSet Slth_GLAArmsDealerCommandSet
   12 = Command_UpgradeGLACamoNetting
   13 = Command_SetRallyPoint
   14 = Command_Sell
-  15 = Command_ConstructGLAVehicleCombatBikeTerrorist ;??? is this for script then?
+  ;15 = Command_ConstructGLAVehicleCombatBikeTerrorist ;??? is this for script then?
 End
 
 CommandSet Slth_GLADemoTrapCommandSet


xezon:
Not clear to me what is going on. Looks like original developer does not know either.

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -3065,7 +3066,7 @@ CommandSet Slth_GLAArmsDealerCommandSet
   12 = Command_UpgradeGLACamoNetting
   13 = Command_SetRallyPoint
   14 = Command_Sell
-  15 = Command_ConstructGLAVehicleCombatBikeTerrorist ;??? is this for script then?
+  ;15 = Command_ConstructGLAVehicleCombatBikeTerrorist ;??? is this for script then?
```
commy2, line 3069: This is how the AI builds demo bikes as Stealthgen. Removing it may break AI scripts, as groups will wait indefinitely for members to be built.

xezon: Ok something needs to be done here though, because if I am not mistaken player could also build demo bikes then if he changed the key mapping for it to something other than Y.

Perhaps we need to remove it from AI script.

commy2: > if I am not mistaken player could also build demo bikes then 

No, player can only click the first 14 buttons and only use shortcuts of the first 14 buttons. Button 15 (and 16+?) are AI only.

>Perhaps we need to remove it from AI script.

Why? It's curious that AI can build an entirely inaccessible unit, but it's really just a way for AI to get a Terrorist on a Combat Bike. I see no need to change it, unless you want to go through the trouble of making AI mount Terrorists on Combat Bikes manually, and get that done in a reliable way. Don't fix what ain't broken. This is working as intended. Maybe add some doc about what it does though.

xezon: Hmm. I thought I have seen Mods using extended Command Sets with additional buttons, up to max 18. I can test what happens when setting custom key for Stealth demo bike.

commy2: The button has:

> Options       = SCRIPT_ONLY

It would not appear even when extending the control bar beyond 14, just like the MOAB doesn't appear in the Generals Power unlock screen.

commy2: Btw, multiple campaign missions (e.g. China03) use this technique to get the AI to use Terror Bikes.

xezon: Ah ok very well then. Then this is obsolete and we can just put a better comment to avoid confusion.

PR #2106, opened 2023-07-13
Fix: Key conflict of Neutron Shells warhead with SELECT_MATCHING_UNITS (E)

xezon:
* Refers to #2091

This change fixes the key conflict of Neutron Shells warhead with SELECT_MATCHING_UNITS (E) for all languages.


xezon:

PR #2102, opened 2023-07-12
Fix: Boss General key conflicts in English localization

xezon:
* Relates to #2091

This change fixes 13 key conflicts in English localization.

* The Boss Nuke Missile can now be constructed with key I and no longer conflicts with Patriot Missile (M).
* The Boss Nuke Missile can now be launched with key I and no longer conflicts with Nationalism (N).
* The Boss Sneak Attack can now be placed with key S and no longer conflicts with Carpet Bomb (T).
* The Boss Gattling Tank can now be produced with key A and no longer conflicts with Avenger (G).
* The Boss Land Mines can now be researched with key M and no longer conflicts with Laser Missiles (L).
* The Boss Arm the Mob can now be researched with key O and no longer conflicts with Neutron Mines (M).
* The Boss Composite Armor can now be researched with key C and no longer conflicts with Particle Cannon (P).
* The Boss Anthrax Beta can now be researched with key X and no longer conflicts with Buggy Ammo (A).

## Issue log

```
Boss_AmericaDozerCommandSet has key conflict with
  6 = Boss_Command_ConstructAmericaPatriotBattery : CONTROLBAR:ConstructAmericaPatriotBattery "Patriot &Missile System"
  11 = Boss_Command_ConstructChinaNuclearMissileLauncher : CONTROLBAR:ConstructChinaNuclearMissileLauncher "Nuclear &Missile"

Boss_ChinaCommandCenterCommandSet has key conflict with
  2 = Command_ChinaCarpetBomb : CONTROLBAR:CarpetBomb "Carpe&t Bomb"
  7 = Command_SneakAttack : CONTROLBAR:SneakAttack "Sneak A&ttack"

Boss_ChinaCommandCenterCommandSetUpgrade has key conflict with
  2 = Command_ChinaCarpetBomb : CONTROLBAR:CarpetBomb "Carpe&t Bomb"
  7 = Command_SneakAttack : CONTROLBAR:SneakAttack "Sneak A&ttack"

Boss_ChinaAirfieldCommandSet has key conflict with
  8 = Command_UpgradeAmericaLaserMissiles : CONTROLBAR:UpgradeAmericaLaserMissiles "&Laser Missiles"
  12 = Command_UpgradeChinaMines : CONTROLBAR:UpgradeChinaMines "&Land Mines"

Boss_ChinaWarFactoryCommandSet has key conflict with
[...]

PR #2119, opened 2023-07-16
Add separate buttons and labels to all major Boss commands

xezon:
* Follow up for #2102

This changes adds new dedicated buttons and labels for all major Boss commands. This allows full key mapping customization independent of the three factions it borrows features from.

User facing nothing changes as is.

The downside is a lot of new code redundancy - the usual.


xezon:
Verified with script that there are no broken links between command sets, command buttons and strings.

xezon:
Looked at most things in Boss Command Sets and spotted no errors.

xezon:
Documentation added.

xezon:
Looks good.

PR #1528, opened 2023-01-07
Fix: Wrong exploded death animation of supply box carrying GLA Worker

xezon:
**This change has been superseded by https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2120**

* Fixes #1468

PR #2120, opened 2023-07-16
Fix: Worker Has Wrong Death Animation When Carrying Boxes

### LINKED ISSUE #1468: Worker With Supply Plays Wrong Animation When Blown Up
When worker is carrying supply and building while it has supply, it plays wrong blown up animation.  Using mine detector while have supply is ok.

For fix delete "AliasConditionState =  DYING CARRYING" under   "ConditionState = DYING" .

Original:

https://user-images.githubusercontent.com/103841664/201549092-1c005f94-d62d-4766-9362-1a5c72a4b5a4.mp4

Fixed:

https://user-images.githubusercontent.com/103841664/201549116-ad98b93a-d38d-4a14-a48b-2dcbd1e54c5d.mp4

Original Builder:

https://user-images.githubusercontent.com/103841664/201549136-0cfc496a-1d57-4959-8866-177107fd24a2.mp4

Fixed Builder:

https://user-images.githubusercontent.com/103841664/201549148-ca5675b4-27fe-4a6d-a92e-89dc8cf6593c.mp4

commy2:
#### Situation in 1.04:

- Worker displays wrong death animation when blown up while carrying a box as described here: #1468

#### Situation in current patch main branch

- After the change in #1528, the Worker displays the correct death animation when blown up, but the wrong animation when shot while carrying a box. This worked previously in 1.04

Repro:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/059f4bc9-88df-4d7d-87b3-4c63d655b986

#### Situation with this fix

- The correct death animation is played in all cases:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/72cfde48-5247-4050-a180-746d035db4af

- close #1468

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2120_worker_death_animation.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2120_worker_death_animation.yaml
@@ -14,7 +14,6 @@ labels:
   - v1.0
 
 links:
-  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1528
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2120
 
 authors:
--- a/Patch104pZH/Design/Changes/v1.0/2120_worker_death_animation.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2120_worker_death_animation.yaml
@@ -1,5 +1,5 @@
 ---
-date: 2023-01-08
+date: 2023-07-16
 
 title: Fixes wrong exploded death animation of GLA Worker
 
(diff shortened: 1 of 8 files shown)


commy2:
Imho, the ideal fix would be if the engine reset the CARRYING flag on death, but w/e.

xezon:
1528_worker_death_animation.yaml file needs to be updated.

xezon:
Instead of renaming the updated yaml file, I suggest to create a txt file for `2120_worker_death_animation.txt` and write in it the existing yaml file. This way we can find files for both pull numbers. I have used this approach for several other changes already. This just makes it a bit easier to find things.

The old date also does not need to be updated.

commy2:
Even though this PR completely undoes all changes in #1528?

xezon:
In that case what you can also do is edit the opening text of #1528 and note that it has been reverted or superseeded, so anyone looking at it knows it has no relevance anymore.

xezon (approved):

PR #2121, opened 2023-07-16
Fix: Stinger Trooper Upgrade Icons

commy2:
- In the Generals Challenge vs Stealthgen / Kassad, there are a few Stinger Troopers placed outside of the Stinger Site.
- https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1086 makes them selectable, which makes them way less annoying to kill
- However, with them being selectable it becomes apparent that they are missing upgrade icons for AP Rockets.

With this PR:

- AP Rockets icon will be added to all Stinger Troopers. 
- Suicide icon will be added to the unused Demogen Stinger Troopers (the Demo Stinger Site uses vGLA Troopers). 
- Anthrax Gamma icon will be added to the Toxgen Stinger Troopers.
- Camo Netting icon will be removed from the single player only GLA02 Stinger Troopers. That upgrade does not transfer from the Site to the Troopers anyway, so it would always be greyed out.

This has essentially no effect in multiplayer as these units are mot buildable outside of Stinger Sites and are not selectable on any map by default.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2121_stinger_soldier_upgrade_cameos.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2121_stinger_soldier_upgrade_cameos.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-07-16
+
+title: Adds missing upgrade icons to Stinger Soldiers
+
+changes:
+  - fix: All Stinger Soldiers are missing AP Rockets upgrade icon.
+  - fix: Toxin General Stinger Soldier is missing Anthrax Gamma upgrade icon.
+  - fix: Unused Demo General Stinger Soldier is missing Demolitions upgrade icon.
+  - fix: Single player only GLA02 Stinger Soldier has inapplicable Camo Netting upgrade icon.
+
+labels:
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2121
+
+authors:
+  - commy2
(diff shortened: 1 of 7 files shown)


Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
```diff
@@ -13402,6 +13402,14 @@ Object Chem_GLAInfantryStingerSoldier
   SelectPortrait         = SUStinger_L
   ButtonImage            = SUStinger
 
+  ; Patch104p @bugfix commy2 16/07/2023 Add missing upgrade icon for AP Rockets.
```
xezon: Please also add `(#2121)` to comments.

Patch104pZH/Design/Changes/v1.0/2121_stinger_soldier_upgrade_cameos.yaml
```diff
@@ -0,0 +1,21 @@
+---
+date: 2023-07-16
+
+title: Adds missing upgrade icons to Stinger Soldiers
+
+changes:
+  - fix: All Stinger Soldiers are missing AP Rockets upgrade icon.
```
xezon: The way I describe changes here is the new state of game. In this case: "All GLA Stinger soldiers now show the AP Rockets upgrade icon". So reader does not read broken state, but new state of affairs. For consistency in style I recommend to write it alike.

xezon (approved):

PR #2124, opened 2023-07-17
Fix: Set consistent name for EMP Patriot Battery in German language

xezon:
This change sets the correct name of the EMP Patriot Battery in the German language. It is now consistent with "Patriot-Batterie". In voice lines it is also described as "Patriot Batterie".


xezon:

PR #2129, opened 2023-07-19
Add donation page

xezon:


xezon:
QR codes show and work. Looks good.

PR #2130, opened 2023-07-20
Implement USA Tomahawk Storm

commy2:
Model's already there, so why not.

shot_20230720_210911_3


Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -3916,6 +3916,7 @@ CommandSet SupW_AmericaDozerCommandSet
   8  = SupW_Command_ConstructAmericaCommandCenter
   9  = SupW_Command_ConstructAmericaFireBase
   11 = SupW_Command_ConstructAmericaWarFactory
+  12 = SupW_Command_ConstructAmericaTomahawkStorm
```
commy2: rm

MTKing4:
I don't know why we haven't added this already, all the other SWG Superweapons are already there/fixed so why not this?

Good addition

xezon:
> I don't know why we haven't added this already

Because this is not user facing on regular maps.

commy2:
Yeah, it may be outside the scope of the project. Don't want to drift into mod territory. However, this is the most canonical implementation I can come up with of a model that already sits in the game files. If EA didn't hate their customers, I could totally see them finishing this thing in a patch and delivering this state (or better).

xezon:
It is fine to finish up half baked original content. It would also be ok to add new props for map creators if it has reasonable demand and fits the original specs for the game.

MTKing4:
> Because this is not user facing on regular maps.

Neither is Laser Cannon, Cruise Missile & ICBM and we fixed those, these Additions will be good for fun maps

xezon:
Observations:

## Tomahawk missiles on the pad are damaged

shot_20230721_192215_3

## Target Cursor is Toxic

shot_20230721_192255_4

## Tomahawk uses SCUD Storm voice events

When the Tomahawk Storm is ready to fire it says "Scud Storm ready".
When the Tomahawk Storm fires it says "Scud Storm launched".

commy2:
> Tomahawk missiles on the pad are damaged

Yeah, it's strange, but that is the texture used in all states, even pristine. I also made that the model of the launched rockets. No good? Would have to hexedit the model to solve this.

> Target Cursor is Toxic

There is a hard coded limit for cursor types. No free ones left over. Would have to reuse one, but it would have to be one with that diameter. I don't know which else to use.

> Tomahawk uses SCUD Storm voice events

ICBM uses "Nuclear Missile ready". It depends on the Enum used in SpecialPower.ini. There is none for Tomahawk Storm, and there are obviously no voice lines. Could maybe use one without voice lines linked in engine like `SUPR_SPECIAL_CRUISE_MISSILE`. Would have to test if reusing some potentially breaks the button in the shortcut bar.

xezon:
It would be good if grounded missile always matches the launched missile. And that pristine missiles are undamaged. This issue was also not yet fixed for SCUD Launcher and Tomahawk. If it is not possible to launch a different missile in damaged state, then I think it would be best to always draw undamaged missiles.

xezon:
> SUPR_SPECIAL_CRUISE_MISSILE

```c++
enum SpecialPowerType
{
    SPECIAL_INVALID,
    SPECIAL_DAISY_CUTTER,
    SPECIAL_PARADROP_AMERICA,
    SPECIAL_CARPET_BOMB,
    SPECIAL_CLUSTER_MINES,
    SPECIAL_EMP_PULSE,
    SPECIAL_NAPALM_STRIKE,
    SPECIAL_CASH_HACK,
    SPECIAL_NEUTRON_MISSILE,
    SPECIAL_SPY_SATELLITE,
    SPECIAL_DEFECTOR,
    SPECIAL_TERROR_CELL,
    SPECIAL_AMBUSH,
    SPECIAL_BLACK_MARKET_NUKE,
    SPECIAL_ANTHRAX_BOMB,
    SPECIAL_SCUD_STORM,
    SPECIAL_DEMORALIZE_OBSOLETE,
    SPECIAL_CRATE_DROP,
    SPECIAL_A10_THUNDERBOLT_STRIKE,
    SPECIAL_DETONATE_DIRTY_NUKE,
    SPECIAL_ARTILLERY_BARRAGE,
    SPECIAL_MISSILE_DEFENDER_LASER_GUIDED_MISSILE,
    SPECIAL_REMOTE_CHARGES,
    SPECIAL_TIMED_CHARGES,
    SPECIAL_HELIX_NAPALM_BOMB,
    SPECIAL_HACKER_DISABLE_BUILDING,
    SPECIAL_TANKHUNTER_TNT_ATTACK,
    SPECIAL_BLACKLOTUS_CAPTURE_BUILDING,
    SPECIAL_BLACKLOTUS_DISABLE_VEHICLE_HACK,
    SPECIAL_BLACKLOTUS_STEAL_CASH_HACK,
    SPECIAL_INFANTRY_CAPTURE_BUILDING,
    SPECIAL_RADAR_VAN_SCAN,
    SPECIAL_SPY_DRONE,
    SPECIAL_DISGUISE_AS_VEHICLE,
    SPECIAL_BOOBY_TRAP,
    SPECIAL_REPAIR_VEHICLES,
    SPECIAL_PARTICLE_UPLINK_CANNON,
    SPECIAL_CASH_BOUNTY,
    SPECIAL_CHANGE_BATTLE_PLANS,
    SPECIAL_CIA_INTELLIGENCE,
    SPECIAL_CLEANUP_AREA,
    SPECIAL_LAUNCH_BAIKONUR_ROCKET,
    SPECIAL_SPECTRE_GUNSHIP,
    SPECIAL_GPS_SCRAMBLER,
    SPECIAL_FRENZY,
    SPECIAL_SNEAK_ATTACK,
    SPECIAL_CHINA_CARPET_BOMB,
    EARLY_SPECIAL_CHINA_CARPET_BOMB,
    SPECIAL_LEAFLET_DROP,
    EARLY_SPECIAL_LEAFLET_DROP,
    EARLY_SPECIAL_FRENZY,
    SPECIAL_COMMUNICATIONS_DOWNLOAD,
    EARLY_SPECIAL_REPAIR_VEHICLES,
    SPECIAL_TANK_PARADROP,
    SUPW_SPECIAL_PARTICLE_UPLINK_CANNON,
    AIRF_SPECIAL_DAISY_CUTTER,
    NUKE_SPECIAL_CLUSTER_MINES,
    NUKE_SPECIAL_NEUTRON_MISSILE,
    AIRF_SPECIAL_A10_THUNDERBOLT_STRIKE,
    AIRF_SPECIAL_SPECTRE_GUNSHIP,
    INFA_SPECIAL_PARADROP_AMERICA,
    SLTH_SPECIAL_GPS_SCRAMBLER,
    AIRF_SPECIAL_CARPET_BOMB,
    SUPR_SPECIAL_CRUISE_MISSILE,
    LAZR_SPECIAL_PARTICLE_UPLINK_CANNON,
    SUPW_SPECIAL_NEUTRON_MISSILE,
    SPECIAL_BATTLESHIP_BOMBARDMENT,
    SPECIAL_POWER_COUNT,
};
```

xezon:
> There is a hard coded limit for cursor types. No free ones left over. Would have to reuse one, but it would have to be one with that diameter. I don't know which else to use.

If we cannot change it now then ok nothing we can do.

commy2:
>cursor type

Ah, not "Enum", I was thinking about RadiusCursorType from CommandButton, which is an enum, where each value is hard coded to link to an object in InGameUI. E.g. SCUDSTORM -> ScudStormRadiusCursor.

I changed it to something neutral:

shot_20230724_190250_2

commy2:
Reusing the same enum does break the shortcut bar as "expected" (both SPECIAL_PARTICLE_UPLINK_CANNON):

shot_20230724_191016_1

Button is unusable and a 2 is shown, despite there being only 1 Particle Cannon.

This is only an issue if two superweapons are used by the same faction that share a shortcut button, which rules out SUPR_SPECIAL_CRUISE_MISSILE, but I just changed it to SPECIAL_BLACK_MARKET_NUKE, which should be safe.

shot_20230724_192759_1

The only issue is that now no notification plays at all when the weapon is ready or fired (by the enemy).

commy2:
I replaced the damaged missiles with [undammaged](https://community.bistudio.com/wiki/setDammage) ones, except in badly damaged state to mirror the Tomahawk Launcher.

<details><summary>script used</summary>

```py
from pathlib import Path

for file in Path(".").glob("*.W3D"):
    with open(file, "rb") as f:
        data = f.read()

    data = data.replace(
        b"avtomahawk_md.tga\0",
        b"avtomahawk_m.tga\0\0",
    )

    with open(file, "wb") as f:
        f.write(data)
```

</details>

xezon:
Is it possible to add custom sound events for SPECIAL_BLACK_MARKET_NUKE ?

commy2:
Don't think so. Seems to be list of hard coded DialogEvent entries in speech.ini that are bound to a few SpecialPower enums.

xezon (approved):
Looks good from my side. Just need to remove it from Command Set.

As for Audio event, can be follow up task once we can technically fix it. It may be possible to assemble voice lines for this. If not, perhaps we can just add generic ready and alert sounds for it.

As for tomahawk missiles texture, can also be follow up task when we tackle Scud Launcher and Tomahawk Launcher.

On that note, Scud Storm does not show damaged rocket texture. Perhaps that is the way to go.

shot_20230725_182004_1

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
commy2:
```suggestion
```

Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
```diff
@@ -3931,6 +3931,7 @@ CommandSet SupW_AmericaDozerCommandSet
   9  = SupW_Command_ConstructAmericaFireBase
   10 = SupW_Command_ConstructAmericaCommandCenter
   11 = SupW_Command_ConstructAmericaWarFactory
+  12 = SupW_Command_ConstructAmericaTomahawkStorm
```
commy2:
```suggestion
```

commy2:
Removed the buttons and made the missiles undamaged in all states.

commy2:
merge?

I would prefer if it played the Scud Storm notifications instead of being silent, even if they don't fit very well. Laser Cannon and ICBM play "Particle Cannon" and "Nuclear Missile" respectively.

xezon:
Merge as you like. You can re-add the Scud Storm sound if you like it better. We have to revisit that sound either way if we want to polish this thing further, but it is not important because it is not user facing for the most part.

Issue #2126, opened 2023-07-17
Rockets Emerge From Below Bunker, Possibly Colliding With Terrain And Destroying Bunker


commy2:
### 1.04

If a Bunker is filled with rocketeers, some rockets will spawn from the center of the Bunker and if the terrain is rough, they will hurt the Bunker. This is jank.

- Bunker (xChina and Inf) only have 4 FIREPOINT's. The rest default to 0,0,0.

### Remedy

- add FIREPOINT05 ... FIREPOINT08 (FIREPOINT05 ... FIREPOINT13 for Inf) to 0,0,z=7,697. This provides enough slots from the center of the building such that the rocketeers don't have to fire from the wrong side of the Bunker.

commy2:
Gallery:

![shot_20230721_122307_6](https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/1a4267b3-69de-41e6-a71d-bcf1f5035324)
Minigunner 1 and 2 fire from proper positions. Minigunner 3 to 10 fire from below the Bunker.

---

![shot_20230721_122341_7](https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/a81a37b4-2c58-44e8-b13b-0982bb8942d6)
It is safe to put 2 Tank Hunters inside a Bunker, even on rough terrain, because two fireports will (pretty much) always be closer to the enemy than the default position below the Bunker.

---

![shot_20230721_122402_10](https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/4c3bc3b2-1def-45d5-9d66-3199d4a081a9)
The third Tank Hunter will begin firing from below the Bunker and damage it.

xezon:
This is tragic.

commy2:
It doesn't have to be this way.

![shot_20230721_130850_1](https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/671b4a0e-6597-4af1-821e-87e9075f50cd)

With additional firepoints at (0,0,z>0), the shots now don't spawn from below the Bunker and avoid the terrain - no more self kills.

xezon:
Other than adding more fire ports to models, this needs fixing from code side. Spawning rockets at 0,0,0 should really only be the last resort. It would be better if fire ports are used multiple times to avoid spawning at 0,0,0.

commy2:
[...]

PR #2133, opened 2023-07-21
Fix: Add more fire ports to China Bunkers to help prevent passengers from damaging it near uneven terrain

commy2:
* Relates to #2126

This change adds 8 (16 for Infgen) additional fireports to China Bunkers. This prevents them from killing themselves on rough terrain in most situations if filled with more than 2 Tank Hunters.

shot_20230726_162556_2

shot_20230726_163014_5

This does not solve the bug with all GarrisonContain module using objects where shots emerge from 0,0,0 if attacked from certain angles. That can only be solved with engine changes.


xezon:
This is amazing fix.

Patch104pZH/Design/Changes/v1.0/2133_bunker_self_kills.yaml
```diff
@@ -0,0 +1,24 @@
+---
+date: 2023-07-21
+
+title: Fixes the bug where Tank Hunters inside Bunkers frequently destroy their Bunker
```
xezon: "Fixes bug where China Tank Hunters destroy a Bunker near uneven terrain when firing from inside of it"

perhaps to clarify where bug happens?

commy2:
Unfortunately, they are stubborn. Needs more investigation to get it totally right.

shot_20230721_141937_1

ExiLe-ZH:
would be an amazing fix when it all works properly

xezon:
> Unfortunately, they are stubborn. Needs more investigation to get it totally right.

What is the problem? On first image everything looked right.

commy2:
> What is the problem? On first image everything looked right.

Troops still use 0,0,0 if that is closer than any of the other unused firepoints. I think it may not be possible to prevent them from doing this, unless one adds enough firepoints closer to potential targets from all angles. Note that targets are at z=0 on even terrain, so 0,0,0 has a natural advantage too. I think we need to add enough fire points on the slits in the geometry. It is probably possible to do this for 5 guys, but probably impossible for Infgen Bunker when attacked from behind, unless many points sit (almost?) on top of each other.

MTKing4:
Can't you just delete the default fireport at 0,0,0? And if hardcoded we do it with thyme?

commy2:
It appears to be hard-coded with no work-around. Thyme. Ideally one could place some fallback firepoint that can be used by multiple troops. Maybe it is enough to just place enough points along the geometry. Will have to experiment more with it.

xezon:
Then we need to put enough ports at the windows and door.

xezon:
We can paint small windows in the door so that it also looks plausible to shoot from there.

xezon:
I do wonder, is this also a problem with other bunker structures, such as Palace? Does it have enough ports in all directions to never shoot from center?

And why is this not a problem for vehicles? Would a Humvee not also shoot from íts center if does not have enough ports on either side?

commy2:
> I do wonder, is this also a problem with other bunker structures, such as Palace?

It is in 1.04 in certain animation states (e.g. on certain maps or after Fortified Structures). In patch, Palace has enough firepoints.

> And why is this not a problem for vehicles? Would a Humvee not also shoot from íts center if does not have enough ports on either side?

Good question. Needs investigation. I know it is an issue with the GC Stealthgen Bus, which has a disguise ability similar to the Bomb Truck. If you disguise into certain models, e.g. Toxin Tractors, they will kill themselves due to the lack of non 0,0,0 firepoints on the models. (Could probably fix that too by giving that Bus a dummy model with extra firepoints.)

Maybe it has to do with TransportContain vs GarrisonContain.

commy2:
Issue does not exist for Humvee:

shot_20230724_181340_14

Issue persists after changing the Bunker model to Humvee:

shot_20230724_181216_8

Issue does not exist when switching Bunker to TransportContain:

shot_20230724_183205_3

I conclude:

- that GarrisonContain is the cause of this issue
- the only way to solve it without Thyme is to add a sufficient number of firepoints to the models that cover all directions

commy2:
The issue also exists for the civilian Bunker (e.g. 4 times on TD). If attacked from behind, the shots spawn at 0,0,0 and may collide with the terrain to hit the bunker.

xezon:
The simplest fix from code side would be to allow reuse fire ports if there are not enough individual ones. This would mean that multiple units can shoot from the same port. There is a visual element to rocket ports on buildings. I think it shows a small barrel. Therefore individual positions are preferred, but reused ports would work too. I think we can do both fixes. Fix in Thyme for general coverage and fix in models where we find it.

commy2:
I updated the OP.

The only feasible solution I see is to add sufficiently many additional firepoints. This seems to work pretty well from my testing. I used a script to space out the points evenly, because I'm a nerd with no augenmaß.





Yellow points are the original firepoints. Red are additions for vChina (top) and Infgen Bunkers (bottom).

<details><summary>script used</summary>

```py
from itertools import islice, pairwise, starmap

import matplotlib.pyplot as plt
import numpy as np

fps = np.array([
    # x         y           z
    [ 3.928,     10.802,    7.697],
    [-8.217,      6.510,    7.697],
    [-8.331,     -6.190,    7.697],
    [ 3.469,    -10.168,    7.697]
])

def dist(A, B):
    return np.linalg.norm(A - B)

def angle(A, B):
    return np.rad2deg(np.arctan2(B[1] - A[1], B[0] - A[0]))

def fromto(A, B):
    diff = B - A
    return diff / dist(A, B)

def zrot(A, theta):
    theta = np.deg2rad(theta)
    vec = np.array([
        [np.cos(theta), -np.sin(theta), 0],
        [np.sin(theta),  np.cos(theta), 0],
        [0, 0, 1]
    ])
    return np.dot(vec, A)

def hpos(A, B):
    c = dist(A, B)
    h = c / 2 * np.tan(np.deg2rad(72/2))
    hdir = zrot(fromto(B, A), 90)
    hvec = h * hdir

    # edge A: FP1 + (FP2 - FP1) / 2 + hvec
    return A + (B - A) / 2 + hvec

FP1, FP2, FP3, FP4 = fps
A, B, C = starmap(hpos, pairwise(fps))
E = FP1 - (A - FP1)
D = FP4 - (C - FP4)

vcfps = [
    FP1 + (E - FP1) / 2,
    FP1 + (A - FP1) / 2,
    FP2 + (A - FP2) / 2,
    FP2 + (B - FP2) / 2,
    FP3 + (B - FP3) / 2,
    FP3 + (C - FP3) / 2,
    FP4 + (C - FP4) / 2,
    FP4 + (D - FP4) / 2,
]

igfps = [
    FP1 + (E - FP1) / 3,
    FP1 + (E - FP1) / 3 * 2,
    FP1 + (A - FP1) / 3,
    FP1 + (A - FP1) / 3 * 2,
    FP2 + (A - FP2) / 3,
    FP2 + (A - FP2) / 3 * 2,
    FP2 + (B - FP2) / 3,
    FP2 + (B - FP2) / 3 * 2,
    FP3 + (B - FP3) / 3,
    FP3 + (B - FP3) / 3 * 2,
    FP3 + (C - FP3) / 3,
    FP3 + (C - FP3) / 3 * 2,
    FP4 + (C - FP4) / 3,
    FP4 + (C - FP4) / 3 * 2,
    FP4 + (D - FP4) / 3,
    FP4 + (D - FP4) / 3 * 2,
]

plt.plot(*islice(zip(A, B, C, D, E, A), 2), alpha=0.5)
plt.scatter([0], [0], c="orange", marker="o")
plt.scatter(*zip(FP1, FP2, FP3, FP4), c="orange", marker="o")
plt.scatter(*zip(*vcfps), c="red", marker="o")
# plt.scatter(*zip(*igfps), c="red", marker="o")
plt.show()
```

</details>

MTKing4:
Doesn't that mean when enemies approach from the door side they will still shoot from point 0,0,0?

xezon:
I think we should allow them shooting from door too. Makes the setup much easier.

commy2:
> Doesn't that mean when enemies approach from the door side they will still shoot from point 0,0,0?

Didn't happen for me with vChina Bunker, but Infgen Bunker when full. Works on the front though, which is much better already.

> I think we should allow them shooting from door too. Makes the setup much easier.

Yeah, but it makes no sense. Imo, instead GarrisonContain should work like TransportContain, where 0,0,0 is only used if no other free firepoint exists, instead of when 0,0,0 happens to be closer. It is an error in some algorithm most likely where 0,0,0 are the initial values - probably.

xezon:
I will paint window in door. Then you can add fire ports there.

commy2:
I don't think that is necessary or helpful. It has to be fixed in engine anyway, because its not feasible to go through all buildings like this.

This Bunker has 1 fireport in the backside, but you can fill so many guys inside that it starts damaging itself when firing backwards.

shot_20230726_205053_4

shot_20230726_205515_5

xezon:
Too late. Windows painted. They are good for at least 4 ports. #2154

commy2:
shot_20230727_114152_1

shot_20230727_114243_8

shot_20230727_114701_13

shot_20230727_114704_14

xezon:
Nice work. Looks promising. I will test later.

xezon:
2 missiles will emerge from below the 10 man bunker when shooting in this direction:

shot_20230727_184053_11

Adding more ports may become too cramped. We can accept current state and fix the rest in code.

MTKing4:
I don't understand this comment, which two missiles? And how are we supposed to see which ones are emerging from below the bunker from a top down view? I don't see an issue in this picture.

commy2:
> And how are we supposed to see which ones are emerging from below the bunker from a top down view? 

It's not the best view, but if you count the barrels, it's only 8 (2 are obscured at the center of the building). You can also see a faint trail in the middle.

Patch104pZH/Design/Changes/v1.0/2133_bunker_self_kills.yaml
```diff
@@ -0,0 +1,24 @@
+---
+date: 2023-07-21
+
+title: Fixes bug where China Tank Hunters destroy their Bunker near uneven terrain.
```
xezon: Title sentences never end with "."
The other sentences do.

Patch104pZH/Design/Changes/v1.0/2133_bunker_self_kills.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+changes:
+  - fix: Projectiles from units inside Bunkers no longer spawn from below the Bunker.
+
+subchanges:
+  - fix: Adds 4 extra fire ports to the China Bunker.
+  - fix: Adds 8 extra fire ports to the Infantry General Bunker.
```
xezon: Needs updated numbers.

Patch104pZH/Design/Changes/v1.0/2133_bunker_self_kills.yaml
```diff
@@ -0,0 +1,24 @@
+---
+date: 2023-07-21
+
+title: Fixes bug where China Tank Hunters destroy their Bunker near uneven terrain.
+
+changes:
+  - fix: Projectiles from units inside Bunkers no longer spawn from below the Bunker.
```
xezon: Needs clarification that this is not always true.

xezon (approved):

PR #2141, opened 2023-07-23
Change text key mapping format

### LINKED ISSUE #2115: Key mappings are hard to read in texts
Key mappings are hard to read in texts. Uppercase characters at begin of words are ok, but lowercase characters can be difficult to see.

## Ok

shot_20230715_154935_2

## Not good

shot_20230715_154800_1

xezon:
**Merge with Rebase**

* Fixes #2115

This changes the text key mapping format for all languages.

Before it used a mix of in-text key mappings and trailing key mappings in parentheses or brackets.

All key mappings in all languages are now streamlined to trailing key mappings behind a colon.

The new trailing key mappings have the following advantages:

1. Key mapping is always uppercase and therefore larger and better to see
2. Key mapping is always at the end of the name, making it easier to look for
3. Key mapping never needs to be bound to the characters of the words

The colon separator between the title text and the key mapping was chosen because:

1. It is an ascii character
2. It has a small visual footprint and does not add bloat to the title text
3. It is consistent with other parts of the tool tip text, such as "Cost: 100"
4. It is perfectly centered height-wise on various resolutions, contrary to the dash, parentheses and brackets
5. It does not make the key letter look cramped, contrary to the parentheses and brackets

## Patched 1024 x 768

shot_20230723_103258_2

## Patched 1920 x 1080

shot_20230723_103212_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2141_text_key_mapping_format.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2141_text_key_mapping_format.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-07-23
+
+title: Improves visibility of text key mappings in all languages
+
+changes:
+  - tweak: All in-text key mappings are now trailing upper-case key mappings behind a colon.
+  - tweak: All trailing key mappings in parantheses are now trailing upper-case key mappings behind a colon.
+
+labels:
+  - design
+  - gui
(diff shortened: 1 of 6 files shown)


xezon:
Documentation added. Looks good. Could not spot any mistakes. All the refactoring was done with script.

Issue #2095, opened 2023-07-10
Some key mappings are confusing


xezon:
Some key mappings are confusing. Examples:

```
&Assault Troop Transport
&Troop Crawler
```
```
Attack &Outpost
Li&stening Outpost
```
```
Anthra&x Gamma
&Anthrax Beta
```
```
Carpe&t Bomb
Nu&ke Bomber
```
```
&Land Mines
Neutron &Mines
```
Do they need to be this way? Streamline in Core bundle? Streamline in Optional bundle?

MTKing4:
Optional

commy2:
Should offer full key customization option and not doctor ourselves with keys people are used to. Exceptions are key conflicts of course like the Satellite Scan Stop s or Neutron Shells mode e.

ExiLe-ZH:
Fully agree.

xezon:
Would be nice if 

CONTROLBAR:SuicideAttack
CONTROLBAR:DetonateBombTruck
CONTROLBAR:DetonateFakeBuilding
CONTROLBAR:Detonate
CONTROLBAR:DetonateCharges
CONTROLBAR:DetonateNuke

had same key mappings.

xezon:
It would nice if GLA Fake Buildings have same key mappings as normal GLA buildings.

PR #2142, opened 2023-07-23
Fix: Streamline key mappings of GLA fake structures with regular GLA structures

xezon:
* Relates to #2095

This change fixes all key mapping mismatches between Fake GLA buildings and regular GLA buildings.

Affected languages are French, Spanish, Italian, Polish.


xezon:

PR #2148, opened 2023-07-26
Fix: Add consistent wording in Polish tool tip strings

xezon:
This change fixes polish tool tip texts. They now no longer say "Good against" or "Strong against" or "Effective against" but always "Strong against" like English does.


xezon:

PR #2150, opened 2023-07-26
Fix: English tool tip texts

xezon:
**Merge with Rebase**

This change fixes English tool tip texts. The capitalization of words is now consistent. And it always says "Strong vs. ..." instead of "Strong against ..." (1).


xezon:

PR #2151, opened 2023-07-26
Fix: Add consistent wording in Spanish tool tip strings

xezon:
This change fixes Spanish tool tip texts. They now no longer say "Strong against" or "Powerful against" but always "Powerful against".


xezon:

PR #2152, opened 2023-07-26
Add Range And Damage Bonus To Infantry Inside Overlord Bunkers

### LINKED ISSUE #2131: China Overlord Bunker does not grant GARRISON bonus to passengers
In Zero Hour, when units like rocketeers are place inside buildings, they gain a GARRISON bonus which gives them 133% range and 125% damage.

The GARRISON bonus does not apply when units are placed inside vehicles. There is one exception to this: GARRISON bonus is granted to units inside a Bunker Helix.

unit / building | GARRISON bonus
---- | --- 
Humvee | no
Outpost | no
Bunker Overlord | no
Bunker Helix | yes
Battle Bus | no
Fire Base | yes
(Inf) Bunker | yes
Palace | yes
any civilian building | yes

This oddity is likely an oversight, but it makes the Helix viable, as with this bonus they can engage even long range AA weapons like Quads with good micro and deal a lot of damage.

The Bunker Overlord is mostly unviable in comparison. The only time you maybe want to use them is against Crusaders/Laser tanks. I think this is mostly because of the underwhelming range and damage of the garrisoned rocketeers.

It is odd that putting a Bunker on an Overlord negates the range and damage bonus of the static base defense Bunker, especially while it does not do this for the Helix Bunker.

Comparison:

shot_20230721_100120_4

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/issues/2131

This change grants GARRISON bonus (133% range, 125% damage) to infantry inside Overlord Bunkers. They already gain this bonus in regular and Helix Bunkers in 1.04, but not inside Overlord Bunkers.

shot_20230721_102229_1

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
@@ -899,7 +899,7 @@ Object ChinaTankOverlordBattleBunker
     InitialHealth   = 100.0
   End
 
-  ; Patch104p @balance commy2 21/07/2023 Change from Transport to HelixContain to grant GARRISON bonus. (#XXXX)
+  ; Patch104p @balance commy2 21/07/2023 Change from Transport to HelixContain to grant GARRISON bonus. (#2152)
   Behavior = HelixContain ModuleTag_03
     Slots                 = 5
     PassengersAllowedToFire = Yes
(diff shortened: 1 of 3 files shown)


commy2:
Hmm, force-fire works, but the guys won't fire on enemies anymore :/

MTKing4:
What do you mean? They don't fire at all? Not even auto fire when in range? What about dircet order attack? Those all work in helix so what is preventing it here?

commy2:
> Not even auto fire when in range? What about dircet order attack?

Neither.

> Those all work in helix so what is preventing it here?

Helix Bunker enables firing. Overlord Bunker puts a Bunker with 5 slots into the Overlord (which has 1 invisible slot). Seems like HelixContain is not compatible with that setup.

ReLaX82:
I like the idea, making bunkers on OLs useful.
Can't help with fixing these bugs though.

PR #2154, opened 2023-07-26
Change: Add windows to ntdoor04l

xezon:
* Follow up for #2133

This change draws windows on ntdoor04l textures of the China Bunker. This allows for two fire ports at the door.

## Patched

shot_20230726_205409_1


xezon:
The doors appear to be exclusive for China Bunker. I could not find them on other China buildings.

Optimized window dimensions and visuals. Documentation added. Looks good.

PR #1271, opened 2022-09-23
Fix: avraptor textures

xezon:
This change fixes avraptor textures. The damaged texture changes are very minor, therefore are optional.

### avraptor

* Removes obsolete alpha channel

### avraptor_d (Optional)

* Removes strong green color tint
* Removes "butterfly" scorch marks

### avraptor_d1 (Optional)

* Blends texture with avraptor_d for consistent damage look

PR #2157, opened 2023-07-28
Fix: avraptor textures (2)

xezon:
* Follow up for #1271

### avraptor_d, avraptor_d1

* Moves to core bundle

### avraptor_d1b

* Blends texture with avraptor_d for consistent damage look


xezon:

PR #2162, opened 2023-07-29
Fix line break discrepancies in all control bar strings for all languages

xezon:
**Merge with Rebase**

This change fixes line break discrepancies in all control bar strings for all languages.

* Removes superfluous line breaks from the end of sentences in all languages
* Removes superfluous line breaks after 'Produktionsstätte:' in German tool tip strings
* Removes superfluous line breaks in various tool tip strings for all languages
* Removes superfluous line breaks in various control bar strings for all languages
* Adds missing line breaks to various tool tip strings for all languages


xezon:
Documentation added and updated. Looks good.

PR #2164, opened 2023-07-29
Fix: Battle Bus with Demolitions upgrade does explode when killed

### LINKED ISSUE #2161: Demogen Demolitions Battle Bus no longer explodes when killed
After this commit https://github.com/TheSuperHackers/GeneralsGamePatch/commit/c365514211591781b9245226bf9e5e263aa18483, the Battle Bus with Demolitions upgrade no longer explodes when killed.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/edae5d1a-0d87-4ffa-9b6e-cb503d58f4a5

In 1.04, it explodes, but deals no damage. It worked correctly after https://github.com/TheSuperHackers/GeneralsGamePatch/pull/102. I am unsure what the issue was that https://github.com/TheSuperHackers/GeneralsGamePatch/commit/c365514211591781b9245226bf9e5e263aa18483 solved. I did not notice any double explosions previously.

commy2:
- fix #2161 

Should work now. Damage appears to be correct in all cases. Much simpler than previous implementation. Clarified some comments.

With this, Bus Demolitions interaction can be described as:

- no boom before Demolitions upgrade
- big damage when suicided
- bunker mode disables suicide button
- small damage when bunker mode is destroyed or abandoned (only with upgrade).

I don't see any double explosions.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/102_demo_battle_bus_suicide_damage.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/102_demo_battle_bus_suicide_damage.yaml
@@ -16,6 +16,7 @@ labels:
 links:
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/102
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1504
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2164
 
 authors:
   - commy2
--- a/Patch104pZH/Design/Changes/v1.0/2164_demo_battle_bus_suicide_damage.txt
+++ b/Patch104pZH/Design/Changes/v1.0/2164_demo_battle_bus_suicide_damage.txt
@@ -1,0 +1,1 @@
+102_demo_battle_bus_suicide_damage.yaml
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -22436,7 +22436,7 @@ Object Demo_GLAVehicleBattleBus
   ;  DeathTypes = NONE +SUICIDED
   ;End;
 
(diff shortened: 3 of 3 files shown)


xezon:
The weak explosion scorch is bigger than the suicide explosion scorch. Can you please check why?

Left suicide, right death explosion.

shot_20230729_162824_1

commy2:
Because you put a unit inside, so it's 2+ explosions?

xezon:
Yes it spawns multiple scorches. It should not be that way.

To test, put `exscorch01.psd` in `Patch104pZH\GameFilesOptional\Art\Textures\GenerateMip`

[exscorch01.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/12208847/exscorch01.zip)

shot_20230730_113302_1

commy2:
If you put units inside, each of them will spawn a scorch.

shot_20230730_115832_1

shot_20230730_115843_2

Dunno what to do about that.

xezon:
Ok I see. This appears to happen only when killed, not suicided.

shot_20230730_121652_4

And it does not happen with Terrorist unit. So perhaps we should add MASKED exempt status to suicide weapons of all demo units to avoid this? We did add that to Terrorist only.

commy2:
We probably should, otherwise the Terrorist exploit can still be used with Demogen units.

Different PR for that?

xezon:
As you like.

commy2:
Different PR.

xezon:
See, there is an unlimited supply of things to fix :-)

xezon:
This was not meant to be closed, no?

commy2:
Nah, github just too smart.

Can be merged now tho.

xezon:
I will retest

commy2:
Need to rebase onto current main branch though.

xezon:
Documentation updated.

Amount of scorches looks proper now.

shot_20230731_182007_1

xezon (approved):

PR #2167, opened 2023-07-29
Fix: Remove superfluous mentions of 'enemy' in tool tip strings

xezon:
This change removes superfluous mentions of 'enemy' in tool tip strings.

Affected tool tips are

* CONTROLBAR:ToolTipUSABuildStealthFighter
* CONTROLBAR:ToolTipUSABuildTomahawk
* CONTROLBAR:ToolTipChinaBuildInfernoCannon
* CONTROLBAR:AirF_ToolTipUSABuildStealthFighter

Unaffected tool tip is

* CONTROLBAR:ToolTipGLABuildSaboteur

Most tool tips do not mention enemy forces, because the associated unit could be used against any forces, including allies and civilians. Only the GLA Saboteur can attack enemy buildings only, and therefore the mention of "enemy" is ok.


commy2:
"obsolete" implies those mentions were useful at any point. :)

xezon:
Renamed "obsolete" to "superfluous" in all mentions. Documentation added. Looks good.

PR #2170, opened 2023-07-29
Fix: Boss Aurora Countermeasures Evasion Rate

commy2:
### 1.04

- Almost all airplanes, including all Aurora variants except the Boss General's have a 30% missile evasion rate after Countermeasures upgrade
- Boss Aurora has 50% for whatever reason. This seems to be an oversight, because missile evasion rate was nerfed from 50% in 1.02.

### this change

- All Aurora have the same 30% missile evasion rate as almost all other planes.


xezon (approved):

commy2:
Doc ok?

xezon:
i dont think it needs worldbuilder label

PR #2172, opened 2023-07-29
Fix: MOAB B2 Bomber Countermeasures Evasion Rate

commy2:
### 1.04

- Almost all airplanes, including the Airforce General B2 Carpet Bomber have a 30% missile evasion rate after Countermeasures upgrade
- The MOAB Bomber has 50% for whatever reason. This seems to be an oversight, because missile evasion rate was nerfed from 50% in 1.02.
- This error likely happened, because the object is placed in a strange file that only contains the B2 and a Chinese Bomber that has no Countermeasures.

### this change

- All B2 Bomber variants have the same 30% missile evasion rate as almost all other planes.

This is technically a nerf I suppose, but I do not think the MOAB Bomber even needs the extra missile evasion, considering that will not be shot down by Stingers or Patriots unless there are MASSIVE amounts of them. The only real impact I see is trying to MOAB massive blobs of TOW vees.

It outruns Rocketeer missiles.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2172_moab_countermeasures_missile_evasion_rate.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2172_moab_countermeasures_missile_evasion_rate.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-07-29
+
+title: Fixes MOAB B2 Bomber Countermeasures missile evasion rate
+
+changes:
+  - fix: Reduces MOAB B2 Bomber missile evasion rate after Countermeasures upgrade from 50% to 30%.
+
+labels:
+  - bug
+  - minor
+  - nerf
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2172
+
+authors:
+  - commy2
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SpecialPowerObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SpecialPowerObjects.ini
@@ -163,6 +163,7 @@ Object AmericaJetB3
     TriggeredBy           = Upgrade_AmericaCountermeasures
   End
 
+  ; Patch104p @bugfix commy2 29/07/2023 Change EvasionRate from 50% to 30% in line with other units. (#xxxx)
   Behavior                = CountermeasuresBehavior ModuleTag_12
     TriggeredBy           = Upgrade_AmericaCountermeasures
     FlareTemplateName     = CountermeasureFlare
@@ -173,7 +174,7 @@ Object AmericaJetB3
     DelayBetweenVolleys   = 1000  ; Time between flare volleys
     NumberOfVolleys       = 5     ; Number of times the volleys will fire before reloading
     ReloadTime            = 0  ; After all volleys launched, then reloading must occur. If 0, then reloading occurs at airstrip only.
-    EvasionRate           = 50%   ; With active flares, the specified percentage will be diverted.
+    EvasionRate           = 30%   ; With active flares, the specified percentage will be diverted.
     ReactionLaunchLatency = 100   ; Reaction between getting shot at and the firing of the first volley of countermeasures.
     MissileDecoyDelay     = 200   ; A reported missile that has been determined to hit a decoy will wait this long before acquiring countermeasures.
   End
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/SpecialPowerObjects.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/SpecialPowerObjects.ini
@@ -163,7 +163,7 @@ Object AmericaJetB3
     TriggeredBy           = Upgrade_AmericaCountermeasures
   End
 
-  ; Patch104p @bugfix commy2 29/07/2023 Change EvasionRate from 50% to 30% in line with other units. (#xxxx)
+  ; Patch104p @bugfix commy2 29/07/2023 Change EvasionRate from 50% to 30% in line with other units. (#2172)
   Behavior                = CountermeasuresBehavior ModuleTag_12
     TriggeredBy           = Upgrade_AmericaCountermeasures
     FlareTemplateName     = CountermeasureFlare


xezon:
Not sure if desired. It costs a lot of cash to upgrade this thing and it does not even deal more damage than the default Daisy Cutter bomb. It does have 50% more damage radius however. The plane is one of the main perks of MOAB + Countermeasures, which I think is a 6000 cash investment in total?

commy2:
$5000. It still does get the +60% armor to Quads, Gats and Avengers, as well as +10% to missiles that hit anyway.

xezon:
I suggest gather feedback from others for this. See what they say. I am neutral on this. I think the 50% evasion rate is nothing the player can really notice in an actual match.

MTKing4:
~It has always been insanely hard to kill while being fast to get enough shot resulting in it dropping it bombs before dying, this mainly nerfs AFG carpet (which is welcome if anything) and USAs late game.~

From what i read it makes sense to be reduced, but will it have a big impact on the usefulness of MOAB? I doubt it, Carpet bomb seems pretty evasive already

xezon:
Exile:
> Doesn't seem like the biggest nerf, plane is very fast and hard to take down already.
I'm quite neutral on it.

Jundiyy:
Same, neutral on it.

xezon:
Ok that is odd situation. No one cares about Moab Bomber :D

MTKing4:
We should remove neutral as an option :D

Jundiyy:
Well, my preference would be to keep it but I don't have an argument at the moment. So I'll stay neutral lol.
Now that I think about it, since there isn't any damage increase but just radius perhaps the extra evasion is good to keep.

xezon:
@commy2, are you ok with close?

commy2:
No. At least try it out first.

MTKing4:
So we have 2 Neutrals, 2 Approved, and 1 Against. It reads like an approval to me

xezon:
It needs some data. How many rocket men or Stingers or Patriots does it take to shoot this thing down before and after this change? And what is the Rationale for this, other than streamlining the evasion rate across all Countermeasure modules? Is the plane too strong and need to be not as strong?

MTKing4:
It definitely is too strong, it can escape a line of stingers like they were nothing, and streamlining of the strength is also nice, it's kinda odd that Carpet B2 and MOAB B2 are literally the same unit yet they're reflected differently in each genpower

xezon:
Would be nice to see this in action. Could we perhaps create a test map where we can send bomber over enemy Stingers and other Rockets?

MTKing4:
Yep we can, I'll do one later tonight

MTKing4:
Yep we can, I'll do one later tonight

MTKing4:
# Original

MOAB First, Carpet Second.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/2f32fb1e-b91a-462b-9e78-ec9ef73ed4cf

xezon:
That looks like Original. Can you also show with this branch?

MTKing4:
ok

commy2:
When testing like this, one has to ensure that Stinger Sites are prepared equally for both test runs. Stinger Sites will engage air units at a greater distance if they aimed at air units recently due to a weapon switching bug/oversight/quirk.

MTKing4:
ok i will redo test with different runs for each plane for patch and orignal

MTKing4:
# Original

## MOAB


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/ecf82492-96d6-4e06-b062-a071e9113783

## Carpet


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/e3f43656-0c9c-41d7-8220-9fc94dde6130


# Patched

## MOAB


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/f1654724-c71a-4d9e-8769-f87cda3b4ee8



## Carpet


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/43fb0749-4a0d-46e9-a562-aae2aee7388a

xezon:
I do not see the Countermeasure flares. Does the player have it?

MTKing4:
No

xezon:
We cannot use the attached videos. Planes need to have Counter Measures, because this change only touches that.

MTKing4:
Oh

Didn't know that

I'll do it again 😁

MTKing4:
# Original

## MOAB


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/e41daabb-5d33-47dd-b22b-695b4fe22eff



## Carpet


https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/d8bbf19f-3c7c-46ee-8ceb-5e3847e81f85




# Patched

## MOAB



https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/a02a262b-2d9a-42dc-8243-85c7648e9083




## Carpet



https://github.com/TheSuperHackers/GeneralsGamePatch/assets/83613147/61d1df71-f2ea-486d-bd28-93450ead00e6



Replays Attached for better Observation, Basically they are passing through 3 lines of 8 Stingers each, after that there's a line of quads and a line of humvees
[MvC CM Test.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/12408126/MvC.CM.Test.zip)

xezon:
It is a good test, but the result is not presented in a good way. It would be good to simply have one image each that shows where exactly original and patched Bomber dies.

It looks like the discrepancy for the Carpet Bomber is not that big, but notable, and for the MOAB Bomber the difference is really significant.

xezon:
Can you send the test map?

MTKing4:
I will make a Pr for it later tonight

xezon:
After a bit of testing I realized that the random seed is a problem. With randomization we get unreliable test results. The random seed is set on a new game start, but not on game restart.

There are 2 ways to test this

1) Force fixed random seed(s) with Thyme
2) Add before and after Bomber to faction so they can be tested in one game session (with restart)

xezon:
I suggest to close this. It is not worth the effort to investigate in all detail. MOAB Bomber is not considered overpowered. This Evasion Rate discrepancy is unlikely to bother anyone because it is difficult to observe the rocket miss difference between 30% and 50%.

commy2:
> This Evasion Rate discrepancy is unlikely to bother anyone

It bothers me.

> it is difficult to observe the rocket miss difference between 30% and 50%.

That's just more reason to do it.

xezon:
> It bothers me.

This is not good enough reason to change in this case.

> That's just more reason to do it.

I meant just in the context of missile misses, not in terms of plane reach. Discrepancy in reach certainly is observable, but unreliable to compare due random seed as mentioned above.

commy2:
> This is not good enough reason to change in this case.

The reasons to change this are to make the upgrade boost consistent with almost any other plane in 1.04 aside from units that only appear on the very last mission of the GC or on stuff like Casino maps, and to especially make it consistent with another General's Power that is delivered by, at least visually, the exact same plane.

> Discrepancy in reach certainly is observable, but unreliable to compare due random seed as mentioned above.

If its so unreliable and luck based then it especially should not also depend on what bombs exactly are dropped by the plane.

xezon:
> The reasons to change this are to make the upgrade boost consistent with almost any other plane in 1.04 aside from units that only appear on the very last mission of the GC or on stuff like Casino maps, and to especially make it consistent with another General's Power that is delivered by, at least visually, the exact same plane.

Logically this reasoning is sound and if MOAB was overpowered then this probably would be ok to change, but there is currently no known evidence that MOAB is overpowered and so it is also reasonable to just leave the setup as is.

> If its so unreliable and luck based then it especially should not also depend on what bombs exactly are dropped by the plane.

There definitely is a reach advantage observable in the test map. But it is unreliable to paint it because it comes at random variations.

Here is the discrepancy of the first comparison I did.

shot_20230824_193305_1

MTKing4:
You are restarting the game after each bomber right?
There shouldn't be that much difference

> When testing like this, one has to ensure that Stinger Sites are prepared equally for both test runs. Stinger Sites will engage air units at a greater distance if they aimed at air units recently due to a weapon switching bug/oversight/quirk.

MTKing4:
What random seed? I don't see any

commy2:
https://en.wikipedia.org/wiki/Pseudorandom_number_generator

MTKing4:
I know what random seed means :3
Was asking where is it in the MOAB test, because his test seems to be identical to mine

xezon:
Every match starts with new random seed and random seed determines outcome of "random" events. Have 30% miss chance? That may be 25% or 35% on small sample size depending on the random seed that was chosen.

MTKing4:
Can you see this in code? Or just by observation?
Is it related to stinger troopers orientation? Or maybe response time, commy said they all equally oriented at the start of each match so it must be something else?

commy2:
50% of all missiles miss. Whether any particular missile misses is determined by the PRNG. Does the first missile miss? That is determined by the seed.

xezon:
Close?

commy2:
Merge :)

xezon:
The current rationale is

> This is technically a nerf I suppose, but I do not think the MOAB Bomber even needs the extra missile evasion, considering that will not be shot down by Stingers or Patriots unless there are MASSIVE amounts of them. The only real impact I see is trying to MOAB massive blobs of TOW vees.

It appears to not convincingly explain why the MOAB should have a weaker plane. One must put its difficult requirements into perspective:

* Command Center (2000)
* Rank 5
* 1 Generals Point
* Strategy Center (2500)
* MOAB Upgrade costing 4000
* Countermeasures Upgrade costing 2000

In relation, the Airforce Carpet Bomber is much easier to obtain, and therefore a weaker plane is also ok.

* Rank 1
* 1 Generals Point
* Strategy Center (2500)
* Countermeasures Upgrade costing 2000


Can we prove that the MOAB Bomber does not need the extra missile evasion? Are all USA's overpowered with original MOAB? What about USA Superweapon General? Are we ok with nerfing it?

If we cannot come up with convincing arguments pro change then we need to close this.

commy2:
> It appears to not convincingly explain why the MOAB should have a weaker plane.

It shouldn't have a weaker plane. It just shouldn't have a stronger plane for no discernible reason either. Arguing about how this nerfs MOAB somehow is completely missing the point and the realities of the game. Watch any long 1v1 where USA gets MOAB. Or watch Dominators FFA. No one ever gets Countermeasures to make the MOAB bomber invulnerable. Because it doesn't even need the extra evasion, since it already is so fast that it is guarenteed to hit. Except against ungodly amounts of AA like in https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2172#issuecomment-1693728596 which you will never see in a serious game where the winner hasn't already been decided and they aren't just messing around.

MTKing4:
It's true, no one has ever built Countermeasures to make MOABs stronger, they don't even know it does, and many times they'll use it without having an airfield because they don't need one except AFG and SWG

I don't have strong feelings about this change, but i think it will mostly be fine to normalize

xezon (approved):
Go ahead and merge it if you like. I think it is not necessary. It is ok if MOAB plane is stronger.

PR #2173, opened 2023-07-29
Fix: Lone Stinger Soldiers float above ground when moving while reloading

commy2:
### 1.04

Lone Stinger Troopers float over the ground when moving while reloading.

### patch

Lone Stinger Troopers use move animation while moving and reloading

Notes:

Lone Stinger Troopers only appear in the Generals Challenge vs Kassad, or on custom maps.


commy2:
https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/1d7b25b8-aca4-41f7-8299-94810b518afd

commy2:
ready

xezon (approved):

PR #2175, opened 2023-07-30
Fix: infantry stuck in walk animation when dying

### LINKED ISSUE #2174: Stunwalking infantry displays wrong death animation
When an Outpost is popped, passenger may be forced to walk. If they die in this state, they display a frozen walk animation instead of the proper death animation.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/a520fb11-628f-4c6d-86ee-f02d36739e31

This seems to be because the engine forces them to walk, which applies MOVING flag. They are MOVING and DYING, which is not supposed to happen. Because MOVING is always defined first, they use that animation instead of the death one.

Solution is to go through all the infantry units in the game and add AliasConditionState DYING MOVING under DYING state.

commy2:
- fixes https://github.com/TheSuperHackers/GeneralsGamePatch/issues/2174

Purely visual

- [x] Ranger
- [x] MD
- [x] Pathfinder
- [x] Burton
- [x] CIA Agent
- [x] Biohazard
- [x] USA Officer
- [x] Pilot
- [x] Red Guard
- [x] Minigunner
- [x] Tank Hunter
- [x] Hacker
- [x] Lotus
- [x] China Officer
- [x] Ambassador
- [x] Worker
- [x] Rebel
- [x] Tox Rebel
- [x] RPG Solider
- [x] Tox RPG Soldier
- [x] Mr. T
- [x] Jarmen Kell
- [x] Saboteur
- [x] Hijacker
- [x] Stinger Trooper
- [x] Sniper
- [x] misc units
- [x] test with random units

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2175_broken_infantry_death_animations.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2175_broken_infantry_death_animations.yaml
@@ -1,5 +1,5 @@
 ---
-date: 2023-07-29
+date: 2023-07-30
 
 title: Fixes a bug where infantry remains stuck in walk animation when dying
 
--- a/Patch104pZH/Design/Changes/v1.0/2175_broken_infantry_death_animations.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2175_broken_infantry_death_animations.yaml
@@ -1,0 +1,22 @@
+---
+date: 2023-07-29
+
+title: Fixes a bug where infantry remains stuck in walk animation when dying
+
+changes:
+  - fix: All infantry units now always display death animation when dying while moving.
+
+labels:
+  - boss
+  - bug
+  - china
+  - gla
+  - minor
+  - usa
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2175
+
+authors:
+  - commy2
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -4214,12 +4214,14 @@ Object AirF_AmericaInfantryPilot
     End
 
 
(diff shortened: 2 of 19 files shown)


commy2:
Intermediate state. Fix applied to USA units, except Burton:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/240f4a72-b4c2-47f2-9431-e75722f70579

as expected, fix works as Rangers die normally while Burton remains stuck

commy2:
ready

xezon:
What about all the civilian guys? And the Cine and Angry Mob guys? Would it hurt to add?

commy2:
> civilian guys

If there is a way to reproduce the issue on them, sure. But how would they end up in a transport?

> Angry Mob

Same, can't enter vehicles.

> Cine

No point, they're hopelessly outdated already in 1.04. Most are CCG state. In general, don't mess with cinematics. They have specific changes for the cutscene they appear in. If the death animation were an issue in those cutscenes, it would've been fixed or worked around already.

xezon:
I would have treated all infantry setups equally. Reason being, consistent setups are easier to manage and also avoid wrong setups if a Mod creator decides to use some of the untreated objects as new faction units. For example someone may add a Molotov soldier to GLA.

xezon (approved):

commy2:
Well, it definitely should be added to the UNSoldier.

Will also treat the civies and maybe Mobsters. Still don't like messing with cinematics though.

commy2:
- [x] UNSoldier
- [x] MogadishuFemaleCivilian01
- [x] MogadishuFemaleCivilian02
- [x] MogadishuMaleCivilian01
- [x] MogadishuMaleCivilian02
- [x] MogadishuMaleCivilian03
- [x] GenericFemale01
- [x] GenericFemale02
- [x] GenericMale01
- [x] GenericMale02
- [x] AmericanFarmer01
- [x] AsianFarmer01
- [x] AsianFarmer02
- [x] AsianFarmer3
- [x] HomelessGuy :(

commy2:
Mob should get their own PR. They have different issues anyway.

PR #2176, opened 2023-07-30
Fix: Streamline Nuke Missile and Scud Storm tool tip strings for all languages

xezon:
This change streamlines Nuke Missile and Scud Storm tool tip strings for all languages (except Arabic). They now say "Strong against everything" like English does.


xezon:
Documentation added. Good.

PR #2177, opened 2023-07-30
Fix: Add status requirement to GLA Demolition death weapons

commy2:
Same as https://github.com/TheSuperHackers/GeneralsGamePatch/pull/764, but for Demogen.
May also fix https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2164#issuecomment-1656742863


commy2:
Unused Demogen Saboteur does not support Suicide ability. Probably another issue.

xezon:
Perhaps can be added to 764_terrorist_crash.yaml ?

commy2:
done

xezon (approved):

PR #2180, opened 2023-07-30
Fix: Use Chinese colon character consistently

xezon:
**Merge with Rebase**

This change replaces the ascii colon ':' character with the unicode colon '：' in Chinese strings to make their appearance consistent.


xezon:
Documentation added.

xezon:
The colon looks like this. Looks good.

shot_20230730_191305_1

PR #2182, opened 2023-07-31
Implement: Missing animation for Secret Service unit

### LINKED ISSUE #2079: AmericaInfantrySecretService missing some animation states
Incomplete implementation in 1.04

Custom map / worldbuilder unit only.

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/2079

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/c4ebc147-c2ff-4fa8-a55c-a83b8f8dfd7f

Suitcase disappears while moving, but better than floating above ground I suppose.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2182_missing_secret_service_animations.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2182_missing_secret_service_animations.yaml
@@ -1,0 +1,20 @@
+---
+date: 2022-07-31
+
+title: Fixes missing animations of Secret Service unit
+
+changes:
+  - fix: Add walk and death animations to Secret Service.
+
+labels:
+  - bug
+  - civilian
+  - minor
+  - v1.0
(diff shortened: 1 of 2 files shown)


xezon:
Perhaps remove the suit case?

commy2:
Could do, but without suitcase, he's just like CIA Agent:

shot_20230731_230030_1

But without a gun. Making him even more pointless.

xezon:
Ok. Perhaps we can give him animations from Lotus or Hackers? Since they have suitcases?

commy2:
It doesn't work like that:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/4c3ea9e2-21e7-46ab-8bd4-de1b528c391e

It's either this, making him unmovable (and unkillable?) or nothing.

xezon (approved):

PR #2184, opened 2023-07-31
Add: Prison

commy2:
Also for worldbuilder

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
@@ -4,7 +4,7 @@ date: 2022-07-31
 title: Adds unused GLA Prison to World Builder
 
 changes:
-  - fix: Add Prison to world builder.
+  - feature: The GLA Prison can now be placed in the world builder.
 
 labels:
   - enhancement
--- a/Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
@@ -1,0 +1,20 @@
+---
+date: 2022-07-31
+
+title: Adds unused Prison
+
+changes:
+  - fix: Add Prison to world builder.
+
+labels:
+  - enhancement
+  - minor
+  - gla
+  - v1.0
+  - worldbuilder
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2184
+
+authors:
+  - commy2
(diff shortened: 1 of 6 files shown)


xezon:
It appears to be missing content.

commy2:
Fixed. Detention Camp has Intelligence, but this one was pointless, so I gave it Sat Hack 1 basically.

xezon:
There is also an `add-prisons` branch.

commy2:
It can be deleted. Better to integrate them separately. Honestly, add-prisons only exist, because I was too lazy to use an USB to transfer the code from desktop to laptop.

Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2022-07-31
+
+title: Adds unused Prison
```
xezon:
```suggestion
title: Adds unused GLA Prison to World Builder
```

Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+changes:
+  - fix: Add Prison to world builder.
```
xezon:
```suggestion
  - feature: The GLA Prison can now be used in Mods and Maps.
```

xezon (commented):
### Damage transition effects are very large

Perhaps can use the ones of the Black Market.

shot_20230808_200623_4

### Command Set

Tool tip says it can spy on enemies, but actually it allows to build a Scorpion Tank.

shot_20230808_200445_1

### Just a Spy Ability?

Functionality looks a bit underwhelming. Perhaps can add some more gimmicks? On the other hand, perhaps up to Mod Maps makers what they want to do with it.

### No snow version

Has no snow version. As designed?

shot_20230808_202259_7

commy2:
> Tool tip says it can spy on enemies, but actually it allows to build a Scorpion Tank.

Scorpion is for showcasing the door. Should be taken out before merge. It has Sathack 1, hence spying on enemies.

> Has no snow version. As designed?

No models for that. Would have to double check if textures exist, but then models would have to be made. Maybe that's overkill?

xezon:
> Maybe that's overkill?

Up to you.

Patch104pZH/Design/Changes/v1.0/2184_detention_camp.yaml
commy2:
```suggestion
  - feature: The GLA Prison can now be placed in the world builder.
```
It can be used in mods already.

xezon (approved):

PR #2189, opened 2023-08-01
Fix: USA Hellfire Drone spawns Scout Drone debris when destroyed

commy2:
self explanatory

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2189_hellfire_drone_debris.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2189_hellfire_drone_debris.yaml
@@ -4,7 +4,7 @@ date: 2023-08-02
 title: Fixes wrong USA Hellfire Drone debris when shot down
 
 changes:
-  - fix: Hellfire Drone now uses unique debris when destroyed instead of Scout Drone debris.
+  - fix: The destroyed USA Hellfire Drone now shows unique debris instead of the Scout Drone debris.
 
 labels:
   - bug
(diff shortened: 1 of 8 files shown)


xezon:
Ah nice. This reminds me that I was also not yet done with Hulk fixes. All the vehicle crush hulk bugs are yet to be done.

commy2:
good to go. Comments exist for the second variant (with more debris parts)

Patch104pZH/Design/Changes/v1.0/2189_hellfire_drone_debris.yaml
```diff
@@ -0,0 +1,19 @@
+---
+date: 2023-08-02
+
+title: Fixes Hellfire drone debris when shot down
```
xezon:
```suggestion
title: Fixes wrong USA Hellfire Drone debris when shot down
```

Patch104pZH/Design/Changes/v1.0/2189_hellfire_drone_debris.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+changes:
+  - fix: Hellfire Drone now uses unique debris when destroyed instead of Scout Drone debris.
```
xezon:
```suggestion
  - fix: The destroyed USA Hellfire Drone now shows unique debris instead of the Scout Drone debris.
```

xezon (approved):

PR #2190, opened 2023-08-02
Add: Use snow textures for Supply Pile and Supply Docks

commy2:
Textures already in CCG/ZH. Had to export a model using it for the boxes:

shot_20230802_101911_1

Please double check it looks good.

Pile:

shot_20230802_091127_1

Looks really cool, but it has to be considered that especially Supply Piles are harder to see now:

shot_20230802_091355_1


xezon (approved):
Looks good. Needs documentation.

commy2:
ok?

PR #2191, opened 2023-08-02
Fix: Stinger Soldiers can fire while Stinger Site is being sold (bad fix)

commy2:
Taken from NPM.

The point is to kill the Troopers when model state switches to SOLD.

This is a bad idea for several reasons:

- they can still fire for a variable amount of time due to how FireWeaponCollide works
- this will kill or injure other units near the Stinger Site when it is sold, which makes no sense at all
- EVA noise "unit lost" will be played when Stinger Site is sold

In conclusion, the Stinger Site is weirder with this than without this. A proper fix would need engine support. I advice to close this.


xezon:
What was this trying to fix?

commy2:
Stinger Soldiers can fire while Stinger Site is being sold

xezon:
If its bad fix, then close.

PR #2192, opened 2023-08-02
Fix: Add implementation of Partisans and Militia Man

commy2:
This implements the unused models for the broken Partisan worldbuilder unit. Could be placed on a fun map as third ai faction that attacks all/some players, similar to The Forgotten in C&C or Goblins/Wargs/Cave Trolls in BFME. It also implements the Milita Man model.

~I gave them AK Mob weapons for spice.~

I gave them custom dual uzi, shotgun, semi-automatic rifle weapons depending on the model.

For some reason, they all have parachute animations.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2192_militia_man.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2192_militia_man.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-07
+
+title: Adds Militia Man
+
+changes:
+  - feature: Adds armed Milita guy to worldbuilder.
+
+labels:
+  - civilian
+  - enhancement
+  - minor
+  - v1.0
+  - worldbuilder
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2192
+
+authors:
+  - commy2
--- a/Patch104pZH/Design/Changes/v1.0/2192_partisans.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2192_partisans.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-07
+
+title: Implements Partisans
+
+changes:
+  - feature: The armed Partisans are now fully functional worldbuilder units.
+
+labels:
+  - civilian
+  - enhancement
+  - minor
+  - v1.0
+  - worldbuilder
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2192
+
+authors:
+  - commy2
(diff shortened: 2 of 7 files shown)


commy2:
Found this unimplemented model too:

shot_20230807_230251_1

commy2:
exploded death of MilitaMan needs fixing.

xezon:
Nice stuff. My 2 cents:

### 1. No Militia guy potrait

Militia guy could also have that Partisan Portrait. Better than none. Because if he has none, it will show an empty slot in transporters and houses.

shot_20230808_183612_2

### 2. Muzzle flash of Militia guy is long

Muzzle flash of Militia guy looks a bit long. Model issue I think. Not a concern for this change.

shot_20230808_183724_12

### 3. No voices

Do other playable Civilian units like this have voices? If yes then this perhaps should too. Perhaps can just use GLA Rebel voices.

### 4. Shotgun Partisan fires machine gun

This guy has a shotgun like animation, but fires machine gun. Perhaps we can replace weapon with single shot to match the animation?

shot_20230808_184317_16

### 5. Enemy AI Militia guy does not attack Player

Perhaps also happens with the Partisan guys. I could not see, because USA just spawns the Partisan guys at the start.

shot_20230808_185036_18

### 6. Gun effects look different from faction solder guns

When Rangers & Co shoot then they will have lens flare and bullet case effects. Partisans have actual light emitters instead. Perhaps is better to make them look consistent if possible.

shot_20230808_185514_41

commy2:
Yeah, it's because I gave them the Mob Ak47s. Could implement something more fitting for the respective models.

commy2:
> Do other playable Civilian units like this have voices? If yes then this perhaps should too. Perhaps can just use GLA Rebel voices.

Unarmed civs never have voices. The only other armed civ is UN Soldier. Would have to check, but iirc, he does have neither voice nor portrait. UN Humvve and Milita Tank do not have voice or portraits either.

MTKing4:
> ### 1. No Militia guy potrait
>
>Militia guy could also have that Partisan Portrait. Better than none. Because if he has none, it will show an empty slot in >transporters and houses.

Maybe do it with AI

commy2:
I'm not so sure about the current Partisan portrait either. It's a GLA dude, possibly higher up. Doesn't really match the Partisans.

commy2:
- should automatically attack now
- grant xp when killed (20)
- each one has a weapon according to the model

xezon (approved):
Good job with the weapons. Looks and sounds excellent. Bugs are fixed. Portrait does not look like a good match. Missing voice makes them not pleasant to control.

Remove the debug stuff before commit.

PR #2193, opened 2023-08-02
Fix: Remove tiny debris pieces from exploded USA Scout Drone

xezon:
This change removes 2 tiny debris pieces from the exploded USA Scout Drone, which were barely visible. It now spawns 3 instead of 5 pieces on death, as much as Battle Drone.


commy2 (approved):

xezon:
Documentation added. Good.

PR #2194, opened 2023-08-02
Fix: Infinitely circling plane and heli wrecks

commy2:
* Follow up for TheSuperHackers/GeneralsGamePatch#1018

Same as https://github.com/TheSuperHackers/GeneralsGameCode/issues/62, but for all planes.


Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -17675,7 +17675,7 @@ Object AirF_AmericaVehicleChinook
     Mass = 50.0
   End
   Behavior = HelicopterSlowDeathBehavior ModuleTag_10
-    DestructionDelay                = 99999999        ; the destruction delay
+    DestructionDelay                = 10000 ; Patch104p @bugfix commy2 02/08/2023 Fix immortal plane wreck continuing to circle by deleting it after 10 seconds. ((#2194))
```
xezon: Comments (())

xezon: Are 10 seconds certainly enough to not delete all these before final death? Some of these things tumble a bit.

commy2:
```py
    Gravity = 64  # GameData.ini
    PreferredHeight = 100  # RaptorJetLocomotor
    FallHowFast = 1.1  # 110.0% in JetSlowDeathBehavior

    from math import sqrt
    t = sqrt(2 * PreferredHeight / (Gravity * FallHowFast))
    print(t)  # 1.69 seconds
```
It takes ~1.7 seconds for a plane to fall from preferred height above flat terrain.

xezon: The Chinooks and Comanches and Helix tumble when falling. But it takes less than 10 seconds I think. But we better check to make sure.

commy2: Helix/Chinook/Comanche: FallHowFast 0.12, PrefferedHeight 100, gives me 5.1 seconds

commy2: NPM is using 10 seconds in HelicopterSlowDeathBehavior. I think NPM has been playtested a lot to say the least.

xezon: ok

commy2: In 10 seconds, Helix etc. will fall 380 ft, which is 3.8 times the preferred height.

xezon:
1018_dead_aurora_bug.yaml needs updating/renaming.

xezon (approved):
Very good.

PR #2195, opened 2023-08-03
Fix: Errors in China Infantry Mini Gunner and Troop Crawler tool tip strings

xezon:
This change fixes errors in Infantry Mini Gunner and Troop Crawler tool tip strings.

Mini Gunner strings no longer say "Strong vs air", but "Strong vs aircraft", like all other tool tips.

And Mini Gunner Troop Crawler strings no longer claim to be "Weak vs aircraft". Instead they are now labeled "Strong vs aircraft".


xezon:

PR #2196, opened 2023-08-03
Fix: Promoted Crusaders, Laser and Microwave tanks spawn Pilots when crushed

commy2:
From NPM tasks. Makes sense to me. Paladin doesn't spawn pilots either when crushed. Only way to crush is an Overlord, or on fun maps, getting hit by the avalanche object.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2196_crushed_tank_pilots.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2196_crushed_tank_pilots.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-08-03
+
+title: Fixes promoted Crusader, Laser and Microwave tanks spawn pilots when crushed.
+
+changes:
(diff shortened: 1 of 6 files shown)


Patch104pZH/Design/Changes/v1.0/2196_crushed_tank_pilots.yaml
```diff
@@ -0,0 +1,21 @@
+---
+date: 2023-08-03
+
+title: Fixes promoted Crusader, Laser and Microwave tanks spawn pilots when crushed.
```
xezon: Dot not desired in title.

xezon (approved):
Good.

Patch104pZH/Design/Changes/v1.0/2196_crushed_tank_pilots.yaml
commy2:
```suggestion
title: Fixes promoted Crusader, Laser and Microwave tanks spawn pilots when crushed
```

PR #2197, opened 2023-08-03
Fix: Laser turret has no snow textures

commy2:
shot_20230803_201905_1

- Note: All Patriots lack ES texture and use DS instead. I made the same choice here. If ABPatriot_ES texture version is ever added, this model as well as the other ones can be adapted by hexedit 3 liner.
- Note 2: The round plate below the Turret is using the Crusader texture (AVLeopard), which has no snow version. Looks fine though.


Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
```diff
@@ -16204,6 +16204,7 @@ End
 
 
 
+; Patch104p @bugfix commy2 03/08/2023 Apply snow textures on snow maps. (#2197)
```
xezon: Comment is above wrong object.

xezon (approved):

Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
commy2: lol how did that happen

PR #2208, opened 2023-08-05
Fix strings for Mouse:FireBomb, Mouse:Defector

xezon:
This change fixes strings for Mouse:FireBomb, Mouse:Defector.


xezon:
Requires no special documentation. Text is not user facing. Perhaps debug or legacy.

xezon:
Fine.

PR #2209, opened 2023-08-05
Fix: Listening Outpost headlights barely visible from some angles, turned on while stationary

commy2:
* Follow up for #261

This pull request fixes a model issue that causes the headlight to be barely visible when viewed from some angles. 

### after from left (working)

no change after from right

- This pull request also changes the headlight when stationary. It logically makes no sense for a stealthed vehicle to have the lights turned on at night. Therefore, with this PR, the Outpost now turns off the lights when parked:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/bf006c11-f61f-4b46-bfb4-e14b893925e3

This includes some randomization to make it look properly when mass moving Outposts.


xezon:
May I suggest to apply the new lamps toggle separately, together with the sentry drone?

commy2:
I think I will fix each individually, or maybe some in batches like all the Vee variants.

Fixing headlights first, then adding the light toggle wouldn't work for Outpost, as both require mutually exclusive model setups (bone name HEADLIGHT is hard-coded to turn on/off depending on time of day and ignored manual override).

xezon:
Ok. How do we want to document lamp fixes?

Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
```diff
@@ -2719,24 +2719,80 @@ Object ChinaVehicleListeningOutpost
...
+      AnimationMode = ONCE
+      AnimationSpeedFactorRange = 0.3 0.6
+      ShowSubObject = ManualLight
+    End
+
+    ; damaged
+    ConditionState = REALLYDAMAGED ; headlight off
+      Model = NVLOutpost_D
+      TransitionKey = Trans_LightOffDamaged
+      HideSubObject = ManualLight
```
xezon, line 2763: Same as in the other change, perhaps this name should be UPPERCASE?

Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
```diff
@@ -4965,23 +4965,80 @@ Object Tank_ChinaVehicleListeningOutpost
...
+
+    TransitionState = Trans_LightOnDamaged Trans_LightOffDamaged ; turn off headlight
+      Model = NVLOutpost_D
+      Animation = NVLOutpost_D.NVLOutpost_D
+      AnimationMode = ONCE
+      AnimationSpeedFactorRange = 0.3 0.6
+      ShowSubObject = ManualLight
     End
 
+    ; kaput
```
xezon: destroyed

xezon (commented):
Needs documentation. I suggest merge documentation with 261_outpost_smoke_effect.yaml

Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaVehicle.ini
commy2: Model forces bones to be uppercase, but I find pascal case to be much easier to read. You can make out the individual words for starters. I've been using pascal case for this in many prs before, and even EA does it everywhere. Recent example: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2212/files#diff-ebb635495973f844c35841b5ee18e09a56ea37b2d2f3bfb967018a861ca19492L16321-R16325

xezon: ok

xezon (approved):

PR #2210, opened 2023-08-05
Fix: Avenger headlight barely visible from some angles

commy2:
* Follow up to #158

### before

shot_20230805_124231_1

shot_20230805_124239_2

### after

shot_20230805_124129_1

shot_20230805_124355_1


xezon (approved):
Needs documentation. I suggest to merge documentation with 158_avenger_model.yaml

PR #2212, opened 2023-08-05
Fix: Model issues with Scorpion tank

commy2:
## 1.04

shot_20230805_190623_1

- Chassis is not deformed when damaged, unlike Scorpion in CCG
- Some parts on the turret and barrel use undamaged texture

## this patch

shot_20230805_191819_1

- Chassis uses deformed model from CCG
- All undamaged textures have been replaced with the damaged versions

### ini configuration

- use correct muzzle flash when scrapped up
- fix recoil animation only played for every second shot

### UVLiteTank

- both muzzle flash (basic and upgraded gun) now are visible from any side

### UVLiteTank_D

- both muzzle flash (basic and upgraded gun) now are visible from any side
- barrel, missile racks and turret shields now use the UVLiteTank_D.dds texture
- Chassis is replaced with the deformed chassis from CCG

### UVLiteTank_D1

- threads no longer turn


xezon:
Did you remove Basic_Sphere from the file?

Also, I suggest to save blend file with compression. Then it is much smaller.

https://blender.stackexchange.com/questions/45664/how-do-i-reduce-my-file-size

commy2:
I don't see any basic spheres in the model.

> Also, I suggest to save blend file with compression. Then it is much smaller.

👍

xezon:
Nice. Do you plan to append D1 fixes to this pull too?

commy2:
ye

xezon:
Nice. You can also fix the muzzle flash bugs in all Scorpion models (Two Sided).

MTKing4:
Basic Sphere is present in all of the 3 .blend files

image

commy2:
Basic spheres deletes. All lgtm now. Ok to merge?

I think it's fine to use Blender plugin now as long as import is used on models without sections that contain multiple textures (such as the Scorpion tank).

xezon:
I would prefer holding off merging blend files until we have made pass on W3D to blend situation (Mod Builder).

PR #2220, opened 2023-08-07
Fix: Spectre wreck is silent when crashing on the ground

commy2:
### 1.04

- Spectre has no sound when impacting on the ground, unlike any other plane

### patch

- Spectre has generic explosion sound

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2220_spectre_wreck_impact_sound.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2220_spectre_wreck_impact_sound.yaml
@@ -7,6 +7,7 @@ changes:
   - fix: The Spectre now uses a generic explosion sound when falling to the ground.
 
 labels:
+  - audio
   - boss
   - bug
   - minor
--- a/Patch104pZH/Design/Changes/v1.0/2220_spectre_wreck_impact_sound.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2220_spectre_wreck_impact_sound.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-07
+
+title: Fixes issue with Spectre wreck being silent when crashing onto the ground
+
+changes:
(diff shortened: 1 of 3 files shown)


Patch104pZH/Design/Changes/v1.0/2220_spectre_wreck_impact_sound.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2023-08-07
+
+title: Fixes issue with Spectre wreck being silent when crashing onto the ground
+
+changes:
+  - fix: The Spectre now uses a generic explosion sound when falling to the ground.
+
+labels:
```
commy2, line 9:
```suggestion
labels:
  - audio
```

xezon:
I tested crash sound. Scud missile does not sound good.

I suggest to derive sound from Helix crash sound. Helix size is closest in size to Spectre.

I tested this and it sounds good.

```
; Patch104p @feature commy2 08/08/2023 Adds a copy of ComancheCrash, but a bit louder and deeper.
AudioEvent SpectreCrash
  Sounds = vcomcraa
  Limit = 3
  PitchShift = -10 -10
  Volume = 90
  VolumeShift = -10
  Priority = normal
  Type = world shrouded everyone
End
```

Patch104pZH/GameFilesEdited/Data/INI/SoundEffects.ini
```diff
@@ -7980,3 +7980,13 @@ AudioEvent SpectreHowitzerWeapon
 End
 
 
+; Patch104p @feature commy2 09/08/2023 Adds a copy of ComancheCrash, but a bit louder and deeper.
+AudioEvent SpectreCrash
```
xezon: I suggest to move this definition near `AudioEvent ComancheCrash`, since it is similar.

xezon (approved):

PR #2223, opened 2023-08-11
Fix: Chemical Suits for Pilots

### LINKED ISSUE #404: USA Pilot has non-functional Chemical Suit upgrade and crashes the game
USA Pilot has non-functional Chemical Suit upgrade.

GeneralCamo:
> I may have figured out a method to grant Pilots the benefits of Chemical Suits. Use a different upgrade, that spawns a decal under the pilot and grants them an armor boost. That upgrade is granted by an object that is spawned from the strategy center when Chemical Suits is researched. This avoids the issue with `ArmorUpgrade` crashing the game with `Upgrade_AmericaChemicalSuits` when a vetted pilot spawns. The only side effect I can see is that due to how these decals would work (FS_FAKE), the chem suit decal on the pilot would not be visible to enemy players.

Crashes the game when a vetted infantry spawns from a vehicle:
```
  Behavior                = ArmorUpgrade ModuleTag_Armor01
    TriggeredBy           = Upgrade_AmericaChemicalSuits
  End
```

Does not crash the game:
```
  Behavior                = ArmorUpgrade ModuleTag_Armor01
    TriggeredBy           = Upgrade_DifferentOne
  End
```

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/404 chemsuits not found

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/a31aba34-1145-4c3d-86ca-b651567b92e4

This is a real decal, will be seen by the enemy.

Remaining problem is getting the replacement upgrade to work, meaning, make the armor set actually change for the Pilot.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AmericaInfantry.ini
@@ -1415,7 +1415,7 @@ Object AmericaInfantryPilot
   ButtonImage            = SAPilot
 
   ; Patch104p @bugfix hanfield 03/09/2021 Commented out Chemical Suit and Advanced Training cameos since it never benefited from either.
-  ;UpgradeCameo1 = Upgrade_AmericaChemicalSuits
+  UpgradeCameo1 = Upgrade_AmericaChemicalSuits
   ;UpgradeCameo2 = Upgrade_AmericaAdvancedTraining
   ;UpgradeCameo3 = NONE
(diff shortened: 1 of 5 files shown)


xezon:
Wouldn't it be better to simply fix the crash in Thyme and use the regular INI implementation for it?

commy2:
Yes it would.

xezon:
Currently Chemical Suits are disabled, so the crash is gone. This is already good. Up to you if you want this in now, but it would then have to be changed again when the Thyme fix comes.

PR #2226, opened 2023-08-12
Fix: Set correct names for China Troop Crawler and Assault Troop Crawler in English, Italian languages

xezon:
This change corrects the names of the China Troop Crawler and Assault Troop Crawler in English and Italian languages.


xezon:

PR #2230, opened 2023-08-13
Add faction decal for Boss General

xezon:
This change adds a faction decal texture for Boss General.

zhca_nbboss


commy2:
shot_20230813_113032_1

Doesn't show for me. Guess the second texture is required for China CC.

xezon:
How do you change texture?

commy2:
Hexedit.

xezon:
From what to what did you change?

commy2:
Infantry. It has the longest filename.

xezon:
I mean from what texture name to what texture name did you change?

commy2:
ZHCA_NBInftry.tga -> ZHCA_NBBoss.tga (with 0 byte padding)

xezon:
I do not know what is wrong. The texture builds fine. Check if texture exists in `Patch104pZH\.Build\RawBundleItems\CoreTextures\Art\Textures\zhca_nbboss.tga`

xezon:
Can you share w3d file?

xezon:
I noticed that the alpha channel was inverted compared to the original decals. I have flipped it.

PR #2074, opened 2023-07-08
Fix: Infantry death modules

xezon:
**Merge with Rebase**

This change fixes and streamlines death modules of

* (USA) AmericaInfantrySecretService
* (China) ChinaInfantrySecretPolice
* (China) ChinaAmbassador
* (China) ChinaInfantryOfficer
* (China) ChinaInfantryAgent
* (Civilian) MogadishuMaleCivilian01
* (Civilian) MogadishuMaleCivilian02
* (Civilian) MogadishuFemaleCivilian02
* (Civilian) MogadishuFemaleCivilian01
* (Civilian) GenericFemale01
* (Civilian) AmericanFarmer01
* (Civilian) AsianFarmer01
* (Civilian) HomelessGuy
* (Civilian) AsianFarmer02
* (Civilian) AsianFarmer3
* (Civilian) GenericMale02
* (Civilian) GenericFemale02
* (Civilian) Partisan01
* (Civilian) Partisan02
* (Civilian) Partisan03

All infantry (except GLA Terrorist and CINE infantry) now have the same setup and will show burned, toxic and other special deaths properly.

The added setup already respects the modified LASERED setup of #2071.

PR #2231, opened 2023-08-13
Fix: Broken Exploded death animation for certain civilian units

### LINKED ISSUE #2221: Broken civilian exploded death animations
This PR: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2074/files added "missing" death modules to infantry, including civilians. Many of these civilians do not have any "flailing in the air" death animations, so with those new modules, it now looks like they're jumping into the air upright when exploded, which looks very very wrong.

Propose deleting the EXPLODED death modules that apply FlingForce/FlingPitch, since no supporting animation exists.

commy2:
- continuation of https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2074
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/2221

- ChinaAmbassador has fling animations that were not implemented (the commented ones where mispelt)
- GenericFemale02 has fling animations that were not implemented
- MogadishuMaleCivilian01, MogadishuMaleCivilian02 and MogadishuFemaleCivilian01 do not have fling animations, so I removed the FlingForce/Pitch entires, which makes exploded death equivalent to regular death (same as Milita Man in #2192)

This means every infantry unit, except three Arab Civilians, and the Milita guy support fling death.


xezon (approved):
👍

PR #2236, opened 2023-08-13
Fix: Issues with USA, China and CIA Officers

commy2:
- replaces close https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2073


Patch104pZH/Design/Changes/v1.0/2236_officers_and_cia_agent.yaml
```diff
@@ -0,0 +1,39 @@
+---
+date: 2023-08-15
+
+title: Fixes issues with America and China Officer as well as CIA Agent
```
xezon: Perhaps make 3 separate files for these since they have a number of changes each?

commy2: done

xezon (approved):
I briefly tested in game and it looks good. I noticed there are issues with CIA voice quality. I will fix that.

PR #2237, opened 2023-08-16
Fix: INFANTRY KindOf on unused Toxic Infantry

commy2:
Follow-up on #2072. INFANTRY flag remained on two unused copies of ToxicInfantry. Remove for completeness and consistency.

Also, curiously, despite both of them being named "Gamma", they're blue dabadee dabadie


xezon:
yaml needs update

commy2:
There is none. There's just a txt palceholder.

xezon:
75_dead_target_bug.yaml

xezon (approved):

PR #2243, opened 2023-08-17
Fix: Improve detonate and suicide strings for French, Spanish, Chinese, Brazilian languages

xezon:
This changes improves the detonate strings for French, Spanish, Chinese, Brazilian languages. They are now written more consistently and closer to what the English text says.


xezon:

PR #2244, opened 2023-08-17
Change: Streamline key mappings for all detonation related buttons, except suicide

xezon:
* Relates to #2095

This change streamlines key mappings for all detonation related buttons, except suicide. Previously mappings were all over the place. Now they are all simply bound to "D", which makes them easier to use and learn.

### Note

Changes are moved to Optional bundle with #2247

## Patched (Core)

Det. = Detonate

| Language   | Disguise | Det. Charges | Det. Nuke | Det. Demo Trap | Det. Bomb Truck | Det. Fake Building |
|------------|----------|--------------|-----------|----------------|-----------------|--------------------|
| US         | D        | D            | N         | D              | N               | N                  |
| DE         | T        | D            | O         | D              | D               | D                  |
| FR         | D        | P            | N         | F              | F               | D                  |
| ES         | D        | D            | N         | D              | N               | N                  |
| IT         | M        | D            | N         | D              | N               | N                  |
| KO         | C        | D            | N         | D              | N               | N                  |
| ZH         | C        | D            | N         | D              | N               | N                  |
| BP         | C        | D            | N         | D              | I               | A                  |
| PL         | C        | D            | Y         | J              | N               | N                  |

## Patched (Optional)

Det. = Detonate

| Language   | Disguise | Det. Charges | Det. Nuke | Det. Demo Trap | Det. Bomb Truck | Det. Fake Building |
|------------|----------|--------------|-----------|----------------|-----------------|--------------------|
| US         | C        | D            | D         | D              | D               | D                  |
| DE         | T        | D            | D         | D              | D               | D                  |
| FR         | C        | D            | D         | D              | D               | D                  |
| ES         | C        | D            | D         | D              | D               | D                  |
| IT         | M        | D            | D         | D              | D               | D                  |
| KO         | C        | D            | D         | D              | D               | D                  |
| ZH         | C        | D            | D         | D              | D               | D                  |
| BP         | C        | D            | D         | D              | D               | D                  |
| PL         | C        | D            | D         | D              | D               | D                  |

PR #2247, opened 2023-08-19
Change: Streamline key mappings for all detonation related buttons, except suicide (2)

xezon:
* Relates to #2095
* Follow up for #2244

This change demotes all key mapping changes of #2244 to Optional bundle pack.

Reason being, the new key mappings could mess with legacy players. For example if a player with English localization would press D to use the disguise ability of the Bomb Truck, then it would suddenly detonate and potentially cause upset.


xezon:
Documentation updated. Good.

PR #2251, opened 2023-08-20
Change: Increase bonus firepower of China Frenzy to 30% flat, increase max duration from 30 to 40 s, decrease recharge from 240 to 180 s

### LINKED ISSUE #709: China Frenzy Generals Power is very rarely used
China Frenzy Generals Power is very rarely used. Players typically prefer to pick the full Artillery Barrage. Check if it can be better and more attractive.

## Stats
```
WeaponBonus = FRENZY_ONE    DAMAGE 110%
WeaponBonus = FRENZY_TWO    DAMAGE 120%
WeaponBonus = FRENZY_THREE  DAMAGE 130%

SpecialPower SuperweaponFrenzy
  ReloadTime = 240000

Object Frenzy_InvisibleMarker_Level1
  BonusDuration = 10000
  BonusRange = 200

Object Frenzy_InvisibleMarker_Level2
  BonusDuration = 20000
  BonusRange = 200

Object Frenzy_InvisibleMarker_Level3
  BonusDuration = 30000
  BonusRange = 200
```

## Proposal 1
Decrease reload time.

## Proposal 2
Increase bonus duration.

## Proposal 3
Increase bonus range.

## Proposal 4
Increase bonus damage.

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/issues/709

This change buffs the Frenzy ability at all levels.

### The bonus is now always +30% firepower. 

- 10% and even 20% are not meaningful enough bonuses, especially for a temporary ability.
- It is now immediately obvious what bonus applies: Glows red = +30% firepower.
- Frenzy is now useful right from the start. Upgraded levels only increase bonus duration.

### Duration of the bonus doubles with every level of Frenzy

- To make level 3 Frenzy somewhat appealing, the duration is now 40 seconds.
- The progression is now 10, 20 and 40 seconds. This follows to rule of many abilities to double in effectiveness on each level (Rebel Ambush 3 spawns 4 times as many Rebels as Ambush 1; Paradrop 3 spawns 4 times as many Rangers as Paradrop 1; Cash Hack: $1000, $2000, $4000, Cash Bounty: 5%, 10%, 20% etc.)

### The recharge time is now 3 minutes

- Unlike Artillery Barrage or Carpet Bombing, which are useful immediately after they recharged, Frenzy is situational (only useful if combat happens), so this inherent disadvantage is priced in into the recharge time now.

---

The Frenzy scan bug was previously removed: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/593

---

Things left to consider: 

- Infantry General is the only sub-faction that starts out with Rank 1 Frenzy 1. Infantry General is also considered to be the strongest China sub-faction over all. It may be necessary to move Frenzy 1 to rank 3 for all factions if the buffed Frenzy 1 is deemed too effective to be available from the very beginning.
- This can be done in a follow-up change

---

Side effects:

- This change fixes a bug where different levels of Frenzy can be stacked on top of each other in team games for increased bonuses.

---

## Original

| Stat               | Frenzy 1 | Frenzy 2 | Frenzy 3 |
|--------------------|----------|----------|----------|
| Damage Bonus       | +10%     | +20%     | +30%     |
| Bonus Duration (s) | 10       | 20       | 30       |
| Recharge Time (s)  | 240      | 240      | 240      |

## Patched

| Stat               | Frenzy 1 | Frenzy 2 | Frenzy 3 |
|--------------------|----------|----------|----------|
| Damage Bonus       | +30%     | +30%     | +30%     |
| Bonus Duration (s) | 10       | 20       | 40       |
| Recharge Time (s)  | 180      | 180      | 180      |
| Overall Buff %     | +275%    | +87%    | +67%     |

# Rationale

The original Frenzy Power is rarely used. China Players typically prefer the Artillery Powers (3), Carpet Bomber (1), Mines (1) and EMP Pulse Bomb (1). Infantry General uses Frenzy (1) sometimes to scan at the start of the match, even though the Frenzy Power is not designed or advertised as a scan ability.

The new patched Frenzy Power makes it considerably more attractive to use due its favorable progression with higher bonus damages in the lower Frenzy levels, its longer bonus duration at the last level and the overall reduced recharge time.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
@@ -14,7 +14,7 @@ labels:
   - v1.0
 
 links:
-  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/xxxx
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2251
 
 authors:
   - commy2
--- a/Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
@@ -1,16 +1,19 @@
 ---
 date: 2023-08-20
 
-title: Buffs Frenzy General's Power ability
+title: Increases effectiveness of China Frenzy ability
 
 changes:
-  - tweak: All levels of Frenzy now grant +30% fire power.
-  - tweak: Effect duration increased at Frenzy level 3: Progression is now 10, 20 and 40 seconds.
+  - tweak: All levels of China Frenzy now grant +30% fire power.
+  - tweak: Effect duration increased at Frenzy level 3. Progression is now 10, 20 and 40 seconds.
+  - tweak: Recharge time reduced to 3 minutes.
 
 labels:
   - buff
   - china
-  - minor
+  - controversial
+  - design
+  - major
   - v1.0
 
 links:
--- a/Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
@@ -270,7 +270,7 @@ End
 ;-----------------------------------------------------------------------------
 SpecialPower SuperweaponFrenzy
   Enum                = SPECIAL_FRENZY
-  ReloadTime          = 240000   ; in milliseconds. min is 2x door/open close time!
+  ReloadTime          = 180000   ; in milliseconds. min is 2x door/open close time!
   RequiredScience     = SCIENCE_Frenzy1
 ;  InitiateSound       = FireArtilleryCannonSound
   PublicTimer         = No
@@ -284,7 +284,7 @@ End
 ;-----------------------------------------------------------------------------
 SpecialPower Early_SuperweaponFrenzy
   Enum                = EARLY_SPECIAL_FRENZY
-  ReloadTime          = 240000   ; in milliseconds. min is 2x door/open close time!
+  ReloadTime          = 180000   ; in milliseconds. min is 2x door/open close time!
   RequiredScience     = Early_SCIENCE_Frenzy1
   InitiateAtLocationSound = FrenzyActivate
   PublicTimer         = No
@@ -297,7 +297,7 @@ End
 ;-----------------------------------------------------------------------------
 SpecialPower SuperweaponCashHack
   Enum              = SPECIAL_CASH_HACK
-  ReloadTime        = 240000   ; in milliseconds
+  ReloadTime        = 180000   ; in milliseconds
   RequiredScience   = SCIENCE_CashHack1
   PublicTimer       = No
   SharedSyncedTimer   = Yes
--- a/Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
@@ -266,7 +266,7 @@ SpecialPower SuperweaponArtilleryBarrage
 End
 
 ; Patch104p @bugfix commy2 23/10/2021 Fix Frenzy special power revealed map.
-
+; Patch104p @balance commy2 20/08/2023 Decrease ReloadTime from 240000. (#2251)
 ;-----------------------------------------------------------------------------
 SpecialPower SuperweaponFrenzy
   Enum                = SPECIAL_FRENZY
--- a/Patch104pZH/Design/Changes/v1.0/xxxx_frenzy_buff.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/xxxx_frenzy_buff.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-20
+
+title: Buffs Frenzy General's Power ability
+
+changes:
+  - tweak: All levels of Frenzy now grant +30% fire power.
+  - tweak: Effect duration increased at Frenzy level 3: Progression is now 10, 20 and 40 seconds.
+
+labels:
+  - buff
+  - china
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/xxxx
+
+authors:
+  - commy2
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/System.ini
@@ -556,13 +556,14 @@ Object Frenzy_InvisibleMarker_Level1
     InitialHealth = 1
   End
 
+  ; Patch104p @balance commy2 20/08/2023 Change bonus to +30% damage for all Frenzy levels. (#xxxx)
   Behavior = WeaponBonusUpdate ModuleTag_02
     RequiredAffectKindOf = CAN_ATTACK ; Must be set
     ForbiddenAffectKindOf = STRUCTURE ; Must be clear
     BonusDuration = 10000             ; How long effect lasts
     BonusDelay = 100000               ; How often to pulse (short lifetime will trump, of course)
     BonusRange = 200                  ; Keep in line with radius cursor size in special power template
-    BonusConditionType = FRENZY_ONE   ; And which bonus to give
+    BonusConditionType = FRENZY_THREE ; And which bonus to give
   End
 
   Behavior = DeletionUpdate ModuleTag_03 ; Not LifetimeUpdate, since I can't die.  This will DestroyObject me.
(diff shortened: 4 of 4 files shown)


Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2023-08-20
+
+title: Buffs Frenzy General's Power ability
+
+changes:
+  - tweak: All levels of Frenzy now grant +30% fire power.
+  - tweak: Effect duration increased at Frenzy level 3: Progression is now 10, 20 and 40 seconds.
```
xezon: Excess ":" symbol will lead to problems during yaml parse.

Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2023-08-20
+
+title: Buffs Frenzy General's Power ability
+
+changes:
+  - tweak: All levels of Frenzy now grant +30% fire power.
```
xezon: Does not document decreased recharge time.

Always add faction name before faction unit/structure/upgrade/ability to make it easy to recognize for noobs.

Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2023-08-20
+
+title: Buffs Frenzy General's Power ability
```
xezon: I suggest to word it "Increases effectiveness of China Frenzy ability"

Patch104pZH/Design/Changes/v1.0/2251_frenzy_buff.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+changes:
+  - tweak: All levels of Frenzy now grant +30% fire power.
+  - tweak: Effect duration increased at Frenzy level 3: Progression is now 10, 20 and 40 seconds.
+
+labels:
+  - buff
+  - china
+  - minor
```
xezon: Needs controversial, design, major labels.

Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
```diff
@@ -270,7 +270,7 @@ End
 ;-----------------------------------------------------------------------------
 SpecialPower SuperweaponFrenzy
   Enum                = SPECIAL_FRENZY
-  ReloadTime          = 240000   ; in milliseconds. min is 2x door/open close time!
+  ReloadTime          = 180000   ; in milliseconds. min is 2x door/open close time!
```
xezon, line 273: Needs documentation in code.

xezon (commented):
Changes look good to me. Reasonable improvements for a currently unused ability. This is a controversial change, so this pull request requires a rationale text to explain why this is done.

xezon:
I added tables and rationale text to the opening post.

xezon (approved):

commy2:
For anyone that's bad with numbers:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/5f2d4ad4-0d8a-4a84-9a7c-91db2d2284a9

With this buffed version of Frenzy, a Frenzy 1 Minigunner (vet 1) is now equally as strong as an Elite Training Minigunner (vet2).

kABUSE1:
> Different levels of Frenzy can be stacked on top of each other in team games for increased bonuses.

Did I read that right?
In 1.04 frenzy would stack for team members but in 1.04p you desire to remove this behaviour?

Do you by chance know how the math works when stacking?

xezon:
He wrote only different levels would stack. So it is inconsistently stacking, which is not good. It should either stack uniformly or not at all.

commy2:
>  In 1.04 frenzy would stack for team members but in 1.04p you desire to remove this behaviour?

Frenzy only stacks if different levels are used at once on the same units. This is only possible in team games, where one player picks Frenzy X, the other Frenzy Y, where X != Y.

> Do you by chance know how the math works when stacking?

Multiplicative as they are completely different bonuses. 

Frenzy 1 + Frenzy 2 = +32% damage
Frenzy 1 + Frenzy 3 = +43% damage
Frenzy 2 + Frenzy 3 = +56% damage
Frenzy 1 + Frenzy 2 + Frenzy 3 = +72% damage (3 players)

It's very illogical, but if you want to create the highest dps unit in the game, you need to stack all three levels of Frenzy lol

MTKing4:
I like the change, though i do feel that 10 and 20 seconds duration is too little that can be buffed higher, or just reduce the timer to 2 minutes is better, since as you mentioned, it's too situational, so there will be a lot of idle time where it's ready but not used, I'd really like that to be 2 minutes

xezon:
2 minutes might be a bit too little. With Level 3 Frenzy this would mean that it could be used for max 33% of play time. Originally max 12.5%. And in current proposal max 22.2%.

commy2:
Big duration disincentivises upgrading it to level 2 or 3. From player acceptance perspective, it's probably better to underbuff something initially than to overbuff it. It can always be tweaked further after people actually play with it.

MTKing4:
Yeah forget about the duration, i think reducing cooldown is the better approach, and actually incentivises level 2 and 3, maybe 2:30 if not 2 minutes

But yeah i get it that we shouldn't overbuff, I'm interested into what others think though

xezon:
Current total buffs in percentages:

Level 1 Frenzy +275%
Level 2 Frenzy +87%
Level 3 Frenzy +67%

So overall Frenzy is now +159% better than original.

ExiLe-ZH:
I approve big buffs to make it a good alternative to meta genpowers.

Although I have to say level 2 or 3 aren't that attractive, since level 1 is already so good now. 

Remember that currently level 1 is already picked sometimes (mostly for the scan ability ironically), however even with the big % increase level 2 and 3 were never picked and the incentive only becomes smaller with this change?

xezon:
The Level 2 Frenzy is +100% better than Level 1 Frenzy.
The Level 3 Frenzy is +300% better than Level 1 Frenzy.

If Level 2 and 3 Frenzy were considered to be of no use, then this would imply that the increased duration makes no practical difference. Is that true?

ExiLe-ZH:
Originally higher levels increased both damage and duration, but this wasn't enough for players to sacrifice meta powers. Now it only increases duration, that doesn't improve the situation (for level 2 and 3).

xezon:
Well, on paper the Level 3 Frenzy still much better than Original. How high would this need to go and how can you determine this?

ExiLe-ZH:
Honestly it's tough to determine this evidence based. I know for myself I would definitely pick level 1 more often with this change, but the extra duration alone wouldn't make the cut for me to spent more genpoints. 2-3 genpoints is a **_huge_** investment/sacrifice.

xezon:
Another way to scale this would be with fixed time and increasing damage boost.

xezon:
I fixed the improvement percentages in opening post. Here is model with same improvement percentages, but instead of longer times they give bigger damages.

| Stat               | Frenzy 1 | Frenzy 2 | Frenzy 3 |
|--------------------|----------|----------|----------|
| Damage Bonus       | +15%     | +30%     | +60%     |
| Bonus Duration (s) | 20       | 20       | 20       |
| Recharge Time (s)  | 180      | 180      | 180      |
| Overall Buff %     | +275%    | +87%     | +67%     |

commy2:
I feel like making people chose Frenzy 1 and using it for the intended purpose already is a victory. 

Truth though is the way Frenzy behaves, it is not really suited to be an upgradable power. Making Frenzy 1 any worse than OP suggests (< 30% boost) means it's not worth it, which in turn means it will never be chosen except to reach Frenzy 2 or 3. 

If you make 2 points of Frenzy necessary to have any tangible benefit (30%), then that comes at the cost of people skipping Cluster Mines, Cash Hack, Emergency Repair or maybe EMP instead. I'd rather they chose 1 level of Frenzy and the other powers, skipping maybe one point in Artillery Barrage, than have them sink it all into some overpowered form of Frenzy 2 or 3.

xezon:
The disadvantage of the increasing damages model would be that other players would not recognize the Frenzy level and therefore would not be able to anticipate the damage level early. This is already a problem in Original. The way to mitigate that would be to add different shades of red for each level, but it appears that Frenzy color is hard coded. I could not find it in INI.

Another thing we can do to Buff Frenzy is raising these in Gattling Attachments:

```
  ; Patch104p @bugfix hanfield 01/09/2021 Added WeaponBonus parameters for Frenzy that nullify the bonus but allow it to color the Gattling Cannon.
  WeaponBonus           = FRENZY_ONE DAMAGE 90%
  WeaponBonus           = FRENZY_TWO DAMAGE 80%
  WeaponBonus           = FRENZY_THREE DAMAGE 70%
```

On that note, it needs to be adapted for this change anyway. All to 70%.

commy2:
> On that note, it needs to be adapted for this change anyway. All to 70%.

The implementation in the OP does not use FRENZY_ONE or FRENZY_TWO bonuses. Instead, FRENZY_THREE is granted on all levels, which, while confusing, at least leaves two hard-coded and limited buffs for other use, be it in this project in some other feature / fix (speculative), or a custom map or mod based on the patch.

The red glow of Frenzy is indeed hard-coded as far as I can tell.

xezon:
Ah right. Then we can just think about changing them to 100%.

commy2:
Or removing them if you want Gattling towers on Overlords to benefit from Frenzy.
Why doesn't it work on base defenses anyway? I know it's rather small, and it will never work on Bunkers, but it could at least buff Gatlings to make it ever so slightly more versatile.

MTKing4:
I think we need to have much bigger upgrades from level 1 to 2 and from 2 to 3 for people to consider them over other genpowers, i think it would be beneficial to give the player more options and they decide how to play it

PR #2253, opened 2023-08-20
Fix: Cash Bounty does not mention Command Center requirement in General's promotion menu

commy2:
shot_20230820_110959_1

This line disappears if the player owns a Command Center.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2253_cash_bounty_info.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2253_cash_bounty_info.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-08-20
+
+title: Add dynamic "Requires Command Center" line to GLA Cash Bounty tooltip
+
+changes:
+  - fix: The tooltip for GLA Cash Bounty now mentions the Command Center requirement if no Command Center is owned by the player.
+
+labels:
+  - boss
+  - enhancement
+  - gla
+  - minor
+  - text
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2253
+
+authors:
+  - commy2
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -4138,9 +4138,11 @@ CommandButton Command_PurchaseScienceRebelAmbush3
 End
 
 
+; Patch104p @balance commy2 20/08/2023 Add dynamic "Requires Command Center" line to tooltip. (#2253)
 CommandButton Command_PurchaseScienceCashBounty1
   Command           = PURCHASE_SCIENCE
   Science           = SCIENCE_CashBounty1
+  Object            = GLAHoleCommandCenter ; something with Command Center as prerequisite
   ButtonImage       = SSCashBounty
   ButtonBorderType  = UPGRADE ; Identifier for the User as to what kind of button this is
 End
@@ -4148,6 +4150,7 @@ End
 CommandButton Command_PurchaseScienceCashBounty2
   Command           = PURCHASE_SCIENCE
   Science           = SCIENCE_CashBounty2
+  Object            = GLAHoleCommandCenter
   ButtonImage       = SSCashBounty2
   ButtonBorderType  = UPGRADE ; Identifier for the User as to what kind of button this is
 End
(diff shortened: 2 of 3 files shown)


xezon:
What happens when player has Cash Bounty 1, builds Command Center, sells Command Center, then looks at Cash Bounty 2. Will the given information be correct? Will player need another Command Center after selecting Cash Bounty 2?

commy2:
The text will show that you require a Command Center.

You do in fact require a Command Center to upgrade Bounty from 1 to 2 or 2 to 3:

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/0f2924a3-d49c-429f-98a1-5d395a4a4cb2

xezon (approved):
Nice.

kABUSE1:
Are we sure that we want this behaviour at all?

It would be more logical if you only receive cash bounty if you own a command center when scoring the kill.

xezon:
It is a legitimate question. The current design is odd. Ideally it would either be fully passive always or always require the Command Center.

Is there any other Generals Power that acts like Cash Bounty does?

commy2:
> Is there any other Generals Power that acts like Cash Bounty does?

I can't think of any. I don't think there is.

xezon:
How about we just remove this requirement altogether? Just make it a passive ability always? The advantage of that would be that it would not require learning this odd requirement. And it is also close to the Original where it just requires dropping a 0% scaffold briefly.

commy2:
I suppose CashBounty module could be added to Default/Object.ini:DefaultThingTemplate. That way, any object can handle the cash bounty logic.

Pretty much every other General's Power is used by the Command Center though. Exceptions are unit unlocks and unit veterancy, and the AFG Carpet Bomber. The ideal would be if the power just turned off if no Command Center is left as kABUSE1 suggested. But even in that case, this "Requires: Command Center" line makes sense.

xezon:
It would be a quite big design change if the Cash Bounty was no longer a purely passive reward. Design wise it would be much better if it was bound to DefaultThingTemplate or Command Center. The current implementation is a bit confusing.

kABUSE1:
Agree, quite a big change.

Scaffold cc makes no sense. I only see two valid options:
1. Make it fully passive
2. Make it require a fully constructed, alive CC at all times (no hole)

I think option 2 is okay, as cash bounty is currently very, very powerful. Might as-well delay it a bit.

xezon:
We already changed it in Patch to this #595. It means full Command Center is required, but Cash Bounty level will remain active even if Command Center is demolished. The extra confusing part then starts when selecting a new Cash Bounty level, because then a new Command Center is required.

commy2:
Here's a patch where Command Center is no longer requried at all: https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2263

MTKing4:
I don't think this is a good idea, we're buffing GLAs unnecessarily

I'd much rather having Cash Bounty only work when there's a CC up, and not work otherwise 

Also we're adding a huge favoritism over an already highly favored genpower, people will now pick cash bounty over the other options even more than before because it's the only genpower that doesn't require a CC and can get its value instantly, in a meta where GLA builds CCs very late and or often times forgotten, with a change like this, CC will likely not be built until level 5

xezon:
> I don't think this is a good idea, we're buffing GLAs unnecessarily

It is not a buff. Not even controversial. It just removes the useless mechanic of placing a temporary scaffold to activate it.

> I'd much rather having Cash Bounty only work when there's a CC up, and not work otherwise

This would be a definite nerf and controversial.

MTKing4:
Maybe not a direct buff, but buff nonetheless because it increases the chances of a player getting cash for kills, for example when you reach level 3 but don't have extra cash for CC, you can benefit from cash bounty regardless, and much earlier than before, in 1.04 that was not possible

> This would be a definite nerf and controversial.

True, We could make it at least require 1 CC built after getting the promotion then it's unlocked for good even after upgrades, I'm fine with that as well

xezon:
> Maybe not a direct buff, but buff nonetheless because it increases the chances of a player getting cash for kills, for example when you reach level 3 but don't have extra cash for CC, you can benefit from cash bounty regardless, and much earlier than before, in 1.04 that was not possible

Ok this makes sense. Then buff and controversial labels need to be added.

xezon:
Cash Bounty is a really good upgrade, because it incentives the player to pick fights, instead of just sit back and do nothing. So in a way we would like it to remain an attractive option, as it should make the game better for it.

PR #2254, opened 2023-08-20
Fix: Add tool tip variant to GLA Demo Scud Storm for all latin languages

xezon:
This change adds better tool tip strings to GLA Demo Scud Storm for all latin languages. It now indicates higher damage output by stating it shoots high explosive warheads.


xezon:

PR #2255, opened 2023-08-20
Tweak: Partisan and Milita Man Guard radius

commy2:
This change reduces the Guard radius from Rebel level to Red Guard and Ranger level for the Partisans and the Militia Man, because I think that matches the units better than what I initially chose.


xezon (approved):

commy2:
shot_20230820_125827_2

The Guard radius is now only slightly larger than the weapon range, instead of much larger.

We should probably look into Rebel Guard range btw.

PR #2257, opened 2023-08-20
Fix: Set correct names for GLA Booby Trap and GLA Demo Trap in Brazilian language

xezon:
This change corrects the names for GLA Booby Trap and GLA Demo Trap in Brazilian language.


xezon:
Improved. Documentation added. Looks good.

PR #2258, opened 2023-08-20
Fix: Add object name variant to GLA Toxin Demo Trap for latin languages

xezon:
This change adds a text variant for GLA Toxin Demo Trap for latin languages.


xezon:
We could make the Toxin Demo Trap look slightly different by INI hiding some of the red demo charges from the barrel. This would also be consistent with its lower damage output.

Edit: It is one Mesh. Not possible without making new model.

xezon:

PR #2261, opened 2023-08-20
Fix: Resolve errors in tool tip strings of China Helix for all latin languages

xezon:
This change improves the wording in tool tip strings of China Helix for all latin languages. The China Infantry Helix tool tip text no longer claims to be upgrade-able with all base defense structures. And the Polish tool tip string no longer claims that the Bunker passengers can shoot at each other.


xezon:

PR #2263, opened 2023-08-20
Change: Removes Command Center scaffold requirement from GLA Cash Bounty activation

commy2:
* Replacement for #595

This change removes the Command Center scaffold requirement for GLA Cash Bounty. It makes the interaction with Cash Bounty more intuitive.

> Tested, works. Now controlling any object at all can manage Cash Bounty.


xezon:
I am ok with this change. It makes the feature simpler user facing. And legacy players dropping the Command Center scaffolds would see no change, because the scaffold would just change nothing.

If we ever think Cash Bounty needs a nerf, then we could think about making it a Command Center requirement.

We need to properly document this before merging.

commy2:
Done.

Patch104pZH/GameFilesEdited/Data/INI/Default/Object.ini
```diff
@@ -118,7 +118,21 @@ Object DefaultThingTemplate
     End
   End
 
-
+  ; Patch104p @tweak commy2 20/08/2023 Remove Command Center scaffold requirement from Cash Bounty.
```
xezon: `@fix` or `@bugfix` I would say

Patch104pZH/Design/Changes/v1.0/2263_cash_bounty_activation.yaml
```diff
@@ -0,0 +1,23 @@
+---
+date: 2023-08-21
+
+title: Fixes GLA Cash Bounty activation
+
+changes:
+  - fix: GLA Cash Bounty no longer requires a GLA Command Center scaffold to be placed down to active or upgrade
```
xezon:
```suggestion
  - fix: The GLA Cash Bounty no longer requires a GLA Command Center scaffold or structure to activate or upgrade. It works entirely passive now.
```

Patch104pZH/Design/Changes/v1.0/2263_cash_bounty_activation.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - boss
+  - bug
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/595
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1764
```
xezon: I do not think we need the old links in the change log. The old ones are effectively irrelevant in regards to this change.

Patch104pZH/Design/Changes/v1.0/2263_cash_bounty_activation.yaml
```diff
@@ -0,0 +1,23 @@
+---
+date: 2023-08-21
+
+title: Fixes GLA Cash Bounty activation
```
xezon:
```suggestion
title: Removes Command Center scaffold requirement from GLA Cash Bounty activation
```

xezon (commented):
Also delete `1764_cash_bounty_dummy_spawn_interval.txt` file.

Patch104pZH/GameFilesEdited/Data/INI/Default/Object.ini
commy2:
```suggestion
  ; Patch104p @fix commy2 20/08/2023 Remove Command Center scaffold requirement from Cash Bounty.
```

Patch104pZH/Design/Changes/v1.0/2263_cash_bounty_activation.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - boss
+  - bug
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/595
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1764
```
commy2:
```suggestion
```

xezon (approved):

MTKing4:
I prefer the CC requirement approach, this one buffs GLAs unnecessarily

xezon:
Buff and Controversial labels need to be added to yaml file. (Done Pending)

ExiLe-ZH:
> I prefer the CC requirement approach

Same, GLA (depending which one) hasn't got many incentives to build a CC earlier on. Seems only fair needing a CC for such good genpower.

MTKing4:
2 people with this approach and 2 against, shouldn't merge this because it's undecided

commy2:
For the record, I have no preference. I will note though that making the 100% CC constructed a requirement is the bigger departure from 1.04.

xezon:
Definitely. The current fix is good. It just removes the stupid scaffold prerequisite. In most cases it makes no tangible difference, because player will drop scaffold and cancel is straight away, which incurs no financial loss.

commy2:
I made a pull that adds a 100% Command Center requirement to upgrade Cash Bounty based on current main branch.

https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2315

Discussion seems to be all over the place already. Don't start debating there too. We need some place to collect all ideas on this perhaps.

PR #2266, opened 2023-08-21
Fix: Resolve Suitcase Nuke name inconsistencies in Italian, Polish strings

xezon:
This change resolve Suitcase Nuke name inconsistencies in Italian, Polish strings.


xezon:
Requires no special documentation I think.

PR #2269, opened 2023-08-22
Fix: Add object name variant for GLA Demo Jarmen Kell

xezon:
This change adds an object name variant for the GLA Demo Jarmen Kell.


xezon:

PR #2270, opened 2023-08-22
Fix: Add tool tip variant to GLA Demo Jarmen Kell

xezon:
This change adds better tool tip strings to GLA Demo Jarmen Kell for all languages. The sentences were copied from Burton and they highlight the bomb abilities.


xezon:

PR #2271, opened 2023-08-22
Fix: Resolve errors in tool tip strings of USA Colonel Burton for all languages

xezon:
This change resolves errors in tool tip strings of USA Colonel Burton for all languages.

It no longer contains the "Build Limit: 1" sentence. It is not necessary. The other heroes do not have it too.

The Korean string now lists the strengths.

## Original

shot_20230822_174625_1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2153_korean_tool_tip_text.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2153_korean_tool_tip_text.yaml
@@ -5,6 +5,7 @@ title: Fixes errors in Korean tool tip strings
 
 changes:
   - fix: The wording in Korean tool tip strings is more consistent now.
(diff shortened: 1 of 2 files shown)


commy2:
I kind of like the Build Limit: 1 line and would like to see it on all other objects that have this limit: Heros, Strat and Internet Centers.

It is not made clear anywhere else that only 1 of these can be built at once.

It should be placed right below the Cost though. First line in str.

xezon:
There is an auto string "Unit already exists" after the limit is reached, but not before. If we want to have another limit string then I think this should be code driven and not hardcoded in STR, because the limits can be changed in INI.

commy2:
Yes, automatically adding these would be ideal. Same as build times.

xezon:

PR #2272, opened 2023-08-22
Tweak: Paradrop Rangers and Missile Defenders in batches

commy2:
* Follow up for #2026
* Follow up for #2056

With this change, the Rangers and Missile Defenders are created in batches of 5, because all Rangers, then all MDs looks strange to me.


xezon:
What about this one? #2057

commy2:
Alzheimers. I don't remember that.

Although spawning 1-1-1-1-1-1 is equally strange looking. Can we agree on 2-2-1 -- 2-2-1 or something instead?

xezon:
The disadvantage of opening 10000 Pulls :D

Yeah not sure what pattern is best here.

commy2:
(3-2) times n is pretty straight forward. Paradrop 1 is just first unit type, then second unit type; Paradrop 2 and 3 are just multiples of that.

xezon:
Do you prefer 3-2-3-2 over 1-1-1-1-1-1-1-1-1 ?

commy2:
Yes, definitely. 1-1-1-1-1 is just as bad as all-rest

xezon (approved):
Ok.

xezon:
I like more mix better.

PR #2279, opened 2023-08-23
Tweak: Make Supply Drop Zone Cargo Planes attackable again

commy2:
- continuation of https://github.com/TheSuperHackers/GeneralsGamePatch/pull/836

NO_ATTACK_FROM_AI + UNSELECTABLE combined with SELECTABLE + NO_SELECT (don't question it) gives the, in my opinion, desired behavior.

The planes are not automatically attacked. They are not attacked by guard mode either, including air guard mode. This hasn't changed. 

However, with this change, the planes can now be attacked by left clicking on them. This can be used to deny the drop (if you manage to sneak behind the enemy base), or to earn a small amount of experience from the kill, or to annoy the USA player with "unit under attack" EVA message.

PR #2277, opened 2023-08-23
Fix: USA Spy Drone can no longer be selected, but attacked, attempt 2

commy2:
* Fixes #115
* Relates to #1338

This change makes the USA Spy Drone no longer selectable, but still attackable. It behaves like vehicle drones.

PR #2280, opened 2023-08-23
Tweak: Make Command Centers grant the global Dummy Upgrade

commy2:
This avoids issues with further use of the DummyUpgrade where it may not apply if the structure was captured instead of constructed, or given to the player by a map script. It also enables the use with units that are not constructed, but spawned (and thus are unable to grant upgrades) such as the Spy Drone or the Cargo Plane.

The Command Center guarantees the dummy upgrade from the start of the map
- in skirmish
- in multiplayer
- in generals challenge

The dummy upgrade may be missing:
- in the campaign
- on custom maps without spots

I have not found a campaign mission where this may be an issue.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/AirforceGeneral.ini
@@ -8911,6 +8911,11 @@ Object AirF_AmericaCommandCenter
     TriggeredBy = Upgrade_AmericaMOAB
   End
 
+  ; Patch104p @fix commy2 27/08/2023 We always need this. Used to trigger various upgrade modules like ArmorUpgrade or StatusBitsUpgrade. (#2280)
+  Behavior = GrantUpgradeCreate ModuleTag_GrantDummyUpgrade
+    UpgradeToGrant = Upgrade_DummyUpgrade
+  End
+
   Geometry            = BOX
   GeometryMajorRadius = 60.0
   GeometryMinorRadius = 70.0
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -1405,6 +1405,11 @@ Object Boss_CommandCenter
     TriggeredBy = Upgrade_ChinaEMPMines
   End
 
+  ; Patch104p @fix commy2 27/08/2023 We always need this. Used to trigger various upgrade modules like ArmorUpgrade or StatusBitsUpgrade. (#2280)
+  Behavior = GrantUpgradeCreate ModuleTag_GrantDummyUpgrade
+    UpgradeToGrant = Upgrade_DummyUpgrade
+  End
+
   Geometry            = BOX
   FactoryExitWidth    = 25
   GeometryMajorRadius = 60.0
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChemicalGeneral.ini
@@ -1114,6 +1114,11 @@ Object Chem_GLACommandCenter
     ReallyDamagedParticleSystem3 = Bone:None RandomBone:No PSys:StructureTransitionMediumShockwave
   End
 
+  ; Patch104p @fix commy2 27/08/2023 We always need this. Used to trigger various upgrade modules like ArmorUpgrade or StatusBitsUpgrade. (#2280)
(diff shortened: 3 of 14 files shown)


xezon:
Need check in Thyme.

xezon:
I do wonder, can we perhaps move the `CashBountyPower` modules to a persistent dummy object that is spawned through the StartingThings object for GLA factions?

xezon:
Here is a list of all objects that will use DefaultThingTemplate in Original game.

[templates.txt](https://github.com/TheSuperHackers/GeneralsGamePatch/files/12432801/templates.txt)

xezon:
I suggest dropping this change and adding `GrantUpgradeCreate` module to objects addressed in #2277, #2279. This will reduce cost.

commy2:
Cash Bounty needs a second look anyway:

shot_20230826_112056_1

Seems to be happening when capturing preplaced empty vehicles.

commy2:
- Cash Bounty module re-added to Command Center. Handles almost all maps correctly, except GLA05, where the player is able to unlock Bounty without owning a Command Center

- [ ] GLA05

I am still unsure what to do about DummyUpgrade. Without granting it by everything, it is not guaranteed in the campaign. Can't we just leave it in DefaultThingTemplate until we get a Thyme solution? I don't see how it breaks performance. Unlike Cash Bounty, it doesn't get called after map start at all except when a unit leaves the factory or a building completes to 100%.

xezon:
DummyUpgrade we can grant in all objects that need it. Like it was before.

commy2:
How do you guarantee that the player does not capture a Supply Drop Zone with a Rebel they obtained from a player that surrendered?

Altohugh I suppose that can only happen in MP.

xezon:
Is the upgrade effect revoked after capture? If yes, can you give example?

commy2:
> Is the upgrade effect revoked after capture?

GrantUpgradeCreate is not called when you capture a building or obtain it from a team mate that surrendered. There is no way to guarentee that the owner of a SDZ will have the upgrade.

Best we can do is: 

- Give DummyUpgrade to StartingThings. That covers all MP maps with spots and skirmishes, as well as the player in the GC
- Give DummyUpgrade to all CCs. That covers the AI enemy in the GC, and _probably_ all or at least most campaign missions.

There could still be a custom mission made that breaks it.

Just make a spotless map where the player starts with a random building, Capture Buildings upgrade and a Rebel. Enemy has an SDZ. You can capture the SDZ without ever getting the DummyUpgrade.

commy2:
I think DummyUpgrade could also be given to all capture units. Red Guard, Rebel, Ranger and Lotus. That would only leave surrendering as possibility.

commy2:
And Combat Bike.

commy2:
Would also break if a custom map gives Rebel Ambush to a non-CC building. aaaa

xezon:
> GrantUpgradeCreate is not called when you capture a building or obtain it from a team mate that surrendered. There is no way to guarentee that the owner of a SDZ will have the upgrade.

I do not understand this problem.

```
  Behavior = GrantUpgradeCreate ModuleTag_MicrowaveTankFix01
    UpgradeToGrant = Upgrade_DummyUpgrade
  End
  Behavior = ArmorUpgrade ModuleTag_MicrowaveTankFix02
    TriggeredBy = Upgrade_DummyUpgrade
  End
```
In this sample, isn't the whole case of the Dummy upgrade to simply update the Armor on creation? So the armor is updated after creation. If someone now captures this structure, it will not revert the Armor, or would it?

commy2:
For Armor yes. We need DummyUpgrade for Supply Drop Zone planes.

xezon:
Just add

```
  Behavior = GrantUpgradeCreate ModuleTag_MicrowaveTankFix01
    UpgradeToGrant = Upgrade_DummyUpgrade
  End
```

to the plane object.

commy2:
That's not called. GUC is only called when 

1.) the object its preplaced on the map
2.) you construct the object from a production building
3.) the object is a building that just finished 100% construction

xezon:
Hmm ok sucks. Perhaps add it to all faction structures? Would that be sufficient?

commy2:
Had a walk. My idea A is DefaultThingTemplate. My idea B is:

add it to:
- All CCs
- Red Guard, Rebel, Ranger, Lotus, Rebel Combat Bikes
- Reactors (guarantees ArmorUpgrade)
- SDZ (guarantees ArmorUpgrade)
and ignore the edge-cases. This covers > 99.9% of all played games I expect. You have to be trying to break it.

Eventually, use a Thyme solution. Or just not require these hacky workarounds.

commy2:
Done. This is required for https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2277 and https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2279

xezon (approved):

PR #2282, opened 2023-08-24
Add: Red Guard Bayonet Attack

commy2:
Fixes the issues with the Bayonet Attack. It will be removed from command set before merge, but afterwards can be added easily in custom maps.


Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
```diff
@@ -2379,6 +2379,19 @@ CommandButton Command_ChinaInfantryRedGuardCaptureBuilding
   DescriptLabel           = CONTROLBAR:ToolTipChinaRedGuardCaptureBuilding
 End
 
+; Patch104p @feature commy2 24/08/2023 Add button for Bayonet Attack. (#xxxx)
```
xezon: xxxx

xezon (dismissed):
Nice.

Patch104pZH/GameFilesEdited/Data/INI/Object/TankGeneral.ini
```diff
@@ -1413,8 +1422,11 @@ Object Tank_ChinaInfantryRedguard
   EditorSorting       = INFANTRY
   TransportSlotCount  = 1                 ;how many "slots" we take in a transport (0 == not transportable)
   WeaponSet
-    Conditions = None
-    Weapon = PRIMARY RedguardMachineGun
+    Conditions          = None
+    Weapon              = PRIMARY     RedguardMachineGun
+    Weapon              = TERTIARY    RedguardBayonet
```
xezon, line 1427: This means the Bayonet weapon could now be used in Network match when client adds the button to command set. Wouldn't we want to comment this out as well to avoid this?

commy2: That would make it harder to enable the button on custom map.

I don't think it is worth it to enable this for cheating. It's not very good due to the small range and Red Guards being kind of meh in general.

In general, button abuse should be fixed in Thyme. E.g. auto Laser Lock exploit is still possible and that is much more powerful than a melee attack.

xezon: Ok.

xezon (approved):

PR #2284, opened 2023-08-24
Fix: Toxin General RPG-Soldier and USA05 Toxin RPG-Soldier missing muzzle flash when firing at ground targets

commy2:
They have the muzzle fire when aiming at airborne targets, but not when aiming at ground targets due to a bug.

Note that there is a exhaust glow associated with RPG missiles that are missing from the Stinger missiles they fire, so it still does not has all the fire effects of the other/regular RPG Soldiers.


xezon (approved):
Needs documentation.

commy2:
Doc ok?

xezon (approved):

PR #2285, opened 2023-08-24
Tweak: Make USA05 GLA infantry no longer weak to rocketeer infantry

commy2:
This mainly makes the USA05 campaign mission enemy infantry no longer crumble against Missile Defenders.

This armor is not used on any multiplayer unit. Is is used by the Biohazard Tech and the USA05 Dr. Thrax infantry. It is also referenced by some Infantry General units, but looked behind an undefined upgrade and thus unreachable.

I am pretty sure this definition is supposed to be there, but was forgotten.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2285_usa05_campaign_enemy_infantry.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2285_usa05_campaign_enemy_infantry.yaml
@@ -1,0 +1,19 @@
+---
+date: 2023-08-25
+
+title: Fixes armor of enemy infantry on the USA05 campaign map.
+
+changes:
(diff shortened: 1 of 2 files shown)


xezon (approved):

Patch104pZH/Design/Changes/v1.0/2285_usa05_campaign_enemy_infantry.yaml
```diff
@@ -0,0 +1,19 @@
+---
+date: 2023-08-25
+
+title: Fixes armor of enemy infantry on the USA05 campaign map.
```
commy2:
```suggestion
title: Fixes armor of USA05 campaign map enemy infantry against rocket infantry
```

PR #2286, opened 2023-08-24
Tweak: Add Anthrax Bomb and Rebel Ambush from USA05 GLA allies to shortcut bar

commy2:
In USA05, you're given GLA allies. You can call Rebel Ambush level 3 and Anthrax Bomb from the GLA Command Center.

This is annoying though, because the powers do not appear on the shortcut bar. You will need to select the Command Center regularly to see if the powers have been recharged and to use them.

shot_20230825_000855_1


xezon (approved):
Needs yaml doc.

PR #2287, opened 2023-08-25
Fix: Burton Knife Attack animation skip exploit

commy2:
PER_ATTACK is a bad choice here, because it will be skipped if Burton already engages the target with his other weapon. This can be abused to insta stab infantry.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/9bd07aff-dc05-43d4-8346-e4f06a249efc

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2287_burton_insta_stab.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2287_burton_insta_stab.yaml
@@ -1,7 +1,7 @@
 ---
 date: 2023-08-25
 
-title: Fix USA Colonel Burton Knife Attack animation skip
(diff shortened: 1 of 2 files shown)


Patch104pZH/Design/Changes/v1.0/2287_burton_insta_stab.yaml
```diff
@@ -0,0 +1,19 @@
+---
+date: 2023-08-25
+
+title: Fix USA Colonel Burton Knife Attack animation skip
```
xezon:
```suggestion
title: Fix USA Colonel Burton instant Knife Attack exploit
```

xezon (approved):
Nice.

PR #2290, opened 2023-08-25
Fix: Injured members slow down the Angry Mob, can never catch up

commy2:
I think it would be better if injured Mob members didn't move slower.
There is no way for GLA to heal them before vet 2, aside from shooting them and waiting for their replacements.
They slow down the entire Mob and may cause Mob members to be killed as they get too far away from the center.

---

Only ways to heal are:
- Propaganda
- Ambulance
- Hospital
- Reaching vet 2

Barracks, Tunnels or other objects that heal passengers do not work, as Mobs can't enter them.
Jank way of healing members: shoot them, wait for replacement.
But it's not obvious which Mobsters are injured, as they have no individual healthbars and have no injured standing/idle animations.

The best way to figure out who's injured is to walk the Mob, and observe who's lagging behind. But this is also a bit unituitive, as Mob members have 3 different speed, depending on whether they are in front, close to, or behind the invisible nexus.

Also: Killing injured members with weapons with splash damage may injure healthy Mobsters.

---

There is a MobNexusContain module with "HealthRegen%PerSec" parameter, but it's not used anywhere and I expect it doesn't work.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
@@ -4,7 +4,7 @@ date: 2023-08-26
 title: Fixes movement speed of injured Angry Mob members
 
 changes:
-  - fix: Injured members no longer slow down the GLA Angry Mob.
+  - fix: Injured members no longer slow down the GLA Angry Mob. This way they no longer lag behind and die off at a distance.
 
 labels:
   - buff
--- a/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
@@ -1,7 +1,7 @@
 ---
 date: 2023-08-26
 
-title: Fixes movement speed of injured Angry Mob members
+title: Increases movement speed of injured GLA Angry Mob members to prevent them from falling behind and dying
 
 changes:
   - fix: Injured members no longer slow down the GLA Angry Mob. This way they no longer lag behind and die off at a distance.
--- a/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
@@ -8,6 +8,8 @@ changes:
 
 labels:
   - buff
+  - bug
+  - controversial
   - gla
   - minor
   - v1.0
--- a/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
@@ -1,0 +1,19 @@
+---
+date: 2023-08-26
+
+title: Fixes movement speed of injured Angry Mob members
+
+changes:
+  - fix: Injured members no longer slow down the GLA Angry Mob.
+
+labels:
+  - buff
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/xxxx
+
+authors:
+  - commy2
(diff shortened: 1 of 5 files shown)


Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+title: Fixes movement speed of injured Angry Mob members
+
+changes:
+  - fix: Injured members no longer slow down the GLA Angry Mob.
+
+labels:
+  - buff
+  - gla
+  - minor
```
xezon: Requires bug and controversial labels.

Patch104pZH/GameFilesEdited/Data/INI/Locomotor.ini
```diff
@@ -2789,11 +2789,12 @@ Locomotor AngryMobNexusLocomotor
   StickToGround = Yes       ; walking guys aren't allowed to catch huge (or even small) air.
 End
 
+; Patch104p @tweak commy2 26/08/2023 Set SpeedDamaged equal to Speed. (#xxxx)
```
xezon:
```suggestion
; Patch104p @fix commy2 26/08/2023 Set SpeedDamaged equal to Speed for all Angry Mob locomotors to prevent damaged members lagging behind and dying off at a distance. (#2290)
```

Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
```diff
@@ -0,0 +1,19 @@
+---
+date: 2023-08-26
+
+title: Fixes movement speed of injured Angry Mob members
```
xezon:
```suggestion
title: Increases movement speed of damaged GLA Angry Mob members to prevent movement glitches
```

Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+changes:
+  - fix: Injured members no longer slow down the GLA Angry Mob.
```
xezon:
```suggestion
  - fix: Injured members no longer slow down the GLA Angry Mob. This way they no longer lag behind and die off at a distance.
```

xezon (approved):
> This is why it starts at 50% health

I was unable to observe this. When produced with GLA, Angry Mob starts at 100% health.

commy2:
> I was unable to observe this. When produced with GLA, Angry Mob starts at 100% health.

Oh, nvm. This only happens with preplaced Mobs, such as StartingUnitN or in the campaign.
shot_20230826_101032_1

Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
commy2: Damaged? I think it's better to say "injured" when talking about infantry.

commy2: > to prevent movement glitches

Hmm, sounds very unspecific. Them dying off when falling behind, is that a movement glitch? More like the expected response to bad movement parameters.

commy2:
```suggestion
title: Increases movement speed of injured GLA Angry Mob members to prevent them from falling behind and dying
```

Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
```diff
@@ -0,0 +1,21 @@
...
+labels:
+  - buff
+  - bug
+  - controversial
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/xxxx
```
xezon: xxxx

Patch104pZH/Design/Changes/v1.0/xxxx_angry_mob_movement.yaml
xezon, line 1: xxxx_angry_mob_movement.yaml

PR #2292, opened 2023-08-26
Fix: Red Guard animation issues

commy2:
- enable injured standing and moving transition
- fix standing receive injury animation
- fix missing dust trails when injured

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/468cc3bc-8d2e-416f-a6f9-47bcff3387f1

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2292_mini_gunner_animations.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2292_mini_gunner_animations.yaml
@@ -1,0 +1,22 @@
+---
+date: 2023-08-27
+
+title: Improves and fixes issues with the China Mini Gunner animations
+
+changes:
+  - fix: China Mini Gunner had the wrong animation when getting injured while standing.
+  - tweak: China Mini Gunner now has short transition animations when starting and stopping movement while injured.
+  - fix: China Mini Gunner now spawns dust trails when moving while injured.
+
+labels:
+  - bug
+  - china
+  - enhancement
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2292
+
+authors:
+  - commy2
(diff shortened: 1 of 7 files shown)


commy2:
Preserving preattack animation + https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1123 seems not feasible with current engine.

xezon:
When injured Red Guard starts moving, there is a small hick up in the transition. I assume this cannot be fixed without touching animations?

commy2:
Ah, this is beause the movement animation itself starts on a random frame, instead of the first frame.

Could fix that, however, it would mean that mass moving injured red guards have a synched up animation instead of a shifted phase one, if you can follow.

xezon:
Oh

xezon (approved):

commy2:
### Variant 1

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/0c5ff02f-0d5b-456b-817b-46c2f62217fc

- Movement loop starts on random frame, has hickup between starting-to-move and moving loop


### Variant 2

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/deb0b6a7-9a46-41bc-bebc-e85d9bf03b41

- Movemnt loop not random, no hickup between starting-to-move and moving loop

commy2:
I'm for Nr. 2

commy2:
lgtm

MTKing4:
I prefer the random star, looks more realistic , the hickup is barely noticeable, in fact, i still don't see it

commy2:
I want to cry.

xezon:
Variant 2 looks good to me.

MTKing4:
Variant 2 is ugly and much more noticeable, walking uniformly while injured is just lol

Variant 1, i barley just noticed the hiccup, it's very minuscule, not noticable without thorough investigation and even can be imagined sometimes, i started looking at variant 2 after and i started doubting myself that i saw hiccups there too, that's how minuscule it is, also the randomization is very cool and i wouldn't wanna get rid of it for that

commy2:
Rarely will you mass move injured units though. It is more likely you have one of them. The stutter can be large or small, depending on where the RANDOMSTART decides to begin the animation.

MTKing4:
I've yet to see the stutter be highly visible, i don't see it as an issue here.

Perhaps take another opinion on this

PR #2293, opened 2023-08-26
Fix: Boss Supply Truck and Civilian Toxic Supply Truck out of sync with patch changes

commy2:
Boss Supply Truck is missing the mass fix...

Toxic Supply Truck is a fully functional civilian copy / reskin of the Supply Truck and it should inherit all changes made to the regular Supply Truck.


xezon (approved):

commy2:
shot_20230826_122140_7

PR #2302, opened 2023-08-27
Change: Hide tower on Radar Van until Radar Van Scan upgrade

commy2:
shot_20230827_150805_1

shot_20230827_150820_2

- [ ] Don't spawn round radar piece when destroyed without upgrade
- [ ] Spawn radar tower scaffold piece when destroyed with upgrade (just disappears in 1.04)

Looks odd to me. Not sure I like.


xezon:
I different variant of this would be to just disable the blinking lights on the tower.

MTKing4:
If you can get this to work it would be very nice

PR #2303, opened 2023-08-27
Change: Add initial delay to Suicide ability

### LINKED ISSUE #1184: Demo and Toxin GLA Ambush are very strong
Demo and Toxin GLA Ambush are very strong in comparison to regular Ambushes. Does it need change?

Stubbjax
> The reason Rebel Ambush + Demolitions is typically banned in rule-based games is because it is absolutely outrageous that $10,000+ of stuff can be destroyed in the blink of an eye, with almost no chance for counterplay. Destructive powers / abilities should always provide the opponent with some opportunity for anticipation / avoidance / denial, and otherwise feel incredibly frustrating and unfair to play against. The specialised ambush variants unfortunately ignore this rule, and are very poorly considered / designed. (A far better design would be if the Suicide ability had an initial cooldown like the capture upgrade, so that opponents had some time to react. The Toxin variant could just spawn fewer Rebels.)

### Proposal 1

Add short cooldown to Demo suicide ability.

### Proposal 2

Spawn fewer Toxin Rebels. 3/6/12 Rebels (from 4/8/16)

### Proposal 3

Make Toxin Rebels unable to shoot for a few seconds.

### Proposal 4

Spawn fewer Demo Rebels.  2/4/8 Rebels (from 4/6/10)

commy2:
- close https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1184

This adds a configurable delay to the Suicide ability from Demogen.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/6576312/1eb1d336-8de9-4b42-9528-09c398e516a9

The goal is to nerf Demo Rebel Ambush vs armies.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -13261,7 +13261,7 @@ Object Demo_GLAInfantryRebel
     TriggeredBy = Demo_Upgrade_SuicideBomb
   End
 
-  ; Patch104p @balance commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
+  ; Patch104p @feature commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
   Behavior = OCLSpecialPower ModuleTag_SuicideAttack
     SpecialPowerTemplate = Demo_SuperweaponSuicideAttack
     OCL                  = Demo_OCL_TertiarySuicide
@@ -13791,7 +13791,7 @@ Object Demo_GLAInfantryJarmenKell
     PreTriggerUnstealthTime   = 5000 ; in milliseconds
   End
 
-  ; Patch104p @balance commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
+  ; Patch104p @feature commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
   Behavior = OCLSpecialPower ModuleTag_SuicideAttack
     SpecialPowerTemplate = Demo_SuperweaponSuicideAttack
     OCL                  = Demo_OCL_TertiarySuicide
@@ -14084,7 +14084,7 @@ Object Demo_GLAInfantryTunnelDefender
     TriggeredBy = Demo_Upgrade_SuicideBomb
   End
 
-  ; Patch104p @balance commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
+  ; Patch104p @feature commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
   Behavior = OCLSpecialPower ModuleTag_SuicideAttack
     SpecialPowerTemplate = Demo_SuperweaponSuicideAttack
     OCL                  = Demo_OCL_TertiarySuicide
@@ -14397,7 +14397,7 @@ Object Demo_GLAInfantryStingerSoldier
     TriggeredBy = Demo_Upgrade_SuicideBomb
   End
 
-  ; Patch104p @balance commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
+  ; Patch104p @feature commy2 28/08/2023 Add special power module that fires tertiary suicide weapon. (#2303)
   Behavior = OCLSpecialPower ModuleTag_SuicideAttack
     SpecialPowerTemplate = Demo_SuperweaponSuicideAttack
     OCL                  = Demo_OCL_TertiarySuicide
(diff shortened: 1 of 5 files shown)


commy2:
The delay is currently set to 3 seconds. 

I tested it with all Demogen units and it works. Second test. I tried putting units in some situations where Suicide fails in 1.04 / is buggy, such as while firing (Technical) or while stuck inside buildings (Rebel). All those issues had been fixed previously. With this change, they do not re-appear. Units behave responsive and suicide when ordered immediately.

This change also has extra logic for the Toxin Tractor that frees up the TERTIARY weapon slot, while keeping the ability to suicide. This freed slot can be used for https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1113.

commy2:
Documentation added. I tested Rebel Ambush specifically, and Rebels now need to wait 3 seconds to be able to explode.

Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
```diff
@@ -1002,3 +1002,10 @@ SpecialPower SuperweaponSatelliteHackTwo
   AcademyClassify     = ACT_SUPERPOWER ;Considered a powerful special power that a player could fire. Not for simpler unit based powers.
   ShortcutPower       = Yes     ;Capable of being fired by the side-bar shortcut.
 End
+
+; Patch104p @balance commy2 28/08/2023 Adds special power that handles initial delay for Demolitions suicide.
+;-----------------------------------------------------------------------------
+SpecialPower Demo_SuperweaponSuicideAttack
+  Enum                = SPECIAL_LAUNCH_BAIKONUR_ROCKET
```
xezon: This SPECIAL_LAUNCH_BAIKONUR_ROCKET, why was this chosen and what does it do? What would happen if there was another Special power with SPECIAL_LAUNCH_BAIKONUR_ROCKET? Would this still work?

Patch104pZH/GameFilesEdited/Data/INI/ObjectCreationList.ini
```diff
@@ -10254,6 +10254,37 @@ ObjectCreationList Nuke_OCL_NuclearTankDeathDetonation
...
+    WeaponSlot = TERTIARY
+    NumberOfShots = 1
+  End
+End
+
+; -----------------------------------------------------------------------------
+ObjectCreationList Demo_OCL_TertiarySuicideUpgrade
+  ; Special variant for Toxin Tractor. Step 1, upgrade weapon set.
+  CreateObject
+    ObjectNames       = Demo_SuicideWeaponSetUpgradeDummy
```
xezon: Can remove the white space here.

Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
```diff
@@ -20148,6 +20235,24 @@ Object Demo_GLAVehicleToxinTruck
 
 End
 
+; Patch104p @balance commy2 27/08/2023 Upgrades Tractor weapon set to unlock tertiary suicide.
```
xezon: Perhaps label these as `@feature`, because this goes beyond simple balancing, and use `@balance` only where it says `ReloadTime = 3000`.

Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
```diff
@@ -1002,3 +1002,10 @@ SpecialPower SuperweaponSatelliteHackTwo
...
+  ReloadTime          = 3000 ; in milliseconds
+End
```
xezon, line 1011: Perhaps also set `PublicTimer = No` here. The other Special Powers do that.

xezon (commented):
I tested and it appears to not work on vehicles.

Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
commy2: I don't think that's necessary. It is the default, and it is an obvious default. Besides, PublicTimer also toggles the initial delay for some reason. If it were set to Yes, it would start without initial charge. Other SpecialPower do not reference it either. When writing new objects like this, instead of copying existing ones, I think we should stick to as few parameters as possible. Less distractions.

Patch104pZH/GameFilesEdited/Data/INI/SpecialPower.ini
commy2: I added a comment to clarify.

commy2:
> I tested and it appears to not work on vehicles.

Which vehicle? Which map?

shot_20230828_172939_2
shot_20230828_172944_3

xezon:
I tested on !TestAllFactions with Demo General.

commy2:
shot_20230828_174428_1
shot_20230828_174430_2

This map? Can't reproduce.

commy2:
Also works for pre-placed units:

shot_20230828_175104_1

Keep in mind that it will not work in loaded savegames. The special power module would not be added.

commy2:
I did forget the base defenses tho. Fixed.

xezon:
I got latest and now it works.

I see there is a little glitch with the command set after suicide. After pressing the suicide button, it shows the suicide ability reloading for a short moment before hiding the command set. Any way to avoid that?

commy2:
I tried a few things without success. I don't know how that would be possible. Sry.

xezon:
Perhaps we can upgrade the command set to an empty one?

xezon:
The Jarmen Kell and RPG Trooper are silent and without Demo explosion effect on Suicide.

commy2:
> The Jarmen Kell and RPG Trooper are silent and without Demo explosion effect on Suicide.

The same appears to happen on main branch.

commy2:
Unrelated. Fixed in https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2307.

MTKing4:
Very nice! This hits 7 birds with one stone :D

PR #2307, opened 2023-08-28
Fix: RPG and Jarmen are not playing explosion effect (visuals)

commy2:
black magic

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2307_missing_infantry_demo_suicide_effects.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2307_missing_infantry_demo_suicide_effects.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-29
+
+title: Fixes missing explosion effects for suicided GLA Demo RPG-Soldier and Jarmen Kell
+
+changes:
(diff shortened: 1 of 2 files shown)


Patch104pZH/Design/Changes/v1.0/101_demo_infantry_death_voice.yaml
```diff
@@ -14,6 +14,7 @@ labels:
 
 links:
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/101
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2307
```
xezon: This looks not accurate. This change fixes missing suicide death effects of Demo RPG Trooper and Jarmen Kell.

commy2: You're right. I just tested 1.04, and the bug exists there as well (not really surprising, we don't break stuff ;)

So this PR should not be marked "Patch104p" and get its own yaml.

commy2:
Should also check Stinger Soldier.

commy2:
We need to check in general what happens if you blow them up with EXPLODED weapons. Looks like no effects present.

xezon (approved):
very good tricks

PR #2308, opened 2023-08-29
Fix: Incorrect suicide setup of GLA Demo Stinger Site

xezon:
This change fixes the incorrect suicide setup of the GLA Demo Stinger Site.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2308_demo_stinger_death_damage.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2308_demo_stinger_death_damage.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-08-29
+
+title: Fixes death explosion damage of upgraded GLA Demo Stinger Site
+
+changes:
+  - fix: The upgraded GLA Demo Stinger Site no longer deals 20 times more damage than it should when killed.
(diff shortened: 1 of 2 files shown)


xezon:
The `Demo_DestroyedWeapon` does not work on Stinger Site.

xezon:

Patch104pZH/Design/Changes/v1.0/2308_demo_stinger_death_damage.yaml
```diff
@@ -0,0 +1,20 @@
+---
+date: 2023-08-29
+
+title: Fixes death explosion damage of upgraded GLA Demo Stinger Site
+
+changes:
+  - fix: The upgraded GLA Demo Stinger Site no longer deals 20 times more damage than it should when killed.
+  - fix: The upgraded GLA Demo Stinger Site now properly deals damage when killed by poison.
```
commy2, line 8: Sorry for pointing this out so late, but SwallowDamageTypesIfSlavesNotExisting ensures that Stinger Site does not take damage from POISON damage, and all POISONED_X death type weapons deal POISON damage. The fix is good, but the effect is theoretical.

xezon: Fixed.

PR #2309, opened 2023-08-29
Remove commented suicide modules

xezon:
This change removes several commented suicide modules which we do not need to see.


commy2 (approved):
taking out the trash

PR #2310, opened 2023-08-29
Change: Make GLA Demo Stinger soldiers explode on death

xezon:
This change makes the GLA Demo Stinger soldiers explode on death after Demolitions Upgrade like regular units.


xezon:
I suspect we do not need this. Angry Mobs also do not explode on kill, so there is also inconsistency in other areas.

commy2:
I'm not sold on this either. Stinger Troopers are special enough to not follow the rules.

Issue #1739, opened 2023-02-19
The disabled binary shower particle effect causes a memory leak and eventual crash


Stubbjax:
The disabled binary shower particle effect causes a memory leak and eventual crash.

![image](https://user-images.githubusercontent.com/11547761/219927756-7c1d9292-1244-482e-9f6a-82dff4a87ad9.png)

`DisabledEffectBinaryShower0` is the "0" effect, which spawns a `SlaveSystem` of the `DisabledEffectBinaryShower1` "1" effect as part of the overall disabled building / vehicle particle effect. Removing the `SlaveSystem` will solve the leak, but it doesn't look as good without the "1" effect and it does not solve the underlying issue. This particular leak is unlikely to come up during typical gameplay as it is so small and the abilities are uncommon, but it all adds up over time and it may be indicative of other memory issues (playing many games / watching many replays will gradually increase memory usage). There are 60 slave systems in `ParticleSystem.ini`.

```
ParticleSystem DisabledEffectBinaryShower0
  SlaveSystem = DisabledEffectBinaryShower1
End
```

It should be noted that `SpecialAbilityHackerDisableBuilding` handles / overrides the `SystemLifetime` of its `DisableFXParticleSystem` via the `EffectDuration` field (which also controls how long the target is disabled), which may interfere with the slave system's lifetime. Particle system generation frequency is tied to how often the disabled condition is triggered on the target. Altering these values will affect gameplay.

```
Behavior = SpecialAbilityUpdate ModuleTag_05
  PersistentPrepTime = 333 ; NOTE! This drives how often the disable effect gets triggered
                           ; This is also how often a new particle system effect is spawned
  EffectDuration = 2000
End
```

https://user-images.githubusercontent.com/11547761/219955698-e57d4a73-ebf3-49f7-b1d2-29286d922b4c.mp4

---

Note: Triggering either of the (Super) Hacker's Disable Building / Vehicle Hack abilities will also leak memory.

Note: Applying a high negative vertical drift velocity will result in the particle drifting into the ground and culling itself.

xezon:
Perhaps this can be fixed by setting slave `SystemLifetime = 70` to `SystemLifetime = 0`.

Stubbjax:
The particles won't ever destroy if `SystemLifetime = 0`. Each instance of `DisabledEffectBinaryShower0` will only create a single instance of `DisabledEffectBinaryShower1`.

xezon:
[...]

Issue #1519, opened 2023-01-06
Game performance gets slower over time


xezon:
* Similar to #946

Reproducible on map

https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Maps/Any/!TestAllFactionsBuildingsUnitsPromo

xezon:
I narrowed it to USA Microwave. Just build a bunch of them and wait.

What did we do to Microwave?

![shot_20230106_194754_1](https://user-images.githubusercontent.com/4720891/211078726-ad4f4766-7849-46f4-b3e5-bec0cd80de2b.jpg)

MTKing4:
Texture improvements, other than that idk

xezon:
@commy2 
Any ideas?

xezon:
Possibly related to particles. When frame rate is low, many particles no longer spawn, perhaps because more priority effects use up all particle slots.

![shot_20230128_205942_1](https://user-images.githubusercontent.com/4720891/215288331-7b1cc135-4992-4a21-b60c-349a8a29e073.jpg)

xezon:
Ok I think I found the offending change.

xezon:
For some reason the `MicrowaveFlare` effect never despawns. It is essentially a broken effect. Here it is being visualized. Investigating...

![shot_20230128_215621_2](https://user-images.githubusercontent.com/4720891/215290777-9eb4b82c-72b7-45ec-8d0b-a51197bcffa3.jpg)

xezon:
Ok it is `SystemLifetime = 0`. Zero appears to be synonymous for infinite. Setting it to 1 fixes it.

![shot_20230128_215839_1](https://user-images.githubusercontent.com/4720891/215290863-15f6cdbe-8c54-47a6-9900-8dd7ba056991.jpg)

xezon:
In my understanding, `SystemLifetime = 0` is only ok if the parent object is short lived. If lifetime is uncapped, then particle will spawn every frame and live forever.

On quick inspection, there are at least 2 suspicious effects:

JetBurning
WaterSpraySplash

commy2:
Which commit introduced this?

xezon:
> Which commit introduced this?

39f063905bdefcadece7f0249f87901913f03682

PR #2311, opened 2023-08-29
Improve the particle effect for destroyed GLA Demo units

xezon:
This change improves the particle effect for destroyed GLA Demo units. It now looks more similar to the regular Demo Suicide explosion, but with slightly smaller effects and without the explosion pillars.

shot_20230830_193613_3

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/1076_demo_units_death_particles.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/1076_demo_units_death_particles.yaml
@@ -14,6 +14,8 @@ labels:
 
 links:
   - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/1076
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2311
 
 authors:
   - commy2
+  - xezon
--- a/Patch104pZH/Design/Changes/v1.0/2308_demo_stinger_death_damage.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2308_demo_stinger_death_damage.yaml
@@ -5,7 +5,6 @@ title: Fixes death explosion damage of upgraded GLA Demo Stinger Site
 
 changes:
   - fix: The upgraded GLA Demo Stinger Site no longer deals 20 times more damage than it should when killed.
-  - fix: The upgraded GLA Demo Stinger Site now properly deals damage when killed by poison.
 
 labels:
   - bug
--- a/Patch104pZH/Design/Changes/v1.0/2311_demo_units_death_particles.txt
+++ b/Patch104pZH/Design/Changes/v1.0/2311_demo_units_death_particles.txt
@@ -1,0 +1,1 @@
+1076_demo_units_death_particles.yaml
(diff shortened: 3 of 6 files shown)


xezon:
Image added. Documentation added. Looks good.

commy2:
I want to try this first. The difference should be very noticable, since it does vastly different amounts of damage. 1.04 already is not very clear about this.

xezon:
They are distinct, also thanks to the diagonal line lol.

shot_20230830_200223_11

commy2:
Yea, glow helps. Right one looks kind of big though. It does pathetic damage, so it really should just be a puff. The donut adds to the size and probably too much.

xezon:
It does not deal much damage, but it has huge damage radius. Bigger than suicide radius. Also keep in mind it needs to fit on vehicles. If it is too small, it will look unfitting.

I made a bit smaller now. Check.

commy2:
will do

xezon:
commy2
> Re visuals: I think the donut could be smaller. I mean, same radius obv, but smaller wall thickness. Or maybe the radius could be decreased if it is also decided that the secondary damage radius should be 50 instead of 70.

xezon
> We can reduce the radius to match or undercut suicide radius yes. Does not make sense that it is larger than suicide.

Edit:
* Done with TheSuperHackers/GeneralsGamePatch#2420

Float1ngFree:
I've noticed that there's a continuous effects performance degradation the longer the game goes vs demo. Some visual & transition effects stop displaying the more demo units you kill & the more explosion they produce. I've tried playing with ParticleSystems `Priority` - changing from `CRITICAL` & `WEAPON_EXPLOSION` to `DEATH_EXPLOSION` but that didn't solve the issue. 

Seems like effects are accumulating and prevent other effects from displaying fully. Similar to when you have too many firestorms simultaneously.

xezon:
Can you try to figure out which particle system is causing it? It needs its own bug report. Likewise we know that `DisabledEffectBinaryShower0` is causing problems (#1739).

Float1ngFree:
> Can you try to figure out which particle system is causing it? It needs its own bug report. Likewise we know that `DisabledEffectBinaryShower0` is causing problems (#1739).

I've tried switching back to `WeaponFX_DemoSuicideDynamitePackDetonationPlusFire` but the result is the same.
Commented out `ExplosionSmall` and `DemoExplosionSmokeLite` -- a bit better but still accumulates.
Set `SystemLifetime` for these new small ParticleSystems similar to `WeaponFX_DemoSuicideDynamitePackDetonationPlusFire` but no difference.

I'm sorry, I don't have much experience dealing with ParticleSystem items... You could try his on your side, just killing Demo AI units at his base (after demo upgrade) and observe how effects seem to stuck more and more...

BTW, in https://github.com/TheSuperHackers/GeneralsGamePatch/issues/1519 you've turned questionable ParticleSystem into green for debugging -- can you please share how that could be done? I want to debug this as well and check out couple more effects I have my doubts about.

xezon:
Is this performance bug Patch specific or in Original game?

Float1ngFree:
> Is this performance bug Patch specific or in Original game?

Seems to be Patch, original plays all animations no matter how many units/buildings you kill in at demo base...

PR #2312, opened 2023-08-29
Fix: Add key mappings to Battleship Bombardment and Aircraft Carrier Raptor Attack

xezon:
This change adds key mappings to Battleship Bombardment and Aircraft Carrier Raptor Attack.


xezon:
Documentation added. Good.

PR #2313, opened 2023-08-29
Fix: Optimize suicide explosion effects of GLA Demo units

xezon:
This change simplifies suicide explosion effects. All `FX_TerroristSuicide` effects are removed. Death weapons use just one death effect now.


xezon:
Why is this draft? What is missing?

PR #2314, opened 2023-08-30
Add: Destroy ability to scrap piles

### LINKED ISSUE #648: Scrap crates can be removed by placing scaffolds

commy2:
- fix https://github.com/TheSuperHackers/GeneralsGamePatch/issues/648

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -8966,3 +8966,13 @@ CommandButton Command_ConstructTechRadioStation
   ButtonBorderType        = BUILD ; Identifier for the User as to what kind of button this is
   DescriptLabel           = OBJECT:TechRadioStation
 End
+
+; Patch104p @bugfix commy2 30/08/2023 Adds button to blow up scrap piles. (#2314)
+CommandButton Command_Destroy
+  Command           = FIRE_WEAPON
+  WeaponSlot        = PRIMARY
+  TextLabel         = CONTROLBAR:Destroy
+  ButtonImage       = SUUseDeto
+  ButtonBorderType  = ACTION ; Identifier for the User as to what kind of button this is
+  DescriptLabel     = CONTROLBAR:TooltipDestroy
+End
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandSet.ini
@@ -5875,3 +5875,8 @@ CommandSet Boss_ChinaNuclearMissileCommandSetUpgrade
  12 = Boss_Command_UpgradeEMPMines
  14 = Command_Sell
 End
+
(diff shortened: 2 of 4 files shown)


xezon:
Salvage Crate now shows health bar and text.

shot_20230830_164554_2

commy2:
Healthbar is required to show which object is selected. I don't think there is a way to not show object texts for selectable objects, and I think it is ok.

MTKing4:
Nice! Does this work as intended? Can be collected with GLA and destroyed with all factions?

@ImTimK what do you think we test that?

commy2:
Well, it works as is my interpretation of what you intended :)

Any faction can see the scraps with this. Only GLA can collect. Only owner of the wreck can delete it.

MTKing4:
Very Nice, Commy OP!

ExiLe-ZH:
Always up for testing

PR #2315, opened 2023-08-30
Change: Make 100% completed Command Center required to upgrade Cash Bounty, 2

commy2:
In case it is decided to go this route again, 1 PR fix with all the necessary changes.

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/CommandButton.ini
@@ -4151,9 +4151,11 @@ CommandButton Command_PurchaseScienceRebelAmbush3
 End
 
 
+; Patch104p @feature commy2 30/08/2023 Adds dynamic "Requires Command Center" line to tooltip. (#2308)
 CommandButton Command_PurchaseScienceCashBounty1
   Command           = PURCHASE_SCIENCE
   Science           = SCIENCE_CashBounty1
+  Object            = CashBountyPrerequisiteDummy
   ButtonImage       = SSCashBounty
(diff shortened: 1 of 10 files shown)


xezon:
Is the Cash Bounty gone after all Command Centers are demolished?

commy2:
No, the last level of Cash Bounty that was unlocked with a Command Center remains in effect forever. To upgrade it to the next level later however, a new Command Center is required. If that were to change, Thyme intervention would be required.

The Command Center requirement will be noted on the General's Promotion menu.

xezon:
I don't like this, because it is complicated. We had this discussion before. The way it is in main branch is a much simpler design.

xezon:
Can we close this?

MTKing4:
The ideal solution imo is that it you only need the CC fully built once and you get the promotion, upgrading from then on doesn't need the CC, and that requires thyme, so just wait for thyme

xezon:
The way powers and unlocks typically work is that the player needs to own a certain asset. With the design presented in this change the player would not need to keep the asset, but just acquire it once. This is counter-intuitive. The natural expectation, based on all other unlocks, is that the Cash Bounty would be removed again once the Command Center is demolished.

commy2:
> The ideal solution imo is that it you only need the CC fully built once and you get the promotion, upgrading from then on doesn't need the CC, and that requires thyme, so just wait for thyme

I haven't tried, but as far as I can tell, this would not require Thyme. It should be totally possible to make Cash Bounty 2 and 3 only require that Cash Bounty 1 was previously activated by a Command Center.

MTKing4:
Oh nice, if you could make this work then the argument that it's complicated will be obselete :D

xezon:
No need to change again at this time.

PR #2319, opened 2023-09-03
Change: Add overlord expansion sound effect to GLA Bomb Truck upgrades

xezon:
This changes adds the overlord expansion sound effect to GLA Bomb Truck upgrades. This makes it consistent with Overlord and Helix upgrade sounds, which are conceptually identical.


xezon:

PR #2323, opened 2023-09-03
Fix: Add additional information to tool tip strings of China Overlord and Helix Speaker Tower for all languages

xezon:
This change adds additional information to tool tip strings of China Overlord and Helix Speaker Tower for all languages. The sentences are copies of the Speaker Tower building tool tip.


xezon:

PR #2324, opened 2023-09-04
Fix: Improve wording in tool tip strings of China Overlord and Helix upgrade for all latin languages

xezon:
This change improves the wording and reduces inconsistencies in tool tip strings of China Overlord and Helix upgrade for all latin languages.


xezon:

PR #2326, opened 2023-09-06
Fix: Speed up briefings and speech subtitles

xezon:
**Merge with Rebase**

This change speeds up the military briefing texts for all languages.

And it significantly reduces the briefing delays for Spanish and Italian because these languages use subtitles for speeches and need to have fast prints to be able to properly show texts in time.

Korean and Chinese use subtitles too.


xezon:
Looks good to me. Tested in English, Spanish, Italian, Chinese, Korean.

PR #2317, opened 2023-09-02
Fix: Punctuation inconsistencies in various strings for all languages (2)

xezon:
**Merge with Rebase**

* Follow up for #2316

This change fixes punctuation inconsistencies in various strings for all languages.

Affects

* DIALOGEVENT
* MAP
* GUI
* TOOLTIP
* MD_GLA05-MilitaryBriefing
* MDGLA02
* MDGLA03
* HELP
* GENERIC
* LOSE
* ACADEMY
* SCRIPT
* WOL
* Buddy
* QM

Was reviewed by hand without algorithm. Likely did not catch everything, but the bulk of it.

The rules are:

* English, Russian sentences to do not end with a dot before text end and before line feed.
* German, French, Spanish, Italian, Korean, Chinese, Brazilian, Polish sentences do end with a dot before text end and before line feed.
* Arabic untouched as usual.

PR #2329, opened 2023-09-07
Fix: Optimize line breaks in various strings for all languages

xezon:
* Follow up for #2317

This change optimizes line breaks in various strings for all languages.

* Mission briefing and speech subtitle strings will no longer print off the screen
* Fixes a bunch more punctuation issues not covered by #2317 
* Adds missing line breaks at begin of speech subtitle strings for Spanish, Italian, Korean, Chinese languages
* Now uses the 3 dot character for Korean, Chinese languages consistently
* Disables speech subtitle strings for unsupported languages (except Arabic)
* Adds missing hint titles to a few map strings
* Removes superfluous line feeds in gui strings for German language


xezon:
Should probably split this up a bit.

xezon:
Split off with #2337

PR #2330, opened 2023-09-07
Fix: Quote symbol errors in Chinese language strings

xezon:
This change fixes quote symbol errors in Chinese language strings.


xezon:
Documentation added. Good.

PR #2331, opened 2023-09-07
Fix: Add missing line breaks to the begin of Korean dialogevent strings

xezon:
This change adds the missing line breaks to the begin of Korean dialogevent strings. Otherwise the first line prints in bold font, which is not supposed to happen for speech subtitles.


xezon:
Changes were applied with script. Documentation added. Looks good.

PR #2332, opened 2023-09-07
Fix: Use three dot character consistently in Korean, Chinese language strings

xezon:
This change applies the three dot character consistently in Korean, Chinese language strings.


xezon:
Documentation added. Looks good. Perhaps needs another pass on Chinese regarding the spacing near dots.

PR #2333, opened 2023-09-07
Fix: Punctuation inconsistencies in various strings for all languages (2)

xezon:
**Merge with Rebase**

* Follow up for #2317

This change fixes more punctuation inconsistencies in various strings for all languages.


xezon:

PR #2334, opened 2023-09-07
Fix: Add missing like breaks at the begin of Spanish, Italian, Chinese dialogevent strings

xezon:
* Follow up for #2331

This change add missing like breaks at the begin of Spanish, Italian, Chinese dialogevent strings.


xezon:
Documentation added. Good.

PR #2335, opened 2023-09-07
Fix: Disable dialogevent subtitles for all unsupported languages

xezon:
This change disables dialogevent subtitles for all unsupported languages. Excludes Arabic.


xezon:
No special documentation needed.

PR #2336, opened 2023-09-07
Fix: Add missing hint titles to map strings

xezon:
This change adds missing hint titles to map strings.


xezon:

PR #2337, opened 2023-09-08
Fix: Optimize line breaks in various strings for all languages

xezon:
**Merge with Rebase**

This change optimizes line breaks in various strings for all languages.

Mission briefing and speech subtitle strings will no longer print off the screen.

Affects

* MAP
* SCRIPT
* HELP
* MDGLA02
* MDGLA03
* MD_GLA05
* DIALOGEVENT


xezon:
This line break stuff is difficult to get right. I completed multiple review passes. Many strings are well fixed and optimized now, but I probably missed some better optimizations.

xezon:
Documentation added. Good.

PR #2340, opened 2023-09-10
Fix: Add missing string formats to language strings

xezon:
* Follow up for #2216

This change adds missing string formats to language strings. And fixes a few styling issues.

Mainly affects:

* WOL:BuddyOffline
* Version:BuildTime
* Version:BuildMachine
* Version:BuildUser


xezon:
Documentation added. Good.

PR #2341, opened 2023-09-10
Fix: Punctuation inconsistencies in various strings for all languages (3)

xezon:
* Follow up for #2316
* Follow up for #2317

This change fixes punctuation inconsistencies in various strings for all languages.

Affects

* NETWORK:
* LAN:


xezon:
Documentation added. Good.

PR #2342, opened 2023-09-10
Fix: Improve tool tip and help text of USA MOAB

xezon:
* Follow up for #2146

This change improve the tool tip and help text of USA MOAB for all languages.


xezon:

PR #2345, opened 2023-09-11
Fix: Miscellaneous string fixes

xezon:
A bunch of string fixes and tweaks.


xezon:

PR #2351, opened 2023-09-12
Fix: Strength and weakness inconsistencies in tool tip strings

xezon:
This change fixes some strength and weakness inconsistencies in tool tip strings.

* Merge after #2346


xezon:
Documentation added. Good.

PR #2352, opened 2023-09-12
Fix: Improve tool tip strings of GLA Battle Bus

xezon:
This change improves the tool tip strings of the GLA Battle Bus for all languages.

It no longer lists strengths and instead mentions that the passengers can shoot out of it, unlike the GLA Technical.


xezon:

PR #2353, opened 2023-09-12
Fix: Improve tool tip strings of GLA Rocket Buggy

xezon:
This change improves the tool tip strings of the GLA Rocket Buggy. All languages now list the same strengths.


xezon:

Issue #2097, opened 2023-07-11
GLA Saboteur is ineffective against Production facilities and Supply centers


ExiLe-ZH:
- SW reset: Very good
- Genpowers reset: Very good
- Forcing low power: Very good
- Stealing $1K from Supply Center: 
Mediocre (the enemy loses $1K, but you only gain $200).
- Delay production for 30 seconds: Not great at all.

I would suggest buffing production delay to 45 seconds, 1 minute is probably too much again and might evolve a new Sabotech spam meta?

Perhaps a cashsteal buff is also justified? At $1500 you will gain $700 max, depending on how much the enemy floats (you can lose your investment aswell). It's more than Black Lotus Cash Steal however the Saboteur can only be used once.

---

### Current Saboteur Stats:

- Supply Center: Steal $1000
- Supply Drop Zone, Black Market: Steal $1000
- Barracks/War Factory/Airfield: Disable for 30 seconds
- Internet Center: Disable for 15 seconds (equals $750 cash not earned for 8 vet 3 Hackers)
- Particle Cannon, Nuclear Missile Silo, Scud Storm: Reset timer
- Command Center: Reset Genpower timers
- Sabotage Fake Building: Explode

### 1.04 Saboteur Stats:
- same as above, except
- No stealing from Supply Drop Zone or Black Market (can be exploited to tell real and fake Black Markets apart as vGLA)
- Stealthgen Saboteur cannot disable Internet Centers
- Stealthgen Saboteur cannot enter Fake Buildings (can be exploited to tell real and fake buildings apart)

---

### Proposal 1

Increase Barracks/Factory sabotage delay to 45 seconds (from 30 seconds)

* Applied with #2358

### Proposal 2

Steal $1200 instead of just $1000

* Applied with #2355

### Proposal 3

Reduce cost from $800 to $600 ($500?). Reduce total build time from 30 seconds to 20 seconds

### Proposal 4

Increase Internet Center disable time by calculating fair time based on steal cash amount equivalent (estimate with 8 hackers).

* Applied with #2357

### Proposal 5

Disable Internet Center for 45 seconds (from 15, denies maximum of $2250, $1631.25 for 8 "average" Hacker)

commy2:
Either reduce cost to something like $500 or make the cash steal at least $2000 (+$1200 max for player). He also has a 30 second build time, so worth looking into that (Overlord build-time is 24 seconds for example, Aurora 32 sec).

[...]

PR #2355, opened 2023-09-13
Change: Increase steal cash amount of GLA Saboteur from 1000 to 1200

xezon:
* Refers to #2097

This change increases the steal cash amount of the GLA Saboteur from 1000 to 1200.

This bumps the max gain factor from 2.5 to 3.0.

### Original Saboteur reward

1x Saboteur, entering enemy Supply Center

```
-800 build cost
+1000 cash reward
+1000 enemy cash loss

= +200 self cash reward
= +1200 total cash gain

Max cash gain factor 2.5 aka +150%
```

### Patched Saboteur reward

1x Saboteur, entering enemy Supply Center

```
-800 build cost
+1200 cash reward
+1200 enemy cash loss

= +400 self cash reward
= +1600 total cash gain

Max cash gain factor 3.0 aka +200%
```

### Terrorists kill reward

4x Terrorist, killing enemy War Factory

```
-800 build cost
+2000 enemy cash loss
+15 or more seconds production delay
+200 XP

= +0 self cash reward
= +1200 total cash gain

Common cash gain factor 2.5 aka +150%
```

1x Technical + 4x Terrorist, killing enemy War Factory

```
-800 build cost for Terrors
-500 build cost for Technical
+2000 enemy cash loss
+15 or more seconds production delay
+200 XP

= +0 self cash reward
= +700 total cash gain

Common cash gain factor 1.54 aka +54%
```

# Rationale

The GLA Saboteur is seldom used to steal cash from an opponent. The max gain factor of 2.5 is not more attractive in comparison to the Terrorists.

The Terrorists have the same gain factor, but they also delay enemy income or production and give experience points for kills.

The Saboteur is stealth, but is unarmed and has no utility other than sabotaging enemy structures. It needs to have a valuable reward, otherwise it is not attractive to invest in. It is a gamble unit, similar to the Terrorist and Bomb Truck. The max steal cash value is not guaranteed: If the target player has less cash on the bank, then the Saboteur will withdraw just as much cash is available, anywhere from 0 to 1200. The Saboteur requires an Arms Dealer to build and has a relatively long build time of 30 seconds (without Power).

Raising the max gain factor from 2.5 to 3.0 appears reasonable to make the Saboteur more attractive to use.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
@@ -7,6 +7,7 @@ changes:
   - tweak: Increases the steal cash amount of the GLA Saboteur from 1000 to 1200.
 
 labels:
+  - buff
   - controversial
   - design
   - gla
--- a/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
@@ -1,0 +1,20 @@
+---
+date: 2023-09-13
+
+title: Increases steal cash amount of GLA Saboteur from 1000 to 1200
+
+changes:
+  - tweak: Increases the steal cash amount of the GLA Saboteur from 1000 to 1200.
+
+labels:
+  - controversial
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2355
+
+authors:
+  - xezon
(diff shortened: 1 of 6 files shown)


xezon:
After correcting the information above, I no longer believe this change is necessarily required. Saboteur reward already is good and comparable to Terrorists with max gain factor of 2.5. This change raises max gain to 3.0. We can do it, but perhaps Saboteur will be too good?

The logical counter play here for USA are Sentry Drones and for China are Mines.

ExiLe-ZH:
This is a good change I think (please read my comment in the main topic).

xezon:
MTKing
> look good to me

xezon:
Cash gain values corrected. Edit: Reverted, because it was correct before.

xezon:
How about we set Cash Steal amount to 1100 instead of 1200? Then cash gain factor is 2.75, which is close to the average cash gain factor 2.71 of #2357.

xezon:
commy2:
> $1200 is good.

xezon:
Good.

PR #1940, opened 2023-05-09
Fix: Misspelled speech in dialog events

xezon:
* Relates to #176

This change fixes misspelled speech in dialog events. Some or all of these events may be unused in the game.

Affects

* EvaGLA_CashStolen
* EvaGLA_VehicleStolen
* EvaChina_CashStolen
* EvaUSA_BaseDefensesOffLine
* EvaUSA_CashStolen
* Cin_SquadronFlyBy02
* Taunts_Turtle087
* EvaGLA_FundsTransferred

PR #2356, opened 2023-09-13
Disable Cash Stolen EVA event

xezon:
* Follow up for TheSuperHackers/GeneralsGamePatch#1940
* Refers to TheSuperHackers/GeneralsGameCode#195

This change explicitly disables the Cash Stolen EVA event. This restores how it works originally user facing.


xezon:

PR #2027, opened 2023-06-22
Change: Increase armor of China Internet Center against Aurora bomb by 30%

xezon:
* Follow up for #1989
* Relates to #223
* Relates to #754

This change increase the armor of the China Internet Center against the Aurora bomb by 30%.

The Internet Center becomes more difficult to take down with Aurora bombers.

| Structure                          | No. Auroras to kill | No. Alpha Auroras to kill |
|------------------------------------|---------------------|---------------------------|
| China Propaganda Center            |  1                  |  1                        |
| China Airfield                     |  2                  |  1                        |
| China War Factory                  |  2                  |  2                        |
| China Command Center               |  5                  |  3                        |
| Original Internet Center           |  3                  |  2 (Hackers alive)        |
| **Patched Internet Center (this)** |  4                  |  3 (Hackers dead)         |

## Original Damages

| Weapon                                | Internet Center Damage | Hackers killed       |
|---------------------------------------|------------------------|----------------------|
| **USA Alpha Aurora**                  | 58%                    | No                   |
| **USA Aurora**                        | 40%                    | No                   |

## Patched Damages

| Weapon                                | Internet Center Damage | Hackers killed       |
|---------------------------------------|------------------------|----------------------|
| USA Carpet Bomb + Lvl 2 A10           | 100%                   | Yes                  |
| USA Lvl 1 Spectre + Lvl 2 A10         | 100%                   | Yes                  |
| USA Lvl 2 Spectre + Lvl 1 A10         | 100%                   | No                   |
| GLA Demo Scud Storm                   | 100%                   | No                   |
| GLA Lvl 3 Demo Rebel Ambush           | 100%                   | No                   |
| USA Lvl 3 Spectre                     | 99%                    | No                   |
| USA Carpet Bomb + Lvl 1 Spectre       | 99%                    | No                   |
| China Carpet Bomb + Lvl 2 Artillery   | 98%                    | No                   |
| Patched GLA Anthrax Gamma Scud Storm  | 92%                    | No                   |
| GLA Scud Storm                        | 92%                    | No                   |
| USA Lvl 3 A10                         | 90%                    | No                   |
| China Nuke Missile                    | 85%                    | No                   |
| USA Particle Cannon                   | 85%                    | No                   |
| USA Carpet Bomb + Lvl 1 A10           | 79%                    | No                   |
| USA Lvl 1 Spectre + Lvl 1 A10         | 77%                    | No                   |
| China Lvl 3 Artillery                 | 76%                    | No                   |
[...]

PR #1989, opened 2023-06-04
Change: Increase armor of China Internet Center against Particle Cannon beams by 25% and Explosion attacks by 28.5%

xezon:
* Resolves #223
* Relates to #754

This change increases the armor of the China Internet Center against Particle Cannon beams by 25% and Explosion attacks by 28.5%.

With this change it is no longer possible to kill the pristine Internet Center and all its Hackers with a single weapon strike. The GLA Demo Scud Storm and GLA Demo Rebel Ambush can kill the pristine Internet Center, but will not be enough to kill the Hackers.

Internet Center armor against weapons not of PARTICLE_BEAM and EXPLOSION is unchanged.

| Weapon                                | Internet Center Damage | Hackers killed       |
|---------------------------------------|------------------------|----------------------|
| Original GLA Anthrax Gamma Scud Storm **(1)** | 100%                   | Yes (by the Anthrax) |
| USA Carpet Bomb + Lvl 2 A10           | 100%                   | Yes                  |
| USA Lvl 1 Spectre + Lvl 2 A10         | 100%                   | Yes                  |
| USA Lvl 2 Spectre + Lvl 1 A10         | 100%                   | No                   |
| GLA Demo Scud Storm                   | 100%                   | No                   |
| GLA Lvl 3 Demo Rebel Ambush           | 100%                   | No                   |
| USA Lvl 3 Spectre                     | 99%                    | No                   |
| USA Carpet Bomb + Lvl 1 Spectre       | 99%                    | No                   |
| China Carpet Bomb + Lvl 2 Artillery   | 98%                    | No                   |
| Patched GLA Anthrax Gamma Scud Storm  | 92%                    | No                   |
| GLA Scud Storm                        | 92%                    | No                   |
| USA Lvl 3 A10                         | 90%                    | No                   |
| China Nuke Missile                    | 85%                    | No                   |
| USA Particle Cannon                   | 85%                    | No                   |
| USA Carpet Bomb + Lvl 1 A10           | 79%                    | No                   |
| USA Lvl 1 Spectre + Lvl 1 A10         | 77%                    | No                   |
| China Lvl 3 Artillery                 | 76%                    | No                   |
| USA Lvl 2 Spectre                     | 72%                    | No                   |
| USA Lvl 2 A10                         | 61%                    | No                   |
| China Lvl 2 Artillery                 | 51%                    | No                   |
| China Carpet Bomb                     | 48%                    | No                   |
| USA Carpet Bomb                       | 48%                    | No                   |
[...]

PR #2357, opened 2023-09-13
Change: Increase Internet Center sabotage duration of GLA Saboteur from 15000 to 60000 ms

xezon:
* Refers to #2097
* Related to #2355

This change increases the Internet Center sabotage duration of the GLA Saboteur from 15000 to 60000 ms.

This bumps the average gain factor on disabling 8 Hackers from 0.68 to 2.71. The gain factor can be lower or higher depending on the veterancy and amount of Hackers inside the Internet Center.

Saboteur build cost is 800.

### Math

Hackers give 5, 6, 8, 10 cash per tick, averaging at 7.25 cash per tick.

Internet Center tick is 1.6 seconds. 7.25 cash / 1.6 seconds = 4.53125 cash per second.

8 hackers * 4.53125 cash per second = 36.25 cash per second.

For fair sabotage value we use 2000, as much as original max cash gain.

2000 cash / 36.25 cash per second = 55 seconds.

We can round it up to 60 seconds.

## Original Internet Center sabotage

With a sabotage duration of 15 seconds.

| Hackers  | Sabotage cash value | Sabotage real value | Sabotage gain |
|----------|---------------------|---------------------|---------------|
| 8x Vet 0 | 375                 | -425                | 0.47 (-53%)   |
| 8x Vet 1 | 450                 | -350                | 0.56 (-44%)   |
| 8x Vet 2 | 600                 | -200                | 0.75 (-25%)   |
| 8x Vet 3 | 750                 | -50                 | 0.94 (-6%)    |

## Patched Internet Center sabotage

With a sabotage duration of 60 seconds.

| Hackers  | Sabotage cash value | Sabotage real value | Sabotage gain |
|----------|---------------------|---------------------|---------------|
| 8x Vet 0 | 1500                | +700                | 1.88 (+88%)   |
| 8x Vet 1 | 1800                | +1000               | 2.25 (+125%)  |
| 8x Vet 2 | 2400                | +1600               | 3.00 (+200%)  |
| 8x Vet 3 | 3000                | +2200               | 3.75 (+375%)  |

## Other things that can disable for reference

| Object                   | DisabledDuration |
|--------------------------|------------------|
| EMPPulseEffectSpheroid   |  30000           |
| LeafletContainer         |  20000           |
| EMPPatriotEffectSpheroid |  10000           |

## Patch China Economy buffs for reference

China has gotten a lot of buffs for its economy so far. Therefore this buff for GLA against China should be no major concern for China, but make the game content richer by giving the GLA Saboteur more valuable utility.

* #768
* #769
* #770
* #771
* #772
* #773
* #1533
* #1989
* #2027

# Rationale

In original game, it is not worthwhile to send the Saboteur into an Internet Center, because the maximum return for the sabotage duration of 15 seconds is lower than the actual cost of the Saboteur. Therefore the cash gain is negative. The only tangible advantage is the pausing of upgrade research and Satellite Hack spying for those 15 seconds.

This change increases the sabotage duration by +300% to achieve an average gain factor of 2.71 against a fully occupied Internet Center, which is similar to the max cash gain factor of 2.5 or 3.0 with the Supply Center (#2355).

The Internet Center can be easily defended from Saboteurs with the Mines upgrade.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
@@ -4,7 +4,7 @@ date: 2023-09-13
 title: Increases steal cash amount of GLA Saboteur from 1000 to 1200
 
 changes:
-  - tweak: Increases the steal cash amount of the GLA Saboteur from 1000 to 1200.
+  - tweak: Increases the steal cash amount of the GLA Saboteur from 1000 to 1200. The maximum cash gain factor against a Supply Center increases from 2.5 to 3.0.
 
 labels:
   - buff
--- a/Patch104pZH/Design/Changes/v1.0/2357_saboteur_internet_center_sabotage_duration.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2357_saboteur_internet_center_sabotage_duration.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-09-13
+
+title: Increases Internet Center sabotage duration of GLA Saboteur from 15 to 60 seconds
+
+changes:
+  - tweak: Increases the Internet Center sabotage duration of the GLA Saboteur from 15 to 60 seconds. The average cash gain factor against an Internet Center with 8 Hackers increases from 0.68 to 2.72. Internet Center upgrade research and Satellite Hack I & II are paused for the same sabotage duration.
+
+labels:
+  - buff
+  - controversial
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2357
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -17456,7 +17456,7 @@ Object Demo_GLAInfantrySaboteur
 
   ; Patch104p @bugfix commy2 10/10/2021 Add ability to enter Supply Drop Zones and Black Markets.
   ; Patch104p @tweak xezon 13/09/2023 Changes steal cash amount from 1000. (#2355)
-  ;   Max gain factor is 3x: -800 build cost +1200 reward +1200 enemy loss = 1600 gain.
+  ;   Max cash gain factor is 3x: -800 build cost +1200 reward +1200 enemy loss = 1600 gain.
 
   Behavior = SabotageSupplyDropzoneCrateCollide   SabotageTag_02
     BuildingPickup  = Yes
@@ -17483,6 +17483,8 @@ Object Demo_GLAInfantrySaboteur
     BuildingPickup = Yes
   End
   ; Patch104p @tweak xezon 13/09/2023 Changes sabotage duration from 15000 to achieve fair value. (#2357)
+  ;   Cash gain factor varies with amount of Hackers and their veterancy level(s).
+  ;   Average cash gain factor with 8 Hackers is 2.72x: -800 build cost (+36.25 cash/second * 60 seconds) = 1375 gain.
   Behavior = SabotageInternetCenterCrateCollide   SabotageTag_08
     BuildingPickup = Yes
     SabotageDuration = 60000
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -17482,9 +17482,10 @@ Object Demo_GLAInfantrySaboteur
   Behavior = SabotageFakeBuildingCrateCollide     SabotageTag_07
     BuildingPickup = Yes
   End
+  ; Patch104p @tweak xezon 13/09/2023 Changes sabotage duration from 15000 to achieve fair value. (#2357)
   Behavior = SabotageInternetCenterCrateCollide   SabotageTag_08
     BuildingPickup = Yes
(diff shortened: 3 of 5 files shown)


xezon:
After further investigation I have bumped the Internet Center sabotage duration from 15 to 60 seconds. This is more in line with the scale of #2355.

MTKing4:
Can you evacuate hackers from Disabled Internet Centers? Because if you can this whole Advantage will be nullified.

ExiLe-ZH:
> Can you evacuate hackers from Disabled Internet Centers? Because if you can this whole Advantage will be nullified.

well you said everything needs to be counterable, there you go ^^

On a real note though, first of all your hackers will be exposed when evacuating, secondly you will lose hacktime from various factors, like your reaction time, the walking and animations time etc. So is it even worth evacuating?

MTKing4:
This is different, because the attacker is successful but is potentially not getting a reward or at least very little, but for example the cash steal already got their reward when they enter the building. You cannot return your $1000 after it was stolen

> On a real note though, first of all your hackers will be exposed when evacuating, secondly you will lose hacktime from various factors, like your reaction time, the walking and animations time etc. So is it even worth evacuating?

The question is, is it worth sabotaging? Because there's definitely a higher probability of near complete mitigation of the attack, making the enemy lose a small amount for relocating will not be worth the investment

That being said, it can be a nice combo tactic from the enemy, forcing hackers out with sabo then drop an anthrax bomb :D

Of course this all depends if my assumption about evacuating is true at all.

xezon:
You can double check to verify, but I am pretty certain that the sabotaged structure will be disabled and it will not be possible to click buttons, just like you won't be able to build a tank on a sabotaged Arms Dealer.

The way to counter Saboteur is by upgrading Mines or placing appropriate units nearby.

shot_20230915_084252_2

commy2:
That picture is from a Microwave tank, which behaves differently from e.g. EMP bomb.

Hackers outside the IC will also hack 16.7% slower, earning less money and experience.

What does Sabotage actually disable?
Does hacking an IC also reset / freeze / stop Sat Hack I and II?

xezon:
Realistically no one will test this, so these questions cannot be answered at this time.

ExiLe-ZH:
> The question is, is it worth sabotaging? Because there's definitely a higher probability of near complete mitigation of the attack, making the enemy lose a small amount for relocating will not be worth the investment

Exactly, it's high risk targetting the IC, because it's likely to be mined as it's a high value target. High risk should justify an high reward.

> That being said, it can be a nice combo tactic from the enemy, forcing hackers out with sabo then drop an anthrax bomb :D

Yep, or like an ambush whatever.
 
> Of course this all depends if my assumption about evacuating is true at all.

True, I got time tonight to test this.

xezon:
Anything else to add?

commy2:
I would like to know how this interacts with Sathack I & II.

xezon:
Cash gain values corrected. Edit: Reverted, because it was correct before.

xezon:
> I would like to know how this interacts with Sathack I & II.

Ok so here is what Saboteur does in Internet Center

# General, Hackers

* Disables upgrade buttons
* Pauses upgrade progress
* Pauses all hackers
* Does not disables evacuation

When evacuating Hacker(s), then Hacker(s) are disabled and cannot move. Hacker(s) can move after sabotage duration has ended.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/8da4e2a0-f0dc-404c-8824-9bf06ba31a86

# Satellite Hack I

The Satellite Hack I is currently bugged in Patch with #2416, but the Command Center reveal is disabled for the whole duration of the Internet Center sabotage.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/ddb90a8f-2c1b-4b71-8e08-5c8a34063b43

# Satellite Hack II (Patched Ability)

The Satellite Hack II map reveal is disabled during Internet Center sabotage - if it was activated prior the sabotage. During sabotage, Satellite Hack II cannot be activated. The Satellite Hack II cooldown timer will not be reset on sabotage.

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/44d0c89c-ce36-406c-9b09-cfd86a0964f9

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/ad82b908-861b-471a-b6b3-067921730052

xezon:
Full information was appended to post above. Saboteur vs Internet Center is very effective. It does not just shutdown income generation, but also disables Satellite Hack I & II spying for the given duration. With the increased sabotage duration, this advantage is increased.

Are we good with this?

xezon:
FiRe^ArM
> China's late game income is already bad. Doing it would worsen it even further.

xezon:
> China's late game income is already bad. Doing it would worsen it even further.

China Economy and Internet Center buffs were added to opening post for reference.

xezon:
Jundiyy
> 1 min does feel quite a long time, and since Hackers hack faster in the IC now it means a bigger loss than original (other than inf gen). Maybe 30 to 45 secs?

xezon:
> 1 min does feel quite a long time, and since Hackers hack faster in the IC now it means a bigger loss than original (other than inf gen). Maybe 30 to 45 secs?

As outlined above, compared to the original Supply Center steal cash amount (1000 self reward + 1000 enemy loss = 2000), the fair Internet Center sabotage duration on average is 55 seconds. Therefore even 30 seconds are still not worth it. 45 seconds are not bad, but it is possible that players would still not risk the Saboteur for this time. The amount of success also depends on the amount and veterency of Hackers inside the Internet Center. If is filled with 8x Vet0 or 8x Vet1 Hackers, then it is still more valuable to sabotage the Supply Center for cash instead.

xezon:
commy2
> I think the IC change is good.

xezon:
Documentation added. Good.

PR #2358, opened 2023-09-13
Change: Increase Factory sabotage duration of GLA Saboteur from 30000 to 45000 ms

xezon:
* Relates to #2097

This change increases the Factory sabotage duration of the GLA Saboteur from 30000 to 45000 ms. Does not apply to Power Plant, Internet Center, Command Center, Fake Building.

The Saboteur build time is 30 seconds without power. The build cost is 800. Requires an Arms Dealer to build.

### Other things that can disable for reference

| Object                   | DisabledDuration |
|--------------------------|------------------|
| EMPPulseEffectSpheroid   |  30000           |
| LeafletContainer         |  20000           |
| EMPPatriotEffectSpheroid |  10000           |

# Rationale

The GLA Saboteur is seldom used to deactivate enemy production facilitates. The new +50% sabotage duration provides more incentive to invest in and micro manage the Saboteur to try disable production facilities.

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2358_saboteur_factory_sabotage_duration.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2358_saboteur_factory_sabotage_duration.yaml
@@ -1,0 +1,21 @@
+---
+date: 2023-09-13
+
+title: Increases Factory sabotage duration of GLA Saboteur from 30 to 45 seconds
+
+changes:
+  - tweak: Increases the Factory sabotage duration of the GLA Saboteur from 30 to 45 seconds. Applies to unit production facilities only.
+
+labels:
+  - buff
+  - controversial
+  - design
+  - gla
+  - minor
+  - v1.0
+
+links:
+  - https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2358
+
+authors:
+  - xezon
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/DemoGeneral.ini
@@ -17472,9 +17472,10 @@ Object Demo_GLAInfantrySaboteur
     BuildingPickup  = Yes
     StealCashAmount = 1200
   End
+  ; Patch104p @tweak xezon 13/09/2023 Changes sabotage duration from 30000. (#2358)
   Behavior = SabotageMilitaryFactoryCrateCollide  SabotageTag_06
     BuildingPickup = Yes
-    SabotageDuration = 30000
+    SabotageDuration = 45000
   End
 
   ; Patch104p @bugfix commy2 10/10/2021 Fix this factions Saboteur cannot enter Fake buildings and Internet Centers.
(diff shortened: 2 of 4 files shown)


xezon:
I think 45 seconds is still not great, but certainly better than 30 seconds. Since Saboteur build time is 30 seconds, one simply trades 30 seconds against 45 seconds, while going through the hassle of micro-managing the Saboteur through the enemy defenses and investing 800 cash. If the opponent has a multitude of production facilities, then it will require significant investment to shutdown the production for any meaningful duration.

ExiLe-ZH:
Seems like a safe change to me, it's an improvement atleast.

Perhaps 60 seconds is also justified, not entirely sure.

Also for this one an XP reward could be considered.

xezon:
MTKing
> look good to me

xezon:
Keep the 45000, or increase to 60000, matching the Internet Center sabotage duration with #2357?

Jundiyy:
It depends, 30s is still quite long when used with an attack. It means no production whilst there are units in your base. GLA will only delay infantry production which isn't bad as RPGs are usually used from Tunnels, whereas the enemy loses vehicle production for defending. 
Maybe seeing it in action might help make a decision. 

Perhaps a better change would be to increase the Saboteur's health a little.

xezon:
commy2
> I have only two worries, but they may not be substantiated.
> 1) Countering Saboteurs required bunkering down somewhat (with in base scouts / Tunnels), which may not be the style of play that should be encouraged.
> 2) Countering Saboteurs as China requires the War Factory, which is exactly the building that is being shut down. So it could be disabled perpetually by more Saboteurs.

xezon:
> Countering Saboteurs required bunkering down somewhat (with in base scouts / Tunnels), which may not be the style of play that should be encouraged.

There is some truth. It is similar to the GLA Stealth Hijacker, where one needs to be careful to not get the Dragon Tank stolen right as it leaves the War Factory. A smart player will try to anticipate these things and put countermeasures in place, such as Mines, Outpost, Gattling Helix.

> Countering Saboteurs as China requires the War Factory, which is exactly the building that is being shut down. So it could be disabled perpetually by more Saboteurs.

Unclear to me if this is feasible. Saboteur has a build time of 30 seconds and requires the Arms Dealer. By the time it arrives at the door step of the opposition, there should have been ample time to prepare for it, either by building Mines, Outpost or Gattling Helix. If the Saboteur was even remotely efficient shutting down factories, then we would be able to see it being used in top level matches of the original game. But in reality the Saboteur is almost never used for anything other than resetting a Superweapon timer.

Also: On paper the China Infantry Hacker is overpowered because it can walk stealthed into the enemy base and disable buildings for as long as it is alive. But in practice this never happens. I have my doubts that bumping sabotage from 30s to 45s is much of an improvement for the Saboteur, because it can be stopped cheaply given a bit of preparation. Plus just because Saboteur is successful does not mean it will have much of an impact. The opposition could have 3 production facilities but not even enough cash to sustain constant production of all. There are probably some events where shutting down production will be absolutely a game changer, but for those cases I would imagine that one could attest "well played Sir" instead of claiming Saboteur is overpowered. And if the new Saboteur is exploitable, then we can easily dial back.

xezon:
Documentation added. Good.

PR #2359, opened 2023-09-13
Fix: Punctuation inconsistencies in radar strings for all languages (4)

xezon:
* Follow up for #2316
* Follow up for #2317
* Follow up for #2341

This change fixes punctuation inconsistencies in radar strings for all languages.

Affects

* RADAR:


xezon:
Documentation added. Good.

PR #2362, opened 2023-09-15
Fix: Set EXPLODED death type for weapon of USA Missile Defender

xezon:
This change sets the expected EXPLODED death type for the weapon of the USA Missile Defender. It is now consistent with weapons of Tank Hunter and RPG Trooper.

The major consequence is that the missiles will now explode the GLA Terrorists and therefore trigger their suicide weapon. It will also trigger the suicide weapon of the GLA Demo Bike before Demolitions Upgrade.

This is a fix but also nerf on USA.

## Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/9f71ba02-2f89-4265-a88b-e1a7358c885b

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2355_saboteur_steal_cash_amount.yaml
@@ -4,7 +4,7 @@ date: 2023-09-13
 title: Increases steal cash amount of GLA Saboteur from 1000 to 1200
 
 changes:
-  - tweak: Increases the steal cash amount of the GLA Saboteur from 1000 to 1200. The maximum cash gain factor against a Supply Center increases from 2.5 to 3.0.
+  - tweak: The GLA Saboteur can now steal up to 1200 cash. The maximum cash gain factor against a Supply Center increases from 2.5 to 3.0.
 
 labels:
   - buff
--- a/Patch104pZH/Design/Changes/v1.0/2357_saboteur_internet_center_sabotage_duration.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2357_saboteur_internet_center_sabotage_duration.yaml
@@ -4,7 +4,7 @@ date: 2023-09-13
 title: Increases Internet Center sabotage duration of GLA Saboteur from 15 to 60 seconds
 
 changes:
-  - tweak: Increases the Internet Center sabotage duration of the GLA Saboteur from 15 to 60 seconds. The average cash gain factor against an Internet Center with 8 Hackers increases from 0.68 to 2.72. Internet Center upgrade research and Satellite Hack I & II are paused for the same sabotage duration.
+  - tweak: The GLA Saboteur can now sabotage the China Internet Center for 60 seconds. The average cash gain factor against an Internet Center with 8 Hackers increases from 0.68 to 2.72. Internet Center upgrade research and Satellite Hack I & II are paused for the same sabotage duration.
 
 labels:
   - buff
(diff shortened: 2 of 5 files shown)


xezon:
I do wonder, will this have any other gameplay consequences?

penfriendz:
Looks good to me.

MTKing4:
Seems Fair to me

xezon:
The Demo Bikes are also affected - before Demo Upgrade.

Fixing this bug means that Demo General could spam Demo Bikes against USA before Demo Upgrade and see higher success rates.

The Humvee TOW missile and Hellfire Drone missile will also trigger the suicide weapons of Terrorists and Demo Bikes.

### Original

USA Missile Defender will **NOT** trigger suicide weapon of Demo Bike before Demo Upgrade.
USA Missile Defender will trigger suicide weapon of Demo Bike after Demo Upgrade.

China Tank Hunter will trigger suicide weapon of Demo Bike before Demo Upgrade.
China Tank Hunter will trigger suicide weapon of Demo Bike after Demo Upgrade.

GLA RPG Trooper will trigger suicide weapon of Demo Bike before Demo Upgrade.
GLA RPG Trooper will trigger suicide weapon of Demo Bike after Demo Upgrade.

### Patched

USA Missile Defender will trigger suicide weapon of Demo Bike before Demo Upgrade.
USA Missile Defender will trigger suicide weapon of Demo Bike after Demo Upgrade.

China Tank Hunter will trigger suicide weapon of Demo Bike before Demo Upgrade.
China Tank Hunter will trigger suicide weapon of Demo Bike after Demo Upgrade.

GLA RPG Trooper will trigger suicide weapon of Demo Bike before Demo Upgrade.
GLA RPG Trooper will trigger suicide weapon of Demo Bike after Demo Upgrade.

penfriendz:
Seems like biggest difference will be humvees vs demo bikes pre upgrade.

But I'm totally fine with that from a balance perspective. Now it finally makes sense!

ExiLe-ZH:
This depends on how good (pre-upgrade) dbs actually become vs vees.

commy2:
It does not change dps. death type is not the same as damage type.

penfriendz:
> This depends on how good (pre-upgrade) dbs actually become vs vees.

Not really. Humvees already kill dbs with ease. I never find upgraded dbs unfair vs USA and every time I lose a vee to them I only blame myself.
I feel sorry for China. I don't think I've ever seen their units move. At least humvees can.
There will be no difference to mid level players or above.

ExiLe-ZH:
Don't disagree thats it's usually very avoidable to get hit in the game dynamic we currently have. That's the dynamic where USA already has multiple vees before the first bike is produced, also often supported with S&D.

With this change the dynamic changes, where the first bike is ready before the first vee. This might totally change the numbers game and this is where it might get hard, having to stop big blobs without S&D.

xezon:
Yes. It will likely lead to scenarios where Demo does nothing else than spam Demo Bikes to hunt Humvees and Dozers. USA would need very good defense or put Humvee with Hellfire into Chinooks.

MTKing4:
This is good because it buffs the demo's early weakness and nerfs its lategame OPness with the crushed detonation change

xezon:
Pre-Demo Upgrade game is buffed yes. Crushed damage is also a buff, because in 1.04 Demo Terrorist can be crushed without explosion, but in Patch they will explode with weak damage.

MTKing4:
Normalizing demo Terrorist with other terrorists, (or slightly weaker version of that) doesn't read as significant of a buff compared to the significant nerf they're getting for their demo bikes, so overall it's more of a nerf

xezon:
What nerf?

MTKing4:
#2369

xezon:
#2369 was closed.

MTKing4:
I know but my point was that it would mitigate the nerf USA is getting from this change

xezon:
These changes are mutually exclusive.

MTKing4:
Oh, right, don't mind me then

xezon:
Any additional comments on this?

PR #2363, opened 2023-09-15
Fix: GLA Battle Bus wreck no longer shows salvage upgrades

xezon:
The GLA Battle Bus wreck no longer shows salvage upgrades.


xezon:

PR #2364, opened 2023-09-15
Fix: Remove destruction delay from GLA Battle Bus death

xezon:
This change removes the random destruction delay from the GLA Battle Bus. Reason being, it looks odd when the passengers die but the final death explosion effect is slightly delayed behind.


xezon:

PR #2365, opened 2023-09-16
Fix: Prevent radar dish animation reset on damage transition of GLA Radar Van, China Overlord, Listening Outpost

xezon:
This change prevents the radar dish animation to reset on a damage transition of the GLA Radar Van.


xezon:
Also on Overlord, ~~ECM~~, Outpost.

xezon:

PR #2366, opened 2023-09-16
Change: Improve death effects of GLA Quad Cannon, Radar Van

xezon:
This change improves death effects of the GLA Quad Cannon and Radar Van. The destruction delays are increased and a new initial death explosion is added.

The increased destruction delay has been added to better accommodate the Demolition Upgrade explosion. If there is too little of a gap between these, then it is more difficult to see and hear the demo explosion.

And the initial death explosion makes it easier to see when the vehicle actually dies. It has an effect that is smaller and sound that is quieter than that of tanks.


xezon:

PR #2374, opened 2023-09-30
Improved Spanish localization (WIP)

ElTioRata:
Leaving it as draft for now since there are over 6000 lines to be reviewed.


xezon:
Nice.

Be careful with increasing lengths of GUI strings. Sometimes you may need to check if a string still fits the field or button it is contained in.

For example this one, where Spanish is the longest:

```
GUI:GameName
US: "Game Name:"
DE: "Spiel:"
FR: "Partie :"
ES: "Nombre de partida:"
IT: "Nome partita:"
KO: "게임 제목:"
ZH: "遊戲名稱："
BP: "Nome do Jogo:"
PL: "Nazwa sesji:"
RU: "Название игры:"
AR: ": ﺔﺒﻌﻠﻟا ﻢﺳإ"
END
```

ElTioRata:
How do I check that?

xezon:
You can launch Spanish Patch with Mod Builder.

Run https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/Scripts/BuildInstallRunWithGui.bat, then deselect English and select Spanish in the UI. Hit Execute.

You need Spanish game install, or Spanish language pack.

I can also test once you are done.

ElTioRata:
I found out that there are several dialogs from base Generals that are reused in the Zero Hour, looks like the translators took some liberties because a bunch of subtitles don't really match the lines of the Spanish audios from the base game. A fix for this may come in separate future PR because I'd have to listen to all of the audios and look for the subtitles that refer them (currently I have no idea of how to look for those references). This could also be the case for other languages that didn't received further dubbing work on ZH.

xezon:
Note that strings starting with asterisks (*) are hidden in game.

ElTioRata:
No worries, I will translate them either way if someone wants to re-enable them in the future.

xezon:
@ElTioRata How long do you think this Spanish update will take you?

I am asking because when I am done with the texture polishing I would like to start merging Ukrainian into the generals.str. This will be a big change and perhaps it will be better to have your change in before that, to avoid potential complications when merging, although it should technically be alright.

ElTioRata:
You can merge the changes now if you want to add other localizations. I'll submit further changes in another PR.

xezon:
It will still take a while. I will bump again when the time comes.

PR #2375, opened 2023-10-08
Fix: China TNT Bombs now hit structures with their primary damage reliably

### LINKED ISSUE #2373: TNTStickyBomb is placed too far away from buildings and applies secondary damage most of the time
* Perhaps related to #954

TNTStickyBomb applies different amount of damages depending on count and timing.

https://www.youtube.com/watch?v=hmqaU-HF9uU

```
Object TNTStickyBomb
  ;This determines how long the bomb lives before it explodes (via SlowDeathBehavior)
  Behavior = LifetimeUpdate ModuleTag_02
    MinLifetime = 10000   ; min lifetime in msec
    MaxLifetime = 10000   ; max lifetime in msec
  End

  Behavior = SlowDeathBehavior ModuleTag_04
    DestructionDelay    = 1
    FX                  = INITIAL WeaponFX_TNTStickyBombDetonation
    Weapon              = FINAL   TNTDetonationWeapon
  End

  Body = HighlanderBody ModuleTag_NewBody
    MaxHealth = 1.0
    InitialHealth = 1.0
  End

  ...
End
```

```
Weapon TNTDetonationWeapon ;Created by tankhunters
  PrimaryDamage = 500.0
  PrimaryDamageRadius = 10.0
  SecondaryDamage = 150.0
  SecondaryDamageRadius = 50.0
  AttackRange = 5.0
  DamageType = EXPLOSION
  DeathType = EXPLODED
  WeaponSpeed = 99999.0
  ProjectileObject = NONE
  RadiusDamageAffects = SELF ALLIES ENEMIES NEUTRALS
  DelayBetweenShots = 0                   ; time between shots, msec
  ClipSize = 1                        ; how many shots in a Clip (0 == infinite)
  ClipReloadTime = 0               ; how long to reload a Clip, msec
  AutoReloadsClip = No
  FireSound = BombTruckDefaultBombDetonation
End
```

xezon:
* Fixes #2373

With this change the primary damage of the China TNT Bomb can now hit structures reliably. Before they would hit with their secondary damage most of the time, because the TNT Bombs would be located too far away from the target structure.

Placing the TNT Bomb on an approaching vehicle still works before the instigator is crushed. Damage on vehicles is unchanged, because the TNT Bomb is placed at the center of the vehicle, meaning it will apply primary damage always.

The reduced ability range requires Tank Hunters to walk closer to their target before the TNT bomb is placed. The increased primary damage radius gives the TNT bomb more reach.

## Damages

The damage surplus of a correctly placed TNT Bomb is significant. Buildings will now suffer +233% explosion damage from a single TNT Bomb.

```
Weapon TNTDetonationWeapon ;Created by tankhunters
  PrimaryDamage = 500.0
  PrimaryDamageRadius = 15.0
  SecondaryDamage = 150.0
  SecondaryDamageRadius = 50.0
...
End
```

## Original

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/e295d28f-18ce-407d-9475-dd272375b67e

## Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/1a4271a7-2f32-4e51-8dbc-bf82a5f9fc9b

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/BossGeneral.ini
@@ -18950,9 +18950,11 @@ Object Boss_InfantryTankHunter
     UpdateModuleStartsAttack = Yes
     InitiateSound = TankHunterVoiceTNT
   End
+
+  ; Patch104p @bugfix xezon 08/10/2023 Changes start ability range from 5.0 to place the bomb closer to structures so that the primary damage can hit it more reliably. (#2375)
   Behavior = SpecialAbilityUpdate ModuleTag_08
     SpecialPowerTemplate = SpecialAbilityTankHunterTNTAttack
-    StartAbilityRange = 5.0
+    StartAbilityRange = 0.0
     PreparationTime = 0
     SpecialObject = TNTStickyBomb
     MaxSpecialObjects = 8
--- a/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Object/ChinaInfantry.ini
@@ -1164,9 +1164,11 @@ Object ChinaInfantryTankHunter
     UpdateModuleStartsAttack = Yes
     InitiateSound = TankHunterVoiceTNT
   End
+
+  ; Patch104p @bugfix xezon 08/10/2023 Changes start ability range from 5.0 to place the bomb closer to structures so that the primary damage can hit it more reliably. (#2375)
   Behavior = SpecialAbilityUpdate ModuleTag_08
     SpecialPowerTemplate = SpecialAbilityTankHunterTNTAttack
-    StartAbilityRange = 5.0
+    StartAbilityRange = 0.0
     PreparationTime = 0
     SpecialObject = TNTStickyBomb
     MaxSpecialObjects = 8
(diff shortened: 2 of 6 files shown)


xezon:
This is a massive buff for all China.

penfriendz:
This is one of those buff not really buff kinda things. For people who knew how to take advantage of this, this is a massive convenience and increase in consistency. For people that don't, they can now do what only people with insider knowledge could do, and for these people the impact is greatest.

In any case, let's get a table of bombs to kill for each building and pick a damage value which is sensible.

xezon:
I have now also increased the PrimaryDamageRadius from 10 to 15, which helps to apply primary damage in some edge cases.

MTKing4:
> Placing the TNT Bomb on an approaching vehicle still works before the instigator is crushed. Damage on vehicles is unchanged, because the TNT Bomb is placed at the center of the vehicle, meaning it will apply primary damage always

I Assume this is not tested thoroughly, it needs to

Besides the big damage buff to buildings, i predict it to buff vs clumped units as well, since now the range is higher, you might kill 2 or 3 units with one TNT

xezon:
> I Assume this is not tested thoroughly, it needs to

I think I tested it against AI, which always goes for crush against infantry and it worked. But we can retest.

> Besides the big damage buff to buildings, i predict it to buff vs clumped units as well, since now the range is higher, you might kill 2 or 3 units with one TNT

I think I tested this too and natural distance between vehicles is wider. But we can restest. It would be useful to find out what the regular min distance between vehicles is (there is one to avoid collision).

PR #2376, opened 2023-10-08
Change: Decreases primary damage of China TNT Bomb from 500 to 325

xezon:
* Follow up for #2375
* Relates to #2373

This change decreases the primary damage of China TNT Bomb from 500 to 325. This makes it weaker against both vehicles and buildings, however, the bug fix of #2375 makes the TNT Bomb much better against structures.

Why 325 damage? Because 150 + ((500 - 150) / 2) = 350

Compared to Original, practical damage vs buildings is +175 and damage vs vehicles is -175. Damage output of TNT Bomb is essentially redistributed from vehicles to buildings. This makes the TNT Bomb equally useful against structures and vehicles.

Vehicles now have better chances to survive the TNT Bomb.

## Practical difference

| Weapon                   | Effective Damage vs vehicles | Effective Damage vs buildings |
|--------------------------|------------------------------|-------------------------------|
| Original TNT Bomb        | 500                          | 150                           |
| Patched TNT Bomb (#2375) | 500                          | 500                           |
| Patched TNT Bomb (this)  | 325                          | 325                           |

## TNT Bombs vs buildings

It now takes 7 patched TNT Bombs to kill a strong structure such as a strong house or Oil Derrick.

| Weapon                   | Count to kill Oil Derrick |
|--------------------------|---------------------------|
| Original TNT Bomb        | 14                        |
| Patched TNT Bomb (#2375) | 4                         |
| Patched TNT Bomb (this)  | 7                         |

## TNT Bombs vs vehicles

| Patched TNT Bomb target      | Damage      |
|------------------------------|-------------|
| USA Dozer                    | killed      |
| USA Humvee (2 Star)          | killed      |
| USA Humvee (3 Star)          | 95% damage  |
| USA Tomahawk (3 Star)        | killed      |
| USA Crusader                 | 70% damage  |
| GLA Technical (3 Star)       | killed      |
| GLA Scorpion                 | 90% damage  |
| GLA Quad Cannon              | killed      |
| GLA Quad Cannon (1 Star)     | 95% damage  |
| China Overlord               | 30% damage  |
| China Dragon Tank            | killed      |
| China Dragon Tank (1 Star)   | 99% damage  |
| China Truck                  | killed      |
| China ECM Tank               | killed      |
| China Gattling Tank          | killed      |
| China Gattling Tank (1 Star) | 95% damage  |
| China Battle Master          | 90% damage  |

## TNT Bomb damage compared to other bombs

| Weapon                  | Primary Damage | Secondary Damage | PrimaryDamageRadius | SecondaryDamage |
|-------------------------|----------------|------------------|---------------------|-----------------|
| Original C4 Charge      | 2000           | 150              | 25                  | 75              |
| SuicideBikeBomb         | 700            | 100              | 20                  | 50              |
| Original Demo Trap      | 600            | 400              | 25                  | 50              |
| SuicideDynamitePack     | 500            | 300              | 18                  | 50              |
| Original TNT Bomb       | 500            | 150              | 10                  | 50              |
| Patched TNT Bomb (this) | 325            | 150              | 10                  | 50              |
| SuicideBomb             | 300            | 60               | 10                  | 50              |
| Original Booby Trap     | 200            | 50               | 5                   | 15              |
| Original Fake Structure | 200            | 50               | 50                  | 200             |
| CrushedDynamitePack (Patch) | 110            | 66               | 18                  | 50              |
| Demo_DestroyedWeapon    | 50             | 10               | 60                  | 70              |

## And some other weapons for reference

| Weapon                  | Primary Damage | Secondary Damage | PrimaryDamageRadius | SecondaryDamage |
|-------------------------|----------------|------------------|---------------------|-----------------|
| AuroraBombWeapon        | 400            | 0                | 20                  | 0               |
| ChinaCarpetBombWeapon   | 300            | 0                | 75                  | 0               |
| CarpetBombWeapon        | 300            | 0                | 50                  | 0               |
| NapalmBombWeapon        | 75             | 40               | 5                   | 30              |
| HumveeMissileWeapon     | 30             | 0                | 5                   | 0               |

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Weapon.ini
@@ -2638,9 +2638,10 @@ Weapon ChinaInfantryTankHunterMissileLauncher
   ProjectileCollidesWith = STRUCTURES
 End
 
+; Patch104p @tweak xezon 08/10/2023 Changes primary damage from 500.0 to reduce bomb effectiveness against buildings after bug fix.
 ;------------------------------------------------------------------------------
 Weapon TNTDetonationWeapon ;Created by tankhunters
-  PrimaryDamage = 500.0
+  PrimaryDamage = 325.0
   PrimaryDamageRadius = 10.0
   SecondaryDamage = 150.0
   SecondaryDamageRadius = 50.0


xezon:
This redesign appears to strike an interesting balance. Together with the bug fix, TNT Bombs become more useful against buildings. They are still strong against vehicles, but no longer a safe kill against some vehicles.

It creates an interesting dynamic where promoted tanks (Dragon Tank, Gattling Tank) from China Tank General will now be able to survive TNT Bomb attacks.

ExiLe-ZH:
I'm not sure I like the nerf against vehicles tbh.

MTKing4:
I don't like where this is going, agree with Exile, we don't want to nerf it vs Vehicles

xezon:
The problem is that it is just too strong versus buildings with the fix.

MTKing4:
Maybe find the lowest number that can still kill all vehicles that is lower than 500 so that it is a nerf to building attacks only

xezon:
Which vehicle should die to TNT Bomb? All vehicles is impossible, because Overlord will not die to TNT Bomb unless we raise damage.

ExiLe-ZH:
it seems too strong vs buildings indeed,perhaps design new weapon/armor parameter?

xezon:
Not possible to have vehicle and building damages.

MTKing4:
> Which vehicle should die to TNT Bomb? All vehicles is impossible, because Overlord will not die to TNT Bomb unless we raise damage.

All Vehicles that can die to TNTs in 1.04, so yeah that execludes the Overlord/Emperor

xezon:
Which unit would that be? 3 Star Paladin with Composite Armor?

MTKing4:
I'm not sure if that died to a TNT before, also we gave it the double composite armour didn't we?

Float1ngFree:
How about canceling https://github.com/TheSuperHackers/GeneralsGamePatch/pull/2375 and taking advantage of secondary damage instead?

You can decrease the primary radius to 1-5, or 0, to assure that it hits vehicle only, while buffing the secondary damage to say 200 and that will be applied to structures. Obviously, the secondary damage radius should be then reduced as well, 50 is too much for this IMO.

xezon:
> You can decrease the primary radius to 1-5, or 0, to assure that it hits vehicle only, while buffing the secondary damage to say 200 and that will be applied to structures. Obviously, the secondary damage radius should be then reduced as well, 50 is too much for this IMO.

It is a reasonable suggestion. With such a low primary damage radius, would the TNT certainly be unable to hit the structure, even if the Tank Hunter would manage to place it while standing inside the structure somehow?

Decreasing the secondary damage range would not be ideal. Also it means that the TNT charge will be a definite buff, no disadvantages, perhaps too strong?

> Maybe find the lowest number that can still kill all vehicles that is lower than 500 so that it is a nerf to building attacks only

To find this value we would need to find the strongest tank that is not the Overlord. Though looking at Crusader and 70% damage, means 350 * (1/0.7) = 500. So it could not be lower than 500 at all then.

If we want to fix #2375, then I suggest find a reasonable TNT primary damage value between 325 and 500. In my opinion 350 is very generous. It still is an overall buff to Tank Hunters in combination with #2375, because 350-150=200 and 350-500=-150. +50 damage difference in favor of buildings.

A fundamental question here is: Are you willing to trade vehicle damage for building damage? If not, then we cannot have fix #2375.

Is the status quo TNT charge perfect vs vehicles? Not too strong, not too weak?

Float1ngFree:
> It is a reasonable suggestion. With such a low primary damage radius, would the TNT certainly be unable to hit the structure, even if the Tank Hunter would manage to place it while standing inside the structure somehow?

Dunno, needs testing. Definitely works as I've described for "walk-and-plant" use though. I think that, as I've described in Aurora issue, the 'PrimaryDamageRadius = 0.0' would prevent applying main damage to building because of how game handles 0.0 radius. Just change a few lines and give it a try.

> Decreasing the secondary damage range would not be ideal. Also it means that the TNT charge will be a definite buff, no disadvantages, perhaps too strong?

Imo, 50 damage radius is too much and not logical: some big explosions, like Aurora bomb, have half of that! It is as big as US carpet bomb, think about that... 25-30 secondary radius would be about right. After all, the main purpose of the skill is the single target "I'm about to get crushed so F/U" damage.

The disadvantage is what it always was - having to walk in. This buff won't make it much different: same single- target damage, while secondary damage is mostly for buildings. If player can't isolate a unit with bomb out of the group, especially with lower damage radius - that's 100% skill/attention issue.

xezon:
Ok that is a fair point with damage range comparison. Could you perhaps collect damage ranges of other comparable bombs, such as C4, so we can see how it fits in?

Edit: Nevermind I can do it and add to opening post.
Edit: Done

To me it looks like bomb damage ranges are all over the place. There is no system, which means we can probably do as we please. I am ok with trying to design the secondary damage for buildings, primary damage for vehicles - if it works well. But we need to be sure that we really only want to buff Tank Hunters then. Infantry General, one of the best factions, will benefit the most.

Float1ngFree:
> Infantry General, one of the best factions, will benefit the most.

That is a part of bigger problem - rocket infantry being too op overall. Just too much DPS per $ and then you have all the extra bonuses for garrisoned, horde, prop, etc... at the same time taking no damage from shells and outranging AA vehicles. But that's another topic.

As per my suggestion, increasing secondary damage from 150 to 200 won't be too much of a buff but a reminder and encouragement for people to think about it more when they have their remaining Tank Hunters close to enemy's buildings. In most cases, just attacking would yield more DPS than time spent planting a bomb, unless it is a structure that has high armor vs `INFANTRY_MISSILE` damage type, but even then, walking in and out just takes too long... So probably only if you're close already or about to die.

Float1ngFree:
@xezon Tested with TH enter & exist neutral building with quick "stop" command to remain inside of the structure: 

- The 0.0 radius doesn't work if TH is "inside" the building -- the structure receives primary damage.
- Otherwise, TH attempts to attach the bomb at the very edge of the object radius and that applied secondary damage only.

Sad, but on the other hand it quite makes sense - the closer you are to the building, the more time you spend on better placement of the charge - the more damage it should do. Just that the basic damage spread is too wide.

This is how I have it in my project:
image

- 360 is my hp for Scorpion and that's also vet 1 quad/gatt. 3 bombs leave Overlord with 20 hp - one missile.
  - 500 always seemed too much for me, especially given that the ability comes at no cost. 1000 dmg in just 2 bombs, even for units only - is OP.
  - I previously had it at 400 (battlemaster, vet 2 quad/gatt), but with test above I thought that's a bit too much & too lazy (put in some rocket damage guys!)
- 240 secondary makes it 2/3 of primary damage. Radius reduced from 50 to 30.
  - 150 was too low and too wide (50 radius = 100 diameter - that's ranger's attack range). Just confusing and inconsistent.
  - I previously had it at 200, but a 50% damage drop depending on bomb placement seems too harsh, so both changes bring it closer together for more logical output.

Float1ngFree:
Here are some screenshots of my damage setups in action:

360 primary damage on China Barracks (1000 hp) - altered command to be able to use ability on ally objects for testing. Stopped TH exiting barracks to plant bomb "inside" of the object:
image

240 secondary when planting bomb from the outside:
image

Placing a TNT on a vehicle puts it on top of it, **regardless of vehicle size OR TH approach angle:**
image
image

360 primary damage to Overlord (1100 hp) and BM (400 hp):
image
image

PR #2395, opened 2024-05-30
Fix: cbapt01 night textures

xezon:
This change further improves Civilian cbapt01 night textures.

### cbapt01_n, cbapt01_ng, cbapt01_dn, cbapt01_dng

* Recreates night lights with better visuals


xezon:
Looks ok.

PR #2397, opened 2024-05-30
Remove all test maps from this repository

xezon:
This change removes all test maps from this repository.

The test maps now live here:
https://github.com/TheSuperHackers/GeneralsGameData/tree/main/Maps


xezon:
On second thought, perhaps test maps should be kept tied to project, because if the game content evolves, then some or all maps may only be compatible with this project at some point.

PR #2398, opened 2024-05-30
Fix: cbbaikgrnd_e texture

xezon:
This change fixes the Civilian cbbaikgrnd_e texture.

### cbbaikgrnd_e

* Fixes minor window damage inconsistencies with cbbaikgrnd_d


xezon:
Looks good.

PR #2417, opened 2024-06-09
Fix: Satellite Hack I now reveals new Command Centers reliably

### LINKED ISSUE #2416: Satellite Hack I does not work reliably
Satellite Hack I does not work reliably. It turns on and off at random on a new Command Center build after Sat Hack I research - both during construction and after construction of the Command Center.

Somehow we broke it.

* Introduced with #791

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/2066621f-cb00-4b11-8067-443589b11d12

xezon:
* Fixes #2416
* Follow up for #791


xezon:
commy2
> Looks good to me.

PR #2424, opened 2024-06-13
Fix: Wrong USA Humvee TOW Missile object when shooting at airborne units

### LINKED ISSUE #2421: USA Humvee TOW Missile looks different when shooting at airborne targets
USA Humvee TOW Missile looks different when shooting at airborne targets.

xezon:
* Fixes #2421

This change fixes the wrong USA Humvee TOW Missile object when shooting at airborne units. Originally it used the Patriot missile object instead of the Humvee missile object. The missile behavior is unchanged.

## Patched

https://github.com/TheSuperHackers/GeneralsGamePatch/assets/4720891/7705cf54-62c6-4cdb-ad32-30c829b8cf12

### DIFF
--- a/Patch104pZH/Design/Changes/v1.0/2424_tow_missile_air_object.yaml
+++ b/Patch104pZH/Design/Changes/v1.0/2424_tow_missile_air_object.yaml
@@ -1,0 +1,19 @@
+---
+date: 2024-06-13
+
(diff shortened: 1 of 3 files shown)


commy2 (approved):
lgtm
HumveeMissileAir is new and a carbon copy of PatriotMissile, but with the model and particle effects of the regular HumveeMissile.

PR #2454, opened 2024-07-22
Fix: cbgraybumpcement textures

xezon:
This change fixes Civilian cbgraybumpcement textures.

### cbgraybumpcement, cbgraybumpcement_d, cbgraybumpcement_e

* Removes superfluous black alpha channel


xezon:
Good.

PR #2513, opened 2024-09-25
Add developer comments to generals multi str file

xezon:
This change adds and polishes developer comments to the generals str file. The source of the file is unknown, but the comments are from the original developers.


xezon:
Looks ok to me.

PR #2516, opened 2024-09-27
Change key mapping format for Ukrainian language texts

xezon:
* Follow up for #2141

This change adjusts the key mapping format for Ukrainian language texts like we did for the other languages.


xezon:
Generated with python script. Should be good.

PR #2539, opened 2024-11-06
Italian localization fixes and improvements

xezon:
Italian localization fixes and improvements

Authored by dmgreeny


xezon (commented):
This needs to be reviewed by an Italian native.

PR #2574, opened 2024-12-31
Fix: Added a blocking layer to prevent interaction with the game while in the exit menu

### LINKED ISSUE #2548: Control Bar can be used in exit menu
Control Bar can be used in exit menu.

https://github.com/user-attachments/assets/37ca3ab9-4d16-4af9-a909-28ed56800203

DevGeniusCode:
* Closes #2548

An additional layer was added to both types of exit menus that blocks the entire screen with100% transparency, preventing any interaction with the game behind it. The layer functions as expected, and the menu buttons are still working as intended. Upon returning to the game, the blocking layer is removed.

**Layer Hierarchy**
Before:
image

After:
image


Patch104pZH/GameFilesEdited/Window/Menus/QuitNoSave.wnd
```diff
@@ -6,8 +6,8 @@ STARTLAYOUTBLOCK
 ENDLAYOUTBLOCK
 WINDOW
   WINDOWTYPE = USER;
-  SCREENRECT = UPPERLEFT: 252 114,
-               BOTTOMRIGHT: 547 393,
+  SCREENRECT = UPPERLEFT: 0 0,
+               BOTTOMRIGHT: 800 599,
```
xezon, line 10: This should not be 600 ?

Patch104pZH/GameFilesEdited/Window/Menus/QuitMenu.wnd
```diff
@@ -22,7 +22,7 @@ WINDOW
   TEXTCOLOR = ENABLED:  255 255 255 255, ENABLEDBORDER:  255 255 255 255,
               DISABLED: 255 255 255 255, DISABLEDBORDER: 255 255 255 255,
               HILITE:   255 255 255 255, HILITEBORDER:   255 255 255 255;
-  ENABLEDDRAWDATA = IMAGE: NoImage, COLOR: 2 2 2 190, BORDERCOLOR: 47 55 168 255,
+  ENABLEDDRAWDATA = IMAGE: NoImage, COLOR: 2 2 2 50, BORDERCOLOR: 0 0 0 0,
```
xezon: What does this color change do?

Patch104pZH/GameFilesEdited/Window/Menus/QuitMenu.wnd
```diff
@@ -318,23 +78,311 @@ WINDOW
...
     DISABLEDDRAWDATA = IMAGE: NoImage, COLOR: 0 0 64 255, BORDERCOLOR: 0 0 0 255,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
-                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0;
+                        IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
```
xezon: Can revert the whitespace changes here.

xezon (commented):
Is looks like the files have a number of unnecessary whitespace changes that make the diff very large.

What is the name of the element with 50% transparency?

This change is missing a yaml file in Patch104pZH\Design\Changes\v1.0

Patch104pZH/GameFilesEdited/Window/Menus/QuitNoSave.wnd
DevGeniusCode: Originally it was `799 599` (from 0 to 799 there are 800 pixels), but there was still space at the edges of the screen, so I changed it to 800.

Patch104pZH/GameFilesEdited/Window/Menus/QuitMenu.wnd
DevGeniusCode: The whitespace created because I indented all the objects after adding a new container (the new layer).

Patch104pZH/GameFilesEdited/Window/Menus/QuitMenu.wnd
DevGeniusCode: I didn't change the color, it's the color of the menu window frame, the git compare tool compares the empty color of the new layer with the color of the menu window which remains the same later in the file.

DevGeniusCode:
> What is the name of the element with 50% transparency?

There is no name, this is the first WINDOW, (That's alpha 50, out of 255 that's 19.6 percent.)



> This change is missing a yaml file in Patch104pZH\Design\Changes\v1.0

Done.

xezon (commented):
Visually this does not look quite right.

The whole screen flashes white shortly, instead of just the menu window.

I recommend to make it flash on the window only.

Patch104pZH/GameFilesEdited/Window/Menus/QuitMenu.wnd
xezon: Ok, but the indentation looks wrong now. See:

Original:
```
    DISABLEDDRAWDATA = IMAGE: Buttons-Disabled-Left, COLOR: 128 128 128 255, BORDERCOLOR: 192 192 192 255,
                       IMAGE: NoImage, COLOR: 192 192 192 255, BORDERCOLOR: 128 128 128 255,
                       IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
```

This change:


```
      DISABLEDDRAWDATA = IMAGE: Buttons-Disabled-Left, COLOR: 128 128 128 255, BORDERCOLOR: 192 192 192 255,
                        IMAGE: NoImage, COLOR: 192 192 192 255, BORDERCOLOR: 128 128 128 255,
                        IMAGE: NoImage, COLOR: 255 255 255 0, BORDERCOLOR: 255 255 255 0,
```

DevGeniusCode: You're right, I didn't notice that, sorry, fixed now.

DevGeniusCode:
> Visually this does not look quite right.
> 
> The whole screen flashes white shortly, instead of just the menu window.
> 
> I recommend to make it flash on the window only.

After testing under different lighting conditions, I've noticed screen flickering. I have tried various methods to disable it, but I haven't been able to. I suspect it may be determined by the game itself.

(That is, flash the first WINDOW, and since the layer I added is the first, it now flashes across the entire screen because that is the size of the layer. If we make it the second layer (and the dark background of the menu will be the first layer as it was originally) this will cause a bug where it (the added second layer) will remain on the screen and not disappear when exiting the menu)

xezon:
This flashing is some kind of feature that originally is located on the Exit Menu bounds. This now covers the whole screen. It would be ideal to leave the flashing where it was originally. The new visual is quite obvious and may be considered distracting.

DevGeniusCode:
The flashing is indeed as you described, but it always occurs on the first layer. Originally, the first layer was the size of the menu, not covering the entire screen, which is why the screen didn’t flash.

Now, the menu has been moved to the second layer, and the first layer has been resized to cover the whole screen. As a result, the flashing now covers the entire screen, since that's the current size of the first layer.

I tried swapping the layers so that the menu would remain on the first layer (and the flashing would be limited to the size of the menu), with the blocking layer being the second. However, this causes a bug where, after exiting the menu, the blocking layer remains on the screen.

I suggest leaving it as a draft for now until we can find another way to address the issue.

xezon:
Perhaps a solution is to make that Input Block layer with STATUS = SEE_THRU or HIDDEN

DevGeniusCode:
I noticed the Window transitions file and was able to fix the flash using it.

Patch104pZH/GameFilesEdited/Data/INI/WindowTransitions.ini
```diff
@@ -2258,10 +2258,16 @@ WindowTransition QuitNoSave
...
     WinName = QuitNoSave.wnd:QuitMenuParent
-    Style   = FLASH
+    Style   = WINFADE
     FrameDelay = 0
   END
 
+  Window
+    WinName = QuitNoSave.wnd:MenuBox
+    Style   = FLASH
+    FrameDelay = 1
```
xezon: What does this frame delay do? Before it was 0.

Patch104pZH/GameFilesEdited/Data/INI/WindowTransitions.ini
```diff
@@ -2330,10 +2343,16 @@ WindowTransition QuitFull
...
     WinName = QuitMenu.wnd:QuitMenuParent
-    Style   = FLASH
+    Style   = WINFADE
     FrameDelay = 0
   END
 
+  Window
+    WinName = QuitMenu.wnd:MenuBox
+    Style   = FLASH
+    FrameDelay = 1
```
xezon: What does this frame delay do? Before it was 0.

Patch104pZH/Design/Changes/v1.0/2574_add_blocking_layer_exit_menu.yaml
```diff
@@ -0,0 +1,19 @@
+---
+date: 2024-12-31
+
+title: Added a blocking layer to prevent interaction with the control bar while in the exit menu
+
+changes:
+  - fix: Adds a blocking layer with 19.6% transparency to prevent interaction with the game while in the exit menu.
```
xezon: Is it really necessary to add transparency to the whole background? Can't we just have the blocking UI without any change in visuals? It looks unusual the way the black kicks in while the menu slowly spawns. I think it would be better to leave the original look and just have the input blocked outside the menu.

Patch104pZH/GameFilesEdited/Data/INI/WindowTransitions.ini
DevGeniusCode: The delay defines the order in which the elements appear, and I defined that the menu will appear one delay after the blocking screen. You can see in the buttons that the delay increases from button to button, and thus the buttons appear one after the other and not all at once.
Now that we have made the blocking element completely transparent, we can return it to 0.

DevGeniusCode:
Update blocking layer with 100% transparency.

(You can see for example that opening the options menu from within the game or the exit dialog adds a partially transparent background to the game, just for context.)

PR #2575, opened 2024-12-31
add `-nologo` to disable gentool splash screen

DevGeniusCode:
Add `-nologo` to disable GenTool splash screen on startup for a faster launch experience.


xezon:
It does not make the launch faster, just hides the graphics.

xezon (approved):

Issue #2377, opened 2023-10-08
Some buildings disappear when damaged or garrisoned


xezon:
See videos in TheSuperHackers/GeneralsRankedMaps#378

Looks like we need to make a pass on building model setups. I assume INI setup issues.

[Arctic Arena ZH v5.zip](https://github.com/TheSuperHackers/GeneralsGamePatch/files/12841138/Arctic.Arena.ZH.v5.zip)

xezon:
We can write a Python script that reads all W3DModelDraw module definitions and checks if any model names are misspelled or missing by comparing against the actual available w3d file names.

PR #2577, opened 2025-01-07
Add Models/Textures/Mapped Images Validation Checks

DevGeniusCode:
**This PR introduces comprehensive validation checks for graphical assets, including mapped images, textures, models, and language-specific assets. It also identifies unused assets and ensures that references are correctly accounted for, reporting any missing, duplicate, or invalid items. The changes include:**

1.  **Mapped Image Parsing:**
    *   Scans `.ini` mapped files for texture and image references, ensuring that all referenced images are validated and duplicates are reported.

2.  **Missing Texture/Model Files (General):**
    *   Compares the list of texture and model files from the `.ini` files against a provided CSV file and reports any missing texture or model files.

3.  **Unused Graphical Assets (General):**
    *   Identifies texture, model, and image files present in the file system but without corresponding references in any `.ini`, `.wnd`, or `.w3d` file. These unused assets are reported, indicating potential dead assets.

4.  **Invalid Images in WND Files:**
    *   Scans `.wnd` files for references to images. Missing images that are referenced in WND files but not present in the mapped images are reported.

5.  **Invalid Textures in W3D Files:**
    *   Checks W3D files for invalid textures (textures referenced in W3D files but missing from the texture files list).

6.  **Language-Specific Asset Analysis:**
    *   Analyze assets within language-specific folders.
    *   Identifies unused or invalid textures and models specific to each language based on the game's global INI files and language-specific asset lists.
    *   Generates individual text reports for each language, detailing unused or invalid assets.

7.  **Interactive Main Menu:**
    *   Main script with a user-interactive dialog. Users can choose which function to execute.

8.  **Output Files:**
    *   Generates several output files, including:
        *   `mapped_images_list.txt`: A list of all image mappings.
        *   `duplicate_mapped_images.txt`: A list of duplicate image mappings.
        *   `invalid_textures_in_ini_files.txt`: A list of invalid texture files found in INI files
        *   `invalid_mapped_images_in_ini_files.txt`: A list of invalid mapped images found in INI files
        *    `invalid_models_in_ini_files.txt`: A list of invalid model files found in INI files
        *   `invalid_textures_in_mapped_images_folder.txt`: A list of invalid textures found in the mapped images folder
        *    `invalid_mapped_images_in_wnd_files.txt`: A list of invalid mapped images found in WND files.
        *   `unused_textures.txt`: A list of unused texture files.
        *   `unused_models.txt`: A list of unused model files.
        *   `unused_mapped_images.txt`: A list of unused image files.
        *   `invalid_textures_in_w3d_files.txt`: A list of invalid textures found in W3D files.
        *   Individual files for each language located in /languages/ folder.

9.  **Configurable Scanning Options:**
    *   **Scan Original or Edited Files:** Allows users to select whether to scan the original or edited files. In the case of edited files, the script includes also the original files for the validation check.

10. **Documentation:**
     *  Results of research on relevant asset locations are detailed in `Graphical_Assets_Location.md`. If there are any additional locations or patterns not currently documented in that file, please let me know.


Patch104pZH/Design/Scripts/textures/get_maped_textures_info.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def main():
+    # Select whether to scan the original or edited files.
+    is_original = True
+
+    if is_original:
+        ini_folder_path = r'../../../GameFilesOriginalZH/Data/INI/MappedImages'
+        wnd_folder_path = '../../../GameFilesOriginalZH/Window'
+        txtur_folder_path = None
```
xezon: Prefer normal words over abbreviations.

Patch104pZH/Design/Scripts/textures/get_maped_textures_info.py
xezon, line 1: File name has spelling error.

Patch104pZH/Design/Scripts/textures/generated_original/duplicate_mapped_textures.txt
```diff
@@ -0,0 +1,43 @@
+HandCreatedMappedImages.INI LoadPageHuge
```
xezon: What does duplicate mean exactly? Duplicated MappedImage or multiple MappedImage using the same texture?

DevGeniusCode: Refers to duplicated MappedImage textures name, usually the same texture name in different `ini` files. See for example `SSRadarVanScan`, appearing twice in `SUUserInterface512.INI` file

xezon: Ok can you find out which definitions take precedence? We should then remove (or comment out) all shadowed and therefore superfluous definitions.

DevGeniusCode: Of course, right now I'm just creating the scripts to investigate all the files and their contents :)

DevGeniusCode:
Optimized the script and added a check for invalid textures in W3D files. 
Additionally, I added the `get_textures` function in `w3dfilemanager.py` to extract the textures from W3D files.

Patch104pZH/Design/Scripts/w3d/w3dfilemanager.py
```diff
@@ -34,6 +35,17 @@ def get_or_create_w3d_file(self, file_path: str) -> W3dFile:
         return w3dfile
 
 
+    def get_textures(self, file_path: str) -> list[str]:
+        w3dfile: W3dFile = self.get_or_create_w3d_file(file_path)
+        texture_pattern = re.compile(r'([a-zA-Z0-9_\-]+)\.(tga|dds)', re.IGNORECASE)
+        textures = set()
+
+        for match in texture_pattern.finditer(w3dfile.data.decode('utf-8', errors='ignore')):
```
xezon: utf-8 seems unlikely. Perhaps just use ascii.

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
```diff
@@ -0,0 +1,147 @@
...
+                        texture_and_file.append((tga, texture))
+                        tga_files.add(tga)
+                        if texture not in textures:
+                            textures.add(texture)
+                        else:
+                            duplicate_textures.append((filename, texture))
+
+    return texture_and_file, list(textures), list(tga_files), duplicate_textures
+
+def get_csv_files(csv_file_path, texture_folder_path=None):
```
xezon: I suggest rename this to `get_textures_from_csv`.

Then create another function `get_textures_from_folder`

And then merge their results higher up. That makes the code cleaner.

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
```diff
@@ -0,0 +1,147 @@
...
+def read_file_content(file_path):
+    """Helper function to read file content once and return it."""
+    try:
+        with open(file_path, 'r') as file:
+            return file.read()
+    except FileNotFoundError:
+        print(f"Error: The file {file_path} was not found.")
+        return None
+
+def get_tga_and_textures(folder_path):
```
xezon: This is a confusing name because tga are textures.

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
```diff
@@ -0,0 +1,147 @@
...
+    texture_and_file = []  # List of (tga, texture)
+    textures = set()  # Set of unique textures
+    tga_files = set()  # Set of unique tga files
```
xezon: Why are textures and tga's treated separately? Can't we just use one "textures" for all?

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
```diff
@@ -0,0 +1,147 @@
...
+                    matches = re.findall(r"MappedImage (\S+)\s*Texture\s*=\s*(\S+)", content)
+                    for texture, tga in matches:
+                        texture_and_file.append((tga, texture))
+                        tga_files.add(tga)
+                        if texture not in textures:
+                            textures.add(texture)
+                        else:
+                            duplicate_textures.append((filename, texture))
+
+    return texture_and_file, list(textures), list(tga_files), duplicate_textures
```
xezon: This is quite complicated. Perhaps return a class instead of this tuple.

Patch104pZH/Design/Scripts/textures/generated_edited/missing_tga_files.txt
xezon, line 1: Better: "missing_mapped_textures.txt"

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
xezon, line 1: Perhaps rename to "find_texture_errors.py"

Patch104pZH/Design/Scripts/textures/get_mapped_textures_info.py
DevGeniusCode: The textures and TGA files are treated separately because the INI files contain references to both the texture name (e.g., `PowerBarGreen`) and the actual TGA file (e.g., `SCGameUserInterface512_001.tga`).

Here is an example of how the data is structured in the INI files:

```
MappedImage PowerBarGreen
  Texture = SCGameUserInterface512_001.tga
  TextureWidth = 512
  TextureHeight = 512
  Coords = Left:497 Top:77 Right:499 Bottom:83
  Status = NONE
End
```

In this case, `PowerBarGreen` is the texture name, and `Texture` refers to the actual TGA file. I agree that the distinction might not be immediately clear, and I would be happy to consider any suggestions to make it more intuitive.

xezon: Ok. I suggest rename "texture" to "mapped_image" and "tga_file" to "texture".

Patch104pZH/Design/Scripts/textures/generated_edited/missing_tga_files.txt
DevGeniusCode: I changed the file name to `missing_textures_files` because the name you suggested will be more appropriate for a later stage, when I print the mapped textures present in the INI files but not in the mapping files.
Maybe it would be better to call it `invalid_mapped_textures`? I have to admit that I’m a bit confused myself within the code between the actual TGA files and the mapped textures, as both are textures, and the variable names are getting mixed up.

DevGeniusCode: I think we should use the word "textures" for the files (like TGA\DDS\W3D files) and "images" for the mapped textures (the ones from the mapping process).

DevGeniusCode:
optimized the code and updated various functions/variables names for better clarity.

Patch104pZH/Design/Scripts/textures/generated_edited/missing_tga_files.txt
xezon: `invalid_mapped_textures` would be better then.

xezon (dismissed):

DevGeniusCode:
Added validation for textures/images

1.  **Texture/Image Validation**:
    *   Implements a check for invalid textures and images. This identifies resources referenced in `.ini` files but not found in the file system, indicating broken or missing assets.
    *   Adds a check for unused textures and images. This finds resource files present in the file system but without corresponding references in any `.ini`/`.wnd`/`.w3d` file, suggesting potential dead assets.

2.  **Interactive Main Menu**:
    *   Enhances the main script with a user-interactive dialog. Users can now choose which function to execute: either the new validation checks or the existing texture and image extraction process.

DevGeniusCode:
I think I've now covered all the necessary validation for images and textures.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_images_in_ini_files.txt
```diff
@@ -0,0 +1,11 @@
+4
```
xezon: I expect that this number 4 is difficult to find with a text search. Perhaps add a bit more context to this record to make it identifiable.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
```diff
@@ -0,0 +1,80 @@
+aiblocktile
```
xezon: Could you perhaps add a bit more context to the line, aka what it belongs to?

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
xezon, line 1: Better name would be "unused_mapped_images"

xezon, line 1: Why are all mapped images lower case? I suggest print them out in their original case.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
```diff
@@ -0,0 +1,393 @@
...
+genexpbarbottom1
+genexpbartop1
+glacampaign_b
+glacampaign_g
+glacampaign_s
+globalgen
+goodstatsicon
+helpbox-bottom
+helpbox-middle
+helpbox-top
```
xezon: I know that the helpbox textures are definitely used in game. They are referenced in game code.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
```diff
@@ -0,0 +1,393 @@
...
+honorblitz5
+honorcampaignchina
+honorcampaigngla
+honorcampaignusa
+honorstreak_100
+honorstreak_1000
+honorstreak_500
+honorstreak_b
+honorstreak_g
+honorstreak_s
```
xezon: I expect that these honor medals are referenced in game code. So a lot of mapped images here may actually be used.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures_files.txt
```diff
@@ -0,0 +1,4648 @@
...
+cbngasst_tank
+cbngasst_tank_d
+cbngasst_tank_ds
+cbngasst_tank_e
+cbngasst_tank_es
+cbngasst_tank_n
+cbngasst_tank_r
+cbngasst_tank_rs
+cbngasst_tank_s
+cbngasst_wall
```
xezon: I am confident to say that the majority of the textures listed here are referenced in w3d models. This needs rework.

Patch104pZH/Design/Scripts/textures/generated_original/invalid_wnd_images.txt
xezon, line 1: "invalid_mapped_images_in_wnd_files"

Patch104pZH/Design/Scripts/textures/generated_edited/missing_textures_files.txt
xezon, line 1: I think this file needs a better name.

xezon (commented):
I think the txt file names need an improvement pass to clarify what they refer to.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_images_in_ini_files.txt
DevGeniusCode: The regular expression is a bit lazy, I will improve it.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
DevGeniusCode: Can do this, but it will overload the file, as an invalid name can appear in multiple files. You can easily find them by searching in VSC or another tool.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
DevGeniusCode: There are too many differences due to capitalization, so I converted everything to lowercase to find only the ones that are truly invalid because no files with those names exist.
Will print with the original name

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
DevGeniusCode: Do you have any idea how to find them? Just need to check if these names exist in the dat/exe file

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
xezon: Yes. Use lower case for comparison where applicable, but print original.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures_files.txt
DevGeniusCode: This is after checking the textures in w3d as well, just to be safe I'll check again.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
xezon: The strings are certainly in the executable (game.dat). Not sure if plain ascii text search will suffice but you can try. Make sure the string ends on a 0 (null terminator).

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures_files.txt
xezon: Perhaps there is an error in the search method. I know that these textures are used because I have seen them in the w3d viewer on models.

Make sure to combine Generals 1.08 and Zero Hour 1.04 files because they both provide unique assets.

DevGeniusCode: That's right, I didn't extract the w3d files for Generals 1.08. Will do.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_images_files.txt
DevGeniusCode: I'm not sure how to find this, this is what I've tried so far:
```python
def check_string_in_file(string, file_path):
    with open(file_path, 'rb') as file:
        content = file.read()

        # Convert the string to bytes with a null terminator at the end
        byte_string = string.encode('ascii') + b'\x00'

        if byte_string in content:
            print(f'found')
        else:
            print(f'not found')

string_to_check = "helpbox-middle"
file_path = r"C:\Program Files (x86)\Steam\steamapps\common\Command & Conquer Generals - Zero Hour\game.dat"
check_string_in_file(string_to_check, file_path)
```

Result: `not found`.

But maybe it depends on my interpreter Pycharm?

xezon: One potential problem here is capitalization of string. Either do a case insensitive search or get the search string in the correct capitalization. The mapped image name is "Helpbox-middle" and not "helpbox-middle".

DevGeniusCode:
Improvements and fixes, added checking for unused textures/images in the game.dat file, and added the SCPatchUserInterface512.INI file that was missing from the repo. the reports were updated and also include Generals 1.08 files.

Also, saw your YouTube review of my code. Yes, I know regex isn't everyone's cup of tea, but without it I'd be drowning in manual parsing. It's a necessary tool for this job, Oh, and a little secret: I'm also using AI to craft those regex monsters, so really, we are on the same boat! 😉

Patch104pZH/GameFilesOriginalZH/Data/INI/MappedImages/TextureSize_512/SCPatchUserInterface512.INI
xezon, line 1: This needs to be moved into the GameFilesOriginalCCG folder.

Patch104pZH/Design/Scripts/textures/find_mappd_images_and_textures_errors.py
xezon, line 1: spelling error in file name "mappd"

Patch104pZH/Design/Scripts/textures/find_mappd_images_and_textures_errors.py
```diff
@@ -0,0 +1,404 @@
...
+
+        elif function_choice == "7":
+            print("Calling function: find_unused_textures_and_images_in_game_files")
+            w3d_textures = extract_textures_from_w3d_files(folder_paths['w3d'])
+            unused_textures = unused_assets_in_assets_list(textures_files,
+                                                           ini_textures_scraping + mapped_textures + w3d_textures)
+            wnd_images = extract_images_from_wnd_files(folder_paths['wnd'])
+            unused_images = unused_assets_in_assets_list(images, ini_images_scraping + wnd_images)
+            unused_textures = check_strings_in_dat_file(unused_textures, folder_paths['dat'])
+            unused_images = check_strings_in_dat_file(unused_images, folder_paths['dat'])
```
xezon: Calling this twice means game.dat is loaded and processed in twice as well. Just to keep in mind in case this is slow.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
```diff
@@ -0,0 +1,303 @@
...
+SNPowerTray2_Holder
+SNSuperHack
+SNSuperHack_L
+SNTactNuke
+SNTankPara
+SNTray
+SNUAttack
+SSChevron1S
+SSChevron2S
+SSChevron3S
```
xezon, line 247: I think it is very unlikely that this image is unused. The chevron is certainly used on control bar buttons.

Patch104pZH/Design/Scripts/textures/generated_original/unused_textures.txt
```diff
@@ -0,0 +1,2298 @@
...
+NVGattTank_d4
+NVHelixNG_D
+NVMOutpost
+NVOvrlrd_D5
+NVSSupplyTk_B
+NVSupplyTk_d2
+NVSupplyTk_d3
+NVSupplyTk_d4
+NVSupplyTk_d5
+NVSupplyTk_d6
```
xezon: This corresponds to w3d model names, not texture names. What is the goal of this list? It says "unused_textures" which in my understanding would mean texture files that exist on disk but are referenced nowhere.

Patch104pZH/Design/Scripts/textures/generated_original/invalid_textures_in_ini_files.txt
```diff
@@ -0,0 +1,107 @@
...
+CBHigh01_DSNG
+CBHigh01_ES
+CBHigh01_ESN
+CBHigh01_RS
+CBHigh01_RSN
+CBHigh01_S
+CBHigh01_SG
+CBHigh01_SN
+CBHigh01_SNG
+CBHigh02_DS
```
xezon: The txt file name claims invalid textures, but this is a model name.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
```diff
@@ -0,0 +1,303 @@
...
+Rank_General_GLA
+Rank_General_USA
+Rank_LieutenantElite
+Rank_Lieutenant_China
+Rank_Lieutenant_GLA
+Rank_Lieutenant_USA
+Rank_MajorElite
+Rank_Major_China
+Rank_Major_GLA
+Rank_Major_USA
```
xezon, line 135: I think it is very suspicious that so many of these Rank images are claimed unused.

My guess is that they are used, but they are assembled by combining words, and therefore not found as whole words in the executable file.

So for example there could be "Rank_Major" + "_USA", "Rank_Major" + "_GLA", etc...

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
xezon, line 1: I expect that most textures listed here are actually used, but are not easily found because they are assembled of parts depending on conditions.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures.txt
xezon, line 1: Why is this list so much longer than the other?

Patch104pZH/GameFilesOriginalZH/Data/INI/MappedImages/TextureSize_512/SCPatchUserInterface512.INI
DevGeniusCode: What are the 3 folders actually?

xezon: It is described here:
https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Patch104pZH#readme

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
DevGeniusCode: I suggest leaving 'unused' as I have no idea how to deal with it, and focusing on invalid references.

xezon: I suggest document clearly that the unused mapped images are not entirely trustable because of the caveat above. There probably are a few unused ones but they would need to be found by manually reviewing the list.

Patch104pZH/Design/Scripts/textures/generated_original/unused_textures.txt
DevGeniusCode: Indeed, they are w3d files that are in csv but I didn't find any references to them. Should I separate w3d files from tga/dds files?

Patch104pZH/Design/Scripts/textures/generated_original/invalid_textures_in_ini_files.txt
DevGeniusCode: If I understood correctly, these are w3d file models, and if so, then w3d files with these names should exist, which they don't.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures.txt
DevGeniusCode: The edited list only scanned the edited folder, not the content of files that exist in the original folder such as w3d files extracted from big.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
DevGeniusCode: I think what you see in the game is `SSChevron3L`, meaning the ones with the big L suffix and not the small S suffix.

(By the way, you can view tga files using a suitable extension for VSC.)

Patch104pZH/Design/Scripts/textures/generated_original/unused_textures.txt
xezon: Yes I suggest create a "unused_textures.txt" and "unused_models.txt"

Patch104pZH/Design/Scripts/textures/generated_original/invalid_textures_in_ini_files.txt
DevGeniusCode: For example `CBHangr02_DSNG` with incorrect letter composition, which causes issue #2377

xezon: I don't understand this statement.

So for example we have:
```
Object MogadishuShop01Bib
  ; *** ART Parameters ***
  Draw = W3DModelDraw ModuleTag_01
    ConditionState = NONE
      Model = CBModis03_TR
    End
    ConditionState = DAMAGED
      Model = CBModis03_TRD
    End
    ConditionState = REALLYDAMAGED
      Model = CBModis03_TRD
    End
    ConditionState = RUBBLE
      Model = CBModis03_TRD
    End
  End
```
All these are Models. Not Textures. I suggest put these into a "invalid_models_in_ini_files.txt"

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures.txt
xezon: The GameFilesEdited builds on top of the original. Can you perhaps program it in a way that it does not show false errors by considering the original underlying files?

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
xezon: Ok that is a fair assumption.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_mapped_images.txt
DevGeniusCode: I looked into it a bit, and it seems to use Rank_[level][_army], there string `Rank_%s%s`.

But I'll add a warning at the beginning of the file, there are too many cases to handle in code.

DevGeniusCode:
Update:
Separation of textures and models, improved reports.

Patch104pZH/Design/Scripts/textures/find_mappad_images_and_textures_errors.py
xezon, line 1: File has misspelled "mappad"

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures.txt
```diff
@@ -872,612 +84,19 @@ Manic Aggression
...
-ParaFallSK
-ParaJump
-ParaJumpMSH
-ParaJumpSK
-ParaLand
-ParaLandMSH
-ParaLandSK
 Red Rock
 Rocky Rampage
 Rogue Agent
```
xezon: What are these entries here? They look to be map names, not textures!?

Patch104pZH/Design/Scripts/textures/generated_original/invalid_textures_in_ini_files.txt
```diff
@@ -1,86 +1,11 @@
...
-NBNMissle_A3ENSS
-NONE
-NULL
 None
-PMLitPol01_D
-PMLitPol02_D
-PSAirCarrier_NS
-PSCarRapt_D1B
 SCCScroll
 SMUDGE
```
xezon: I think this refers to some special surface in the executable. Perhaps search game.dat as well.

Patch104pZH/Design/Scripts/textures/find_mappad_images_and_textures_errors.py
```diff
@@ -0,0 +1,473 @@
...
+def write_to_file(output_path, data, format_func=None, warning=False):
+    warning_message = '''
+##########################################
+# WARNING: The list of unused mapped images may not be entirely accurate!
+# Many of the "unused" images are actually used, but are assembled by combining
+# multiple string components (e.g., "Rank_" + "Colonel" + "_USA", "Rank" + "Major" + "_GLA").
+# These combined names might not appear as complete strings in the executable file,
+# causing them to be falsely flagged as unused.
+# Manual review of the list is required to verify the actual usage of some images.
+##########################################
```
xezon: Perhaps omit this line make it look cleaner.

Patch104pZH/Design/Scripts/textures/find_mappad_images_and_textures_errors.py
```diff
@@ -0,0 +1,473 @@
...
+
+    unused_assets = assets_lower - assets_list_lower
+    unused_assets_original_case = [asset for asset in assets if asset.lower() in unused_assets]
+
+    return sorted(unused_assets_original_case)
+
+
+def write_to_file(output_path, data, format_func=None, warning=False):
+    warning_message = '''
+##########################################
```
xezon: Perhaps omit this line make it look cleaner.

Patch104pZH/Design/Scripts/textures/find_mappad_images_and_textures_errors.py
```diff
@@ -0,0 +1,473 @@
...
+    output_folder = 'generated_original' if is_original else 'generated_edited'
+    os.makedirs(output_folder, exist_ok=True)
+
+    # Process the data
+    textures_files_mapped_images_dictionary, images, mapped_textures, duplicate_images = get_mapped_images(
+        folder_paths['images'])
+    textures, models = get_textures_models_from_csv(folder_paths['csv'])
+    textures_files = textures.union(
+        get_files_from_folder(folder_paths['textures'])).union(
+        get_files_from_folder(folder_paths['ani']))
```
xezon: I wonder if we can classify ani files as cursors instead of textures to show them in separate lists.

Patch104pZH/Design/Scripts/textures/generated_edited/unused_textures.txt
DevGeniusCode: come form `maps/Rogue Agent/Rogue Agent.tga`, will filter maps

Patch104pZH/Design/Scripts/textures/generated_original/invalid_textures_in_ini_files.txt
DevGeniusCode: This appears in `Data\INI\ParticleSystem.ini` at ` ParticleName = SMUDGE.tga`,

DevGeniusCode:
Update: fix script name and maps references and remove `ani` files

Patch104pZH/Design/Scripts/textures/find_mapped_images_and_textures_errors.py
```diff
@@ -0,0 +1,476 @@
...
+        function_choices = [str(i) for i in range(1, 8)]
+
+    edited_folder_paths = '../../../GameFilesEdited'
+    edited_folder_paths = {
+        'images': f'{edited_folder_paths}/Data/INI/MappedImages',
+        'ini': f'{edited_folder_paths}/Data/INI',
+        'wnd': f'{edited_folder_paths}/Window',
+        'w3d': f'{edited_folder_paths}/Art/W3D',
+        'ani': f'{edited_folder_paths}/Data/Cursors',
+        'textures': f'{edited_folder_paths}/Art/Textures',
```
xezon: Does this script also consider the language specific assets at
```
Data/English/Art/Textures
Data/English/Art/W3D
Data/German/Art/Textures
Data/German/Art/W3D
...
```
This logic would become more intricate if this script also tests if specific files are missing in specific languages (which does happen in the original).

I think it would not be strictly necessary because we already fixed the language specific asset errors, but if it is no big hassle it could be added to this script.

DevGeniusCode: Yes, I will update this, just to point out that I don't have the language files other than English to research them. I scraped all file paths with relevant extensions from the file, so maybe instead of adding them, it's better to filter out the language files? This way they won't appear in lists of unused files.

xezon: Language packs are here:
https://www.gentool.net/download/languages/

Patch104pZH/Design/Scripts/textures/find_mapped_images_and_textures_errors.py
```diff
@@ -0,0 +1,476 @@
...
+        'csv': '../../../Resources/FileHashRegistry/Generals-108-GeneralsZH-104.csv',
+        'dat': r"C:\Program Files (x86)\Steam\steamapps\common\Command & Conquer Generals - Zero Hour\game.dat"
+    }
+
+    orginal_folder_paths = '../../../GameFilesOriginal'
```
xezon: "original"

Patch104pZH/Design/Scripts/textures/find_mapped_images_and_textures_errors.py
xezon: Note that we also have a GameFilesOptional folder that adds optional assets on top of the core assets (GameFilesEdited). Perhaps it would be a consideration to also add that, but that would add another layer of complexity.
https://github.com/TheSuperHackers/GeneralsGamePatch/tree/main/Patch104pZH/GameFilesOptional

Additionally, we also have assets that are generated via Mod Builder and defined in the json files:
https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/ModBundleCoreItems.json
https://github.com/TheSuperHackers/GeneralsGamePatch/blob/main/Patch104pZH/ModBundleOptionalItems.json

For example:
```json
                            {
                                "source": "Art/Textures/GenerateMip/cbapt01_n.psd",
                                "target": "Art/Textures/cbapt01_sn.dds"
                            },
                            {
                                "source": "Art/Textures/GenerateMip/cbapt01_ng.psd",
                                "target": "Art/Textures/cbapt01_sng.dds"
                            },
                            {
                                "source": "Art/Textures/GenerateMip/cbapt01_dn.psd",
                                "target": "Art/Textures/cbapt01_dsn.dds"
                            },
                            {
                                "source": "Art/Textures/GenerateMip/cbapt01_dng.psd",
                                "target": "Art/Textures/cbapt01_dsng.dds"
                            },
```

DevGeniusCode: I think the additional lists will only contribute to unused lists, while our main concern is fixing broken references that can disrupt the game engine.

xezon: Yes ideally Optional files do not solely fix Core issues, but they could if we made mistakes (and I am sure we did).

DevGeniusCode:
- Adds `analyze_language_files` to scan language folders for unused/invalid assets.
- Generates individual text reports for each language.
- Assumes language-specific assets are not present in global folders.
[The language assets parameter is a dict containing models and textures that are only located inside the language folder.]

- Updated regex in Animation Models to correctly extract model names even when optional parameters are present.
[Results of research on relevant value locations are detailed in `Graphical_Assets_Location.md`. If there are any additional locations or patterns not currently documented in that file, please let me know, as it will help improve the accuracy of the reports.]

Patch104pZH/Design/Scripts/textures/Graphical_Assets_Location.md
xezon, line 1: Perhaps use lowercase in file name like all other files in this folder.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
```diff
@@ -0,0 +1,22 @@
+CBWBrdgeSt_
+EXDots
+EXRedSmokePuff
+EXSateliteHack1
+EXthunderRadar
+EXthunderWave
```
xezon: Why is this not written as "EXthunderWave.tga" as per the original value in the INI file?

Patch104pZH/Design/Scripts/textures/generated_edited/languages/French.txt
```diff
@@ -0,0 +1,24 @@
+Report for language: French
+
+============== Unused Textures (INI) ==============
+
+French install
+SCSmShellUserInterface512
```
xezon: Why do these files have no file extension?

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
```diff
@@ -0,0 +1,27 @@
...
+Skirmish_Loaduserinterface
+UITerdrone01
+ZHCA_PMNFlag
+ZHCA_UITerdrone01
+
+============== Unused Models (INI) ==============
+
+CItoxic
+CItoxic_B
+CItoxic_C
```
xezon: These 3 are textures, not models.

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
```diff
@@ -0,0 +1,22 @@
...
+============== Unused Textures (INI) ==============
+
+Skirmish_Loaduserinterface
+ZHCA_PMNFlag
+
+============== Unused Models (INI) ==============
+
+None
+
+============== Invalid Models Assets ==============
```
xezon: "Invalid Models Assets" reads strange, because it has 2 plural words. Probably should be "Invalid Models"

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
```diff
@@ -0,0 +1,22 @@
...
+
+UITRST_SKNP
+
+============== Invalid Textures in Mapped Images Assets ==============
```
xezon: "Mapped Images Assets" reads strange, because it has 2 plural words. Probably should be "Mapped Images"

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
xezon: "Invalid Models" is a confusing term for this, because UITRST_SKNP.w3d is missing in Chinese. It is not an invalid file, but simply missing. Could you explain why "Invalid" was chosen as the title?

Patch104pZH/Design/Scripts/textures/generated_original/unused_textures.txt
```diff
@@ -0,0 +1,930 @@
...
+PTPalm02b
+PTPine01_hi_B
+PTPine03_B
+PTPine04_s
+PTReed01
+PTXPALM02_C
+PTXPALM03
+PTXPALM03._Btga
+PTXPALM03_B.tga
+PTYucca01
```
xezon: Why are these files listed without file extensions?

Patch104pZH/Design/Scripts/textures/find_graphical_assets_errors.py
```diff
@@ -0,0 +1,591 @@
...
+            write_to_file(os.path.join(output_folder, 'unused_models.txt'), unused_models)
+
+            wnd_images = extract_images_from_wnd_files(folder_paths['wnd'])
+            unused_images = unused_assets_in_assets_list(images, ini_images_scraping + wnd_images)
+            unused_images = check_strings_in_dat_file(unused_images, folder_paths['dat'])
+            write_to_file(os.path.join(output_folder, 'unused_mapped_images.txt'), unused_images, warning=True)
+
+        elif function_choice == "8":
+            print("Calling function: analyze_language_files")
+            analyze_language_graphical_assets(folder_paths['ini'].replace('/Data/INI', ''), languages, language_assets, output_folder,
```
xezon: Graphical assets can be described as "art files", which arguably is simpler to understand because the game folder that contains textures and models is called "Art".

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
DevGeniusCode: True, but they are in the W3D folder instead of the Textures folder, so the function defines it as a model, if it is okay for texture files to be in the models folder, I will update the function to only look for w3d files.

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
DevGeniusCode: The invalid list indicates references that exist in the ini files but no files exist for them. In the original folder, this is indeed an invalid reference, but in the edited folder, a file has been added and now the reference is valid (None).

Patch104pZH/Design/Scripts/textures/generated_original/unused_textures.txt
DevGeniusCode: By the way, these two files also appear here without extensions, here is the full path

```
Art/Terrain/PTXPALM03._Btga.tga
Art/Terrain/PTXPALM03_B.tga.tga
```

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
DevGeniusCode: Since the comparison was always without suffixes, so I did not capture them, I will include the suffixes for the reports.

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
DevGeniusCode: what about `Invalid Languages Models references in INI files`

DevGeniusCode:
Added file extensions and improved naming.

Note that since we only catch with an extension, it was found that the `ParticleSystem.ini` file has an invalid parameter `ParticleName = GlassShard01` because it does not include a file extension (it is currently not included in the files because it is not included in the regular expression)

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
xezon: It is a good question why textures are located in the w3d folder. I do not know what is up with that.

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
xezon: How about "Unrecognized Models in INI files" ?

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
xezon: Would a "Texture.psd" also match for a "Texture.tga" ?

Patch104pZH/Design/Scripts/textures/generated_edited/languages/Brazilian.txt
```diff
@@ -0,0 +1,22 @@
...
+
+============== Unused Languages Models ==============
+
+None
+
+============== Invalid Languages W3D Models references in INI files ==============
+
+None
+
+============== Invalid Languages Textures references in INI Mapped Images ==============
```
xezon: Languages Textures references are 3 plurals in a row. I think only the last word should be plural.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
DevGeniusCode: Indeed, assuming that `Mod Builder` converts it to dds/tga

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
DevGeniusCode: [The current update only includes `*.w3d` files. So if there are texture files in the folder, they are not scanned at all.]

Patch104pZH/Design/Scripts/textures/find_art_files_errors.py
```diff
@@ -0,0 +1,592 @@
...
+                w3d_file_path = os.path.join(root, file)
+                w3d_file_path = os.path.normpath(w3d_file_path)
+                textures_in_w3d = w3d_file_manager.get_textures(w3d_file_path)
+                textures_in_w3d = [os.path.splitext(texture)[0] for texture in textures_in_w3d]
+                textures_in_w3d_files.update(textures_in_w3d)
+
+    return list(textures_in_w3d_files)
+
+
+def extract_art_files_from_ini_files(folder_path):
```
xezon: Perhaps move contents of this function into more smaller functions to decrease its size.

xezon (approved):

Patch104pZH/Design/Scripts/textures/generated_original/languages/Chinese.txt
DevGeniusCode: What about txt file names? I preferred invalid to maintain consistency with the files, like `invalid_models_in_ini_files.txt`

xezon: Ok keep it then. I guess reader can infer the meaning eventually.

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
xezon: These are definitely texture files.

DevGeniusCode: I mean, do I need to scan them? Does the game engine scan them? Because right now my code only scans by folder paths.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
xezon: Typically yes, but Mod Builder can also build a "Texture.psd" as "AnotherTexture.dds", and I assume the script will not catch that.

Perhaps it would be worthwhile to check against files in `Patch104pZH\.Build\RawBundleItems` (the files generated before archiving into big files), but this would require building (all languages) before running the script and likely make it more complicated.

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
xezon: I do not know that. We would need to test if the game can load textures from that folder.

Patch104pZH/Design/Scripts/textures/generated_edited/invalid_textures_in_ini_files.txt
DevGeniusCode: Will the `ModBundleOptionalItems.json` file be sufficient for processing?

xezon: Currently `ModBundleCoreItems.json` and `ModBundleOptionalItems.json`.

You could try search through all files listed in `ModJsonFiles.json` and find texture and model targets that are not `*.ext`. Perhaps that would suffice.

Patch104pZH/Design/Scripts/textures/generated_edited/languages/German.txt
DevGeniusCode: I took the file `Data\English\Art\Textures\Defeated.tga` and replaced it with the French version, it worked.
I moved it to the W3D folder, the English version appeared, meaning it didn't work.
I deleted it, now there is no `Data\English\Art\Textures\Defeated.tga` file at all, it still appeared in the English version, meaning the file is also in `game.dat`, but a hex scan yielded no results. I was confused.

xezon: The texture is not in game.dat

Defeated.tga is still inside original big file that is underlying Patch files.

Reads as if texture is not loaded from W3D folder, which means these 3 files are misplaced in the German language.

DevGeniusCode: I deleted all the big files after extracting.

xezon:
Merge this or is anything left to do?

DevGeniusCode:
Improve `extract_art_files_from_ini_files` and spelling

TODO:
- add `ani` assets.
- add json blender assets, currently is assuming that Mod Builder converts psd it to dds/tga

PR #2643, opened 2025-03-05
Add pristine INI bundle items and packs to exclude controversial changes from build

xezon:
This will allows us to exclude controversial changes from build.

### DIFF
--- a/Patch104pZH/ModBundleCoreItems.json
+++ b/Patch104pZH/ModBundleCoreItems.json
@@ -218,6 +218,34 @@
                 ]
             },
             {
+                "name": "PristineCoreINI",
+                "big": true,
+                "files": [
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceList": [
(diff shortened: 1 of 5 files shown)


xezon:
Next step: Add markers. Perhaps INI scripter can assist. Will be hard labor.

xezon:
Thinking about it more, perhaps this is not the way to go, and instead configuration should be supported by new code feature, where we allow to override INI settings and push them into presets, so that these presets become toggleable in game.

So for example here we want to have a "Non-Controversial" preset. And we also have the "Optional" preset. I think long term this also scales better when code manages this, instead of data. So INI files would not be shipped with variants, but they contain all variants always and code is responsible for variant selection.

This is probably a better way to make configurable data presets.

xezon:
The simplest solution here would be to opt out of all controversial changes that Community Outpost does not want to begin with when transitioning to this Patch base. So just compile these INI changes out and have one version. This requires no features from code and all we need is a list of changes that we want to disable.

Issue #2607, opened 2025-03-03
World Builder: Re-organize the texture tab


toirehantaa:
A better categorization of textures helps newbies to understand the structure and find certain textures more easily. The current subdivision is unclear and some textures are misplaced / are duplicates of others

tintinhamans:
This file seems to be the place to do that: https://github.com/TheSuperHackers/GeneralsGamePatch/blob/773999b5348c85c4b43c87c38d78925e5b6cfe20/Patch104pZH/GameFilesEdited/Data/INI/Terrain.ini#L1-L24

tintinhamans:
Changing/adding terrain class would also require rework here:

https://github.com/TheSuperHackers/GeneralsGameCode/blob/0a05454d8574207440a5fb15241b98ad0b435590/GeneralsMD/Code/GameEngine/Include/Common/TerrainTypes.h#L45-L95

PR #2644, opened 2025-03-06
Small update to Terrain.ini

toirehantaa:
For now, 2 small changes without restructuring, but still worth doing:

1.Removal of a duplicate texture
image
image

2. Swapping these two Rock-Textures so they fit their category (Screenshots show fixed version)
image
image

### DIFF
--- a/Patch104pZH/GameFilesEdited/Data/INI/Terrain.ini
+++ b/Patch104pZH/GameFilesEdited/Data/INI/Terrain.ini
@@ -429,8 +429,8 @@ Terrain RocksType6
   Class = Desert_2
 End
 
-Terrain RocksType7
-  Texture = TXRock07a.tga
+Terrain RocksType5              ;FIELD-RocksType5 is the sand-version of DESERT_2-RocksType7, but placed in FIELD. Swapped. // Patch104p @fix -TanSo- 03/06/2025 Swapped Rock Textures (#2607)
+  Texture = TXRock05a.tga
   Class = Desert_2
 End
 
@@ -918,12 +918,7 @@ Terrain GrassMediumType37g
   Texture = TMGras37g.tga
   Class = GRASS_COBBLESTONE
 End
-
-Terrain GrassMediumType38a
-  Texture = TMGras38.tga
-  Class = GRASS_COBBLESTONE
-End
-
+;GrassMediumType38a removed for being a direct clone of GrassMediumType8  // Patch104p @fix -TanSo- 03/06/2025 Removed Duplicate Texture (#2607)
 Terrain Grass02
   Texture = TGGrass02.tga
   Class = GRASS_COBBLESTONE
@@ -1325,8 +1320,8 @@ Terrain DirtMediumType14
   Class = FIELD
 End
 
-Terrain RocksType5
-  Texture = TXRock05a.tga
+Terrain RocksType7              ;DESERT_2-RocksType7 is the non-sand-version of FIELD-RocksType5, but placed in DESERT_2. Swapped. // Patch104p @fix -TanSo- 03/06/2025 Swapped Rock Textures (#2607)
+  Texture = TXRock07a.tga
   Class = FIELD
 End
 


tintinhamans:
Related #2607

Patch104pZH/GameFilesEdited/Data/INI/Terrain.ini
```diff
@@ -918,12 +918,7 @@ Terrain GrassMediumType37g
   Texture = TMGras37g.tga
   Class = GRASS_COBBLESTONE
 End
-
-Terrain GrassMediumType38a
-  Texture = TMGras38.tga
-  Class = GRASS_COBBLESTONE
-End
-
+;GrassMediumType38a removed for being a direct clone of GrassMediumType8  // Patch104p @fix -TanSo- 03/06/2025 Removed Duplicate Texture (#2607)
```
xezon, line 921: What happens if a map uses this Grass type and this INI definition is removed?

Patch104pZH/GameFilesEdited/Data/INI/Terrain.ini
```diff
@@ -429,8 +429,8 @@ Terrain RocksType6
   Class = Desert_2
 End
 
-Terrain RocksType7
-  Texture = TXRock07a.tga
+Terrain RocksType5              ;FIELD-RocksType5 is the sand-version of DESERT_2-RocksType7, but placed in FIELD. Swapped. // Patch104p @fix -TanSo- 03/06/2025 Swapped Rock Textures (#2607)
+  Texture = TXRock05a.tga
   Class = Desert_2
 End
```
xezon, line 435: I suggest write it like so:

```
; Patch104p @fix -TanSo- 06/03/2025 RocksType5 swapped with RocksType7 (#2607)
; RocksType5 is the sand-version of RocksType7, but was placed in FIELD.
Terrain RocksType5              
  Texture = TXRock05a.tga
  Class = Desert_2
End
```

Date format is DD/MM/YYYY

toirehantaa:
You'll be met with the replacement request window. If you want to open a map that uses GrassMediumType38a, you can simply replace it with GrassMediumType8 in 5 seconds
image

tintinhamans:
> You'll be met with the replacement request window. If you want to open a map that uses GrassMediumType38a, you can simply replace it with GrassMediumType8 in 5 seconds

That's cool in World Builder and all but how does that work in-game? Can you verify?

xezon:
Will every World Builder user know what the replacement for GrassMediumType38a is?

toirehantaa:
> That's cool in World Builder and all but how does that work in-game? Can you verify?

image
Black texture - maybe that is a problem for now. As long as you can't edit the map yourself and replace the texture, this will be shown. It's the same for the iShock Arena Map btw


> Will every World Builder user know what the replacement for GrassMediumType38a is?


Depends on how well versed the user is in WB - but I think there barely is anyone who doesn't know

xezon:
I think the graceful way to deprecate this would be a flag for `GrassMediumType38a` that hides it from World Builder UI, but keeps the definition intact. That is a feature for code, World Builder.

tintinhamans:
Wonder if Terrain `Class` has any effect in-game. If not, then maybe we can maintain compatibility by simply moving these duplicates and such into a new `Class = DEPRECATED` which hides it from World Builder UI.

xezon:
Perhaps Class is used to spawn different particles when units drive over.

tintinhamans:
Then maybe adding a new param such as `WbHidden = True` might be a safer option.

kABUSE1:
> > Will every World Builder user know what the replacement for GrassMediumType38a is?
> 
> Depends on how well versed the user is in WB - but I think there barely is anyone who doesn't know

You must be crazy lol

I'd say the number of worldbuilder users on earth who know by heart that GrassMediumType8 and GrassMediumType38a are the same texture is below 10. Not percent. 10 humans.

I have 19 years of worldbuilder experience and I had no idea.

toirehantaa:
> I have 19 years of worldbuilder experience and I had no idea.

You're a special case kABUSE because you dodge detailing like it's a sickness :D

PR #2678, opened 2025-03-14
Starlifter line is consistent now

dmgreeny:


r0b2g1t (approved):

dmgreeny (commented):
Added a comment. This is not used in game and the change is very minor. I think it can be merged without yaml file.

PR #2715, opened 2025-03-21
Adding Swedish entries in various json

Polypheides:
Adding various entries for the Swedish language in the .json located in /Patch104pZH/

This will not build with current gametextcompiler.

### DIFF
--- a/Patch104pZH/ModBundleCoreLanguageItems.json
+++ b/Patch104pZH/ModBundleCoreLanguageItems.json
@@ -396,6 +396,40 @@
                         }
                     }
                 ]
+            },
+            {
+                "name": "CoreLangSwedish",
+                "big": true,
+                "setGameLanguageOnInstall": "English",
+                "files": [
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceTargetList": [
+                            {
+                                "source": "Data/Swedish/*.ini",
+                                "target": "Data/English/*.ini"
+                            }
+                        ]
+                    },
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceTargetList": [
+                            {
+                                "source": "Data/generals.str",
+                                "target": "Data/English/generals.csf"
+                            }
+                        ],
+                        "params": {
+                            "language": "Swedish",
+                            "excludeMarkersList": [
+                                [
+                                    "//patch104p-optional-begin",
+                                    "//patch104p-optional-end"
+                                ]
+                            ]
+                        }
+                    }
+                ]
             }
         ]
     }
--- a/Patch104pZH/ModBundleCoreAudioItems.json
+++ b/Patch104pZH/ModBundleCoreAudioItems.json
@@ -171,6 +171,21 @@
                         ]
                     }
                 ]
+            },
+            {
+                "name": "CoreAudioSwedish",
+                "big": true,
+                "files": [
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceTargetList": [
+                            {
+                                "source": "Data/Audio/Sounds/English/*.wav",
+                                "target": "Data/Audio/Sounds/English/*.wav"
+                            }
+                        ]
+                    }
+                ]
             }
         ]
     }
--- a/Patch104pZH/ModBundleCorePacks.json
+++ b/Patch104pZH/ModBundleCorePacks.json
@@ -157,6 +157,20 @@
                     "CoreW3D",
                     "CoreWindow"
                 ]
+            },
+            {
+                "name": "CoreSwedish",
+                "itemNames": [
+                    "CoreAudio",
+                    "CoreAudioSwedish",
+                    "CoreINI",
+                    "CoreLangSwedish",
+                    "CoreMaps",
+                    "CoreMisc",
+                    "CoreTextures",
+                    "CoreW3D",
+                    "CoreWindow"
+                ]
             }
         ]
     }
--- a/Patch104pZH/ModBundleFullPacks.json
+++ b/Patch104pZH/ModBundleFullPacks.json
@@ -231,6 +231,27 @@
                     "CoreWindow",
                     "RecoveredTextures"
                 ]
+            },
+            {
+                "name": "FullSwedish",
+                "itemNames": [
+                    "OptionalAudio",
+                    "OptionalAudioSwedish",
+                    "OptionalINI",
+                    "OptionalLangSwedish",
+                    "OptionalTextures",
+                    "OptionalW3D",
+                    "CoreAudio",
+                    "CoreAudioSwedish",
+                    "CoreINI",
+                    "CoreLangSwedish",
+                    "CoreMaps",
+                    "CoreMisc",
+                    "CoreTextures",
+                    "CoreW3D",
+                    "CoreWindow",
+                    "RecoveredTextures"
+                ]
             }
         ]
     }
--- a/Patch104pZH/ModBundleOptionalAudioItems.json
+++ b/Patch104pZH/ModBundleOptionalAudioItems.json
@@ -291,6 +291,33 @@
                         ]
                     }
                 ]
+            },
+            {
+                "name": "OptionalAudioSwedish",
+                "big": true,
+                "files": [
+                    {
+                        "sourceParent": "GameFilesOptional",
+                        "sourceTargetList": [
+                            {
+                                "source": "Data/Audio/Sounds/English/*.wav",
+                                "target": "Data/Audio/Sounds/English/*.wav"
+                            },
+                            {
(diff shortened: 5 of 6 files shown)


Patch104pZH/ModBundleCoreLanguageItems.json
```diff
@@ -396,6 +396,40 @@
...
+                "name": "CoreLangSwedish",
+                "big": true,
+                "setGameLanguageOnInstall": "Swedish",
+                "files": [
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceList": [
+                            "Data/Swedish/*.ini"
+                        ],
+                        "registryList": [
```
xezon: This can be removed here.

Patch104pZH/ModBundleCoreLanguageItems.json
```diff
@@ -396,6 +396,40 @@
...
+                    {
+                        "sourceParent": "GameFilesEdited",
+                        "sourceTargetList": [
+                            {
+                                "source": "Data/generals.str",
+                                "target": "Data/Swedish/generals.csf"
+                            }
+                        ],
+                        "params": {
+                            "language": "Swedish",
```
xezon, line 423: Will this cause an error on build?

xezon (commented):
Looks good to me in principle.

Perhaps you can also create the same change for Ukrainian afterwards (in a separate change)?

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: Lets see if i understand this, should i not add anything to this file or is it these specific lines i don't need here?

if removing line 407-410 how should it look to make sure its correct?
Example 1:
image

or should i use other added languages as template like Arabic if so the Swedish entry would look similar to the Arabic one like this
Example 2:
image

xezon: What I mean is you can remove the

```
"registryList": [
  "Resources/FileHashRegistry/Generals-108-GeneralsZH-104.csv"
]
```

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: it depends on where or how an error would occur, i ran the BuildInstallRunWithGui.bat without issues.

xezon: Select the Swedish pack in the gui and click on Build.

Polypheides: it will build and launch the game with core or full Swedish, but i do have a modified gametextcompiler.exe for Swedish if that is needed for the build process. 

Is there a build log text output i can link or just the command line?

image

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: When using what i have in the commit it will build but removing the lines you suggested the json decoder gives me errors.

So maybe using the Arabic snippet example 2 is better with a source and target? when tested this build as well.

xezon: You need to remove the comma above as well.

Polypheides:
> Looks good to me in principle.
> 
> Perhaps you can also create the same change for Ukrainian afterwards (in a separate change)?

Sure when as soon as we figure out what is needed or not since I'm not a coder i heavily rely on random testing, asking questions and GPT misinformation.

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: Like this
image

Polypheides: After removing the comma FullSwedish
image

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: for the gametextcompiler

https://github.com/xezon/Thyme/commit/88b7c409fd4e400e404d419e70de8499292b70ba

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: I guess we don't want the translation to use the English folder for any added language in the long run, but i noticed that Russian and Arabic uses `"setGameLanguageOnInstall": "English",` and `"sourceTargetList"`

While other language uses `"sourceList":`

So maybe my entry is all wrong and should be like Arabic and Russian.

xezon: That is because the community made Russian and Arabic (and Swedish, Ukrainian) are not yet standalone languages. They rely on English big files to provide the base audio, that the Patch replaces in part. As long as that is the case, the community made languages use "English".

Perhaps we should document that somewhere. We cannot put comments in Json.

Polypheides: would this work in a way?

    "name": "CoreLangArabic",
    "_comment": "This Arabic localization is community-made",
    "big": true,
    "setGameLanguageOnInstall": "English",
    "files":

xezon: Yes that would work.

Patch104pZH/ModBundleCoreLanguageItems.json
xezon: Does it work with old gametextcompiler that does not have Swedish added yet?

Patch104pZH/ModBundleCoreLanguageItems.json
```diff
@@ -396,6 +396,38 @@
                         }
                     }
                 ]
+            },
+            {
+                "name": "CoreLangSwedish",
+                "_comment": "This Swedish localization is community-made",
```
xezon: I think this comment is not necessary here. There already is a wiki page somewhere which could list the original game languages, which then implies which ones are community made.

Here is the wiki page
https://github.com/TheSuperHackers/GeneralsGamePatch/wiki/localization_contribution

xezon (commented):
Looks

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: Removed

Reason i added the comment in the file was for future people so they could know by just browsing the JSON about what language were community-based!

The other JSON PR #2717 i made has all the comments in all the files but i guess i should remove them there as well.

Patch104pZH/ModBundleCoreLanguageItems.json
Polypheides: > Does it work with old gametextcompiler that does not have Swedish added yet?

Short answer: No

Long answer:
I have made a pull request over at [xezon/thyme](https://github.com/xezon/Thyme/pull/3) this will implement Swedish and many other languages.

We would need to add this new version over at https://github.com/TheSuperHackers/GeneralsTools/tree/main/Tools
so this will be fetched when running the setup 

Here is a link for a compiled gametextcompiler.exe https://www.dropbox.com/scl/fi/4jsp12fdb2tzh1y4d1wi1/gametextcompiler.zip?rlkey=dfnlkc5rja8wem32abosoh4rs&st=dt0qyodw&dl=0

xezon: Ok. If Swedish breaks build right now then we need to postpone submit until after new gametextcompiler is hooked up.

Polypheides: > Ok. If Swedish breaks build right now then we need to postpone submit until after new gametextcompiler is hooked up.

Building Swedish and Ukrainian fail, building all the other languages succeed.
```Build completed in 20.207 s
Do Post Build ...
Build Job completed in 21.192 s
```

We need to update the WindowsTools.json with size hash and github link.


The error when building Swedish with the current main branch build progress and compiler:
`
subprocess.CalledProcessError: Command '['C:\\Users\\Film\\Desktop\\GENERALS\\PATCH1~1\\Scripts\\Windows\\.tools\\gametextcompiler.exe', '-LOAD_STR', 'C:\\Users\\Film\\Desktop\\GENERALS\\PATCH1~1\\.Build\\RawBundleItems\\CoreLangSwedish\\Data\\Swedish\\generals.csf.tmp', '-SAVE_CSF', 'C:\\Users\\Film\\Desktop\\GENERALS\\PATCH1~1\\.Build\\RawBundleItems\\CoreLangSwedish\\Data\\Swedish\\generals.csf', '-LOAD_STR_LANGUAGES', 'Swedish']' returned non-zero exit status 2.
`

My additions should fix all this when the compiler gets added to the tool repo.

PR #2728, opened 2025-03-24
Fix Ukrainian string for GUI:Multiplayer

dmgreeny:


xezon:
This is very small change. Is this final?

dmgreeny:
Yes

xezon (approved):
Alright then :-)

PR #2765, opened 2025-05-10
Move Generals INI files into new project folder

xezon:
This change moves the Generals INI files into a new project folder. The project does not do anything yet.


ibrahim785382:
thanx

PR #2772, opened 2025-05-10
[GEN] Add new key mapping for Select All Aircraft to original game languages

xezon:
* Relates to TheSuperHackers/GeneralsGameCode#834
* Relates to #2768

This change adds a new key mapping for Select All Aircraft to the original game languages in Generals.

Select All Aircraft for German is bound to CTRL+Q. Potential key conflicts are not addressed.

## TODO

- [ ] Add documentation

### DIFF
--- a/Patch108pCCG/GameFilesCore/Data/French/CommandMap.ini
+++ b/Patch108pCCG/GameFilesCore/Data/French/CommandMap.ini
@@ -654,6 +654,17 @@ CommandMap SELECT_ALL
   DisplayName = GUI:SelectAll
 END
 
+; Patch108p @feature xezon 10/05/2025 Backports new key mapping from Zero Hour.
+CommandMap SELECT_ALL_AIRCRAFT
+  Key = KEY_W
+  Transition = DOWN
+  Modifiers = NONE
+  UseableIn = GAME
+  Category = SELECTION
+  Description = GUI:SelectAllAircraftDescription
+  DisplayName = GUI:SelectAllAircraft
+END
+
 CommandMap SCATTER
   Key = KEY_X
   Transition = DOWN
--- a/Patch108pCCG/GameFilesCore/Data/German/CommandMap.ini
+++ b/Patch108pCCG/GameFilesCore/Data/German/CommandMap.ini
@@ -654,6 +654,18 @@ CommandMap SELECT_ALL
   DisplayName = GUI:SelectAll
 END
 
+; Patch108p @feature xezon 10/05/2025 Backports new key mapping from Zero Hour.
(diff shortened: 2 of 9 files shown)


Patch108pCCG/GameFilesCore/Data/French/CommandMap.ini
```diff
@@ -654,6 +654,17 @@ CommandMap SELECT_ALL
   DisplayName = GUI:SelectAll
 END
 
+; Patch108p @feature xezon 10/05/2025 Backports new key mapping from Zero Hour.
+CommandMap SELECT_ALL_AIRCRAFT
+  Key = KEY_W
```
Skyaero42, line 659: In French it should be bound to 'Z' because the French use an azerty-keyboard. (The 'select-all' is therefore assigned to 'A' for example)

Patch108pCCG/GameFilesCore/Data/German/CommandMap.ini
```diff
@@ -654,6 +654,18 @@ CommandMap SELECT_ALL
   DisplayName = GUI:SelectAll
 END
 
+; Patch108p @feature xezon 10/05/2025 Backports new key mapping from Zero Hour.
+; Uses CTRL+Q mapping because W is already mapped by Guard mode.
+CommandMap SELECT_ALL_AIRCRAFT
+  Key = KEY_Q
```
Skyaero42, line 660: Where would the guard button be defined? I think it is a unit-specific command, but I cannot find it in the INI. 
Based on the German keyboard layout (qwertz), it does not make sense it is assigned to W and should be changed, rather that assign an illogical key to select all aircraft.

edit: I saw your youtube video and see the issue. I still think we need to change the german key layout to match the keyboard. 'bewachen' could also have hotkey 'b' and still be logical (if 'b' isn't used for something else)

Skyaero42 (changes requested):

Patch108pCCG/GameFilesCore/Data/German/CommandMap.ini
xezon: Well we open Pandoras box if we try to change keys at mass, because one reassignment will lead to new conflicts.

Patch108pCCG/GameFilesCore/Data/French/CommandMap.ini
xezon: Why?

Patch108pCCG/GameFilesCore/Data/German/CommandMap.ini
aliendroid1: I think we should let all keys be assignable from ini, csf, or string files instead of hard coding them

PR #2836, opened 2026-01-12
Fix: Remove suicide Upgrade icon from the GLA Demo BombTruck

### LINKED ISSUE #2751: Bug: Demo General's Bomb Truck has the suicide upgrade icon, but has no benefit from the upgrade
The upgrade icon should be removed.

DrGoldFish1:
The GLA Demo Bomb Truck has the suicide upgrade icon, but has no benefit from the upgrade.
So we remove the Upgrade icon from the Bomb Truck.

* Closes #2751


FireRainV:
Another option could be to increase its damage output when you get the upgrade, but that would make it a balance change.

